Issue #29, opened 2016-05-23
Mimic by Default


ryanleesipes:
This should be the case. Better quality and we maintain the project. Just makes sense.

PR #36, opened 2016-05-23
Use mimic as default TTS, updated mycroft/configuration/default.ini

aatchison:
https://github.com/MycroftAI/mycroft-core/issues/29

Would this be enough?


ryanleesipes:
Need to make sure Mimic is installed as part of the process.

jdorleans:
If we ensure mimic is always been installed, then yes! :+1:

aatchison:
I believe we fixed that with a previous PR

aatchison:
lgtm

PR #40, opened 2016-05-23
Media skills WIP (but quite functioning)

forslund:
I'm putting this back up for review, reference and testing. For me it seem to work just fine but there are loads of improvements that could be made.

# What it does

Adds a base class for all media skills setting up intents for the most common functions (next, previous, stop). (the individual skills need to implement the actual functionality). There is basic IPC to stop a playing skill when another skill is started.

Three skills are included, spotify, local music, swedish radio.

## Mopidy Setup
Skills using mopidy such as spotify and local music requires some configuration of the mopidy media server to work properly.

Firstly mopdiy-spotify requires [libspotify](https://developer.spotify.com/technologies/libspotify/).

Install ```mopidy```, ```mopidy-spotify``` and ```mopidy-local-sqlite``` from pip (or by running ```./dev_setup```)

in *~/.config/mopidy/mopidy.conf* under
``` [spotify] ```

```
enabled=true
username=USERNAME
password=PASSWORD
```

and under
``` [local] ```

```
enabled = true
library = sqlite
media_dir = PATH_TO_YOUR_MUSIC
```

after this is done scan the local collection by running

``` mopidy local scan ```

and then launch mopidy

``` mopidy ```

## The example skills
**Swedish Radio**
- Swedish radio, a skill to tune in to Swedish radio
examples: 

```play p3```
```play p3 from swedish radio```

**spotify**
- a skill to play music from spotify. the skill collects the play lists from the user and will play them if requested. If the spotify skill is specified explicitly it also performs an album search
examples:

```play discover weekly```
```play hello nasty from spotify```

**local music**
- the local music skill browses the local media directory and adds each album found as a play intent.

# TODO

- doc strings
- multiple sound sinks (local, living room, etc)
- probably pep8
- more skills!


aatchison:
Cool! Trying to get the licence headers in so we can all re-base again :P Then I'll see about testing this.

aatchison:
It looks like there is a dependency error. From travis-ci logs:

>     build/temp.linux-x86_64-2.7/spotify._spotify.c:422:28: fatal error: libspotify/api.h: No such file or directory
    compilation terminated.
    error: command 'gcc' failed with exit status 1
    
    ----------------------------------------
Command "/home/travis/virtualenv/python2.7.9/bin/python -c "import setuptools, tokenize;__file__='/tmp/pip-build-ngPiqL/pyspotify/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-Of7yPS-record/install-record.txt --single-version-externally-managed --compile --install-headers /home/travis/virtualenv/python2.7.9/include/site/python2.7/pyspotify" failed with error code 1 in /tmp/pip-build-ngPiqL/pyspotify
The command "VIRTUALENV_ROOT=${VIRTUAL_ENV} ./dev_setup.sh" failed and exited with 1 during .
Your build has been stopped.

>

forslund:
@aatchison, yes. libspotify is an external dependency for mopidy-spotify. Shall I add that installation to dev_setup.sh to fix?

aatchison:
How are you satisfying this requirement? I'm not seeing it in my apt-cache (Ubuntu 16.04) , and `pip install -p mopidy-spotify` fails becasue of the libspotify dependency.

aatchison:
Ahh, I see that I need three mopidy deps. You can add these to the requirements.txt and be good I think; confirming.

forslund:
It's manual download and install so far. See link in description above.

May exist in the spotify apt-repo but I'm uncertain.
Den 23 maj 2016 23:24 skrev "aatchison" <notifications@github.com>:

aatchison:
Oh, NVM then :D

ryanleesipes:
@forslund how would you put libspotify in the dev_setup.sh?

aatchison:
I wonder if this might be better as a skill created with the SDK? That way, you could have separate dependencies and attach to a running Mycroft messagebus. Now sure though, @clusterfudge ?

ryanleesipes:
@aatchison I was considering this too. Mopidy is a great base media player, would almost rather use this that mpg123 for all the media we can. The problem is managing dealing with the plugins - Spotify, Pandora, etc.

I'm interested in how we might handle this. But I'd love to see us figure something out. Also, in regards to Spotify and Pandora (and these other 3rd party media streaming services), we should provide users a way to enter their creds on Cerberus. However, we need to ensure this is done in a safe and secure fashion.

aatchison:
I'm more worried about dependencies that aren't upstream. A PPA doesn't sound so hot either. Hrrm

aatchison:
I'm about to try this!

forslund:
Sorry for the lack of responses, I had to sleep. I'll look into the sdk. It might be an idea to keep the base skill (media) internal so it's easy to inherit from but I'll have to look at how the sdk is handled first. I'll create a script to download and install libspotify in any case and it may be included or left out of the dev_setup.sh as you guys see fit.

forslund:
Seem to have closed this by mistake but that might just be as well. I'm gonna try moving it over to a standalone skillset using the skill-sdk. The main reason is that even if mopidy is properly installed it seem to need some system libs that basically will only work if the venv was created ```--with-system-site-packages``` (I don't think the standard mycroft venv is configured this way). Back to the drawing board.

ryanleesipes:
Can you submit the media skill without Spotify?

On Tue, May 24, 2016, 10:46 AM Åke <notifications@github.com> wrote:

> Seem to have closed this by mistake but that might just be as well. I'm
> gonna try moving it over to a standalone skillset using the skill-sdk. The
> main reason is that even if mopidy is properly installed it seem to need
> some system libs that basically will only work if the venv was created
> --with-system-site-packages (I don't think the standard mycroft venv is
> configured this way). Back to the drawing board.
>
> —
> You are receiving this because you commented.
>
>
> Reply to this email directly or view it on GitHub
> <https://github.com/MycroftAI/mycroft-core/pull/40#issuecomment-221314476>
>
-- 
Ryan Sipes
CTO, Mycroft A.I.
https://mycroft.ai
785-979-6091

clusterfudge:
Yeah, system site packages is purposely not enabled on our venvs, as it creates a leaky sandbox. There was an unfortunate amount of effort that went into installing pocketsphinx and pygtk into a virtualenv, and that's what you'd want to do here. If you can point me to docs/source on the mopidy dependency, I can see what would be involved and offer direction.

forslund:
@ryanleesipes I can submit the media-base class and the swedish radio skill. (as an example skill)

Local music also depends on mopidy but will work as long as mopidy is installed and running outside of mycroft venv

@clusterfudge, the big problem seems to be python-gst-1.0 (that in turn depends on gstreamer). pygtk is also a dependency but that is already solved. I understand the decision to keep the sandbox waterproo. (who likes mud anyway?)

forslund:
@clusterfudge more details:

The mopidy docs (http://mopidy.readthedocs.io/en/latest/devenv/) says 
Step 1. Install system version to get all dependencies
Step 2. Make virtualenv using system site packages
Step 3. Install using ```pip install mopidy```

https://gist.github.com/jegger/10003813 gives a clue to which packages and files are of interest

gstreamer from source doesn't seem to be trivial, they don't provide a tar.gz and their repo is divided into several  modules and might be tricky to build.

Module of interest would probably be gst-python fetched by:
```git clone git://anongit.freedesktop.org/gstreamer/gst-python```

PR #45, opened 2016-05-25
Issues 14 - Listener Improvements and Fixes

jdorleans:
1. Preventing many of the listener "dead state" (avoiding unwanted exceptions to break the loop)
2. Wake up must work for both "mycroft wake up" and "wake up mycroft"
3. To ensure "mycroft" keyword is always detected when it exists
4. To rely on "mycroft" instead of "hey mycroft"
5. To process "ok", "okay" and "allright" as part of the wake word detection
6. Faster, more precise and  keyword extractor
7. Lots of bug fixes! 

Some of the possible test cases:

1. wake up mycroft with: "mycroft wake up" and "wake up mycroft"
- it must wake up no matter the position of the "wake up" keyword
- it must always say it's awake at the first time (before, it'd only say it the second time you try to wake up)

2. try "okay mycroft, what's the weather in lawrence" and similar requests with "alright"
- this ensures we ignore "okay" and "alright" on the sentence

3. "mycroft" keyword should be detected as much as possible even if the sentence does not trigger a real skill/action
- that means the wake word detection is working

4. try all of those from close distance to the unit
- this ensures the changes are working

5. try all of those from a far distance to the unit
- to test the auto gain mic (for those who have one)


mycroft/client/speech/listener.py
```diff
@@ -57,110 +56,6 @@ def run(self):
                     self.emitter.emit("recognizer_loop:ioerror", ex)
 
 
-class WakewordExtractor:
```
clusterfudge: Just to confirm (because it's difficult to tell from the diff alone), this code was just moved, not moved and modified?

mycroft/client/speech/listener.py
```diff
@@ -170,137 +65,135 @@ class AudioConsumer(threading.Thread):
...
-            hyp = self.wakeup_recognizer.transcribe(
-                audio.get_wav_data(), metrics=self.metrics)
-            if hyp and hyp.hypstr:
-                logger.debug("sleeping recognition: " + hyp.hypstr)
-            if hyp and hyp.hypstr.lower().find("wake up") >= 0:
-                SessionManager.touch()
-                self.state.sleeping = False
-                self.__speak("I'm awake.")  # TODO: Localization
-                self.metrics.increment("mycroft.wakeup")
+            self.wake_up(audio)
```
clusterfudge: i think this method name can be more helpful; perhaps self.check_wake_up(audio), as self.wake_up phrased as a verb  suggests the state is being changed unconditionally (as opposed to on the condition that the input matches).

mycroft/client/speech/listener.py
```diff
@@ -170,137 +65,135 @@ class AudioConsumer(threading.Thread):
...
+        elif self.state.skip_wakeword:
+            self.skip_wake_word(audio)
```
clusterfudge: see comment above.

mycroft/client/speech/listener.py
```diff
@@ -170,137 +65,135 @@ class AudioConsumer(threading.Thread):
...
-
-                SessionManager.touch()
-                payload = {
-                    'utterance': hyp.hypstr,
-                    'session': SessionManager.get().session_id,
-                    'pos_begin': int(extractor.range.begin),
-                    'pos_end': int(extractor.range.end)
-                }
-                self.emitter.emit("recognizer_loop:wakeword", payload)
+            self.wake_word(audio, timer)
```
clusterfudge: see comment above

aatchison:
She builds, checks passed. Over my head but I will test tomorrow:)

test/client/audio_consumer_test.py
```diff
@@ -1,171 +1,136 @@
...
-        self.consumer.try_consume_audio()
+        self.consumer.read_audio()
+
         utterances = monitor.get('utterances')
         self.assertIsNotNone(utterances)
         self.assertTrue(len(utterances) == 1)
         self.assertEquals("what's the weather next week", utterances[0])
 
-    def test_wakeword_in_phrase(self):
+    def test_wakeword(self):
```
clusterfudge: You are removing tests here that confirm designed behavior. Why?

clusterfudge:
I have a couple of questions before I can sign off on this. I have concerns about the unit tests that are being removed, as they cover intentionally designed behavior.

test/client/audio_consumer_test.py
```diff
@@ -1,171 +1,136 @@
...
-            audio.sample_width)
-        TRUE_POS_END = 89138 - int(
-            WakewordExtractor.TRIM_SECONDS * audio.sample_rate *
-            audio.sample_width)
-
-        TOLERANCE_RANGE_FRAMES = (
-            WakewordExtractor.MAX_ERROR_SECONDS * audio.sample_rate *
-            audio.sample_width)
+        tolerance = 4000
+        begin = 70000
```
Wolfgange3311999: Should we add a comment on a way to find this number? This will help with any future changes to the wakeword.

test/client/audio_consumer_test.py
```diff
@@ -1,171 +1,136 @@
...
+        end = 92000
```
Wolfgange3311999: I think `begin` and `end` should be renamed. There's not really a clear difference between the name of other variable, `pos_begin`, and `begin`.

mycroft/client/speech/listener.py
jdorleans: The word extractor code was completely rewritten.

test/client/audio_consumer_test.py
jdorleans: Yes, I'll do that. Just as a brief information, those number are not found by running the work extractor anymore, because this is intended to test the extractor itself. So, those values were found using another tool (Audacity, in this case) and figuring out the sample values for begin and end.

jdorleans:
@clusterfudge the removed tests are been covered by the remaining tests. As they were not adding real use cases I decided to simplify this test set and remove them.

clusterfudge:
Can you indicate which tests are covering the removed tests?

test/client/audio_consumer_test.py
```diff
@@ -1,171 +1,144 @@
...
         self.assertTrue(len(utterances) == 1)
         self.assertEquals("what's the weather next week", utterances[0])
 
-    def test_wakeword_in_phrase(self):
+    def test_wakeword(self):
         self.queue.put(self.__create_sample_from_test_file('mycroft'))
+        self.recognizer.set_transcriptions(["silence"])
         monitor = {}
-        self.recognizer.set_transcriptions([
-            "he can do other stuff too", "what's the weather in cincinnati"])
```
jdorleans: There is no real audio testing this 2 phrases and `mycroft` wakeword alone should not result in any remote STT query, because it'll only produce silence after the `word extraction` process. In practice, this test had exactly the same behaviour of the test: `test_wakeword_in_beginning` which uses a suitable audio test with a sentence including the wakeword.

jdorleans: @clusterfudge explanation above.

test/client/audio_consumer_test.py
```diff
@@ -1,171 +1,144 @@
...
 
         self.loop.once('recognizer_loop:utterance', callback)
-        self.consumer.try_consume_audio()
-        utterances = monitor.get('utterances')
-        self.assertIsNotNone(utterances)
-        self.assertTrue(len(utterances) == 2)
-        self.assertEquals("he can do other stuff too", utterances[0])
-        self.assertEquals("what's the weather in cincinnati", utterances[1])
-
-    def test_call_and_response(self):
```
clusterfudge: I do not see the behavior of this test captured anywhere. This test verifies call and response behavior. The expected transcriptions has changed a bit, but the behavior of the state machine transitioning from "waiting for wakeword" to "record audio regardless of wakeword" needs to be verified via test. It has broken several times in the past.

jdorleans:
@clusterfudge I've added the call_response test fixing the second call to ensure the sentence about the weather exists. Please, check it out so we can finish this PR.

clusterfudge:
thanks for getting that test back and for the awesome work! LGTM.

ryanleesipes:
If it is good to go, let's merge and cut a release.

Issue #46, opened 2016-05-25
skill container makes it hard to import files


forslund:
When using the mycroft-skill-container the skill may be launched from any location in the file system this makes things complicated if the skill is split up in different files as imports will be relative to the position in the file system and not the skill.

forslund:
Resolved by #47

PR #47, opened 2016-05-25
Add skill directory to path.

forslund:
To address issue #46 this PR adds the skill directory to the path when loaded. This allows python files placed in the skill directory to be imported easily from the main skill ```__init__.py```.


clusterfudge:
Yup, awesome. My only comment would be to keep the modifications to sys.path to together.

aatchison:
gtg sean?

PR #62, opened 2016-05-27
Fix to Alarm Distortion

ryanleesipes:
Tested with new alarm, it sounded distorted. Looked at implementation of other sounds (news skill), and decided to try with MP3. Changed to using MP3 instead. Everything works properly now, tested on multiple devices.


aatchison:
lgtm

Issue #61, opened 2016-05-27
Listener Improvements Remove Ability to put "Mycroft" at End of Sentence


ryanleesipes:
Upon further testing of Listener Improvements, PR #45 - it appears that the overall listener has improved, but upon being asked, "What time is it Mycroft" or "What's the weather, Mycroft" - the sentence before is ignored and Mycroft prompts the user to "Go Ahead."

ryanleesipes:
Upon more testing it appears as though the new improvements do often cause Mycroft to opt for call and response "Go Ahead", even when it seems as though it should have been able to process the request.

Wolfgange3311999:
I'm assuming this is the same issue I originally encountered when making this:
The end marker is generated first and often is a bit too far into the wake word segment by a small margin. Afterwards, when generating the begin marker, `self.recognizer.is_recognized()` will always return `False` pushing the begin marker to 0 resulting in the first part of the phrase being cut out.

So to fix this, there are two options. One is to save the last end position where `self.recognizer.is_recognized()` returned true and use it when finding `self.begin`. Another option is to push `self.end` back by something like `2*self.precision` before finding `self.begin`. After finding both markers, `self.end` should be set back to the correct value. Note that `self.end` should not simply be set to the end of the audio file as that would cause it to freak out with something like "mycroft, what is mycroft".

ryanleesipes:
Adding assignee @Wolfgange3311999 to help move this along.

Wolfgange3311999:
@ryanleesipes: Does the silence threshold PR fix this 100% of the time?

PR #66, opened 2016-05-27
 Issues 61 - Reducing silence threshold to ignore a sentence

jdorleans:
This PR solves #61.
It turns out the silence threshold of 1 second was big enough to ignore many sentences causing the "Go ahead behaviour" instead of processing the query. We have decided to reduce it to 0.5 seconds.


aatchison:
I still get the go-ahead behavior when I start the request with "Hey Mycroft."
I get a multiple intent parsing error or WA search because it splits: "Hey Mycroft, what time is it?" into: 'haim','time is it.'  Every once in a while it succeeds.

Otherwise:
"Mycroft, what time is it."
"What time is it, Mycroft."
"Hey, Mycroft". "Go ahead." What time is it?"
are all working about 7/10 times on my cruddy laptop mic.

clusterfudge:
are you suggesting that "what time is it" is an utterance < 1s? That seems unnlikely to me.

ryanleesipes:
@clusterfudge, that is exactly what I'm suggesting. Before the threshold was moved to .5, "What time is it, Mycroft?" And "What's the weather, Mycroft?" were failing.

jdorleans:
Ok guys, we have to be careful here and focus on the scope of this task. After many trails we have noticed that reducing the threshold is beneficial and it prevents short sentences to be ignored like: **what's the time**, **what's the weather** or simply *stop* (from Mycroft Stop), *record* etc.

Wolfgange3311999:
What's the length of the utterance "stop" when detected with a high quality mic? Maybe it should be reduced even further.

However, with low quality mics (in my experience), even a short "stop" can take 0.8-1.5 seconds. Perhaps this should be a preference (that ideally gets calibrated)?

aatchison:
I think we need to think about how we might test different values? Could we detect mic quality in regards to noise or range of the recording? Pretty vague I know.

Wolfgange3311999:
Just brainstorming, to calibrate both this value and maybe some others such as the dynamic adjustment ratio in the speech recognition library, after pairing Mycroft could optionally perform a calibration session where if on a computer you set the volume to the maximum in a quiet environment and it outputs a beep of 1 second measuring the extra length after the beep from whatever it is that causes the variance between mics.

However, for the scope of this PR I'd say someone should test `Mycroft stop` a few times on a good mic and find the maximum acceptable silence threshold.

ryanleesipes:
As @Wolfgange3311999 points out, "Mycroft stop" does not work well with this threshold. Other than that this appears to work quite well.

clusterfudge:
The original purpose of this threshold was twofold, one reason no longer applying. 

1) Decrease the number of calls to cerberus, to minimize unnecessary API calls on a paid resource, or decrease load on a service we own (in the future of openSTT)
2) To decrease the number of STT calls being made on the recognizer loop, which dramatically increased interaction latency.

Since the STT calls have been parallelized, 2 is no longer a problem. If we're willing to just eat the cost of the extra API calls, we may as well drop the threshold to 0 (or remove the threshold code altogether). I haven't seen any data in this ticket yet as to what makes a good threshold, and I don't want to keep guessing with unexpected results.

jdorleans:
The threshold is still valid. Every time we remove "Mycroft" wake word from a sentence (using the word extractor algorithm), it will generate 2 sentences and one of them is very likely to be only silence. For instance, the utterance: "Mycroft, what time is it?" will result in the following 2 sentences: 
- S1: "" (silence, mycroft was removed)
- S2: "what time is it"
We do not want to process S1. Then, the silence threshold will helps us to ignore it.

Our new Word Extractor adds less silence to the split sentences than before. Which means, we need to recalibrate the threshold. 

I'll do some new tests with "Mycroft stop", because it seems to be the shortest query we have. Finally, I'll adjust the threshold again and provide a new commit.

clusterfudge:
In your last comment you say "We do not want to process S1". Can you elaborate on which processing we want to avoid? Server-side STT? Local intent determination? Something else?

jdorleans:
We wanna discard S1, since it's only silence and it is not useful for further processing both local and server-side.

clusterfudge:
So, realistically, either S1 or S2 can be the incorrect phrase (this is currently broken and referenced in #61 ). The original problem was that sometimes S1 **or** S2 would not be useful, **and** that negatively impacted user experience (via latency in the main speech loop).

The recognizer as is does not suffer from the previously mentioned latency problems because of parallel STT calls. The behavior of the speech loop is identical whether we drop the audio locally or get back an error/empty response from STT, so there's no change to user experience here. 

My main argument is that we have a magic number that we're changing because it "works better", without any evidence in either direction. My understanding is that we can remove this code and have identical user behavior.  I am a huge fan of gaining correctness (thus improving user experience) by removing complexity over tinkering with magic numbers or introducing more complexity. 

TL/DR; This change claims "improvements", but it's an empty claim. Please make your case with unit tests, or consider a change that can be validated and enforced (via automated tests) long term.

ryanleesipes:
Let's get some unit tests in here, and get this in. Mycroft is seemingly much more attentive now, and adjusting the min_audio_size does fix the problem of him responding with "Go Ahead" vs just doing the action.

We need some tests in here that validate that a change in this regard is the best move (and I agree with @clusterfudge - perhaps we should get rid of this "magic number" altogether). Just let me know what we need, and we'll get it (audio files or whatever they are).

clusterfudge:
@ryanleesipes you appear to be trying to sneak some StopSkill changes into this PR. Please don't.

ryanleesipes:
Yeah, that was a mistake (literally clicked the wrong thing). Needs to be reverted.

ethanaward:
Seems to be working well for me.

ryanleesipes:
Tested, LGTM, Merge.

PR #67, opened 2016-05-27
 Issues 55 - Stopping skills by voice command

jdorleans:
Stop skill is intend to stop all skills by voice command. e.g. 
- Mycroft stop
- Mycroft shut up

**Note:** 
- This skill depends on #66 to work properly.
- Keep in mind this PR is not intended to solve audio cancellation and It will not work correctly in case where there is a sound getting out the speaker (e.g. when NPR is playing).

### DIFF
--- a/mycroft/skills/stop/__init__.py
+++ b/mycroft/skills/stop/__init__.py
@@ -1,0 +1,49 @@
+# Copyright 2016 Mycroft AI, Inc.
+#
+# This file is part of Mycroft Core.
+#
+# Mycroft Core is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Mycroft Core is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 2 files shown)


aatchison:
reviewing

jdorleans:
@aatchison this will not work properly if we do not reduce the silence threshold... solved on PR #66. I'm still waiting that approval. Maybe you can review that one first as @ryanleesipes mentioned both of you were having the same issues with **mycroft** wakeword in the end of a sentence.

aatchison:
Ok, in progress.

aatchison:
Without the slience threshold adjustments, this works pretty well if I use the `./start.sh cli --quiet` option and it also seems to work well If I turn the speaker way down while NPR is playing.

lgtm

clusterfudge:
so, realistically, this change is not going to work well without acoustic echo cancellation. This is something we've covered several times as not solveable without echo cancellation. Last I heard, we were talking about hardware echo cancellation. What happened with that effort?

aatchison:
I believe that we wanted to integrate the events into each skill, so it would be able to stop itself.  The enclosure, voice, cli, or any other future triggers would be nice.
Besides, some skills might not need to play audio on the device to work. Mycroft stop the light show, or Chrome cast would still be useful.
In the past, the knob button was sending an event to the enclosure client, to do thinks like SIGINT mpg123 directly.

jdorleans:
Actually, this skill is totally realistic. We have to remember not all skills requires or uses the speaker for a long time. At this state, the Stop Skill is capable of stopping the alarm, reminder, audio recorder (during recording and playback) and probably few others.

PR #68, opened 2016-05-28
Explicitly install python-setuptools

clusterfudge:
Brings in easy_install. Thought this was a dependency of python-virtualenv, but it may not be (virtualenv may vendor this in)

verified this requirement when trying to run build_host_setup.sh on a clean 16.04 vm.


clusterfudge:
OOOH MULTIPLE ASSIGNEES!

My life is complete.

aatchison:
lgtm.

PR #72, opened 2016-05-29
Suggestion: default config for skills running in skill container

forslund:
This is a suggestion for allowing the skills running in the ```mycroft-skill-container``` to use a default.ini file in the same way as the built-in skills. I believe this would be useful for many third-party skills. 

By default the file ```./config/mycroft.ini```will be used but this can be altered by using the command line argument ```--default-config```.

### DIFF
--- a/mycroft/configuration/config.py
+++ b/mycroft/configuration/config.py
@@ -33,7 +33,11 @@ DEFAULTS_FILE = os.path.join(
 ETC_CONFIG_FILE = '/etc/mycroft/mycroft.ini'
 USER_CONFIG_FILE = os.path.join(
     os.path.expanduser('~'), '.mycroft/mycroft.ini')
-DEFAULT_LOCATIONS = [DEFAULTS_FILE, ETC_CONFIG_FILE, USER_CONFIG_FILE]
+
+DEFAULT_LOCATIONS = [DEFAULTS_FILE]
+SYSTEM_LOCATIONS = [ETC_CONFIG_FILE]
+USER_LOCATIONS = [USER_CONFIG_FILE]
+CONFIG_LOCATIONS = DEFAULT_LOCATIONS + SYSTEM_LOCATIONS + USER_LOCATIONS
 
 
 class ConfigurationLoader(object):
@@ -56,7 +60,7 @@ class ConfigurationLoader(object):
     def load(self):
         """
         Loads configuration files from disk, in the locations defined by
-        DEFAULT_LOCATIONS
+        class member config_locations
         """
         config = {}
         for config_file in self.config_locations:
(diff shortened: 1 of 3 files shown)


clusterfudge:
It looks like the test runner depends on the old behavior (inject legacy config paths and then the test config on top of it).

Can you dig into [this](https://travis-ci.org/MycroftAI/mycroft-core/builds/133742779#L5517)?

forslund:
Sure thing

mycroft/configuration/config.py
```diff
@@ -128,17 +132,20 @@ class ConfigurationManager(object):
     _config = None
 
     @staticmethod
-    def load(*config_files):
+    def load(defaults=CONFIG_LOCATIONS, *config_files):
```
clusterfudge: looking back now, I'm surprised that python allows this. My understanding is that you can't have any non-keyword arguments (and varargs definitely fit the bill for that) after kwargs (defaults= being a kwarg).

clusterfudge:
I may be worthwhile to use a builder pattern here, such as 
```ConfigurationLoader().withDefaults().append(file).appendAll(files).load()```

*note* do not consider the above anything more than half-assed, please make a pretty pattern if you see a better one :smile:

File: mycroft/configuration/config.py (hunk 1 above)
Comment (forslund) [inline mycroft/configuration/config.py]: Yeah, it definitely didn't do what I intended it to do. It might have worked in python3 but I'm not 100% sure of that either.

**Comment by forslund** [comment]
*sigh* I didn't see those errors...but now I seem to get them on master as well? Is [this](https://travis-ci.org/MycroftAI/mycroft-core/builds/133838862#L5580) something I did?

**Comment by forslund** [comment]
Currently it should (hopefully, maybe) be backwards compatible (except for the new errors in the unittest) which was my original intent.

I'll look into the builder pattern thing, the C programmer in me tells me it looks crazy but I'm willing to give it  a shot. Just need to find a reference to study...

**Comment by ryanleesipes** [comment]
Please provide feedback on this @jdorleans.

**Comment by forslund** [comment]
This is now at sort of a half-way point, a builder is used to create the list and is used internally in the ConfigurationLoader but it's not been integrated into the ConfigurationLoader class. Is the format along the lines you expected?

**Comment by jdorleans** [comment]
@forslund that's a very interesting suggestion and we definitely have to provide a flexible and better way for thrid-party skills to integrate their configuration. That said, I'd like to make your changes simpler. I'll PM you so we can talk about some changes.

File: mycroft/configuration/config.py [hunk 2]
```diff
@@ -152,3 +157,75 @@ def get_config():
...
+    """
+    def __init__(self):
+        """
+            Creates an empty ConfigBuilder object.
+        """
+        self.default = []
+        self.system = []
+        self.user = []
+        self.custom = []
+        self.ptr = self.custom
```
jdorleans: What's `self.ptr` for?

mycroft/configuration/config.py
```diff
@@ -152,3 +157,75 @@ def get_config():
...
+        if len(files) > 0:
+            self.system += list(files)
+        self.ptr = self.system
+        return self
+
+    def with_user(self, *files):
+        """
+            Add configs to the user config slot.
+        """
+        if lden(files) > 0:
```
jdorleans: `len` instead of `lden`

mycroft/configuration/config.py
```diff
@@ -152,3 +157,75 @@ def get_config():
         if not ConfigurationManager._config:
             ConfigurationManager.load()
         return ConfigurationManager._config
+
+
+class ConfigBuilder(object):
```
jdorleans: Overall, it seems we can do a simpler code here. The main idea around our configuration management is to read config files in a priority order and replace **the lowest** relevant configs with the **more relevant** ones. From **default** to **user** level in this case. That means we can control an unique array of config files instead of splitting them into categories and adding unnecessary complexity to control those. I'd like to provide a suggestion on top of your changes. @forslund ping me once you have some time, so we can talk about that.

mycroft/configuration/config.py
forslund: It's for the .append() method. Continue adding at the last point, this is strictly not needed.

jdorleans:
Closing this in favor of all new features we have discussed at #112

PR #76, opened 2016-05-31
Issues 73 - Add 'what does wikipedia say about' to wiki skill vocab

ethanaward:
I added 'what does wikipedia say about' to both the vocab file for the wiki skill and to prefixes in the __init__.py file for the wiki skill so that it can be used to trigger a wikipedia search.

### DIFF
--- a/mycroft/skills/wiki/__init__.py
+++ b/mycroft/skills/wiki/__init__.py
@@ -49,8 +49,7 @@ class WikipediaSkill(MycroftSkill):
         self.load_vocab_files(join(dirname(__file__), 'vocab', 'en-us'))
 
         prefixes = ['wiki', 'wikipedia', 'tell me about', 'tell us about',
-                    'who is', 'who was',
-                    'what does wikipedia say about']  # TODO - i10n
+                   'who is', 'who was']  # TODO - i10n
         self.__register_prefixed_regex(prefixes, "(?P<ArticleTitle>.*)")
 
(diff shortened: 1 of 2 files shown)


ryanleesipes:
While we are messing with this, please remove 'who is' and 'who was' from the vocab. We are going to let Wolfram Alpha handle that.

ryanleesipes:
Oh no! These are failing @ethanaward - run ./start.sh unittest to do this locally, so you know whether or not it will fail before pushing.

ryanleesipes:
@jdorleans any idea why this is failing? I can literally see why it is failing, but I don't know how to solve it.

mycroft/skills/wiki/__init__.py
```diff
@@ -49,7 +49,7 @@ def initialize(self):
         self.load_vocab_files(join(dirname(__file__), 'vocab', 'en-us'))
 
         prefixes = ['wiki', 'wikipedia', 'tell me about', 'tell us about',
-                    'who is', 'who was']  # TODO - i10n
+                    'who is', 'who was', 'what does wikipedia say about']  # TODO - i10n
```
jdorleans:
```
mycroft/skills/wiki/__init__.py:52:80: E501 line too long (88 > 79 characters)
The command "pep8 mycroft test" exited with 1.
```

jdorleans:
Personally, I think 80 characters is **too** restrict now days. We get to adjust those PEP8. For the time begin, please, fix it.

PR #81, opened 2016-06-01
Issues 76 - Fix changes to wikipedia skill so that tests pass

ethanaward:
I changed the wikipedia skill so that it triggers on 'what does wikipedia say about'. I also removed 'who is' and 'who was' in the vocab, as those will be handled by Wolfram. This is a new pull request to be handled by Travis.


ryanleesipes:
LGTM, merging.

PR #87, opened 2016-06-02
Issues 86 - Add basic welcome skill

ethanaward:
I added a skill that will have Mycroft say "You're welcome" when you say either "thanks" or "thank you".

### DIFF
--- a/mycroft/skills/welcome/__init__.py
+++ b/mycroft/skills/welcome/__init__.py
@@ -32,11 +32,11 @@ class WelcomeSkill(MycroftSkill):
 
     def initialize(self):
         self.load_data_files(dirname(__file__))
-        
+
         welcome_intent = IntentBuilder("WelcomeIntent").require("WelcomeKeyword").build()
         self.register_intent(welcome_intent, self.handle_welcome_intent)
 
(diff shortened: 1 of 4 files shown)


LongBoolean:
I made a similar skill my self, great minds think alike aye. It may seem like a simple or unimportant type of skill to have, but I do think there are some good things about having a "candid response" skill. (as @clusterfudge once phased it) I think at the moment if you want to nonchalantly refer to mycroft in conversation with out getting "I don't understand your request" you need to put mycroft to sleep. But there are cases where it would be nice to say out loud "Mycroft sure is amazing!" and have a response more like "Thank you for the compliment." Of coarse skills like this don't seem to be a high priority at the moment, but I do think such a skill would help cover up some of the holes and make the mycroft experience feel a little more complete.

ethanaward:
Yeah, it definitely makes Mycroft seem more complete, especially getting more coverage so that he doesn't say "I don't understand your request" to quite so many things. I also did this so that I could get a handle on skill creation with a basic skill, as well as for documentation purposes so that people can see how to create a simple skill.

ryanleesipes:
LGTM, merging.

Issue #92, opened 2016-06-03
Math Skill


Wolfgange3311999:
This should at least be able to do basic operations (+-*/).

ryanleesipes:
@ethanaward please review this and see what you think.

marksev1:
This would be good... Could you maybe expand it with conversion of physical properties or maybe also currencies, or should we look at different skills to do that...Maybe Wolfram Alpha I don't know?

ryanleesipes:
PR for this closed. Closing accompanying issue.

PR #93, opened 2016-06-03
MathSkill

Wolfgange3311999:
Oops, sorry for this new issue created by the PR.
This resolves issue #92 and adds a very basic math skill. However, I don't think this is ready yet. I couldn't find a way to either support decimal numbers or limit adapt so that something like `what is 2 + 3 * 5` won't trigger this intent.

For some reason the following does not work to allow decimals:
`self.register_regex('(?P<Value1>[\d.]+)')`

Nor does something more specific such as the following:
`self.register_regex('(?P<Value1>\d+(\.\d+)?)')`

### DIFF
--- a/mycroft/skills/math/__init__.py
+++ b/mycroft/skills/math/__init__.py
@@ -25,7 +25,8 @@ class MathSkill(MycroftSkill):
         self.register_regex('(?P<Value1>\d+)')
         self.register_regex('(?P<Value2>\d+)')
 
-        intent = IntentBuilder("MathIntent").require("Value1").require("MathKeyword").require("Value2").build()
+        intent = IntentBuilder("MathIntent").require(
+            "Value1").require("MathKeyword").require("Value2").build()
         self.register_intent(intent, self.handle_intent)
 
     @staticmethod
@@ -45,9 +46,11 @@ class MathSkill(MycroftSkill):
         operator_func = self.operators[operator_char]
         result = operator_func(first_val, second_val)
 
-        equation_string = self.__to_string(first_val) + ' ' + operator_char + ' ' + self.__to_string(second_val)
+        equation_string = self.__to_string(first_val) + ' ' + operator_char + \
+            ' ' + self.__to_string(second_val)
         answer_string = self.__to_string(result)
-        self.speak_dialog('read.answer', data={'equation':equation_string,'answer':answer_string})
+        self.speak_dialog('read.answer', data={'equation': equation_string,
+                                               'answer': answer_string})
 
     def stop(self):
         pass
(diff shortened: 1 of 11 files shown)


ethanaward:
Ideally we won't need a math skill if we can get Wolfram working correctly, right? Wolfram should be able to handle most to all of that.

Wolfgange3311999:
Well, I think we still want a math skill. Relying on Wolfram Alpha isn't the best due to the delay (and it shouldn't take more than a second for Mycroft to answer 2 + 2). This provides a very fast answer for simple questions and if we could get it to work only when it actually can provide an answer, I think this should be a core skill. What's everyone's thoughts on what this skill should ideally cover?

Cliabhach:
@Wolfgange3311999 I agree that we should have simple expressions calculated locally. Could we make the regex just check for numbers and math notation, and try to parse it as math if those are found? If it can't be evaluated then we pass the expression to Wolfram Alpha.

I'm thinking e.g. `30 * -312.7 / 10` would be ok locally but `10.45.6` and `10 squared` would fallback to WA.

Does that seem right?

Wolfgange3311999:
@cliabhach Yeah, that sounds like a good idea to just trigger wolfram on failure. I was thinking it might be a good idea to use [this](https://github.com/danthedeckie/simpleeval) for trying to parse it.

G10DRAS:
well you can try 
https://pypi.python.org/pypi/semantic/

specially the MathService. It ill take input like 

```
input = "five plus twenty one and a fifth"
input = "log sin eleven hundred"
input = "fifteen to the eleven point five power"
input = "3 divided by sqrt twenty five"
input  = "five plus twenty one and a fifth"
input  = "Log one hundred and ten"
input  = "log sin eleven hundred"
input  = "fifteen to the eleven point five power"
input  = "sqrt twenty five"
```

clusterfudge:
Had a response written, then read the feedback on this PR, and revising for stuff that wasn't already covered.

wolfram|alpha latency is a problem, but I think the benefits here outstrip the costs. Reimplementing math in each language seems arbitrarily  hard to me, especially if we're just pushing it down to a 3P library (as @G10DRAS is suggesting). If we build it, we want it to be something portable (language-wise). 

My understanding is that we can specify a language query param to wolfram|alpha and get math for free in other languages. That's super awesome to me. Maybe there's a middle ground where we only register intents/vocab if language='en-us', otherwise math queries would fall back to wolfram|alpha?

Wolfgange3311999:
@clusterfudge Yeah, that makes sense; So would it be fine to exclude any equations written as words ("2 + 2" works but "two plus two" triggers wolfram)? Since I think Google's STT always provides equations in symbols.

aatchison:
Closing this PR, as it's more of a conversation about an issue.

PR #95, opened 2016-06-03
Issues 71 - Clean up unit test output

ethanaward:
I created a dialog/en-us directory in the joke skill and created a dummy JokeDialog file inside it. This made it so that the unit test would stop complaining about the folder not existing.

### DIFF
--- a/mycroft/skills/joke/__init__.py
+++ b/mycroft/skills/joke/__init__.py
@@ -34,7 +34,7 @@ class JokingSkill(MycroftSkill):
         super(JokingSkill, self).__init__(name="JokingSkill")
 
     def initialize(self):
-        self.load_vocab_files(join(dirname(__file__),'vocab',self.lang))
+        self.load_vocab_files(join(dirname(__file__), 'vocab', self.lang))
 
         intent = IntentBuilder("JokingIntent").require("JokingKeyword").build()
         self.register_intent(intent, self.handle_intent)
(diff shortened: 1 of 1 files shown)


jdorleans:
Hey @ethanaward, the real problem in here is that Joke Skill is loading all data (dialogs and vocabs) using this function: `self.load_data_files(dirname(__file__))` and it should load only the vocab folders instead since it does not has dialogs in it. 

Another possible improvement should be to automatically detect the existence of dialogs and vocabs inside MycroftSkill and load only what exists instead. I'll create tasks concern to this suggestion. 

For the time been, I'd suggest to load only the `vocab` folder using `load_vocab_files` method. Please, refer to other skills that do the same.

mycroft/skills/joke/__init__.py
```diff
@@ -34,7 +34,7 @@ def __init__(self):
         super(JokingSkill, self).__init__(name="JokingSkill")
 
     def initialize(self):
-        self.load_data_files(dirname(__file__))
+        self.load_vocab_files(join(dirname(__file__), 'vocab', 'en-us'))
```
jdorleans: We have variables from MycroftSkill to the current language in use, please use that one:
`self.load_vocab_files(join(dirname(__file__), 'vocab', self.lang))`

jdorleans: We have variables from MycroftSkill to the current language in use, please use that one:
`self.load_vocab_files(join(dirname(__file__), 'vocab', self.lang))`

jdorleans:
Great, one last change and we're good to merge this. I'll provide a better way to solve the problem of loading missed resources later.

PR #97, opened 2016-06-04
Updated to include curl

UberPanda:
Tried on vanilla ubuntu-gnome, curl is not installed by default so dev_setup.sh does not work.


aatchison:
Lgtm

PR #102, opened 2016-06-05
Feature/hello world

the7erm:
I don't know if this is useful or not.  I wrote this while playing around with mycroft skills.

Please review and add it to your repo if you find it of any use.


ryanleesipes:
Interesting concept @the7erm! 

The "working hard on" is cute, testing now. :stuck_out_tongue_closed_eyes:

ryanleesipes:
My feedback is that some of the simpler vocab might be dangerous for when people quickly want to address Mycroft, such as "Hey Mycroft". If we could have him say "Hello" or "Hi" and then say "Did you need something" and open the listener loop (like what is done typically with: "Mycroft"), that would be really cool. Great idea and skill.

ryanleesipes:
Other possible triggers include "What are you up to?", "What's going on?"

jdorleans:
@the7erm thx for this PR. We have recently incorporated a Hello World Skill #109 that's very similar to what you have proposed. Although we would love to have some sort of MetricSkill, your proposal is a mix of hello word and metrics making it hard to accept.

Then, I'd suggest you create a new PR for MetricSkill only and instead of using the native `ps` binary, use the python `psutil` which is already incorporated into our requirements. This is a very easy-to-use library and you can find a simple, yet good documentation in the official github page: https://github.com/giampaolo/psutil

It'd be awesome to ask mycroft about memory, cpu, disc and many other resources status.

Dark5ide:
The most important parts of this PR are the comments ! Don't lose them. They make me understand how a _Skill_ works. I'm sure it can be useful to add them to the helloworld _Skill_ #109 for all new _Skill Maker_.

PR #106, opened 2016-06-06
Feature/issue 89

the7erm:
Please review.

PR #109, opened 2016-06-07
Feature/issues 86

ethanaward:
I've repurposed the welcome skill to be a hello world skill, as it will be used as an example for the skills documentation. Thanks to @the7erm for the ideas from your own Hello World skill in issue #106. It might be better to spin out your "What are you working on" functionality to its own skill, or make it a new intent in this skill.


jdorleans:
That's great! Merge it.

the7erm:
Cool, I was thinking of making a diagnostics skil.  So that works out.

aatchison:
lgtm then

Issue #63, opened 2016-05-27
Wolfram Removes "Who is" and "Who was" from Query


ryanleesipes:
Had Wikipedia handling this, but Wolfram should handle it better, need to fix it so that "Who is" and "Who was" are not removed from the query.

ryanleesipes:
To elaborate all question type queries have the question parts removed. This makes the Wolfram skill nearly completely useless.

ryanleesipes:
Reassigned to @jdorleans

ryanleesipes:
This should be fixed by #77

ryanleesipes:
Perhaps this is a problem related to #79 with the queries that are failing.

ryanleesipes:
Many queries are still not answerable through our Wolfram Alpha skill, that are answerable through the [web frontend](http://www.wolframalpha.com/). We are doing something wrong in processing these. Reassigned to @ethanaward

ethanaward:
I've isolated this problem to the Wolfram Alpha skill. If I ask Mycroft `who was Abraham Lincoln`, we get the information from Wolfram about who Abraham Lincoln was, but the skill then looks for a "pod" or chunk of information called `Result`, which a search for a person doesn't give you.  

One possible fix would be to have it look for `Result`, then if that doesn't exist, a pod title that all people queries have, and so on. We would want to try to provide complete coverage though, which may be tricky. 

While looking for a way to do this, I've noticed that the titles all of the pods in a person query have `PeopleData` in them, e.g. `id='WikipediaSummary:PeopleData'`. This means we could possibly search through the titles of the response to determine what data we should extract from the response. This could potentially not work if the API changes, however.

ryanleesipes:
@ethanaward this continues, I would like to see if we can troubleshoot the current API to address this. I feel like that is the safest bet, assuming any other API we gain access to through Wolfram will likely be similar.

ryanleesipes:
Seems to have been mostly solved by the Wolfram refactor from @ethanaward - closing ticket.

PR #110, opened 2016-06-07
Issues 63 - Add more cases where Wolfram responds

ethanaward:
This should fix some cases of the errors in #63. The Wolfram skill was only searching for a `Result` pod, which searches for people didn't return. I added some functionality where it now looks for a `Notable Facts` pod if it can't find results, and then a `Basic Information` pod if it can't find that. Ideally, it would just give a wikipedia summary of the person, but Wolfram seems to return that information as an image rather than as text.


jdorleans:
Looks good, I'd like to add some unit tests though. Let's make it part of our process that we'll help us to review and merge faster. Ping me I'll explain you what we need.

ethanaward:
I'm going to close this and submit a new PR, since the history of this one has gotten very muddied.

Issue #96, opened 2016-06-03
Skill Container - Load configuration file from custom locations


jdorleans:
@forslund pointed contained skills should be able to add custom configuration files and be independent from the defaults provided for the built-in skills.

PR #112, opened 2016-06-07
Issues 96 - Skill container should be able to load custom configuration

jdorleans:
Issues #96 
- Skill container can load custom configuration
- Using a more concise configuration standard
- Adding several unit tests to Configuration classes

Test Cases:

1. Pair device 
  - Delete identity file
  - Pair device

2. Load user configuration
  - Create a new configuration at the user level: ~/.mycroft/mycroft.ini
  - Change the user location, for example:
  ```
    [core]
    location = "Toronto, Ontario"
  ```
  - Ask Mycroft about the weather **without** the city name

3. Load the Skill Container
  - Copy the same configuration from step 2 into your home folder: ~/mycroft.ini
  - Start processes: **service** and **voice**
  - Start IntentSkill inside the skill container: `./start.sh skill_container --dir ./mycroft/skills/intent`
  - Start WeatherSkill inside the skill container using your new configuration: `./start.sh skill_container --dir ./mycroft/skills/weather --config ~/mycroft.ini`
  - Ask Mycroft about the weather **without** the city name


forslund:
@jdorleans asked me to review the changes and provide feedback as this is in response to my PR #72.

My big concern is that this doesn't do what I'd like it to do. I feel the resolution of config files should be:

mycroft defaults, container config, system-config, user-config.

where **mycroft defaults** contains the necessary configs that are needed with good defaults

**container config** contains the necessary configurations for the container

**system config** contains system specific changes (maybe supported baudrates, some other configurations the distribution may prefer).

**user config** is the single config file for any configurations the user would like to do to any skills, core skills and container skills alike.

This is the resolution I tried to suggest in my pull request. I understand if you have different opinion in the matter and I can live with any decision you make.

Secondly these changes requires the `--dir` flag to be provided. Ideally the command line with only defaults should be

`./start.sh skill_container path/to/my/skill`
or as it would be with the sdk
`mycroft-skill-container path/to/my/skill`

the `--dir`doesn't provide anything.

jdorleans:
@forslund thx for the feedback, that's really useful. I understand what you mean, although it does not apply for our configuration design. It really does not matter which values are declared in each config file. There is no such thing as a set of values that should be written in the config file X or Y. 

The big idea around our configuration hierarchy is the ability to override a configuration from `default`, to `system`, `user` and now, `skill` level. This is also ensuring access level, since `system` requires root access. Finally, we can differentiate configurations between users by declaring configuration at an user level.

Having that in mind, there is no real advantage on increasing the configuration management complexity by splitting the Configuration into several categories, because they do not have any practical usage. In the end, we only care about the order configuration files will be loaded.

I understand and agree with you we should remove the `--dir` argument. There is no need for it since we have to specify the skill directory anyway. Thanks for pointing that out. I'll include it on this PR too.

forslund:
@jdorleans, I understand you and feel I might have been unclear. I'm not really interested in the categorization, what I want is to provide good defaults in a container default config. I also want the user to be able to override these configs using the standard `~/.mycroft/mycroft.ini` file.

For example the media skills I've been playing around with requires a mopidy server. When I tried to propose these as built in skills to mycroft I added a section in the `mycroft/configuration/defaults/default.ini` config

```
mopidy_url=localhos:6680
```

and if an user has a mopidy server running elsewhere he or she can just add an entry in config in `~/.mycroft/mycroft.ini`

If I provide my own config with *good defaults* as it is right now the user will need to modify it to change the settings according to preference and anything the user enters in `~/mycroft/mycroft.ini` will be overridden.

My opinion is that the user should be able to override all settings from `~/mycroft/mycroft.ini`. If changing multiple configs is necessary when using container skills the usability decreases. This is only my opinions and I can work around it if necessary with a bunch of

```python
try:
    mopidy_url = self.config['mopidy_url']
except:
    mopidy_url = 'http://127.0.0.1:6680' # default value
```

jdorleans:
I think I'm starting understanding what you mean. Right now, we can only override the configuration by passing config files as reference and we do not support a more programmatic way of updating the configuration. 

How do you think the user should be able to change the configuration other than specifying a new config file or going to our portal and changing the configuration online? 

BTW: That seems to me another feature request. This PR is to deal with the need of Skill Containers only. Therefore, I'm approving this in order to progress on the new features requests.

PR #119, opened 2016-06-08
Feature/nest skill

ChristopherRios7:
READY FOR MERGE!


jdorleans:
Tests are failing, please take a look.

mycroft/skills/nest/dialog/en-us/not.paired.dialog
```diff
@@ -0,0 +1 @@
+Your device is not currently registered. To start pairing, login at cerberus.mycroft.ai
```
jdorleans: where are you using this dialog?

mycroft/skills/nest/dialog/en-us/too.cold.dialog
```diff
@@ -0,0 +1 @@
+I am raising the temperature for you
```
jdorleans: where are you using this dialog?

mycroft/skills/nest/dialog/en-us/too.hot.dialog
```diff
@@ -0,0 +1 @@
+I am lowering the temperature for you
```
jdorleans: where are you using this dialog?

mycroft/skills/nest/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+        self.register_intent(thermostat_intent, self.handle_temp_set_intent)
+
+    def handle_temp_set_intent(self, message):
+        try:
+            temp = message.metadata.get("TempNum", None)
+            LOGGER.debug(temp)
+            subprocess.call("nest --user '" + self.nest_user +
+                            "' --password '" + self.nest_password +
+                            "' temp " + temp, shell=True)
+            self.speak_dialog('temp.set', data={'TempNum': temp})
```
jdorleans: Please, use lower case standard. `temperature` or `temp` instead of `TempNum`

mycroft/skills/nest/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+        super(NestSkill, self).__init__(name="NestSkill")
+        self.nest_user = self.config.get('nest_user')
+        self.nest_password = self.config.get('nest_password')
+
+    def initialize(self):
+        self.load_data_files(dirname(__file__))
+        self.register_regex("(?P<TempNum>\d+)")
+
+        thermostat_intent = IntentBuilder("TempIntent")\
+            .require("tempKeyword").require("TempNum").build()
```
jdorleans: UpperCase for those file names: `TempKeyword` instead of `tempKeyword`

mycroft/configuration/defaults/defaults.ini
```diff
@@ -92,3 +92,7 @@ max_time = 600  # in seconds
 notify_delay = 5  # in seconds
 rate = 16000
 channels = 1
+
+[NestSkill]
+nest_user = "default-user"
+nest_password = "default_password"
```
jdorleans: Since those variables are inside NestSkill session there is no real need to use the 'nest' prefix. Then we can simply use `user` and `password` (even `pass`). Suggestion:
```
[NestSkill]
user = "default-user"
password = "default_password"
```

mycroft/skills/nest/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+__author__ = 'crios'
+
+LOGGER = getLogger(__name__)
+
+
+class NestSkill(MycroftSkill):
+    def __init__(self):
+        super(NestSkill, self).__init__(name="NestSkill")
+        self.nest_user = self.config.get('nest_user')
+        self.nest_password = self.config.get('nest_password')
```
jdorleans: Above suggestion:
```
self.user = self.config.get('user')
self.password = self.config.get('password')
```

mycroft/skills/nest/vocab/en-us/tempKeyword.voc
```diff
@@ -0,0 +1,4 @@
+set the thermostat to
+set the nest to
+update the nest to
+update the thermostat to
```
jdorleans: Please, refer to the VolumeSkill to change this vocabulary set. We don't need **the** and **to**.

mycroft/skills/nest/vocab/en-us/TempNum.voc
```diff
@@ -0,0 +1,4 @@
+62
+68
+72
+80
```
jdorleans: Why do we have those 4 values declared here?

mycroft/skills/nest/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+
+        thermostat_intent = IntentBuilder("TempIntent")\
+            .require("tempKeyword").require("TempNum").build()
+        self.register_intent(thermostat_intent, self.handle_temp_set_intent)
+
+    def handle_temp_set_intent(self, message):
+        try:
+            temp = message.metadata.get("TempNum", None)
+            LOGGER.debug(temp)
+            subprocess.call("nest --user '" + self.nest_user +
```
jdorleans: Is `nest` a binary we should previously install on the system? If so, how do we install it? Should this be using a python library instead?

requirements.txt
```diff
@@ -32,4 +32,5 @@ netifaces==0.10.4
 pyjokes==0.5.0
 psutil==4.1.0
 pep8==1.7.0
+python-nest==2.9.2
```
jdorleans: I didn't see this library been used on the code. Is that the binary `nest` from the subprocess call ?

mycroft/skills/nest/vocab/en-us/TempNum.voc
ChristopherRios7: That is to help make regex recognize similar values.

mycroft/skills/nest/vocab/en-us/tempKeyword.voc
ChristopherRios7: I believe that it helps ensure the intent is recognized.

mycroft/skills/nest/__init__.py
ChristopherRios7: I will discuss this with them to ensure this is a good decision, I do not want to confuse Cerberus.

mycroft/skills/nest/__init__.py
ChristopherRios7: Will do!

mycroft/skills/nest/__init__.py
ChristopherRios7: Temperature will not work because of the weather skill. There is an issue with intents bleed in with each other.

mycroft/skills/nest/dialog/en-us/too.hot.dialog
ChristopherRios7: I forgot to remove it. I will delete it.

mycroft/skills/nest/dialog/en-us/too.cold.dialog
ChristopherRios7: I will remove it!

jdorleans:
Chris, let's add some intent parser and unit test for this skill. I can explain you about the unit tests, ping me.

ethanaward:
Seems to work well, LGTM as long as all the needed tests/changes have been made.

seanfitzgeraldsc:
I think this crosses the threshold of "skills that should be core mycroft", because it's only useful if the user has a nest. I'd rather see this implemented as a 3P skill on the skills sdk, but we don't have a good story for how to install 3P skills yet, so that would suck.

Also, this skill shouldn't register intents/vocab if it can't find a nest device or doesn't have nest credentials.

PR #120, opened 2016-06-08
Issues 94 - Add intent unit tests for most skills

ethanaward:
I added unit intent tests for most of the skills that didn't have them, excluding some that seemed to be special cases. The special cases were skills like `stop` and `cerberus_config`. I also did not address the Wolfram tests, as @jdorleans wanted something additional for them.


ethanaward:
Ah, this is failing because of the desktop_launcher tests. Those will be difficult to test, as the `Application` vocab pulls from the current machine's list of installed applications. I might have to just remove those entirely.

jdorleans:
@ethanaward merge this

ethanaward:
I don't have the ability to merge it on my own, I don't think.

PR #124, opened 2016-06-09
Issues 63 - Add more cases where Wolfram responds

ethanaward:
This is a new pull request to clean up the history, as there are several commits in there that shouldn't have been. @jdorleans, this includes a basic first test for the Wolfram functionality.

### DIFF
--- a/mycroft/skills/wolfram_alpha/__init__.py
+++ b/mycroft/skills/wolfram_alpha/__init__.py
@@ -87,7 +87,7 @@ class CerberusWolframAlphaClient(object):
         response = requests.get(url, headers=headers)
         if response.status_code == 401:
             raise CerberusAccessDenied()
-        #logger.debug(response.content)
+        # logger.debug(response.content)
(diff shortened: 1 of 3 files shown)


ethanaward:
This now includes tests for the WolframAlpha skill, as @jdorleans and I discussed.

jdorleans:
Ethan, those unit tests are really great. Thx for that!
I have one last suggestion though. 
All your tests use this structure at the beginning:
```
result = self.format_result("InvalidTitle", "Test")
res = wolframalpha.Result(StringIO(result))
```
Why not create a new method to generate the result directly and make your tests shorter?

For example:

```
def create_result(key, value)
    result = self.format_result(key, value)
    return wolframalpha.Result(StringIO(result))
```
...
```
def test_invalid_pod(self):
    result = self.create_result("InvalidTitle", "Test")
    self.assertEquals(WolframAlphaSkill().get_result(result), None))
```

ethanaward:
I refactored my code as you suggested.

Issue #131, opened 2016-06-10
Mycroft listens on all nics by default 


the7erm:
By default mycroft will listen on `0.0.0.0` if `service_config` is `None` or an empty string.

```python
application.listen(service_config.get("port"), service_config.get("host"))
```
https://github.com/MycroftAI/mycroft-core/blob/master/mycroft/messagebus/service/main.py#L43

This is a security issue because any device on the Lan can access the service.  This also is a problem if the user puts mycroft in a DMZ or directly on the net.

the7erm:
Here's the main reason why I think it's a good idea to force 127.0.0.1 if it's an empty string:
http://www.securityweek.com/configuration-issue-exposes-30000-mongodb-instances-researcher

Mongodb didn't and because of it they jeopardized their reputation as well as millions of users' data. 

I figure if you're smart enough to change it to 0.0.0.0 then you know what you're doing, but the fact that an empty string defaults to this is bad from a security standpoint.

aatchison:
Getting closer: https://github.com/MycroftAI/mycroft-core/pull/132

aatchison:
This was merged.

PR #132, opened 2016-06-10
#131 added if for situations where the host is empty

the7erm:
Added code to force listening to 127.0.0.1 on the client & service if it's empty/None/false ... etc.

### DIFF
--- a/mycroft/messagebus/client/ws.py
+++ b/mycroft/messagebus/client/ws.py
@@ -36,14 +36,19 @@ config = ConfigurationManager.get()
 client_config = config.get("messagebus_client")
 
 
+class EmptyHostException (Exception):
+    def __init__(self):
+        Exception.__init__(self, "Empty host in mycroft.ini "
+                                 "[messagebus_service] section")
+
+
 class WebsocketClient(object):
     def __init__(self, host=client_config.get("host"),
                  port=client_config.get("port"),
                  path=client_config.get("route"),
                  ssl=str2bool(client_config.get("ssl"))):
         if not host:
-            # By default open localhost if host is empty, or None
-            host = "127.0.0.1"
+            raise EmptyHostException
 
         self.emitter = EventEmitter()
         self.scheme = "wss" if ssl else "ws"
--- a/mycroft/messagebus/client/ws.py
+++ b/mycroft/messagebus/client/ws.py
@@ -40,7 +40,7 @@ class WebsocketClient(object):
     def __init__(self, host=client_config.get("host"),
                  port=client_config.get("port"),
                  path=client_config.get("route"),
-                 ssl=str2bool(client_config.get("ssl"))):
+                 ssl=None):
 
         if not host:
             raise ValueError("Missing or empty host in mycroft.ini "
@@ -50,13 +50,17 @@ class WebsocketClient(object):
             raise ValueError("Missing or empty port in mycroft.ini "
                              "[messagebus_client] section")
 
-        if ssl is None or ssl == "":
-            raise ValueError("Missing or empty ssl in mycroft.ini "
-                             "[messagebus_client] section")
-
         if not path:
             raise ValueError("Missing or empty route in mycroft.ini "
                              "[messagebus_client] section")
+
+        if ssl is None:
+            ssl = client_config.get("ssl")
+            if ssl is None:
+                raise ValueError("Missing ssl in mycroft.ini "
+                                 "[messagebus_client] section")
+
+        ssl = str2bool(ssl)
 
         self.emitter = EventEmitter()
         self.scheme = "wss" if ssl else "ws"
--- a/mycroft/messagebus/client/ws.py
+++ b/mycroft/messagebus/client/ws.py
@@ -36,19 +36,27 @@ config = ConfigurationManager.get()
 client_config = config.get("messagebus_client")
 
 
-class EmptyHostException (Exception):
-    def __init__(self):
-        Exception.__init__(self, "Empty host in mycroft.ini "
-                                 "[messagebus_client] section")
-
-
 class WebsocketClient(object):
     def __init__(self, host=client_config.get("host"),
                  port=client_config.get("port"),
                  path=client_config.get("route"),
                  ssl=str2bool(client_config.get("ssl"))):
+
         if not host:
-            raise EmptyHostException
+            raise ValueError("Missing or empty host in mycroft.ini "
+                             "[messagebus_client] section")
+
+        if not port:
+            raise ValueError("Missing or empty port in mycroft.ini "
+                             "[messagebus_client] section")
+
+        if ssl is None or ssl == "":
+            raise ValueError("Missing or empty ssl in mycroft.ini "
+                             "[messagebus_client] section")
+
+        if not path:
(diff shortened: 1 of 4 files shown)


aatchison:
Has anyone tested this? This is an important PR I think.

clusterfudge:
you didn't want to pull out a default port as well? I'm in general not a huge fan of defaults, i'd rather that we failed fast here.

Also, the only way that this happens (I'm pretty sure) is that there's no defaults.ini config file within the system somewhere, meaning we're already in some janky world. In that case, basically all of the config values that are defaults for the constructor are non-existent, so replacing just one of them seems strange to me. Can we just fail-fast instead?

mycroft/messagebus/client/ws.py
```diff
@@ -41,6 +41,10 @@ def __init__(self, host=client_config.get("host"),
                  port=client_config.get("port"),
                  path=client_config.get("route"),
                  ssl=str2bool(client_config.get("ssl"))):
+        if not host:
+            # By default open localhost if host is empty, or None
+            host = "127.0.0.1"
```
jdorleans: Actually, if `host` or any other variable from the config does not exist, we'll face an exception. In order to prevent that we should have a default parameter during the `.get` operation, for instance: 
```
...
host=client_config.get("host", "localhost"),
port=client_config.get("port", 8080),
...
```

mycroft/messagebus/service/main.py
```diff
@@ -39,8 +39,12 @@ def main():
     ]
 
     application = web.Application(routes, **settings)
+    host = service_config.get("host")
+    if not host:
+        # By default listen to localhost if host is empty, or None
+        host = "127.0.0.1"
```
jdorleans: Same as above, in here.

jdorleans:
@the7erm thank you for this PR. Did you find any situation where we miss those configuration values? 

As Sean pointed out, we should really fail if there is no declared values for those params as they are essential to the application to run. Currently, we have a default configuration file which provides those values, but one could override them.

So, I'd say we should probably validate if those values exist during creation and, in case of failure, give a detailed error message to the user as a feedback instead of creating default values. 

What do you think?

mycroft/messagebus/client/ws.py
the7erm: I didn't use `.get()` That will can return an empty string if that's the value, and `None` if it's not defined at all.  So intentionally I did a test for falseness.

Example:
```
>>> d = { "test":"" }
>>> d.get("test", "default")
''
```

the7erm:
> clusterfudge: you didn't want to pull out a default port as well? I'm in general not a huge fan of defaults, i'd rather that we failed fast here.

No, TBH I didn't look into that.  I figured the default ports were exactly what everyone expects them to be, and I can see your point with the tyranny of the default.

> clusterfudge: Can we just fail-fast instead?

Sure, what Exception should raise?  Create a new one, or re-use one?  If I'm making a new one, is it ok if I put it in `mycroft/messagebus/service/main.py`, or is there a specific file that you want all the exceptions in?  I haven't really paid attention to Exception management ... I think too much.

```python
if not host:
   raise WhatExeceptionDoIuseImNew
```

>  jdorleans: Did you find any situation where we miss those configuration values? 

I just checked the ini file out ... I *think* this is the default .ini file.
https://github.com/MycroftAI/mycroft-core/blob/master/mycroft/configuration/mycroft.ini#L8
```
[messagebus_service]
host = ""
port = 8000
route = "/events/ws"
```

The service is empty by default which means tornado serves `0.0.0.0` - all interfaces, from what I've observed.  Most web servers like flask & cherrypy will broadcast on all interfaces by default.

> jdorleans: What do you think? 

I think in the end it's better to fail closed than fail open.  A fast fail is a good idea.  We can raise an exception like clusterfudge suggested.

the7erm:
@clusterfudge @jdorleans please review.

clusterfudge:
Sorry, this got lost in a sea of pr comments (so many prs!). 

I think ValueError is appropriate here, as opposed our new exception (which seems overly specific).

We're failing fast, but still only capturing if the host param is set, none of the others. I'd like to see us call out what is required config (host, port, path, ssl), and error if any of them are set to None, or explicitly define default behavior. Just calling out host as exceptional doesn't pass my sniff test.

mycroft/messagebus/client/ws.py
```diff
@@ -40,7 +40,28 @@ class WebsocketClient(object):
     def __init__(self, host=client_config.get("host"),
                  port=client_config.get("port"),
                  path=client_config.get("route"),
-                 ssl=str2bool(client_config.get("ssl"))):
+                 ssl=None):
```
jdorleans: Please, initialize ssl like the other ones:
`ssl = client_config.get("ssl")`

mycroft/messagebus/client/ws.py
```diff
@@ -40,7 +40,28 @@ class WebsocketClient(object):
...
+        if not port:
+            raise ValueError("Missing or empty port in mycroft.ini "
+                             "[messagebus_client] section")
+
+        if not path:
+            raise ValueError("Missing or empty route in mycroft.ini "
+                             "[messagebus_client] section")
+
+        if ssl is None:
+            ssl = client_config.get("ssl")
```
jdorleans: Update this after initializing ssl on above.

mycroft/messagebus/client/ws.py
```diff
@@ -40,7 +40,28 @@ class WebsocketClient(object):
     def __init__(self, host=client_config.get("host"),
                  port=client_config.get("port"),
                  path=client_config.get("route"),
-                 ssl=str2bool(client_config.get("ssl"))):
+                 ssl=None):
+
+        if not host:
+            raise ValueError("Missing or empty host in mycroft.ini "
+                             "[messagebus_client] section")
+
```
jdorleans: Please, move all those validations to a new method to make the init logic clear. In addition, as their messages are exactly the same you can reuse then in a common method which receives the property name as a parameter e.g.:
```
def validate_param(value, name)
  if not value:
    raise ValueError("Missing or empty %s in mycroft.ini [messagebus_client] section", name)
```

jdorleans: Then you can call:
```
validate_param(host, 'host')
validate_param(port, 'port')
validate_param(path, 'path')
validate_param(ssl, 'ssl')
```

mycroft/messagebus/service/main.py
```diff
@@ -34,13 +34,27 @@ def main():
...
     application = web.Application(routes, **settings)
+    host = service_config.get("host")
+    if not host:
+        raise ValueError("Missing or empty host in mycroft.ini "
+                         "[messagebus_service] section")
+
+    port = service_config.get("port")
+    if not port:
+        raise ValueError("Missing or empty port in mycroft.ini "
+                         "[messagebus_service] section")
```
jdorleans: Same idea as mentioned above... let's extract those validation to a new method.

aatchison:
Where are we on this?

the7erm:
> Where are we on this?
@aatchison Sorry I got sidetracked with a bunch of things today.  I've rebased this to the current master, and it's ready for another review.

PR #141, opened 2016-06-13
Issues 63 - Update formatting for Wolfram

ethanaward:
This should add some functionality for Wolfram. I was using incorrect formatting on my last PR.


ryanleesipes:
@ChristopherRios7 can you test this branch and make sure it is working properly?

ethanaward:
Specifically what this should do is give at least basic information for any person you ask Mycroft about.

jdorleans:
@ethanaward why pod.id is preferable than pod.title in those changes?

ethanaward:
@jdorleans According to the Wolfram API documentation, the pod IDs are less likely to change than the titles, and so are better to sort things by. My current implementation also only works by checking the pod ID, using pod.title was a mistake because I thought that the IDs and titles were the same.

PR #148, opened 2016-06-14
update pypi version

aatchison:
to 0.2.3


aatchison:
lgtm, works fine it seems, unit tests pass

PR #155, opened 2016-06-16
Fix test runner so it only runs once.

clusterfudge:


jdorleans:
Sean, thank you for fixing that. Do you mind to keep the test on the init file though?

clusterfudge:
a module __init__ file is not the proper place for an entry point.

aatchison:
are we gtg?

aatchison:
my 2 cents is yes

PR #166, opened 2016-06-17
Roku skill

ethanaward:
I've added basic functionality for controlling a Roku using Mycroft. Unfortunately, since the Roku API is essentially simulating button preses, there is no way to add a search feature without hardcoding each series of button presses to get to the search screen on each app. 

There are no intent tests for the launch intent because it requires accessing a Roku, which the travis tests won't be able to do. Also, there is only one piece of vocab for the pause intent, as the only other natural way of saying pause in English is stop, which triggers the stop skill.


aatchison:
We are going to wait on this one for a bit.

mycroft/skills/roku_control/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+
+logger = getLogger(__name__)
+
+
+class RokuSkill(MycroftSkill):
+    def __init__(self):
+        super(RokuSkill, self).__init__(name="RokuSkill")
+        self.ip = str(self.config['ip'])
+        self.roku = None
+        if self.ip:
```
jdorleans: It seems this skill cannot work without a IP. In that case, we should throw a ValueError exception and prevent this skill to run. So, let's add a `else` for that.

mycroft/skills/roku_control/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+
+        roku_pause_intent = IntentBuilder("RokuPauseIntent")\
+            .require("RokuKeyword")\
+            .require("RokuPauseKeyword")\
+            .build()
+        self.register_intent(roku_pause_intent,
+                             self.handle_roku_pause_intent)
+
+    def register_apps(self):
+        if self.roku:
```
jdorleans: Once we have failed during initialization, we can remove all those `if self.roku:` verification from the handles.

mycroft/skills/roku_control/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+            for app in self.roku.apps:
+                self.register_vocabulary(app.name, "Application")
+
+    def handle_roku_launch_intent(self, message):
+        if self.roku:
+            app = message.metadata.get('Application')
+            self.speak_dialog("launch.app", {"Application": app})
```
jdorleans: Please, use lower case key names for the metadata.

mycroft/skills/roku_control/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def handle_roku_pause_intent(self, message):
+        if self.roku:
+            self.speak_dialog("pause")
+            self.roku.play()
+        else:
+            self.speak_dialog("no.device")
+
+    def stop(self):
+        pass
```
jdorleans: We probably have to implement `stop` feature for at least the `play` intent. Ask yourself what Roku processes the user can stop.

mycroft/skills/roku_control/vocab/en-us/RokuKeyword.voc
```diff
@@ -0,0 +1,3 @@
+roku
+on roku
+on the roku
```
jdorleans: `roku` should be enough to cover the other variations

ryanleesipes:
Let's pop this out into a third party skill (we can do this in the Mycroft AI repo).

mycroft/skills/roku_control/__init__.py
ethanaward: The only problem here is that when I do this, the intent tests for the skill start failing because the intents are not loaded.

mycroft/skills/roku_control/__init__.py
ethanaward: So to be clear, would this apply when someone says `stop roku`, or when someone says `stop` when the roku skill is doing something? If it's the first, that should just be as simple as adding another `self.roku.play` call, as that will pause something that is currently playing.

jdorleans:
@ethanaward we do still have some test issues here

ethanaward:
@jdorleans Yeah, I realize that. The Roku IP config entry is empty, so the skill is throwing an exception and not loading. Because of this, the intent tests for the skills are failing.

mycroft/skills/roku_control/__init__.py
jdorleans: SOVELD

aatchison:
I'm thinking this should be migrated to the https://github.com/MycroftAI/mycroft-skills repo.

aatchison:
@ethanaward, may I remove this? is it in the skills repo?

aatchison:
@ethanaward I'm closing this in favor of using the mycroft-skills repo.

PR #167, opened 2016-06-17
Alan Pope voice runs by default, no idea what will happen when I pack…

aatchison:
…age it

### DIFF
--- a/mycroft-base-MANIFEST.in
+++ b/mycroft-base-MANIFEST.in
@@ -4,3 +4,4 @@ include mycroft/configuration/defaults/*.ini
 recursive-include mycroft/skills/*/dialog *
 recursive-include mycroft/skills/*/vocab *
 include mycroft/skills/alarm/alarm.mp3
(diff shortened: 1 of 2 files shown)


ryanleesipes:
This does work as intended. And the voice does sound clearer, but also seems quiet. Need to test on unit. If the voice can be understood at a reasonable volume, I will be alright with merging this.

aatchison:
I'm pretty sure that it's not going to get any louder, it's a problem inherent in Mimic, and perhaps the voice  as well (remember to popping)

ryanleesipes:
I also tested on my laptop speaker, which is not the best. I would like to see someone load this PR on a Holmes IV unit to hear what it sounds like there.

aatchison:
As soon as I can I will. You could always do a checkout on the unit. It's going to be considerably less loud than the other sounds mycroft makes though. ...So they will turn the volume all the way up.....

ethanaward:
LGTM, testing on a laptop.

aatchison:
lgt...ethan!

PR #170, opened 2016-06-17
Issues 160 - Fix IP formatting

ethanaward:
This doesn't list interfaces if they don't have an address, and replaces `.` with `dot`.


aatchison:
I'll take look.

aatchison:
lgtm

PR #188, opened 2016-06-20
Issues 187 - Make Wolfram default to pod 2

ethanaward:
First, this refactors the `__find_value` method to search for a string in the pod id, not find an exact match. This allows for more flexible searching.

It also changes the `get_result` method, which is how the skill finds what text from the result to speak out. It now will search for a pod with `Definition` in the name, which will handle several cases. If none of the pod ids have been matched, it will return the text of the second pod received. 

This means that Mycroft should speak some information for every query that actually gets a result back from WA. This is a very big change in functionality, and may require some testing, as it will also increase false positives or useless information that it gets from WA.

### DIFF
--- a/mycroft/skills/wolfram_alpha/__init__.py
+++ b/mycroft/skills/wolfram_alpha/__init__.py
@@ -165,7 +165,7 @@ class WolframAlphaSkill(MycroftSkill):
             return
 
         if result:
-            input_interpretation = self.__find_value(res.pods, 'Input')
+            input_interpretation = self.__find_pod_id(res.pods, 'Input')
             verb = "is"
             structured_syntax_regex = re.compile(".*(\||\[|\\\\|\]).*")
             if parsed_question:
--- a/mycroft/skills/wolfram_alpha/__init__.py
+++ b/mycroft/skills/wolfram_alpha/__init__.py
@@ -125,7 +125,8 @@ class WolframAlphaSkill(MycroftSkill):
                         if not result:
                             result = self.__find_value(res.pods, 'Definition')
                             if not result:
-                                result = self.__find_value(res.pods, 'DecimalApproximation')
+                                result = self.__find_value(
+                                    res.pods, 'DecimalApproximation')
                                 if not result:
                                     result = self.__find_num(
                                         res.pods, '200')
--- a/mycroft/skills/wolfram_alpha/__init__.py
+++ b/mycroft/skills/wolfram_alpha/__init__.py
@@ -115,17 +115,17 @@ class WolframAlphaSkill(MycroftSkill):
             return result
         except:
             try:
-                result = self.__find_value(res.pods, 'Value')
+                result = self.__find_pod_id(res.pods, 'Value')
                 if not result:
-                    result = self.__find_value(
+                    result = self.__find_pod_id(
                         res.pods, 'NotableFacts:PeopleData')
                     if not result:
-                        result = self.__find_value(
+                        result = self.__find_pod_id(
                             res.pods, 'BasicInformation:PeopleData')
                         if not result:
-                            result = self.__find_value(res.pods, 'Definition')
+                            result = self.__find_pod_id(res.pods, 'Definition')
                             if not result:
-                                result = self.__find_value(
+                                result = self.__find_pod_id(
                                     res.pods, 'DecimalApproximation')
                                 if result:
                                     result = result[:5]
@@ -187,7 +187,7 @@ class WolframAlphaSkill(MycroftSkill):
             self.speak("Sorry, I don't understand your request.")
 
     @staticmethod
-    def __find_value(pods, pod_id):
+    def __find_pod_id(pods, pod_id):
         for pod in pods:
             if pod_id in pod.id:
                 return pod.text
--- a/mycroft/skills/wolfram_alpha/__init__.py
+++ b/mycroft/skills/wolfram_alpha/__init__.py
@@ -127,7 +127,9 @@ class WolframAlphaSkill(MycroftSkill):
                             if not result:
                                 result = self.__find_value(
                                     res.pods, 'DecimalApproximation')
-                                if not result:
+                                if result:
+                                    result = result[:5]
+                                else:
                                     result = self.__find_num(
                                         res.pods, '200')
                 return result
--- a/mycroft/skills/wolfram_alpha/__init__.py
+++ b/mycroft/skills/wolfram_alpha/__init__.py
@@ -123,9 +123,12 @@ class WolframAlphaSkill(MycroftSkill):
                         result = self.__find_value(
                             res.pods, 'BasicInformation:PeopleData')
                         if not result:
-                            result = self.__find_value(
-                                res.pods, 'DecimalApproximation')
-                            result = result[:5]
+                            result = self.__find_value(res.pods, 'Definition')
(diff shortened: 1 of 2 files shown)


ethanaward:
@Wolfgange3311999 @isaacnward Please review.

mycroft/skills/wolfram_alpha/__init__.py
```diff
@@ -123,9 +123,13 @@ def get_result(self, res):
                         result = self.__find_value(
                             res.pods, 'BasicInformation:PeopleData')
                         if not result:
-                            result = self.__find_value(
-                                res.pods, 'DecimalApproximation')
-                            result = result[:5]
```
jdorleans: We should probably trucate the decimals after finding a DecimalApproximation value

mycroft/skills/wolfram_alpha/__init__.py
```diff
@@ -123,9 +123,13 @@ def get_result(self, res):
...
+                            result = self.__find_value(res.pods, 'Definition')
+                            if not result:
+                                result = self.__find_value(
+                                    res.pods, 'DecimalApproximation')
```
jdorleans: Comment about should be probably after this result: 
```
if result:
    result = result[:5]
```

mycroft/skills/wolfram_alpha/__init__.py
```diff
@@ -183,7 +187,7 @@ def handle_fallback(self, message):
     @staticmethod
     def __find_value(pods, pod_id):
         for pod in pods:
-            if pod.id == pod_id:
+            if pod_id in pod.id:
```
jdorleans: Why `in` ?

mycroft/skills/wolfram_alpha/__init__.py
```diff
@@ -123,9 +123,13 @@ def get_result(self, res):
...
-                            result = self.__find_value(
-                                res.pods, 'DecimalApproximation')
-                            result = result[:5]
+                            result = self.__find_value(res.pods, 'Definition')
+                            if not result:
+                                result = self.__find_value(
+                                    res.pods, 'DecimalApproximation')
+                                if not result:
+                                    result = self.__find_num(
+                                        res.pods, '200')
```
jdorleans: What is this exactly?

test/skills/wolfram_alpha/__init__.py
```diff
@@ -15,38 +15,51 @@
...
         res = self.create_result("BasicInformation:PeopleData",
-                                 "Born in 1997")
+                                 "Born in 1997", '300')
         self.assertEquals(WolframAlphaSkill().get_result(res), "Born in 1997")
 
     def test_decimal_approximation_pod(self):
-        res = self.create_result("DecimalApproximation", "5.6666666666")
-        self.assertEquals(WolframAlphaSkill().get_result(res), "5.666")
+        res = self.create_result("DecimalApproximation", "5.6666666666", '300')
+        self.assertEquals(WolframAlphaSkill().get_result(res), "5.6666666666")
```
jdorleans: We should truncate the decimal to 3 digits maximum.

mycroft/skills/wolfram_alpha/__init__.py
```diff
@@ -123,9 +123,12 @@ def get_result(self, res):
...
                         if not result:
-                            result = self.__find_value(
-                                res.pods, 'DecimalApproximation')
-                            result = result[:5]
+                            result = self.__find_value(res.pods, 'Definition')
+                            if not result:
+                                result = self.__find_value(res.pods, 'DecimalApproximation')
+                                if not result:
+                                    result = self.__find_num(
+                                        res.pods, '200')
```
Wolfgange3311999: Perhaps there should be a comment explaining why '200' was chosen.

mycroft/skills/wolfram_alpha/__init__.py
ethanaward: It allows for more flexible searching. For example, there are multiple pod IDs from the API that correspond to a definition: Defiinition:WordData, Definition:MathData, etc. Changing this to `in` allows me to find all of them with one search rather than having cases for each.

Wolfgange3311999:
Tested and it works great! Looks good after those changes.

test/skills/wolfram_alpha/__init__.py
ethanaward: I was having troubles with ordering it correctly. Thinking about it, I should be able to put this back in.

mycroft/skills/wolfram_alpha/__init__.py
ethanaward: This searches for the pod with the `position` attribute 200, which corresponds to the second pod in the XML result. This means that Mycroft will return a result much more frequently rather than saying `Sorry, I don't understand your request` as often.

mycroft/skills/wolfram_alpha/__init__.py
```diff
@@ -202,6 +206,13 @@ def process_wolfram_string(text):
         text = re.sub(r"!", r",factorial", text)
         return text
 
+    @staticmethod
+    def __find_num(pods, pod_num):
+        for pod in pods:
+            if pod.node.attrib['position'] == pod_num:
```
jdorleans: Does a `pod` object **always** has `node` and `attrib` so we can get `position` without breaking?

mycroft/skills/wolfram_alpha/__init__.py
```diff
@@ -183,7 +187,7 @@ def handle_fallback(self, message):
     @staticmethod
     def __find_value(pods, pod_id):
```
jdorleans: I think it makes more sense to rename this method to `find_pod_id`, since we're look for ids rather than general values.

mycroft/skills/wolfram_alpha/__init__.py
ethanaward: Yes, pods always have a `node`, and the node is a python ElementTree object, which always has an `attrib` element.

mycroft/skills/wolfram_alpha/__init__.py
ethanaward: Sure, that would be fine.

ryanleesipes:
Testing this.

ryanleesipes:
@ethanaward no response (or delayed), still waiting ~2 minutes on, on "Who is Hillary Clinton?"

ryanleesipes:
Actually, none of these seem to be working, even though it is showing the <queryresult> stuff.

ethanaward:
@ryanleesipes That's odd. Could it be a Cerberus latency problem? Have you tried asking the same things on master?

ryanleesipes:
Seems to be working just fine in master.

ethanaward:
Oh yeah, whoops. That's my bad, I missed a call to the method that I renamed, which broke it.

jdorleans:
Ok, since we don't have enough time to test many use cases manually we are considering to start adding some integration tests. I'd like to go forward with this idea before merging this PR. @ethanaward what do you say?

ethanaward:
@jdorleans Yeah, that would be a good thing to do.

jdorleans:
It turns out we cannot solve our problem right now, because of the lack of infrastructure to integration tests. Then, I'm assuming @ethanaward is a great developer and we'll test a bit more on real life.

PR #196, opened 2016-06-20
added pocketsphinx to requirements.txt, removed compilation from dev_…

aatchison:
…setup.sh

Would anyone mind trying to run dev_setup.sh on a clean system, to see if this is ok?

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -50,24 +50,6 @@ pip install -r requirements.txt
 CORES=$(nproc)
 echo Building with $CORES cores.
 
-# clone pocketsphinx-python at HEAD (fix to a constant version later)
-if [ ! -d ${TOP}/pocketsphinx-python ]; then
-  # build sphinxbase and pocketsphinx if we haven't already
-  git clone --recursive https://github.com/cmusphinx/pocketsphinx-python
-  cd ${TOP}/pocketsphinx-python/sphinxbase
-  ./autogen.sh
-  ./configure
-  make -j$CORES
-  cd ${TOP}/pocketsphinx-python/pocketsphinx
-  ./autogen.sh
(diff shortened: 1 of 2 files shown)


aatchison:
This works well for me, including after packaging.

ryanleesipes:
Can you try a clean install inside the Docker container to verify? I will also try a clean install from somewhere else. @Wolfgange3311999 @ethanaward please also test this.

aatchison:
Yes, please. I'm going to pull down a whole new checkout and nuke my virtualenvs

aatchison:
This is working fine for me, as far as I can tell. I really don't want to mess with the docker container, but I'm going to build some debs like this for arm. (amd64 is fine)

ryanleesipes:
I was fairly certain there was a reason we weren't doing this. @clusterfudge care to refresh my memory?

aatchison:
I believe it had something to do with the versions being outdated on pypi?

seanfitzgeraldsc:
the python-pocketsphinx bindings require the pocketsphinx and sphinxbase native binaries, which are not included in the pip package (i don't think they can be, as they're platform specific). Thusly, they cannot be installed into a virtualenv. I was just looking for the issue on the legacy repo, but apparently that's been kablammoed?

aatchison:
I'm not sure, that makes sense though.

ryanleesipes:
No kablammo. Just changed permissions on everything. I will hook you up with permissions on that repo again (unless it's because you aren't on @clusterfudge)

aatchison:
I wondered why the Jenkins stuff couldn't access anything:D

aatchison:
Obviously this PR isn't going to fly, have to try another approach.

PR #197, opened 2016-06-20
Issues 180 - Add stock skill

ethanaward:
I've added a stock skill that will give you the stock price of a company when you ask for it. For example, `Hey Mycroft, what is the stock price of alphabet` will give you `With ticker symbol GOOGL, the current price of Alphabet Inc is 706.37 dollars per share.

Since we currently don't have dialog functionality for having conversations with Mycroft, this just grabs the first company match for what you give it, or if it can't find a company matching that name, it'll throw an exception and tell you.

### DIFF
--- a/mycroft/skills/stock/__init__.py
+++ b/mycroft/skills/stock/__init__.py
@@ -1,0 +1,76 @@
+# Copyright 2016 Mycroft AI, Inc.
+#
+# This file is part of Mycroft Core.
+#
+# Mycroft Core is free software: you can redistribute it and/or modify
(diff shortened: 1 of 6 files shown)


ryanleesipes:
I will review this one.

ryanleesipes:
This works as it should. There is probably some fancier stuff we can do to make this skill even more awesome (such as adding an handler for questions like "What is Twitter's stock price")

ryanleesipes:
Merging, can iterate on later. @ethanaward please make a ticket to include my proposed improvement (will have to do some fancy footwork to omit the " 's ")

PR #199, opened 2016-06-20
Issues 198 - fix pairing skill

ethanaward:
This changes the syntax of the messagebus client, there appears to be a big difference between calling str2bool in the parameters of `__init__` and calling it in the actual function.


aatchison:
thanks
will review

aatchison:
lgtm

aatchison:
We might wait for the travis build

ryanleesipes:
@aatchison you tested and it is working for you?

ryanleesipes:
This is working properly, however, there is strange behavior when you are trying to pair a device that already has an identity file but is no longer paired.

aatchison:
The reset event should be targeted at this, I think.

PR #202, opened 2016-06-20
Decreased false positives (Fixes #200)

### LINKED ISSUE #200: Too many false positives when detecting wake word

Wolfgange3311999:
I changed the keyword to "hey mycroft" and increased threshold. The threshold increase is necessary to maintain recognition rate since the key phrase is now more syllables. On my machine I notice less false positives even though just "mycroft" still works.

### DIFF
--- a/mycroft/client/speech/local_recognizer.py
+++ b/mycroft/client/speech/local_recognizer.py
@@ -27,7 +27,7 @@ BASEDIR = os.path.dirname(os.path.abspath(__file__))
 
 
 class LocalRecognizer(object):
-    def __init__(self, sample_rate=16000, lang="en-us", key_phrase="mycroft"):
+    def __init__(self, sample_rate=16000, lang="en-us", key_phrase="hey-mycroft"):
         self.lang = lang
         self.key_phrase = key_phrase
         self.sample_rate = sample_rate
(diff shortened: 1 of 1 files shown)


Wolfgange3311999:
We should wait for the sample audio files to verify this change is an improvement (ETA ~1-2 day(s)).

jdorleans:
@Wolfgange3311999 I didn't have the change to look in deep on your changes for the listener before merge, but I'd like to suggest we create as many unit test as possible to support changes for the mic, listener and recognizers. It'd be great to start on this PR.

Wolfgange3311999:
I should note that this change actually increases the recognition rate (only by 2% but the significant part is that it doesn't decrease it). However, I still have to add some sample false positive test audio files to test the other end of the spectrum.

jdorleans:
We're trying to keep the habit of creating Unit Tests for our changes to help us during code review, to identify mistakes and to improve our code quality. Can you provide some unit tests for these changes?

Wolfgange3311999:
Since in this case these changes only deal with pocketsphinx itself I think the audio accuracy test itself has to be used as support for this change.

[In this PR](https://github.com/MycroftAI/mycroft-core/pull/249), with the following two audio files, it is revealed that the changes in this PR help drastically with false positives:

#### [Talking people noise.wav][1] ####
Before: **18** false positives
After: **5** false positives

#### [Let's Play Minecraft with Geoff, Jack, Michael, Gavin and Ray.wav][2] ####
Before: **21** false positives
After: **8** false positives

#### [Sample audio files][3] ####
Before: **24** false negatives out of **81** 
After: **26** false negatives out of **81**

[1]:https://drive.google.com/file/d/0B7VRusOJyGc1VXVSQnpRQnpZZEU/view?usp=sharing
[2]:https://drive.google.com/file/d/0B7VRusOJyGc1YkRQTFpwb2VKZUU/view?usp=sharing
[3]:https://github.com/Wolfgange3311999/mycroft-core/tree/sample-audio-files

PR #203, opened 2016-06-21
Use pip for `build_host_setup.sh.

flexiondotorg:
Running an up to date Ubuntu MATE 16.04 workstation, with other Python development tools installed, I was unable to complete `build_host_setup.sh` due to version conflict with system install `python-setuptools`.

This pull request replaces `easy_setup` with `pip` to complete the upgrade of `virtualenv`.

### DIFF
--- a/build_host_setup.sh
+++ b/build_host_setup.sh
@@ -4,6 +4,7 @@ sudo apt-get install -y \
     git \
     python \
     python-dev \
+    python-pip \
     python-setuptools \
     python-virtualenv \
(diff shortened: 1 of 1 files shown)


aatchison:
Thanks! I wonder why we used easy_install for pip in the first place?

aatchison:
I will test this out to make sure it doesn't break our existing build.

aatchison:
seems fine to me
@clusterfudge ?
@jdorleans ?

PR #204, opened 2016-06-21
Add #! to mycroft.sh

flexiondotorg:
The pull request simply adds a crunch bang to the `mycroft.sh` shell script to prevent the follow error:

    Failed to execute process './mycroft.sh'. Reason:
    exec: Exec format error
    The file './mycroft.sh' is marked as an executable but could not be run by the operating system.


aatchison:
Good call.

PR #206, opened 2016-06-21
Removed the 'I am searching for' message from Wolfram (Fixes #193)

### LINKED ISSUE #193: Mycroft should only repeat the query in wolfram when short to speak
When the estimated time to speak "I am searching for blah blah blah blah" will take longer than the estimated time to get the result from wolfram, only the result should be said rather than the redundant "I am searching for ...".

Wolfgange3311999:
This commit:
 - Removes the "I am searching for" message from wolfram
 - Adds more context when failing to understand the request (to ensure Mycroft transcribed the audio correctly)

This is open for discussion but often having Mycroft say, "I am searching for ..." increases the amount of time to get an answer substantially. If you look on the debug or unit you can tell that Wolfram does actually give a rather fast response. Most of the delay actually arises from the TTS engine processing long outputs.

However, my question is: is this longer period of silence but shorter total time to answer worth it or should we consider adding a spoken phrase but making it shorter than "I am searching for" (Such as "One moment" or "Hmmm").


ethanaward:
The main benefit of `I am searching for` in my mind is feedback, so that users/developers can know that what they said has been correctly parsed. It does make Mycroft seem a bit more robotic and less personal though, especially since the WA skill is the fallback, and so is likely the one that you hear the most from failed queries.

ethanaward:
Also, the response may do a good enough job of giving feedback to the user so that the `I am searching for` is not needed.

Wolfgange3311999:
@ethanaward I thought about that but Mycroft already rephrases the question he heard in the answer. For example, when asking `what's 2 times 20` mycroft responds `2x20 is 40`.

aatchison:
What's the status of this PR?

ryanleesipes:
@ethanaward please have David and Derick review the new behavior to get their feedback. If they think it is a superior user experience, let's merge.

ethanaward:
@Wolfgange3311999 I just tested this, and it's not working quite as intended. The new section for `Sorry, I didn't understand...` needs to also be at the end of the handle_fallback method, in the else statement. That's actually the main point where Mycroft currently says `Sorry, I don't understand your request`.

Wolfgange3311999:
@aatchison Derick and David approved. LGTY?

isaacnward:
LGTM, let's merge

PR #207, opened 2016-06-21
Added back session manager behavior from old listener

Wolfgange3311999:
- Removed random extra variable from when I was experimenting with multithreading
 - Added back some SessionManager lines to replicate old behavior more precisely
I'm not quiet sure where the session_id was ever used when emitted in the payload however just to be safe I kept it in there. This is to be sure there are no differences between the old and new listener that could be causing Cerberus problems.

### DIFF
--- a/mycroft/client/speech/listener.py
+++ b/mycroft/client/speech/listener.py
@@ -119,6 +119,12 @@ class AudioConsumer(threading.Thread):
             self.metrics.increment("mycroft.wakeup")
 
     def process_audio(self, audio):
+        SessionManager.touch()
+        payload = {
+            'utterance': self.mycroft_recognizer.key_phrase,
+            'session': SessionManager.get().session_id,
+        }
+        self.emitter.emit("recognizer_loop:wakeword", payload)
         try:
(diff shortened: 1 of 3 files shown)


aatchison:
Thanks! I'm sure it was implemented that way for a reason, @clusterfudge ?

Wolfgange3311999:
Just to add the details I know (and maybe refresh clusterfudge's memory), the SessionManager essentially holds both the session ID and the last time the session was used. If the last time the session was used is too far back it regenerates this session ID by randomly generating this. The SessionManager is occasionally touched to reset the timestamp and then outputted with the payload where the wake word is broadcasted. Then, in `main.cpp`, just like the other messages, the message with the session_id is transmitted to the `WebSocket`.

aatchison:
Checking for explosives

aatchison:
lgtm, any objections?

Wolfgange3311999:
Well, since there's nothing new at all with the stuff added, merging.

Wolfgange3311999:
opening as I mistakenly closed it a bit ago.

PR #210, opened 2016-06-21
Issues/122

isaacnward:
I updated the spelling skill so that Mycroft displays the word he's spelling on his mouth as he spells it aloud, instead of showing the talking animation. To allow this, I added an "activate_mouth_listeners" method to the enclosure API, which turns the mouth listener associations (such as the mouth automatically talking whenever the unit speaks) on or off, and which can be used for similar enclosure integration in the future.


aatchison:
testin'

aatchison:
lgtm

Issue #209, opened 2016-06-21
Keep user configuration by default


jdorleans:
During new configuration loads, we should always preserve the user configuration from `~/.mycroft/mycroft.ini`, unless there is an explicitly override declaration.

Note: this task might be related to issue #48

ethanaward:
Addressed by #211

PR #211, opened 2016-06-21
Issues 209 - Keep user configuration by default

jdorleans:
Issues #209 
- Preserving user configuration during new configuration load
- It's possible to override this behaviour by changing the `keep_user_config` flag to `False`

### DIFF
--- a/mycroft/configuration/__init__.py
+++ b/mycroft/configuration/__init__.py
@@ -46,9 +46,11 @@ class ConfigurationLoader(object):
         return config
 
     @staticmethod
-    def init_locations(locations=None):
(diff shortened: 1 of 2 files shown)


aatchison:
reviewing

aatchison:
Hmm, this still seems to use the configuration pulled down from cerberus, at least for my location.

jdorleans:
Yes, this should happen since the user configuration is kept only when dealing with local configuration. We wanna override it with the remote configuration, so the remote config can be commonly shared for all devices.

ethanaward:
LGTM

PR #221, opened 2016-06-22
Noticed that FAQ/Common Errors section of README.md said that mycroft.…

jasonehines:
…sh would start the cli, so I added it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,4 +120,4 @@ source ~/.virtualenvs/mycroft/bin/activate
 
 #### When running mycroft, I get the error `mycroft.messagebus.client.ws - ERROR - Exception("Uncaught 'error' event.",)`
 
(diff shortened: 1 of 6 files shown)


ethanaward:
My intention there was to say that you should run ./start.sh service, ./start.sh skills, and either ./start.sh voice __or__ ./start.sh cli. Running both at the same time will have unintended side effects. Making an option in mycroft.sh for running either `service, skills, and voice` or `service, skills and cli` would be useful though, for those who want to test using cli rather than voice.

jasonehines:
@ethanaward are their other unintended side effects besides Mycroft saying everything twice?

ethanaward:
@jasonehines That's the main one afaik.

jasonehines:
@ethanaward okay. I added the quite option and it seems to work as expected.

jasonehines:
I like having the cli --quiet and the voice running at the same time; unless it's going to cause some other issue. That way I have the option to type/paste something in the cli or just talk to Mycroft.

Please give me some feedback on which way to go with this.

ethanaward:
@jasonehines I think your method of doing it should be fine as long as cli is in quiet mode. The only other suggestion I would have would be to make another mode that doesn't run cli in case people don't want to run both.

jasonehines:
Unless I missed a bug or you want to change something, this should be ready.

ethanaward:
LGTM, merged.

PR #222, opened 2016-06-22
Issues 108 - Change regex to be localized

ethanaward:
This adds a new `load_regex_files` method in `MycroftSkill`, as well as supporting functions. It registers regex from files like `load_vocab_files` does for vocab. It is currently not in `load_data_files`, as the majority of skills do not use regex.

It also refactors all of the skills that use the old `register_regex` syntax to use the new localized syntax, with the exception of the `ScheduledCRUDSkill` base class that @jdorleans made.


ethanaward:
I also noticed while making these changes that regex with the same group name will be shared between skills. This means that the `location.rx` file in the weather skill is redundant, as the same thing has already been registered by the `location.rx` file in the time skill. @clusterfudge, I assume this is intended behavior. It can lead to unintended conflicts, however. 

For regex registered in skills, we might want to use more specific names for the entities. As an example of a possible conflict, one skill may define something like `(at|on|in) (?P<Location>.*)`. Someone developing another skill may want something like `in (?P<Location>\w+)`, and be confused when the Location entity is matching things that they do not want.

the7erm:
@ethanaward For some strange reason I thought this was merged, I was writing code that used it.  I'll clone this repo/branch and test it for you.

the7erm:
@ethanaward seems to be working for me.

the7erm:
Hopefully this PR gets approved/merged, I've already written a skill that uses it.  [mycroft-skill-jupiter-broadcasting](https://github.com/the7erm/mycroft-skill-jupiter-broadcasting/tree/master)

mycroft/skills/core.py
```diff
@@ -60,12 +60,27 @@ def load_vocab_from_file(path, vocab_type, emitter):
                                       'alias_of': entity}))
 
 
+def load_regex_from_file(path, regex_type, emitter):
```
clusterfudge: seems like we don't use regex_type

clusterfudge:
looks like we have no test coverage on anything in skills.core or skills.scheduled_skills. Could you add tests for your new code, and open an issue about coverage? I have no issues with the change, would just like some unit tests.

ethanaward:
@clusterfudge How would you recommend writing tests for my new code? I'm having trouble actually getting testable information.

mycroft/skills/core.py
jdorleans: Let's create a unit test for this guy.

mycroft/skills/core.py
```diff
@@ -60,12 +60,27 @@ def load_vocab_from_file(path, vocab_type, emitter):
...
+                        metadata={'regex': line.strip()}))
+
+
 def load_vocabulary(basedir, emitter):
     for vocab_type in os.listdir(basedir):
         load_vocab_from_file(
             join(basedir, vocab_type), splitext(vocab_type)[0], emitter)
 
 
+def load_regex(basedir, emitter):
```
jdorleans: Another UT for this guy too.

jdorleans:
Changes seem totally reasonable. Good job! Let's add some UT for the new methods and we're good to merge.

aatchison:
Agreed, merge pending UT

ethanaward:
@clusterfudge @jdorleans: Unit tests have been added.

test/skills/core.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def check_load_from_file(self, path, emitter, regex_list):
+        load_regex_from_file(path, emitter)
+        self.check_emitter(emitter, regex_list)
+
+    def check_load(self, path, emitter, regex_list):
+        load_regex(path, emitter)
+        self.check_emitter(emitter, regex_list)
+
+    def check_emitter(self, emitter, regex_list):
```
jdorleans: `emitter` is available from `self` you don't need to pass it all way down here to use it. Just use `self.emitter` inside this method and remove `emitter` from this method signature. As a result, will also remove all `emitter` from the other methods signatures and calls. In the end, your code will be much simpler.

test/skills/core.py
```diff
@@ -0,0 +1,99 @@
...
+from mycroft.util.log import getLogger
+
+__author__ = 'eward'
+logger = getLogger(__name__)
+
+
+class MockEmitter(object):
+    def __init__(self):
+        self.types = []
+        self.results = []
```
jdorleans: replace this initialization for reset()

test/skills/core.py
```diff
@@ -0,0 +1,99 @@
...
+        for type in emitter.get_types():
+            self.assertEquals(type, 'register_vocab')
+        if not regex_list:
+            self.assertEquals(emitter.get_results(), regex_list)
+        for value in regex_list:
+            self.assertTrue(value in emitter.get_results())
+        self.emitter.reset()
+
+    def test_load_regex_from_file_single(self):
+        self.check_load_from_file(join(dirname(__file__),
```
jdorleans: Create a new constant on your class level to represent your file path just like you did for emitter:
`path = join(dirname(__file__)`, then replace all `join(dirname(__file__)` to the variable.

test/skills/core.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def reset(self):
+        self.types = []
+        self.results = []
+
+
+class MycroftSkillTest(unittest.TestCase):
+    emitter = MockEmitter()
+
+    def check_load_from_file(self, path, emitter, regex_list):
```
jdorleans: Initialize `regex_list` with a default empty array to make your tests simpler, like: 
`def check_load_from_file(self, path, emitter, regex_list=[]):`

Then, you'll be able to remove all empty [] pass as parameters from the methods calls below.

test/skills/core.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def get_results(self):
+        return self.results
+
+    def reset(self):
+        self.types = []
+        self.results = []
+
+
+class MycroftSkillTest(unittest.TestCase):
```
jdorleans: There is only one missing test it'd be great to have too. That's for the case when **no valid regex** is provided in a .rx file which means the `re.compile()` would fail. Please, add this one.

ethanaward:
@jdorleans The latest commits should address your feedback

test/skills/core.py
```diff
@@ -0,0 +1,92 @@
...
+    emitter = MockEmitter()
+    path = join(dirname(__file__), 'regex_test')
+
+    def check_load_from_file(self, path, regex_list=[]):
+        load_regex_from_file(join(self.path, path), self.emitter)
```
jdorleans: I think what we wanna do in here is join the self.path with filename. Please, rename `path` to `filename`

test/skills/core.py
```diff
@@ -0,0 +1,92 @@
...
+        self.check_emitter(self.emitter, regex_list)
+
+    def check_load(self, path, regex_list=[]):
+        load_regex(path, self.emitter)
+        self.check_emitter(self.emitter, regex_list)
+
+    def check_emitter(self, emitter, regex_list):
+        for type in emitter.get_types():
```
jdorleans: Since `type` is a native built-in python object, it's a good idea to rename this var to something.

test/skills/core.py
```diff
@@ -0,0 +1,92 @@
...
+        self.check_load_from_file('multiple.rx',
+                                  ['(?P<MultipleTest1>.*)',
+                                   '(?P<MultipleTest2>.*)'])
+
+    def test_load_regex_from_file_none(self):
+        self.check_load_from_file('none.rx')
+
+    def test_load_regex_from_file_invalid(self):
+        try:
+            self.check_load_from_file(join(dirname(__file__), 'invalid.rx'))
```
jdorleans: Why are we join invalid.rx with this path, seen that it'll be joined again inside this check method?

ethanaward:
The new comments should also be addressed.

mycroft/skills/core.py
```diff
@@ -60,12 +60,28 @@ def load_vocab_from_file(path, vocab_type, emitter):
...
+    for regex_type in os.listdir(basedir):
+        if regex_type.endswith(".rx"):
```
jdorleans: @ethanaward this '.rx' check should go inside the `load_regex_from_file` method instead.

mycroft/skills/core.py
```diff
@@ -60,12 +60,27 @@ def load_vocab_from_file(path, vocab_type, emitter):
                                       'alias_of': entity}))
 
 
+def load_regex_from_file(path, emitter):
+    with open(path, 'r') as reg_file:
+        for line in reg_file.readlines():
```
jdorleans: Please, see comment below and place the IF '.rx' check before opening the file.

jdorleans:
Ok, before merging there is two more changes to do. 
Firstly, the IF check for '.rx' file extension should change its place. Secondly, the regex test files should be grouped inside an unique test directory `regex_test`, since the `skills` folder is supposed to have skill modules only. That way we improve coherence of the 'skills' folder usage.

Therefore, we can create a `success` and a `failure` sub-folder inside `regex_test`:
```
test/skills/regex_test/success
test/skills/regex_test/failure
```
Then, place the .rx test files which will succeed inside success and the ones which will fail inside failure. Finally, update your test code to use those.

mycroft/skills/core.py
ethanaward: @jdorleans I don't think so. With this check, I'm checking each file in the directory to see if it has a `.rx` extension. If it does, it gets sent to to the `load_regex_from_file` method. 

If the check was in that method, it would result in more `load_regex_from_file` calls with no real benefit.

PR #227, opened 2016-06-23
Feature/deb packaging

aatchison:
I want to merge this into master, so I don't have a side branch. I've moved the publish scripts to publish-core in order for us to keep the standalone stuff.


aatchison:
lgtm

PR #228, opened 2016-06-23
Speak skill

jasonehines:


ethanaward:
@jasonehines: Your intent test is failing. Looking at it, it's because you define the SpeakIntent as requiring a `Words` entity, and then never supply something that can be matched to a `Words` entity in the utterance in your intent test.

jasonehines:
@ethanaward I think it's correct now.

aatchison:
I'm checking it out now.

aatchison:
This is really, fun! lgtm.

PR #229, opened 2016-06-23
Added audio accuracy test

Wolfgange3311999:
This adds a new command, `./start.sh audioaccuracytest` that will run the listener through a set of audio files and verify that it can find the wake word.

To run it, place some audio files with the wake word in `audio-test/data/query-after/`. The folder is under query-after since audio files are also being collected with the query before even though they are not currently being tested since the listener isn't setup to work like that.

### DIFF
--- a/audio-accuracy-test/audio_accuracy_test.py
+++ b/audio-accuracy-test/audio_accuracy_test.py
@@ -14,7 +14,7 @@ __author__ = 'wolfgange3311999'
 logger = getLogger('audio_test_runner')
 
 
-class MockStream(object):
+class FileStream(object):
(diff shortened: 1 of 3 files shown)


aatchison:
Going to check this out, very good.

aatchison:
Wow, this is cool. I'm going to record some more wavs and see if my pirate voice works.

aatchison:
If she builds, lgtm

Wolfgange3311999:
By the way, [here's some sample audio files we put together](https://github.com/Wolfgange3311999/mycroft-core/tree/sample-audio-files) for anyone else who wants to test it.

PR #230, opened 2016-06-23
Added timeout override to listener - fixes #224

### LINKED ISSUE #224: Listener waits too long and too often for more sound
Often when saying short queries in which the listener hasn't detected enough total noise it sits there waiting for more when in fact you are actually done speaking. To resolve this we should decrease the minimum seconds of noise and put a limit to how long it will wait (perhaps 3-4 seconds).

Wolfgange3311999:
Previously Mycroft would wait a long time for noise if what the user said was too short (Or if you said nothing). It has been changed so that there is a maximum of 3 seconds before it gives up.

Another change is that previously the listener's behavior would be influenced by the buffer size. Now, however, it is not since the noise is increased by an amount relative to the buffer size.

### DIFF
--- a/mycroft/client/speech/mic.py
+++ b/mycroft/client/speech/mic.py
@@ -126,11 +126,15 @@ class ResponsiveRecognizer(speech_recognition.Recognizer):
 
     # The minimum seconds of noise before a
     # phrase can be considered complete
-    MIN_LOUD_SEC_PER_PHRASE = 0.2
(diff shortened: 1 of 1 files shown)


ethanaward:
You have one tiny pep8 error.
```
mycroft/client/speech/mic.py:193:80: E501 line too long (87 > 79 characters)
```

aatchison:
Taking a look.

aatchison:
Hmm, looks lgtm.

PR #231, opened 2016-06-23
Update scheduled skill formatting (Fixed #158)

### LINKED ISSUE #158: Alarm skill doesn't speak the date naturally
What I mean by this is if setting an alarm for 5 minutes, it reads the entire date (ie. "set an alarm for Thrusday, June 16th, at 04:05 pm") rather than simply saying the same thing in simpler terms (ie. "set an alarm for 5 minues"). And if the date is a day ahead I think it would be best that it says "Tomorrow at 4:05pm".

ethanaward:
This change makes it so that when a user creates something scheduled on that day, the dialog will be more natural - `Alarm set for 1 hours and 59 minutes from now` rather than `Alarm set for Thursday, June 23 at 7:37 PM`. 

One possible concern with this change is that it makes it harder to internationalize.

### DIFF
--- a/test/scheduled_skills.py
+++ b/test/scheduled_skills.py
@@ -1,0 +1,29 @@
+from datetime import datetime, timedelta
+import unittest
+
+from mycroft.skills.scheduled_skills import ScheduledSkill
+from mycroft.util.log import getLogger
+
+__author__ = 'eward'
+
+logger = getLogger(__name__)
+
+
+class ScheduledSkillTest(unittest.TestCase):
+    def test_formatted_time_today_hours(self):
+        date = datetime.now() + timedelta(hours=2)
+        self.assertEquals(ScheduledSkill(name='test').
+                          get_formatted_time(float(date.strftime('%s'))),
+                          "1 hours and 59 minutes from now")
+
+    def test_formatted_time_today_min(self):
+        date = datetime.now() + timedelta(minutes=2)
+        self.assertEquals(ScheduledSkill(name='test').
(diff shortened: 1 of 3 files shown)


test/scheduled_skills.py
```diff
@@ -0,0 +1,29 @@
...
+
+__author__ = 'eward'
+
+logger = getLogger(__name__)
+
+
+class ScheduledSkillTest(unittest.TestCase):
+    def test_formatted_time_today_hours(self):
+        date = datetime.now() + timedelta(hours=2)
+        self.assertEquals(ScheduledSkill(name='test').
```
jdorleans: Create only one ScheduledSkill obj inside __init__ for your test class and then reuse the object.
```
def __init__():
  self.skill = ScheduledSkill('ScheduledSkillTest')
...
self.assertEquals(self.skill.get_formatted_time(...))
```

test/scheduled_skills.py
```diff
@@ -0,0 +1,29 @@
...
+
+from mycroft.skills.scheduled_skills import ScheduledSkill
+from mycroft.util.log import getLogger
+
+__author__ = 'eward'
+
+logger = getLogger(__name__)
+
+
+class ScheduledSkillTest(unittest.TestCase):
```
jdorleans: This file should be inside the /test/skill/scheduled_skills.py instead of test/scheduled_skills.py

jdorleans:
My only concern about those changes is we are adding an EN dependency which we'll have to deal with some day. However, I'm gonna approve it since we're focusing on EN only atm.

PR #235, opened 2016-06-24
Added new compiled regex for "what's" in wolfram - Fixes #159

### LINKED ISSUE #159: "Whats" is not parsed out when searching WolframAlpha
`what is a dog` results in `I am searching for a dog` whereas `Whats a dog` results in `I am searching for whats a dog`.

Wolfgange3311999:
This is for verbs that don't have a space between the question word and verb.


ethanaward:
Seems to work well. We might want tests for the EnglishQuestionParser class, although I'm not sure whether it's going to be replaced later.

This should fix #159. The only thing I'm not sure about is whether we should cover both `whats` and `what's`, since I don't know whether `whats` is ever returned from the current STT engine.

Wolfgange3311999:
@ethanaward Ok, I've added some tests.

isaacnward:
What's the status of this PR?

Issue #192, opened 2016-06-20
Wolfram Alpha Skill fails when results have 'did you mean?' queries


Wolfgange3311999:
For instance, when asking for "What is a power supply?", Mycroft will respond 
"I don't understand your request". This is because Wolfram only provides optional similar alternatives such as "What is a power". I think that in this scenario Mycroft should automatically send another request to Wolfram with the top similar alternative query and notify the user of this. So for example, 
"Hey Mycroft, what's a power supply"
"I am searching for a power supply"
"Assuming you asked about a power, a power is the possession of controlling influence."

Even though this was wrong in this case, I can see it being better than a simple, "Sorry, I didn't understand your request."

ethanaward:
@Wolfgange3311999 In order to properly implement this to give feedback to the user, I feel like we would need a dialogue feature, so that the user can select an option from possible `Did you mean...`s. I can make a basic version of this that works when there is only one `did you mean`, but that wouldn't cover all cases.

Wolfgange3311999:
Even though it still (intentionally) fails on multilpe `didyoumean` responses, marking as closes since it has been dealt with for when there is only one.

PR #236, opened 2016-06-24
WA skill will now search again if there is one `didyoumean` in the results

ethanaward:
I changed the WA skill so that it will find alternative queries, represented by `didyoumean` elements in the results. If there is exactly one of these, it will automatically search a second time for the alternative query, and tell the user that it is doing so.

This partially addresses #192.

### DIFF
--- a/mycroft/skills/wolfram_alpha/__init__.py
+++ b/mycroft/skills/wolfram_alpha/__init__.py
@@ -116,22 +116,22 @@ class WolframAlphaSkill(MycroftSkill):
             return result
         except:
             try:
-                result = self.find_pod_id(res.pods, 'Value')
+                result = self.__find_pod_id(res.pods, 'Value')
                 if not result:
-                    result = self.find_pod_id(
+                    result = self.__find_pod_id(
                         res.pods, 'NotableFacts:PeopleData')
                     if not result:
-                        result = self.find_pod_id(
+                        result = self.__find_pod_id(
                             res.pods, 'BasicInformation:PeopleData')
                         if not result:
-                            result = self.find_pod_id(res.pods, 'Definition')
+                            result = self.__find_pod_id(res.pods, 'Definition')
                             if not result:
-                                result = self.find_pod_id(
+                                result = self.__find_pod_id(
                                     res.pods, 'DecimalApproximation')
                                 if result:
                                     result = result[:5]
                                 else:
-                                    result = self.find_num(
+                                    result = self.__find_num(
                                         res.pods, '200')
                 return result
             except:
@@ -157,17 +157,17 @@ class WolframAlphaSkill(MycroftSkill):
         try:
             res = self.client.query(query)
             result = self.get_result(res)
-            others = self.find_did_you_mean(res)
+            others = self._find_did_you_mean(res)
         except CerberusAccessDenied as e:
             self.speak_dialog('not.paired')
             return
         except Exception as e:
             logger.exception(e)
-            self.speak("Sorry, I don't understand your request.")
+            self.speak_dialog("not.understood")
             return
 
         if result:
-            input_interpretation = self.find_pod_id(res.pods, 'Input')
+            input_interpretation = self.__find_pod_id(res.pods, 'Input')
             verb = "is"
             structured_syntax_regex = re.compile(".*(\||\[|\\\\|\]).*")
             if parsed_question:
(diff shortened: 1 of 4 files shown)


test/skills/wolfram_alpha/__init__.py
```diff
@@ -9,71 +10,91 @@
...
 
     def test_factorial_process(self):
-        self.assertEquals(WolframAlphaSkill().process_wolfram_string
+        self.assertEquals(self.skill.process_wolfram_string
                           ("Test!"), "Test,factorial")
+
+    def test_find_did_you_mean_exists(self):
+        res = self.create_did_you_mean(['search for power', 'power'])
+        self.assertEquals(self.skill.find_did_you_mean(res),
+                          ['search for power', 'power'])
```
jdorleans: Please, create a var for this array and reuse it on both method calls.

mycroft/skills/wolfram_alpha/__init__.py
```diff
@@ -184,16 +186,40 @@ def handle_fallback(self, message):
...
-            self.speak("Sorry, I don't understand your request.")
+            if len(others) == 1:
+                self.speak("Sorry, I didn't understand " +
+                           utterance + ". Searching for " +
+                           others[0] + " instead")
+                self.handle_fallback(Message('intent_failure',
+                                             metadata={'utterance':
+                                                       others[0]}))
+            else:
+                self.speak("Sorry, I don't understand your request.")
```
jdorleans: @ethanaward let's replace those old EN hard coded sentences dependency to .dialog and then provide more than one feedback.

mycroft/skills/wolfram_alpha/__init__.py
```diff
@@ -184,16 +186,40 @@ def handle_fallback(self, message):
 
             self.speak(response)
         else:
-            self.speak("Sorry, I don't understand your request.")
+            if len(others) == 1:
+                self.speak("Sorry, I didn't understand " +
+                           utterance + ". Searching for " +
+                           others[0] + " instead")
```
jdorleans: In this case, Mycroft understood the utterance, but could not find the valid result. I think we should improve this feedback by changing `Sorry, I didn't understand` to something like: `Sorry, I didn't find an answer to utterance. I'm searching for others[0] instead`. In addition, these should be in a .dialog instead of hard-coded (see also comment below).

ethanaward:
@jdorleans: I pushed changes to address your feedback.

mycroft/skills/wolfram_alpha/__init__.py
```diff
@@ -184,7 +186,15 @@ def handle_fallback(self, message):
 
             self.speak(response)
         else:
-            self.speak("Sorry, I don't understand your request.")
+            if len(others) == 1:
```
jdorleans: Since `others` is an array, we probably should do a verification of > 0 instead of equals 1.
`if len(others) > 0:`

mycroft/skills/wolfram_alpha/__init__.py
```diff
@@ -194,6 +204,22 @@ def __find_pod_id(pods, pod_id):
...
+        return None
+
+    @staticmethod
+    def _find_did_you_mean(res):
+        value = []
+        root = res.tree.find('didyoumeans')
+        if root:
+            for result in root:
+                value.append(result.text)
+        return value
```
jdorleans: Assuming we **always** pick the first item from `others` array, it seems to me this method logic should change.
We just need to return the first found alternative value from `root` instead of adding all possible into an array. Then, the `others` array would become something else which represents that alternative value. 
```
    def _find_did_you_mean(res):
        root = res.tree.find('didyoumeans')
        if not root:
            return None
        return root[0].text
```

mycroft/skills/wolfram_alpha/__init__.py
jdorleans: Assuming changes on suggestion below, we'd have a different verification here. I'd become something like:
`if alternative:` (only one value, not an array anymore)

test/skills/wolfram_alpha/__init__.py
```diff
@@ -9,71 +10,92 @@
...
+
+    def test_find_did_you_mean_exists(self):
+        values = ['search for power', 'power']
+        res = self.create_did_you_mean(values)
+        self.assertEquals(self.skill._find_did_you_mean(res),
+                          values)
+
+    def test_find_did_you_mean_none(self):
+        res = self.create_did_you_mean([])
+        self.assertEquals(self.skill._find_did_you_mean(res), [])
```
jdorleans: After changes, this should verify `None` instead of `[]`

ethanaward:
I changed it so that it will get always get the first `didyoumean` result, even if there is more than one. I'm keeping in the functionality of returning multiple values though, so that they can later be used for dialog functionality.

Issue #238, opened 2016-06-26
Phillips Hue Skill


ChristopherRogers1991:
Add a skill to control Phillips Hue Lights. Specifically:
- Toggle the lights
- Activate scenes
- Dim
- Change color temp

This is currently underway [here](https://github.com/ChristopherRogers1991/mycroft-core/tree/feature/issues-238). Toggling and scenes are currently supported. Dimming and color temp still to come.

ryanleesipes:
Will try this on Monday.

ChristopherRogers1991:
Dimming and color temp management have been added.

Tests and pep8 are passing - should I open a PR? I'm not sure what the process is for new skills. I saw a few open issues for skills, but haven't seen any that were merged (and it looks like [this roku skill](https://github.com/MycroftAI/mycroft-core/pull/166) was going to be "pop[ped]... out into a third party skill," but I have know idea what that means)

If I should open a PR:
* Should I add requirements to the main requirements.txt, or leave a requirements.txt in my skill's directory?
* Should I just use the same copywrite blob from one of the other files?
* Anything else I should know about or do?

aatchison:
Definitely, that's the best point of discussion for integration into the core. It's encouraged! :)

ChristopherRogers1991:
Just realized I broke a lot of stuff while I was refactoring at 4am this morning (probably a bad idea). Hopefully you didn't try it while it was broken.

I want to add one more intent before opening a PR - should be tonight.

Can someone address the open questions, about whether to include my own requirements.txt file, or add my requirements to the core file, and whether to copy/paste the copywrite blurb from one of the other files?

ethanaward:
@ChristopherRogers1991 At the moment, put the requirements into the core requirements.txt file. That's where they should be for a core skill. We still have yet to work out a good 3P skill system for that, so just put it there for now. 

And yeah, copy the GNU Public License and put it in your file.

ChristopherRogers1991:
Thanks @ethanaward. PR is open. Feel free to rip it to shreds. All feed back is welcome.

aatchison:
Wahoo! @ryanleesipes , you own Phillups Hue lights lol

ChristopherRogers1991:
Looks like the tests failed. I just re-ran locally, and it looks like they pass if I have a local mycroft.ini file (~/.mycroft/mycroft.ini), but failed when I removed that.

I'm assuming that means I need to add the block to the core ini file?

[...]

PR #240, opened 2016-06-28
Issues 238 - Philips Hue Skill

ChristopherRogers1991:
#238 - Phillips Hue Skill


mycroft/configuration/mycroft.ini
```diff
@@ -93,3 +93,10 @@ max_time = 600  # in seconds
 notify_delay = 5  # in seconds
 rate = 16000
 channels = 1
+
+[PhillipsHueSkill]
+verbose = False
```
ethanaward: A verbose option doesn't really seem necessary for this. I would recommend just having one bit of dialog feedback.

mycroft/skills/phillips_hue/__init__.py
```diff
@@ -0,0 +1,283 @@
...
+        super(PhillipsHueSkill, self).__init__(name="PhillipsHueSkill")
+        self.brightness_step = self.config.get('brightness_step', '')
+        if self.brightness_step == '':
+            self.brightness_step = 50
+        else:
+            self.brightness_step = int(self.brightness_step)
+        self.color_temperature_step =\
+            self.config.get('color_temperature_step', '')
+        if self.color_temperature_step == '':
+            self.color_temperature_step = 1000
```
ethanaward: I think that since we have defaults in `mycroft.ini`, we don't need to have a fallback for when the config is empty.

mycroft/skills/phillips_hue/__init__.py
```diff
@@ -0,0 +1,283 @@
...
+from adapt.intent import IntentBuilder
+from mycroft.skills.core import MycroftSkill
+from mycroft.util.log import getLogger
+from os.path import dirname
+from phue import Bridge
+from phue import Group
+from phue import PhueRegistrationException
+from phue import PhueRequestTimeout
+import socket
+import time
```
ethanaward: Since we only use sleep from time, we can probably replace this with `from time import sleep`

mycroft/skills/phillips_hue/__init__.py
```diff
@@ -0,0 +1,283 @@
...
+        self.register_intent(increase_color_temperature_intent,
+                             self.handle_intent)
+
+        connect_lights_intent = \
+            IntentBuilder("ConnectLightsIntent") \
+            .require("ConnectLightsKeyword").build()
+        self.register_intent(connect_lights_intent,
+                             self.handle_intent)
+
+    def handle_intent(self, message):
```
ethanaward: How much do we need this separate `handle_intent` function? Could it be better to do exception handling somewhere else?

mycroft/skills/phillips_hue/__init__.py
```diff
@@ -0,0 +1,283 @@
...
+                    self.handle_increase_color_temperature_intent(message)
+                elif message.message_type == "ConnectLightsIntent":
+                    self.handle_connect_lights_intent(message)
+                else:
+                    raise Exception('No matching intent handler')
+            except Exception as e:
+                if isinstance(e, PhueRequestTimeout):
+                    self.speak_dialog('unable.to.perform.action')
+                elif 'No route to host' in e.args:
+                    if not self.config.get('ip', '') == '':
```
ethanaward: Can we just check self.ip here, since it's already been initialized to `self.config.get('ip')`?

mycroft/skills/phillips_hue/__init__.py
```diff
@@ -0,0 +1,283 @@
...
+                        self.speak_dialog('no.route')
+                        return
+                else:
+                    raise
+
+    def handle_turn_off_intent(self, message):
+        if self.verbose:
+            self.speak_dialog('turn.off')
```
ethanaward: As before, I don't think we need these verbose checks.

mycroft/skills/phillips_hue/__init__.py
```diff
@@ -0,0 +1,283 @@
...
+        self.all_lights.on = False
+
+    def handle_turn_on_intent(self, message):
+        if self.verbose:
+            self.speak_dialog('turn.on')
+        self.all_lights.on = True
+
+    def handle_activate_scene_intent(self, message):
+        keyword_len = len(message.metadata['ActivateSceneKeyword'])
+        scene_name = message.metadata['utterance'][keyword_len:].strip()
```
ethanaward: Instead of grabbing everything left in the utterance, register some regex instead. You can create a regex folder like vocab folders:
```
-regex
--en-us
---scene.rx
```

In `scene.rx`, you can define something like `(scene|lights to) (?P<Scene>.*)` to accomplish what you're doing here much more cleanly. Take a look at [the python regex guide](https://docs.python.org/2/howto/regex.html) for more advice.

mycroft/skills/phillips_hue/__init__.py
```diff
@@ -0,0 +1,283 @@
...
+        turn_off_intent = IntentBuilder("TurnOffIntent")\
+            .require("TurnOffKeyword").build()
+        self.register_intent(turn_off_intent, self.handle_intent)
+
+        turn_on_intent = IntentBuilder("TurnOnIntent")\
+            .require("TurnOnKeyword").build()
+        self.register_intent(turn_on_intent, self.handle_intent)
+
+        activate_scene_intent = IntentBuilder("ActivateSceneIntent")\
+            .require("ActivateSceneKeyword").build()
```
ethanaward: When changing to use regex for a scene, you should also add `.require('Scene')` here. Take a look at the [Stock skill](https://github.com/MycroftAI/mycroft-core/blob/master/mycroft/skills/stock/__init__.py) for an example of this.

mycroft/skills/phillips_hue/__init__.py
```diff
@@ -0,0 +1,283 @@
...
+                self.speak_dialog('failed.to.register')
+                return False
+            else:
+                self.speak_dialog('successfully.registered')
+        self.all_lights = Group(self.bridge, 0)
+        return True
+
+    def initialize(self):
+        self.load_data_files(dirname(__file__))
+
```
ethanaward: When changing to use regex, you'll also need to have a `load_regex_files` call here. Look at the [Stock skill](https://github.com/MycroftAI/mycroft-core/blob/master/mycroft/skills/stock/__init__.py) for examples.

mycroft/skills/phillips_hue/__init__.py
```diff
@@ -0,0 +1,283 @@
...
+    def handle_turn_on_intent(self, message):
+        if self.verbose:
+            self.speak_dialog('turn.on')
+        self.all_lights.on = True
+
+    def handle_activate_scene_intent(self, message):
+        keyword_len = len(message.metadata['ActivateSceneKeyword'])
+        scene_name = message.metadata['utterance'][keyword_len:].strip()
+        if scene_name == '':
+            self.speak_dialog('no.scene.name')
```
ethanaward: When we change to use regex, this won't be necessary, since the intent will require a scene to trigger.

mycroft/skills/phillips_hue/__init__.py
```diff
@@ -0,0 +1,283 @@
...
+        if not self.config.get('ip', '') == '':
+            self.speak_dialog('ip.in.config')
+            return
+        if self.verbose:
+            self.speak_dialog('connecting')
+        if self._connect_to_bridge():
+            self.speak_dialog('successfully.connected')
+        else:
+            self.speak_dialog('failed.to.connect')
+
```
ethanaward: Do we need these extra dialogs, since the `_connect_to_bridge` already gives more specific error dialogs?

mycroft/skills/phillips_hue/dialog/en-us/activate.scene.dialog
```diff
@@ -0,0 +1 @@
+Activating scene {{scene}}
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/bridge.not.found.dialog
```diff
@@ -0,0 +1 @@
+I could not find the bridge on your network. Please ensure that it is connected to the network, and try again. If this continues to happen, you may set the I P address of the bridge manually in the Mycroft I N I file.
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/connect.to.bridge.dialog
```diff
@@ -0,0 +1 @@
+Looks like this is the first time I've controlled your lights. Please press the button on the top of the hub, so I can connect.
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/connecting.dialog
```diff
@@ -0,0 +1 @@
+Connecting
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/decrease.brightness.dialog
```diff
@@ -0,0 +1 @@
+Dimming the lights
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/decrease.color.temperature.dialog
```diff
@@ -0,0 +1 @@
+Decreasing the color temperature of the lights
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/failed.to.connect.dialog
```diff
@@ -0,0 +1 @@
+Failed to connect
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/failed.to.register.dialog
```diff
@@ -0,0 +1 @@
+Failed to register with the bridge. Did you push the button on the hub?
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/increase.brightness.dialog
```diff
@@ -0,0 +1 @@
+Turning up the lights
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/increase.color.temperature.dialog
```diff
@@ -0,0 +1 @@
+Increasing the color temperature of the lights
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/no.route.dialog
```diff
@@ -0,0 +1 @@
+Sorry, I could not communicate with the lights. Please verify that the I P address in the mycroft I N I file is correct.
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/ip.in.config.dialog
```diff
@@ -0,0 +1 @@
+It looks like there is an I P address specified in your Mycroft I N I file. Please update the value there, or remove it if you would like me to discover the bridge on my own.
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/no.scene.name.dialog
```diff
@@ -0,0 +1 @@
+Sorry, I didn't hear a scene name.
```
ethanaward: This won't be necessary with the new regex changes.

mycroft/skills/phillips_hue/dialog/en-us/scene.not.found.dialog
```diff
@@ -0,0 +1 @@
+Sorry, I could not find a scene named {{scene}}
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/successfully.connected.dialog
```diff
@@ -0,0 +1 @@
+Successfully connected
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/successfully.registered.dialog
```diff
@@ -0,0 +1 @@
+Successfully registered with the bridge
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/turn.off.dialog
```diff
@@ -0,0 +1 @@
+Turning the lights off
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/turn.on.dialog
```diff
@@ -0,0 +1 @@
+Turning the lights on
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/dialog/en-us/unable.to.perform.action.dialog
```diff
@@ -0,0 +1 @@
+Sorry, I was unable to perform that action. The lights do not seem to be responding. If this happens again, try saying connect my lights.
```
ethanaward: Please, add more than one options for each of these dialogs.

mycroft/skills/phillips_hue/test/intent/sample1.intent.json
```diff
@@ -0,0 +1,7 @@
+{
+  "utterance": "turn off the lights",
+  "intent_type": "TurnOffIntent",
+  "intent": {
+    "TurnOffKeyword": "turn off the lights"
+  }
+}
```
ethanaward: Please, add two intent tests for each intent to provide more coverage for your vocab.

mycroft/skills/phillips_hue/test/intent/sample2.intent.json
```diff
@@ -0,0 +1,7 @@
+{
+  "utterance": "turn on the lights",
+  "intent_type": "TurnOnIntent",
+  "intent": {
+    "TurnOnKeyword": "turn on the lights"
+  }
+}
```
ethanaward: Please, add two intent tests for each intent to provide more coverage for your vocab.

mycroft/skills/phillips_hue/test/intent/sample3.intent.json
```diff
@@ -0,0 +1,7 @@
+{
+  "utterance": "turn down the lights",
+  "intent_type": "DecreaseBrightnessIntent",
+  "intent": {
+    "DecreaseBrightnessKeyword": "turn down the lights"
+  }
+}
```
ethanaward: Please, add two intent tests for each intent to provide more coverage for your vocab.

mycroft/skills/phillips_hue/test/intent/sample4.intent.json
```diff
@@ -0,0 +1,7 @@
+{
+  "utterance": "turn up the lights",
+  "intent_type": "IncreaseBrightnessIntent",
+  "intent": {
+    "IncreaseBrightnessKeyword": "turn up the lights"
+  }
+}
```
ethanaward: Please, add two intent tests for each intent to provide more coverage for your vocab.

mycroft/skills/phillips_hue/test/intent/sample5.intent.json
```diff
@@ -0,0 +1,7 @@
+{
+  "utterance": "decrease the color temperature of my lights",
+  "intent_type": "DecreaseColorTemperatureIntent",
+  "intent": {
+    "DecreaseColorTemperatureKeyword": "decrease the color temperature of my lights"
+  }
+}
```
ethanaward: Please, add two intent tests for each intent to provide more coverage for your vocab.

mycroft/skills/phillips_hue/test/intent/sample6.intent.json
```diff
@@ -0,0 +1,7 @@
+{
+  "utterance": "increase the color temperature of my lights",
+  "intent_type": "IncreaseColorTemperatureIntent",
+  "intent": {
+    "IncreaseColorTemperatureKeyword": "increase the color temperature of my lights"
+  }
+}
```
ethanaward: Please, add two intent tests for each intent to provide more coverage for your vocab.

mycroft/skills/phillips_hue/test/intent/sample7.intent.json
```diff
@@ -0,0 +1,7 @@
+{
+  "utterance": "activate scene relax",
+  "intent_type": "ActivateSceneIntent",
+  "intent": {
+    "ActivateSceneKeyword": "activate scene"
+  }
+}
```
ethanaward: Please, add two intent tests for each intent to provide more coverage for your vocab. Also, for this test, you'll need to add a new `Scene` category in the `intent` section once you make the regex changes.

mycroft/skills/phillips_hue/vocab/en-us/ActivateSceneKeyword.voc
```diff
@@ -0,0 +1,6 @@
+activate scene
+activate seen
+set my lights to
+set scene
+set seen
```
ethanaward: Since the STT engine aims to return one interpretation of each utterance, you shouldn't need to give these alternate spellings.

mycroft/skills/phillips_hue/vocab/en-us/ConnectLIghtsKeyword.voc
```diff
@@ -0,0 +1,8 @@
+connect my lights
+connect the light
+connect to my lights
+connect to the lights
+reconnect my lights
+reconnect the light
+reconnect to my lights
+reconnect to the lights
```
ethanaward: To clean up this vocab, you can register a `LightKeyword.voc` that has lights and light in it, and then change this to be a `ConnectKeyword.voc` that just has connect and reconnect in it. Take a look at the [Volume skill](https://github.com/MycroftAI/mycroft-core/tree/master/mycroft/skills/volume) for a good example of how to do this.

mycroft/skills/phillips_hue/vocab/en-us/DecreaseBrightnessKeyword.voc
```diff
@@ -0,0 +1,6 @@
+dim my lights
+dim the lights
+decrease the brightness of my lights
+decrease the brightness of the lights
+turn down my lights
+turn down the lights
```
ethanaward: Similar what I suggested before, use a `LightKeyword` and `DecreaseBrightnessKeyword` here to clean up the vocab. The [Volume skill](https://github.com/MycroftAI/mycroft-core/tree/master/mycroft/skills/volume) has good examples.

mycroft/skills/phillips_hue/vocab/en-us/DecreaseColorTemperatureKeyword.voc
```diff
@@ -0,0 +1,10 @@
+warm my lights
+warm the lights
+decrease the color temperature of my lights
+decrease the color temperature of the lights
+decreate the temperature of my lights
+decreate the temperature of the lights
+turn down the color temperature of my lights
+turn down the color temperature of the lights
+turn down the temperature of my lights
+turn down the temperature of the lights
```
ethanaward: Similar what I suggested before, use a `LightKeyword` and `DecreaseColorTemperatureKeyword` here to clean up the vocab. The [Volume skill](https://github.com/MycroftAI/mycroft-core/tree/master/mycroft/skills/volume) has good examples.

mycroft/skills/phillips_hue/vocab/en-us/IncreaseBrightnessKeyword.voc
```diff
@@ -0,0 +1,4 @@
+increase the brightness of my lights
+increase the brightness of the lights
+turn up my lights
+turn up the lights
```
ethanaward: Similar what I suggested before, use a `LightKeyword` and `IncreaseBrightnessKeyword` here to clean up the vocab. The [Volume skill](https://github.com/MycroftAI/mycroft-core/tree/master/mycroft/skills/volume) has good examples.

mycroft/skills/phillips_hue/vocab/en-us/IncreaseColorTemperatureKeyword.voc
```diff
@@ -0,0 +1,10 @@
+cool my lights
+cool the lights
+increase the color temperature of my lights
+increase the color temperature of the lights
+increase the temperature of my lights
+increase the temperature of the lights
+turn up the color temperature of my lights
+turn up the color temperature of the lights
+turn up the temperature of my lights
+turn up the temperature of the lights
```
ethanaward: Similar what I suggested before, use a `LightKeyword` and `IncreaseColorTemperatureKeyword` here to clean up the vocab. The [Volume skill](https://github.com/MycroftAI/mycroft-core/tree/master/mycroft/skills/volume) has good examples.

mycroft/skills/phillips_hue/vocab/en-us/TurnOffKeyword.voc
```diff
@@ -0,0 +1,8 @@
+shut my lights off
+shut off my lights
+shut off the lights
+shut the lights off
+turn my lights off
+turn off my lights
+turn off the lights
+turn the lights off
```
ethanaward: Similar what I suggested before, use a `LightKeyword` and `OffKeyword` here to clean up the vocab. The [Volume skill](https://github.com/MycroftAI/mycroft-core/tree/master/mycroft/skills/volume) has good examples.

mycroft/skills/phillips_hue/vocab/en-us/TurnOnKeyword.voc
```diff
@@ -0,0 +1,4 @@
+turn my lights on
+turn on my lights
+turn on the lights
+turn the lights on
```
ethanaward: Similar what I suggested before, use a `LightKeyword` and `OnKeyword` here to clean up the vocab. The [Volume skill](https://github.com/MycroftAI/mycroft-core/tree/master/mycroft/skills/volume) has good examples.

mycroft/skills/phillips_hue/__init__.py
```diff
@@ -0,0 +1,283 @@
...
+            .require("IncreaseColorTemperatureKeyword").build()
+        self.register_intent(increase_color_temperature_intent,
+                             self.handle_intent)
+
+        connect_lights_intent = \
+            IntentBuilder("ConnectLightsIntent") \
+            .require("ConnectLightsKeyword").build()
+        self.register_intent(connect_lights_intent,
+                             self.handle_intent)
+
```
ethanaward: When you address my suggested vocab changes below, you'll need to change these intents to require a `LightKeyword` in addition to each respective keyword.

mycroft/configuration/mycroft.ini
ChristopherRogers1991: I'd like to keep this in, if you're not strongly against it. I live in a small studio apartment, so for me, the lights changing is all the feedback I need. Having Mycroft tell me what I can see him doing is just noise - it's how the Amazon Echo works, and I hate it (trying to turn off the lights before bed results in a jarring 'OK!' any time the volume is left too high).

Someone with a larger home, however, may appreciate the feedback, if, for example, they are trying to turn off the living room lights from the bedroom.

mycroft/skills/phillips_hue/__init__.py
ChristopherRogers1991: I'm very open to changing this (not a fan of the giant if block), but I'm not sure how else to structure things so that I'm not repeating the check at the start (confirming that we've established the initial connection to the bridge), and the error handling at the end.

Do you have any suggestions?

mycroft/skills/phillips_hue/__init__.py
ChristopherRogers1991: self.ip can be set via the value in the config, or by trying to discover the bridge on the network. Here, we must check if it was supplied in the config, because we only want to speak the no.route dialog if the user supplied the ip. This is because with a user supplied ip, we will never try to change it, only alert them if there is an issue using it.

I should probably update the ip if we get here and it's not user supplied though. I think I was a little too focused on handling the case where the ip was user supplied, and not enough on resolving the issue otherwise.

mycroft/skills/phillips_hue/__init__.py
ChristopherRogers1991: Cool, I'll check that out.

I opened [this docs issue](https://github.com/MycroftAI/mycroft-core-documentation/issues/28) to get regex added to the skill creation page.

mycroft/skills/phillips_hue/__init__.py
ChristopherRogers1991: The success message should be spoken here, `_connect_to_bridge` only speaks the success message if it's the first time you're connecting, and need to register with the host. Otherwise it's silent, because it's just being run as part of another request, and there's no need to alert the user that we connected, if they can already see their lights changing.

The failure message here will happen if the ip address and username supplied in the config. The `_connect_to_bridge` method doesn't make any network calls if the ip address is user supplied. It might be better to make a test request against the hub though, and in that method, and handle all the errors/dialog there. I'll look into that.

mycroft/skills/phillips_hue/vocab/en-us/ActivateSceneKeyword.voc
ChristopherRogers1991: I added the alternate spelling after having the commands fail, and seeing that the cause was that 'seen' was used instead of 'scene.' I tested again just now to confirm. Here are the logs:
https://gist.github.com/ChristopherRogers1991/9d9f81958994dbd6212d2ad8d18183a0.

You'll notice that he correctly hears 'scene' the first few times, but when he hears 'seen,' it fails.

mycroft/skills/phillips_hue/__init__.py
ethanaward: Yeah, that does need to be added. Thanks for the issue.

mycroft/skills/phillips_hue/__init__.py
ethanaward: How likely is it that the intent handling functions will throw exceptions? Could it be better to try to connect to/discover the bridge in `__init__` or `initialize`, and then fail if it can't connect?

mycroft/skills/phillips_hue/vocab/en-us/ActivateSceneKeyword.voc
ethanaward: Oh, alright. Still, I think rather than building around flaws in the STT responses, we could try to control them more. Try changing the vocab to something like `set the scene` and `activate the scene` so that seen will be thought of as a less grammatical alternative.

mycroft/configuration/mycroft.ini
ethanaward: I'm not super strongly against it. I'll do some testing with other team members to see what we think.

mycroft/skills/phillips_hue/__init__.py
ChristopherRogers1991: I'm not so concerned with the odds, as with consequences. If we don't have error checking like this, and something goes wrong, the user will hear a generic error about something going wrong with the skill. This error is entirely unhelpful, as it doesn't provide any information on how to resolve the error (e.g. try again, fix your config, re-connect). We also can't do things such as attempt to re-connect automatically, if we get a 'no route to host' exception (as I mentioned I should be doing, in one of my comments yesterday).

Also, I don't think the connection stuff should happen in `__init__`. In general, I don't think it's a good idea to have long running operations (especially network operations) happen in `__init__`. Here, I think it would be particularly bad, because the bridge discovery takes up to 5 seconds, which means for anyone without a hue bridge, their boot time would increase by 5 seconds for nothing. Also, the initial connection requires interaction from the user to register the application with the bridge, which would be strange to handle during bootup.

I didn't see documentation for the `initialize` function, but it looks like it is called in `skills.core.load_skill`, shortly after `__init__`, meaning it would have the same issues.

I'm also not quite sure what you mean by 'fail.' Do you mean raise an exception, so the skill isn't loaded? Speak a dialog to the user indicating the failure? Something else?

mycroft/configuration/mycroft.ini
ChristopherRogers1991: Ok, cool. If I do lose via the democratic process here though, can I ask if there's a practical reason to remove it? The overhead for keeping it seems pretty negligible (it's one extra line per intent type that makes use of it, and an essentially non-existent perf hit from the extra ifs) so I'm curious what your concerns are.

Just as a note, this functionality is important enough to me as a user, I would remove this if I had to to get it into core, but fork and add it back for my own Mycroft. I really, really hate the confirmations from the Echo, and wish I could disable them there.

mycroft/skills/phillips_hue/vocab/en-us/ActivateSceneKeyword.voc
ChristopherRogers1991: If we the best we can do is 'control them more,' I'd still argue for keeping the alternate spellings. Developing and watching the logs, I can see and understand why Mycroft didn't understand my query - he got 'seen' instead of 'scene.' To an end user, however, when Mycroft keeps saying "I am searching for activate seen reading," they're going hear 'seen,' be unable to differentiate it from 'scene,' and assume he just took the wrong action, which is bad.

I'll admit that I don't know how the STT works, so correct me if I'm way off base here, but I feel like altering the phrasing would just be crossing my fingers that it's interpreted more accurately, and that doesn't seem good enough. I'll also admit it's a hack to have the alternate spellings, but it's a hack we know works 100% of the time.

mycroft/configuration/mycroft.ini
ethanaward: I'm fine with keeping it, I think the only thing would be whether it would default to false or true.

mycroft/skills/phillips_hue/vocab/en-us/ActivateSceneKeyword.voc
ethanaward: I agree, making sure the user knows what is going on is very important, One of the big reasons why I don't want to add the extra vocab is to prevent future unnecessary vocab conflicts. 

From what I've seen from the logs, the STT gets multiple interpretations of the given utterance, and then picks the one that it has the highest confidence level for, presumably based on grammatical correctness and some other characteristics. That's why I would try to say something like `activate the scene reading`, where `activate the seen reading` would have a lower confidence level. It would probably be better to test that though, as I don't actually know how significant that difference would be.

mycroft/skills/phillips_hue/__init__.py
ethanaward: For failing: some other skills, such as the Roku one that I developed, throw an exception and don't load if an ip is not in the config file. Thinking about it more though now, your reasoning for keeping exception handling in this method does make more sense. I wish there was a better way to do it than a big conditional block, but I can't see it for now.

mycroft/configuration/mycroft.ini
ChristopherRogers1991: Sweet. I'm fine with either as the default. Let me know if you, or anyone else, has a preference.

mycroft/skills/phillips_hue/__init__.py
ChristopherRogers1991: - [x] remove fallbacks

mycroft/skills/phillips_hue/__init__.py
ChristopherRogers1991: - [x] import only sleep

mycroft/skills/phillips_hue/__init__.py
ChristopherRogers1991: Ok, I refactored the connect/registration code a bit. All the dialog/error handling happens in _connect_to_bridge. I realized that I was wrong before, and there is a network call when it sets self.all_lights, so I was able to use that as the 'test' to confirm that the bridge is connected.

mycroft/skills/phillips_hue/__init__.py
ChristopherRogers1991: This doesn't seem to work properly if I put the regex in a file (structured exactly as you described above). If I register a hard coded string, e.g. `self.register_regex("lights to (?P<Scene>.*)")`, it works, but with the same string in a regex file, he fails to recognize the action, and falls back to searching wolfram alpha:
> "I am searching for set my lights to reading"

I didn't see any other .rx files anywhere (the other skills that use regex, e.g. stock are all hard coded).

Do you have any idea what's going on?

ChristopherRogers1991: Scrolling down to address other feedback, it looks like I need to call load_regex_files. I'll give that a shot, and see if that resolves things.

Any particular reason the regex files are not loaded with the other data files?

ChristopherRogers1991: Ok, that worked. And I found your comment on https://github.com/MycroftAI/mycroft-core/pull/222, about most skills not using the regex. It seems a little strange to me to have `load_data_files` not actually load all the data files though. I'm going to open another issue to discuss that, though, as this does not seem like the place to do it.

ChristopherRogers1991: Opened #256.

ethanaward: For not finding .rx files in other skills, are you sure that you're up to date? That sounds like you don't have the localized regex changes on your local repository. And yeah, I can see loading regex files after checking to make sure they exist.

ChristopherRogers1991: I was not. I realized when I couldn't find the function. Sorry, not used to the github workflow of forking and syncing with and upstream project.

ethanaward: No problem, it's just something that you have to get used to. I would recommend fairly frequently rebasing off of master so that you have the latest changes and can see if there are any conflicts.

mycroft/skills/phillips_hue/__init__.py
clusterfudge: Gonna chime in here since I'm just seeing this. 

I agree this giant conditional block is gross. The intent skill/adapt should be routing to the correct handle_* function. Error handling is typically verbose/boiler plate code, but it's better than this. You can use decorators or a wrapper when registering the handler to do error checking, but not like this.

*not like this*

requirements.txt
```diff
@@ -32,3 +32,4 @@ netifaces==0.10.4
 pyjokes==0.5.0
 psutil==4.1.0
 pep8==1.7.0
+-e git+https://github.com/ChristopherRogers1991/phue#egg=phue
```
clusterfudge: why are we depending on your fork of phue? And if we're going to depend on something via git, it has to be at a tag or sha (pulling from HEAD means that every build gets a potentially different version, which is bad).

mycroft/skills/phillips_hue/__init__.py
ChristopherRogers1991: Great call on use of decorators! I've never defined my own before, so let me know if I've done anything in a less than optimal way, but the code looks much better now, than it did with the giant conditional.

requirements.txt
ChristopherRogers1991: The upstream phue library does not have scene support. I added the functionality to make this skill. I've opened [this PR](https://github.com/studioimaginaire/phue/pull/64) to get the functionality into the upstream library, but until it's merged, we'll have to use mine.

Definitely a good call on using a specific commit. I've updated the requirements.

ChristopherRogers1991:
The unit tests are going to fail with this latest commit. I didn't want to take the time to update them yet, because I'm having some trouble getting Mycroft to take the correct actions.

@ethanaward, in the latest commit, I tried restructuring how I handle all of my keywords, as you'd mentioned in a few of your comments (using an IncreaseKeyword and a LightsKeyword to turn the brightness up, for example).

Since doing this, all of the actions still work for certain phrasings, however, some phrases that I would expect would work do not. For example, if I say "turn up the lights,' the brightness increases as expected, however, if I say "increase the brightness of the lights," he fails to find a matching intent, and searches Wolfram Alpha. Similarly, if I say "turn up the color temperature of the lights," everything works, however, if I say "increase the color temperature of the lights," he gives me the weather report. "Turn down" and "decrease" for both of these intent types works out the same way, with "turn down" working correctly, and "decrease" giving me wolfram alpha or weather.

With the way I have things structured, I would expect to have decent matches, so I'm confused why things are working.

Do you have any ideas? Should I just go back to the longer, more well defined keywords?

ethanaward:
My first guess is some sort of vocab conflict. The weather intent requires only `temperature` as a keyword, so if decrease isn't working for some reason, it would make sense that it would trigger. I'm not entirely sure why that it, though. @clusterfudge , any ideas?

And could you post some logs of what the intent parser is interpreting?

clusterfudge:
It's going to be a bit before I can go deep on this change. My recommendation would be to test the skill in isolation (via the skills container) and see if it works as you intend. It is possible to walk through the intent tests in a debugger, but it's not well documented and I haven't done it in a while.

ChristopherRogers1991:
The links on [this docs page](https://docs.mycroft.ai/skill.creation/skills-sdk) for setting up the skills sdk, and installing dependencies both currently just say they are works in progress. Rather than screw around trying to set them up on my own, I just made a dummy branch, and stripped out all the other skills. Even without the other skills, it does not seem to be creating the intents properly. Logs for the run can be found [here](https://gist.github.com/ChristopherRogers1991/f6bd1cdf2a9c3a8397dbc4f1ce5edc25)

Let me know if there is anything else you need, or anything else I should try.

ChristopherRogers1991:
I've been stepping through with the debugger, and, unless I am misunderstanding the purpose/use of the `IntentBuilder.one_of` function, it looks like the issue is with either the `intent.resolve_one_of` method, or the `intent.find_first_tag` function. 

`find_first_tag` will return (None, None) if the tag `start_token` is not strictly greater than `after_index`. In `resolve_one_of`, `last_end_index` is being set to 0. Because 'decrease' is the first word in the phrase, it's `start_token` value is 0. This causes the intent to be invalid.

If I say any word before 'decrease', it works:

    2016-07-04 23:07:19,773 - Skills - DEBUG - {"message_type": "intent_failure", "context": null, "metadata": {"utterance": "decrease the brightness of the lights"}}
    2016-07-04 23:07:27,527 - Skills - DEBUG - {"message_type": "recognizer_loop:record_begin", "context": null, "metadata": {}} 
    2016-07-04 23:07:30,362 - Skills - DEBUG - {"message_type": "recognizer_loop:record_end", "context": null, "metadata": {}} 
    2016-07-04 23:07:30,364 - Skills - DEBUG - {"message_type": "recognizer_loop:wakeword", "context": null, "metadata": {"session": "4f7a1d71-1d96-41b3-9e6c-f31319573a89", "utterance": "hey mycroft"}}
    2016-07-04 23:07:31,189 - Skills - DEBUG - {"message_type": "recognizer_loop:utterance", "context": null, "metadata": {"session": "4f7a1d71-1d96-41b3-9e6c-f31319573a89", "utterances": ["decrease the brightness of the lights"]}}
    2016-07-04 23:16:27,134 - intent__init__ - ERROR - 
      File "/home/chris/Code/mycroft-core/mycroft/skills/intent/__init__.py", line 47, in handle_utterance
        utterance, num_results=100))
    StopIteration
    2016-07-04 23:16:27,139 - Skills - DEBUG - {"message_type": "intent_failure", "context": null, "metadata": {"utterance": "decrease the brightness of the lights"}}
    2016-07-04 23:16:44,959 - Skills - DEBUG - {"message_type": "recognizer_loop:record_begin", "context": null, "metadata": {}} 
    2016-07-04 23:16:48,697 - Skills - DEBUG - {"message_type": "recognizer_loop:record_end", "context": null, "metadata": {}} 
    2016-07-04 23:16:48,702 - Skills - DEBUG - {"message_type": "recognizer_loop:wakeword", "context": null, "metadata": {"session": "4f7a1d71-1d96-41b3-9e6c-f31319573a89", "utterance": "hey mycroft"}}
    2016-07-04 23:16:50,268 - Skills - DEBUG - {"message_type": "recognizer_loop:utterance", "context": null, "metadata": {"session": "4f7a1d71-1d96-41b3-9e6c-f31319573a89", "utterances": ["anything decrease the brightness of the lights"]}}
    2016-07-04 23:16:50,298 - Skills - DEBUG - {"message_type": "AdjustBrightnessIntent", "context": {"target": null}, "metadata": {"LightsKeyword": "lights", "confidence": 0.34782608695652173, "target": null, "intent_type": "AdjustBrightnessIntent", "BrightnessKeyword": "brightness", "DecreaseKeyword": "decrease", "utterance": "anything decrease the brightness of the lights"}}
    2016-07-04 23:16:51,042 - phue - INFO - Attempting to connect to the bridge...
    2016-07-04 23:16:51,043 - phue - INFO - Using ip: 192.168.1.12:80
    2016-07-04 23:16:51,043 - phue - INFO - Using username from config: PkDulb65AWvewASxg-7NMUC6MmDRBN9E0CVymAA5
    2016-07-04 23:16:51,045 - phue - DEBUG - GET /api/PkDulb65AWvewASxg-7NMUC6MmDRBN9E0CVymAA5/groups/0 None
    2016-07-04 23:16:51,109 - phue - DEBUG - {"name":"Lightset 0","lights":["1","2","3","4","5"],"type":"LightGroup","state":{"all_on":true,"any_on":true},"recycle":false,"action": {"on":true,"bri":212,"hue":15331,"sat":121,"effect":"none","xy":[0.4448,0.4066],"ct":343,"alert":"none","colormode":"xy"}}
    2016-07-04 23:16:51,109 - phue - DEBUG - {'bri': 162}
    2016-07-04 23:16:51,111 - phue - DEBUG - PUT /api/PkDulb65AWvewASxg-7NMUC6MmDRBN9E0CVymAA5/groups/0/action {"bri": 162}
    2016-07-04 23:16:51,215 - phue - DEBUG - [{"success":{"/groups/0/action/bri":162}}]
    2016-07-04 23:16:51,215 - phue - DEBUG - [[{u'success': {u'/groups/0/action/bri': 162}}]]

I don't have enough context on these functions to know if the > in `find_first_tag` should be >=, or if the 0 in `resolve_one_of` should be a -1, but it seems that either of these would resolve the issue.

I'll look into it a bit more, but I would imagine you'll both have more context than I will.

ChristopherRogers1991:
Was pulling up https://github.com/MycroftAI/adapt/blob/master/adapt/intent.py to be more explicit about what I was referencing, and it looks like the issue has already been resolved. Line 47 now has `last_end_index` set to -1. I'll refresh my virtualenv, and see confirm that things are working.

ChristopherRogers1991:
Re-ran dev_setup.sh, and it looks like things are working as expected now.

I'll continue addressing the rest of the feedback.

mycroft/skills/phillips_hue/__init__.py
ChristopherRogers1991: To make things a little more explicit, I made the following:

    @property
    def user_supplied_ip(self):
        return self.config.get('ip') != ''

Now instead of doing things like `if self.config.get('ip') == ''`, I can just do `if self.user_defined_ip`

ethanaward:
Yeah, i was actually wondering if the issue was the one_of function. I had remembered that issue being resolved though, and I didn't think that was it. Is it possible that we either don't have the latest version of Adapt as a requirement, or that last fix is not in the latest release?

mycroft/skills/phillips_hue/__init__.py
ChristopherRogers1991: Scene names are now obtained via regex: 09419a77852ceede6f6c72b2c9bcff59562fb844

ChristopherRogers1991:
@ethanaward, the release being used has the fix. Requirements currently has it set at 0.2.7. I was on 0.2.1.

ChristopherRogers1991:
Ok, at this point, the tests should pass, and I think I've addressed all the feedback, other than the alternate spellings of scene.

For that, @ethanaward, you had suggested changing the phrases to 'activate the scene' and 'set the scene.' While this does seem to work better, these phrases feel awkward, and unnatural to me. Are the alternate spellings definitely a no-go? Assuming so, I am considering alternatives keywords and phrases. If you have other suggestions, I would be happy to hear them.

ethanaward:
I think I'd prefer not having the alternate spellings, yeah. It seems too much like relying on currently broken functionality that may change. It also potentially limits the vocab that other skills can use. If you have other ideas for phrasing that sound better, I think that would work the best.

ethanaward:
Oops, didn't mean to close that.

ChristopherRogers1991:
Hey, I've been looking into adding support for groups (a friend pointed out that it would be an important feature for him), and I'd like to get your guys take on something.

I started looking at regexes ([see the existing support with scenes](https://github.com/MycroftAI/mycroft-core/pull/240/commits/c4f99a70f9ce49703014f354aefd1b3071f2f09e)) to pull out the group name, but I'm realizing that this is likely to get out of hand. I'd like to allow for flexible phrasing, e.g.
* "Turn off the kitchen lights"
* "Shut off the lights in the kitchen"
* "Turn the kitchen lights off"
* "Turn off the lamp"
* "Turn the lamp off"

and I could probably make regular expressions to handle most of the phrasing for most of the operations (dimming, color temp changes, scenes, toggle), but it seems like I'd need _a lot_ of carefully crafted regexes.

A simpler, fairly naive solution would be to check if a group name exists in the utterance, and if so, apply the operation to the group specified, else apply it to the default group. Does this seem reasonable? Or would this be terrible?

grahamreeds:
I opened a issue of 'Generic house automation skill' #253 not seeing this
one while looking through the open features.

Sent from my Nexus 5.
On 7 Jul 2016 6:10 a.m., "ChristopherRogers1991" <notifications@github.com>
wrote:

Hey, I've been looking into adding support for groups (a friend pointed out
that it would be an important feature for him), and I'd like to get your
guys take on something.

I started looking at regexes (see the existing support with scenes
<https://github.com/MycroftAI/mycroft-core/pull/240/commits/c4f99a70f9ce49703014f354aefd1b3071f2f09e>)
to pull out the group name, but I'm realizing that this is likely to get
out of hand. I'd like to allow for flexible phrasing, e.g.

   - "Turn off the kitchen lights"
   - "Shut off the lights in the kitchen"
   - "Turn the kitchen lights off"
   - "Turn off the lamp"
   - "Turn the lamp off"

and I could probably make regular expressions to handle most of the
phrasing for most of the operations (dimming, color temp changes, scenes,
toggle), but it seems like I'd need *a lot* of carefully crafted regexes.

A simpler, fairly naive solution would be to check if a group name exists
in the utterance, and if so, apply the operation to the group specified,
else apply it to the default group. Does this seem reasonable? Or would
this be terrible?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/MycroftAI/mycroft-core/pull/240#issuecomment-230980524>,
or mute the thread
<https://github.com/notifications/unsubscribe/AEy9nIOVEDCYKCE-ChoxbvqyOjrQXS8pks5qTIozgaJpZM4I_q4t>
.

ethanaward:
I'm not sure how groups work with Hue, but is it possible to pull a list of registered groups from the Hue hub and then register all of those as vocab? Then you could just add a `.optionally('Group')` to the intents and matching cases for when there is a group in the utterance.

ChristopherRogers1991:
Hmm... That is a very clever idea. I like it. I worry that things would get stale though - if a user creates a new group, I'd want them to be able to use it right away. We could treat it like a cache, and update periodically, as well as provide an intent to update it manually, but that seems a little less than ideal. I suppose it may be the best option though.

If we do go this route, is there a way to 'unregister' names, in the case that a user deletes a group?

ethanaward:
Looking at the Adapt docs, I don't think there's a way to unregister names, no. You can probably just add error handling in case the group doesn't exist anymore, though. Then whenever a user restarts their Mycroft, the removed groups wont be in the vocab anymore.

ChristopherRogers1991:
To get proper group handling, I ultimately had to do the network calls in the `initialize` function that I was hoping to avoid. It could perhaps be removed again, if I added some persistent config, however, I'm not sure if that's something that is supported/encouraged, or if it's even the best option. For now though, the network calls are necessary to allow groups/scenes to be used immediately after boot (since they need to be registered as vocab, after a successful connection to the bridge is established).

There are a couple negative side-effects to the changes:
* Intent tests no longer cover scenes or groups.
* The unit test script takes much longer, as it attempts to find the bridge for every intent test.

Is there any equivalent of a unit test setUp function for intent tests, that could establish some notional scene/group names, and/or alter config settings? Something like that could mitigate these issues.

Other than that, I think things are pretty much ready to go, unless anyone has any other feedback. I know of one, fairly unlikely edge case - our username on the bridge is deactivated - that is not handled, but it will require updates to the Phue library, as it swallows the error. I'll work on updating the Phue library to allow a means of handling it, however, given that it's unlikely to happen (user would have to consciously do it via the web app Phillips provides, or reset the hub to factory settings), and there's a simple work around (remove the ~/.python_hue file), I don't think it necessarily needs to block the merge.

Let me know what you think.

ethanaward:
For persistent config, would that be as simple as storing an IP in the config, or would it involve something like storing all of the group and scene names? We do have a system set up currently to give each skill a [sandboxed filesystem](https://github.com/MycroftAI/mycroft-core/blob/master/mycroft/filesystem/__init__.py), so you might be able to do something with that. It would especially be nice to be able to cut down on the intent test times.

With regards to a `setUp` function for the intent tests, I don't think something like that exists right now. It would be useful though, I agree.

ChristopherRogers1991:
Hey, thanks for the tip about the FileSystemAccess. It looks like I get that for free with my skill. I'll look at getting that set up, and saving out some of the values. The best thing would probably be to save the ip, id, and username for the hub. If the file exists, I can assume a hub exists, and try to get up-to-date group/scene info. If the file doesn't exist, either there is no hub, or we haven't registered yet. For all the tests, the file should not exist, and we won't waste any time trying to find a bridge, but in the real world, users get current info right from boot (and with knowledge of the hub, the retrieval of the scene/group info is very quick, so it shouldn't affect their boot time much either).

ChristopherRogers1991:
Ok, updated this to use a file to store just the username, as that turned out to be all that was needed. I also stripped out all the config related code from my fork of the phue library, so that it won't clash/cause issues. A significant side benefit to all of this is that the edge case I mentioned before (user reset the hub, or removes the username) is no longer such an issue - reconnecting the lights (or restarting) will recognize that the username is invalid, and generate a new one. The unit tests should also be much faster, as long as the username file does not exist.

Unless there's other feedback for me, I think this should be merge-able.

aatchison:
I would love to merge this now but I think that we need to get the third party skills system worked out first. Thanks @ChristopherRogers1991

ChristopherRogers1991:
Hey @aatchison, thanks for the update. Is there any public discussion about the skills system anywhere? I saw the update [here](https://mycroft.ai/managing-skills-delivery-update/), and that there is a new [skills repo](https://github.com/MycroftAI/mycroft-skills), but haven't seen a description or discussion of the plan.

For me, I'd be proud to have my skill be a part of the core product, but I actually think a store model would make more sense. Developers could publish to a store you host, skills/developers could be rated, it opens a straightforward monetization strategy (similar to Google/Amazon/Apple app stores, allow developers to set a price, and take a percentage cut for the Mycroft team - looks like some other folks had [the same basic idea](https://community.mycroft.ai/t/premium-services/586/4) during the monetization discussion), and perhaps best of all, you're not caught holding the bag if a developer wants to drop support (you seem more responsible for maintaining the skill if it lives in your repo). Thought I'd give my two cents, just in case it's at all valuable. Feel free to ignore if you've got other plans.

ChristopherRogers1991:
Does anyone (@clusterfudge in particular) have any idea why [this intent test](https://github.com/MycroftAI/mycroft-core/pull/240/files#diff-d2376f7b8219f1cae7e91e0eebd26e5cR2) would pass, but using the phrase would fail in practice:

```
2016-09-17 13:47:48,349 - Skills - DEBUG - {"message_type": "recognizer_loop:wakeword", "context": null, "metadata": {"session": "802db0e1-66a7-4be1-a85a-2a94051d2378", "utterance": "hey mycroft"}}
2016-09-17 13:47:49,443 - Skills - DEBUG - {"message_type": "recognizer_loop:utterance", "context": null, "metadata": {"session": "802db0e1-66a7-4be1-a85a-2a94051d2378", "utterances": ["lights to 85%"]}}
2016-09-17 13:47:49,456 - intent__init__ - ERROR - 
  File "/home/chris/Code/mycroft-core/mycroft/skills/intent/__init__.py", line 47, in handle_utterance
    utterance, num_results=100))
StopIteration
2016-09-17 13:47:49,461 - Skills - DEBUG - {"message_type": "intent_failure", "context": null, "metadata": {"utterance": "lights to 85%"}}
2016-09-17 13:47:49,463 - wolfram_alpha__init__ - DEBUG - Could not determine intent, falling back to WolframAlpha Skill!
2016-09-17 13:47:49,467 - Skills - DEBUG - {"message_type": "enclosure.mouth.think", "context": null, "metadata": {}}
2016-09-17 13:47:49,478 - requests.packages.urllib3.connectionpool - INFO - Starting new HTTPS connection (1): cerberus.mycroft.ai
2016-09-17 13:47:52,511 - requests.packages.urllib3.connectionpool - DEBUG - "GET /wolframalpha/v2/query?input=lights+to+85%25 HTTP/1.1" 200 475
2016-09-17 13:47:52,518 - Skills - DEBUG - {"message_type": "speak", "context": null, "metadata": {"utterance": "Sorry, I couldn't find an answer for lights to 85%. Perhaps you meant lights instead?"}}
```

The regex is `(?P<Value>100|[1-9][0-9]?)(\D|$)`, which seems to parse out what I want just fine:

```
>>> x = re.compile(u'(?P<Value>100|[1-9][0-9]?)(\D|$)')
>>> r = x.search('set my lights to 50')
>>> r.group('Value')
'50'
>>> r = x.search('set my lights to 50%')
>>> r.group('Value')
'50'
>>> r = x.search('set the brightness to 50 on my lamp')
>>> r.group('Value')
'50'
>>> r = x.search('set the brightness to 80% on my lamp')
>>> r.group('Value')
'80'
```

But everything with a '%' after the value fails to recognize the intent:

```
2016-09-17 13:56:08,437 - Skills - DEBUG - {"message_type": "recognizer_loop:record_begin", "context": null, "metadata": {}}
2016-09-17 13:56:12,335 - Skills - DEBUG - {"message_type": "recognizer_loop:record_end", "context": null, "metadata": {}}
2016-09-17 13:56:12,341 - Skills - DEBUG - {"message_type": "recognizer_loop:wakeword", "context": null, "metadata": {"session": "802db0e1-66a7-4be1-a85a-2a94051d2378", "utterance": "hey mycroft"}}
2016-09-17 13:56:14,263 - Skills - DEBUG - {"message_type": "recognizer_loop:utterance", "context": null, "metadata": {"session": "802db0e1-66a7-4be1-a85a-2a94051d2378", "utterances": ["set the brightness to 100 for my lamp"]}}
2016-09-17 13:56:14,280 - Skills - DEBUG - {"message_type": "SetBrightnessIntent", "context": {"target": null}, "metadata": {"confidence": 0.3344594594594595, "Group": "lamp", "target": null, "intent_type": "SetBrightnessIntent", "Value": "100", "BrightnessKeyword": "brightness", "utterance": "set the brightness to 100 for my lamp"}}
2016-09-17 13:56:14,281 - phue - DEBUG - {'bri': 254}
2016-09-17 13:56:14,281 - phue - DEBUG - PUT /api/F9uWIDcslLgVb54fG47wxyR7VjT6ZfDVq6LtVeCk/groups/2/action {"bri": 254}
2016-09-17 13:56:14,299 - phue - DEBUG - [{"success":{"/groups/2/action/bri":254}}]
2016-09-17 13:56:14,299 - phue - DEBUG - [[{u'success': {u'/groups/2/action/bri': 254}}]]
2016-09-17 13:56:14,299 - phue - DEBUG - {'on': True}
2016-09-17 13:56:14,300 - phue - DEBUG - PUT /api/F9uWIDcslLgVb54fG47wxyR7VjT6ZfDVq6LtVeCk/groups/2/action {"on": true}
2016-09-17 13:56:14,327 - phue - DEBUG - [{"success":{"/groups/2/action/on":true}}]
2016-09-17 13:56:14,327 - phue - DEBUG - [[{u'success': {u'/groups/2/action/on': True}}]]
2016-09-17 13:56:20,352 - Skills - DEBUG - {"message_type": "recognizer_loop:record_begin", "context": null, "metadata": {}}
2016-09-17 13:56:23,139 - Skills - DEBUG - {"message_type": "recognizer_loop:record_end", "context": null, "metadata": {}}
2016-09-17 13:56:23,147 - Skills - DEBUG - {"message_type": "recognizer_loop:wakeword", "context": null, "metadata": {"session": "802db0e1-66a7-4be1-a85a-2a94051d2378", "utterance": "hey mycroft"}}
2016-09-17 13:56:24,645 - Skills - DEBUG - {"message_type": "recognizer_loop:utterance", "context": null, "metadata": {"session": "802db0e1-66a7-4be1-a85a-2a94051d2378", "utterances": ["set my lights to 10"]}}
2016-09-17 13:56:24,651 - Skills - DEBUG - {"message_type": "SetBrightnessIntent", "context": {"target": null}, "metadata": {"LightsKeyword": "lights", "confidence": 0.3684210526315789, "target": null, "intent_type": "SetBrightnessIntent", "Value": "10", "utterance": "set my lights to 10"}}
2016-09-17 13:56:24,651 - phue - DEBUG - {'bri': 25}
2016-09-17 13:56:24,652 - phue - DEBUG - PUT /api/F9uWIDcslLgVb54fG47wxyR7VjT6ZfDVq6LtVeCk/groups/0/action {"bri": 25}
2016-09-17 13:56:24,672 - phue - DEBUG - [{"success":{"/groups/0/action/bri":25}}]
2016-09-17 13:56:24,673 - phue - DEBUG - [[{u'success': {u'/groups/0/action/bri': 25}}]]
2016-09-17 13:56:24,673 - phue - DEBUG - {'on': True}
2016-09-17 13:56:24,675 - phue - DEBUG - PUT /api/F9uWIDcslLgVb54fG47wxyR7VjT6ZfDVq6LtVeCk/groups/0/action {"on": true}
2016-09-17 13:56:24,702 - phue - DEBUG - [{"success":{"/groups/0/action/on":true}}]
2016-09-17 13:56:24,703 - phue - DEBUG - [[{u'success': {u'/groups/0/action/on': True}}]]
2016-09-17 13:56:28,158 - Skills - DEBUG - {"message_type": "recognizer_loop:record_begin", "context": null, "metadata": {}}
2016-09-17 13:56:32,949 - Skills - DEBUG - {"message_type": "recognizer_loop:record_end", "context": null, "metadata": {}}
2016-09-17 13:56:32,955 - Skills - DEBUG - {"message_type": "recognizer_loop:wakeword", "context": null, "metadata": {"session": "802db0e1-66a7-4be1-a85a-2a94051d2378", "utterance": "hey mycroft"}}
2016-09-17 13:56:34,334 - Skills - DEBUG - {"message_type": "recognizer_loop:utterance", "context": null, "metadata": {"session": "802db0e1-66a7-4be1-a85a-2a94051d2378", "utterances": ["set the brightness of my light - 50%"]}}
2016-09-17 13:56:34,343 - intent__init__ - ERROR - 
  File "/home/chris/Code/mycroft-core/mycroft/skills/intent/__init__.py", line 47, in handle_utterance
    utterance, num_results=100))
StopIteration
2016-09-17 13:56:34,344 - Skills - DEBUG - {"message_type": "intent_failure", "context": null, "metadata": {"utterance": "set the brightness of my light - 50%"}}
2016-09-17 13:56:34,344 - wolfram_alpha__init__ - DEBUG - Could not determine intent, falling back to WolframAlpha Skill!
2016-09-17 13:56:34,346 - Skills - DEBUG - {"message_type": "enclosure.mouth.think", "context": null, "metadata": {}}
2016-09-17 13:56:34,348 - requests.packages.urllib3.connectionpool - INFO - Starting new HTTPS connection (1): cerberus.mycroft.ai
2016-09-17 13:56:40,615 - requests.packages.urllib3.connectionpool - DEBUG - "GET /wolframalpha/v2/query?input=set+the+brightness+of+my+light+-+50%25 HTTP/1.1" 200 534
2016-09-17 13:56:40,630 - Skills - DEBUG - {"message_type": "speak", "context": null, "metadata": {"utterance": "Sorry, I couldn't find an answer for set the brightness of my light - 50%. Perhaps you meant light - 50% instead?"}}
2016-09-17 13:56:40,634 - Skills - DEBUG - {"message_type": "recognizer_loop:audio_output_start", "context": null, "metadata": {}}
2016-09-17 13:56:40,635 - Skills - DEBUG - {"message_type": "enclosure.eyes.blink", "context": null, "metadata": {"side": "b"}}
2016-09-17 13:56:40,957 - Skills - DEBUG - {"message_type": "enclosure.mouth.viseme", "context": null, "metadata": {"code": "4:0.062,3:0.184,0:0.285,2:0.408,0:0.514,4:0.642,0:0.748,3:0.880,2:0.968,3:1.013,3:1.107,3:1.140,5:1.249,0:1.398,3:1.463,3:1.494,0:1.541,3:1.627,0:1.769,3:1.830,3:1.953,2:2.001,5:2.140,6:2.260,2:2.275,3:2.421,0:2.516,3:2.571,3:2.647,0:2.758,4:2.850,2:2.940,0:3.046,3:3.118,3:3.163,0:3.227,3:3.356,0:3.435,5:3.481,4:3.545,0:3.660,3:3.747,0:3.901,3:3.990,5:4.097,0:4.195,5:4.272,3:4.371,0:4.468,4:4.579,2:4.729,3:4.840,0:4.947,3:5.047,3:5.180,4:5.288,"}}
2016-09-17 13:56:46,350 - Skills - DEBUG - {"message_type": "recognizer_loop:audio_output_end", "context": null, "metadata": {}}
2016-09-17 13:56:46,354 - Skills - DEBUG - {"message_type": "recognizer_loop:audio_output_start", "context": null, "metadata": {}}
2016-09-17 13:56:46,357 - Skills - DEBUG - {"message_type": "enclosure.eyes.blink", "context": null, "metadata": {"side": "b"}}
2016-09-17 13:56:46,552 - Skills - DEBUG - {"message_type": "enclosure.mouth.viseme", "context": null, "metadata": {"code": "4:0.150,4:0.208,2:0.244,0:0.321,0:0.470,4:0.532,3:0.615,0:0.701,2:0.728,4:0.818,0:0.915,3:1.004,3:1.078,3:1.140,0:1.318,3:1.407,5:1.514,0:1.641,5:1.720,3:1.819,0:1.918,4:2.029,2:2.179,3:2.322,0:2.404,3:2.479,3:2.554,0:2.613,3:2.685,3:2.807,3:2.840,0:2.972,3:3.101,4:3.272,"}}
2016-09-17 13:56:49,950 - Skills - DEBUG - {"message_type": "recognizer_loop:audio_output_end", "context": null, "metadata": {}}
2016-09-17 13:56:55,300 - Skills - DEBUG - {"message_type": "recognizer_loop:record_begin", "context": null, "metadata": {}}
2016-09-17 13:56:58,806 - Skills - DEBUG - {"message_type": "recognizer_loop:record_end", "context": null, "metadata": {}}
2016-09-17 13:56:58,808 - Skills - DEBUG - {"message_type": "recognizer_loop:wakeword", "context": null, "metadata": {"session": "802db0e1-66a7-4be1-a85a-2a94051d2378", "utterance": "hey mycroft"}}
2016-09-17 13:57:00,358 - Skills - DEBUG - {"message_type": "recognizer_loop:utterance", "context": null, "metadata": {"session": "802db0e1-66a7-4be1-a85a-2a94051d2378", "utterances": ["set my lights to 10%"]}}
2016-09-17 13:57:00,362 - intent__init__ - ERROR - 
  File "/home/chris/Code/mycroft-core/mycroft/skills/intent/__init__.py", line 47, in handle_utterance
    utterance, num_results=100))
StopIteration
2016-09-17 13:57:00,363 - Skills - DEBUG - {"message_type": "intent_failure", "context": null, "metadata": {"utterance": "set my lights to 10%"}}
2016-09-17 13:57:00,364 - wolfram_alpha__init__ - DEBUG - Could not determine intent, falling back to WolframAlpha Skill!
2016-09-17 13:57:00,365 - Skills - DEBUG - {"message_type": "enclosure.mouth.think", "context": null, "metadata": {}}
2016-09-17 13:57:00,367 - requests.packages.urllib3.connectionpool - INFO - Starting new HTTPS connection (1): cerberus.mycroft.ai
2016-09-17 13:57:05,691 - requests.packages.urllib3.connectionpool - DEBUG - "GET /wolframalpha/v2/query?input=set+my+lights+to+10%25 HTTP/1.1" 200 460
2016-09-17 13:57:05,709 - Skills - DEBUG - {"message_type": "speak", "context": null, "metadata": {"utterance": "I couldn't get a response for set my lights to 10%. Maybe you meant lights?"}}
2016-09-17 13:57:05,712 - Skills - DEBUG - {"message_type": "recognizer_loop:audio_output_start", "context": null, "metadata": {}}
2016-09-17 13:57:05,713 - Skills - DEBUG - {"message_type": "enclosure.eyes.blink", "context": null, "metadata": {"side": "b"}}
2016-09-17 13:57:05,942 - Skills - DEBUG - {"message_type": "enclosure.mouth.viseme", "context": null, "metadata": {"code": "4:0.128,0:0.188,3:0.320,2:0.408,3:0.453,3:0.556,3:0.606,3:0.685,0:0.753,3:0.833,0:0.893,2:1.035,0:1.074,3:1.204,4:1.249,0:1.377,3:1.476,3:1.585,5:1.652,6:1.737,2:1.753,3:1.906,0:2.014,3:2.078,4:2.135,0:2.264,3:2.344,0:2.490,3:2.560,3:2.647,3:2.700,0:2.736,3:2.861,0:2.986,3:3.085,4:3.173,2:3.307,3:3.433,0:3.540,3:3.624,3:3.700,4:3.808,"}}
2016-09-17 13:57:09,840 - Skills - DEBUG - {"message_type": "enclosure.eyes.blink", "context": null, "metadata": {"side": "b"}}
2016-09-17 13:57:09,846 - Skills - DEBUG - {"message_type": "recognizer_loop:audio_output_end", "context": null, "metadata": {}}
2016-09-17 13:57:09,850 - Skills - DEBUG - {"message_type": "recognizer_loop:audio_output_start", "context": null, "metadata": {}}
2016-09-17 13:57:09,856 - Skills - DEBUG - {"message_type": "enclosure.eyes.blink", "context": null, "metadata": {"side": "b"}}
2016-09-17 13:57:09,973 - Skills - DEBUG - {"message_type": "enclosure.mouth.viseme", "context": null, "metadata": {"code": "4:0.212,4:0.273,0:0.429,4:0.484,0:0.591,0:0.676,2:0.718,4:0.811,0:0.904,3:0.977,3:1.054,3:1.115,0:1.297,3:1.380,3:1.453,4:1.587,"}}
2016-09-17 13:57:11,623 - Skills - DEBUG - {"message_type": "recognizer_loop:audio_output_end", "context": null, "metadata": {}}
```

Explicitly adding the '%' to the pattern made no difference.

clusterfudge:
This is a case where Adapt is a little bit brittle. I hadn't expected regex entities to be used to parse out part of a token (i.e. the '50' from '50%'), and as such, there's no test cases for it, and totally expectedly it doesn't work :).

I don't have an easy fix for adapt at this time, however I can recommend a change to your code. I was able to get the intent to fire by changing your regex slightly.

from 
```'(?P<Value>100|[1-9][0-9]?)(\D|$)'```
to 
```'(?P<Value>100|[1-9][0-9]?%?)(\D|$)'```

The optional '%' allowed both 50 and 50% to parse correctly, and triggered the intent for me. It would then be on you to strip out the percent symbol in your intent handler. I don't have any good ideas on how to fix this in Adapt right now, but will reference this pull/comment in the event I come up with something.

I also don't have a good explanation as to why your intent test is succeeding when it didn't work as part of mycroft-core; I've been out of the project for a while, and running unit tests in a debugger is currently broken for me. I don't have the time to invest into getting it working again, so I'm just going to leave that as a mystery for the time being.

ChristopherRogers1991:
Thanks! I updated the regex to allow for the optional '%.'

ChristopherRogers1991:
I've opened (this pull request)[https://github.com/MycroftAI/mycroft-skills/pull/3] to the skills repo. Assuming that's where skills are going to be tracked going forward, perhaps we should close this out in favor of that one?

aatchison:
@ChristopherRogers1991 
I was thinking the same. Close if we are in agreement.

ethanaward:
Closing since this has been merged into the skills repo with MycroftAI/mycroft-skills/#3

Issue #247, opened 2016-06-28
skill container --dir should be replaced with positional argument


forslund:
Previously the command to start a skill in a container was:

`./start.sh skill_container SKILL_PATH`

Right now the command needs to be supplemented with a `--dir`:

`./start.sh skill_container --dir SKILL_PATH`

this is clunky as the skill directory is generally needed (the default value will seldom work). I believe the old command line should be brought back.

ethanaward:
Fixed by #248

PR #248, opened 2016-06-28
Issues 247 - Make skill directory a positional argument

forslund:
this eliminates the need of always supplying --dir before the argument
that will basically always be needed. References #247.


jdorleans:
@ethanaward could you please test this?

ethanaward:
Yeah, will do

ethanaward:
LGTM, should be ready to merge.

PR #249, opened 2016-06-28
Added false positive audio accuracy test - Fixes #237

### LINKED ISSUE #237: Add false positive support to audio accuracy test
This should be able to both read in long segments of audio and count the false positives along with reading small audio files that have only a single phrase such as "hey mike ross". I think the result should not be a percentage since the ideal total detection count is 0 (and thus the % detected is x / 0).

Wolfgange3311999:
This adds false positive testing in the audio accuracy test. To use it, place audio files of any length without the wake word in `audio-accuracy-test/without_wake_word` and run `./start.sh audioaccuracytest`. This first scans for files in the now called `with_wake_word/query_after` folder and then scans in the `without_wake_word` directory. I can't think of a logical way to format the result of the false positive test as anything but a count of fails since there's no total expected fail count (beyond 0 which results in undefined).

[Here's a sample audio file](https://drive.google.com/file/d/0B7VRusOJyGc1VXVSQnpRQnpZZEU/view?usp=sharing) of talking from a YouTube video to test with.

### DIFF
--- a/audio-accuracy-test/audio_accuracy_test.py
+++ b/audio-accuracy-test/audio_accuracy_test.py
@@ -1,8 +1,10 @@
+import math
 import os
 import wave
 from glob import glob
 
 import pyee
+import time
 from os.path import dirname, join
 from speech_recognition import AudioSource
 from mycroft.client.speech.local_recognizer import LocalRecognizer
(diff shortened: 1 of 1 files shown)


audio-accuracy-test/audio_accuracy_test.py
```diff
@@ -43,82 +74,145 @@ def close(self):
...
+
+def test_false_negative(directory):
+    file_names = get_file_names(directory)
 
     # Grab audio format info from first file
-    ex_file = wave.open(file_names[0], 'rb')
+    tester = AudioTester(file_frame_rate(file_names[0]))
+
+    def print_status(word, short_name):
+        print("Wake word " + bold_str(word) + " - " + short_name)
```
jdorleans: There is no real advantage of declaring this inner method to print those data. Replace this method to the `print` function directly where they are needed. However, if you have a reusable logic across all your unit tests, you should move this method outside and use it in all unit tests methods instead.

Wolfgange3311999:
@jdorleans Thanks, I've fixed that redundancy.

ethanaward:
What is the status of this PR?

Wolfgange3311999:
@jdorleans LGTY?

jdorleans:
Sorry for taking long on this. That should be fine.

PR #252, opened 2016-06-29
Decreased "wake up" false positives - Fixes #189

### LINKED ISSUE #189: Mycroft is too light of a sleeper
After saying "mycroft sleep", the next time he hears "mycroft" he will wakeup (Even if you didn't say "mycroft wakeup".

Wolfgange3311999:
Previously after telling Mycroft to "go to sleep", the wakeup_recognizer was so sensitive that the next phrase with "mycroft" in it would wake Mycroft up even if it were something like "mycroft go away". So, this change both changes LocalRecognizer to accepting and requiring both the keyword and threshold as parameters to both allow them to be changed and to emphasize that for every new keyword there should correspondingly be a new threshold. Along with that, the default parameter for the keyword was removed since I don't think it makes sense to have LocalRecognizer default to using "mycroft" as the keyword since the class is generalized.

### DIFF
--- a/mycroft/client/speech/listener.py
+++ b/mycroft/client/speech/listener.py
@@ -208,9 +208,9 @@ class RecognizerLoop(pyee.EventEmitter):
 
         # FIXME - channels are not been used
         self.microphone.CHANNELS = channels
-        self.mycroft_recognizer = LocalRecognizer(sample_rate, lang)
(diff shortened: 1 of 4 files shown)


ethanaward:
LGTM

isaacnward:
Merging.

PR #255, opened 2016-07-01
Issues/234: Weather display on faceplate

isaacnward:
I've added functionality to allow the unit to display the weather (specifically, an icon representing the weather conditions, with the temperature next to it). To do this, I've also added a new class in Enclosure: Weather. 

This PR relies on MycroftAI/enclosure#18.


mycroft/client/enclosure/enclosure.py
```diff
@@ -185,7 +189,7 @@ def __register_mouth_events(self):
         self.client.on('recognizer_loop:record_begin', self.mouth.listen)
         self.client.on('recognizer_loop:record_end', self.mouth.reset)
         self.client.on('recognizer_loop:audio_output_start', self.mouth.talk)
-        self.client.on('recognizer_loop:audio_output_end', self.mouth.reset)
+        # self.client.on('recognizer_loop:audio_output_end', self.mouth.reset)
```
jdorleans: Is that supposed to be removed? If so, let's remove instead of comment.

mycroft/client/enclosure/weather.py
```diff
@@ -0,0 +1,47 @@
...
+        self.__init_events()
+
+    def __init_events(self):
+        self.client.on('enclosure.weather.display', self.display)
+
+    def display(self, event=None):
+        img_code, temp = 0, 0
+        if event and event.metadata:
+            img_code = event.metadata.get("img_code", img_code)
+            temp = event.metadata.get("temp", temp)
```
jdorleans: Is that correct to implement a default value of 0 for img and 0 for temp? I mean, if we send 0 and 0 to the enclosure arduino code, what will happen? Taking a step back, it seems we should not event send anything to the unit if we don't have valid values for img and temp. Don't you agree?

mycroft/client/enclosure/weather.py
```diff
@@ -0,0 +1,47 @@
...
+        msg = "weather.display=" + str(img_code) + str(temp)
```
jdorleans: We should use a separator char to differ from img_code and temp. Something like ; for example:
`msg = "weather.display=%d;%d" % (img_code, temp)`
This can result in something like: `"weather.display=02;70.5"`

mycroft/client/enclosure/weather.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def __init_events(self):
+        self.client.on('enclosure.weather.display', self.display)
+
+    def display(self, event=None):
+        if event and event.metadata:
+            img_code = event.metadata.get("img_code")
+            temp = event.metadata.get("temp")
+            msg = "weather.display=" + str(img_code) + str(temp)
+            self.writer.write(msg)
```
jdorleans: @isaacnward we'll need to verify the values like this:
```
if event and event.metadata:
  img_code = event.metadata.get("img_code", None)
  temp = event.metadata.get("temp", None)
  if img_code and temp: 
    msg = "weather.display=" + str(img_code) + str(temp)
    self.writer.write(msg)
```

mycroft/client/enclosure/weather.py
```diff
@@ -0,0 +1,46 @@
...
+        self.__init_events()
+
+    def __init_events(self):
+        self.client.on('enclosure.weather.display', self.display)
+
+    def display(self, event=None):
+        if event and event.metadata:
+            img_code = event.metadata.get("img_code")
+            temp = event.metadata.get("temp")
+            msg = "weather.display=" + str(img_code) + str(temp)
```
jdorleans: We gonna need a separator between `img_code` and `temp`:
`msg = "weather.display=%d;%d" % (img_code, temp)`

isaacnward:
The enclosure PR is ready, so I'm merging.

PR #257, opened 2016-07-03
Add media skills

forslund:
By request by @ethanaward I'm resubmitting the media-skills I've been playing around with. I think he wanted it mainly as a source of discussion. Personally I think the media base class (the one in this PR or another) is the most important part to get included in Mycroft so members of community can create media skills that work well together.

The *Spotify*, *local music* and *gmusic* skills all require the mopidy media-server. The *Swedish radio* skill is mainly there as a test with mpg123 based skill.

# Setup Instructions

## Installing mopidy

Included in the PR is `install-mopidy-debian.sh` and `install-mopidy-redhat.sh` for installing mopidy on the system.

## Setup

After mopidy has been installed it needs to be configured.
open *~/.config/mopidy/mopidy.conf* and modify as you see fit.

`[spotify] `

```
enabled=true
username=USERNAME
password=PASSWORD
```

and under
` [local] `

```
enabled = true
library = sqlite
media_dir = PATH_TO_YOUR_MUSIC
```

after this is done scan the local collection by running

` mopidy local scan `

and start mopidy by running `mopidy`

# Stuff to try

**Swedish radio**
`play p3`
`play p3 from swedish radio`

**Local Music**
`play rock music`
`play something by Cheer`

**Spotify**
`play discover weekly`
`play hello nasty from spotify`

**Common keywords**
`next`
`pause`
`resume`
`stop`
`what are we listening to`

# Questions and and known problems with this PR

## Lacks windows support
Mopidy does currently not work on windows, the maintainers claim that if someone performs the porting and creates a continuous integration setup they are willing to maintain it.

Possible solutions:
* Replace mopidy with an equally good media server with support for the major content providers.
* Refactor the code to allow multiple media servers as backend.
* Wait for windows support in mopidy

## Unittests
I'm working on it

## No pandora skill
Pandora is not available in my region so I haven't had the possibility to create a skill for it but it would be easy to do.

## Lower volume while mycroft is speaking
Lower volume while mycroft is speaking can be triggered a bit late on fast computers. I think it's the REST-api that's a little slow, possible solutions:
* replace post with MPD
* for localhost playback pulseaudio control could be used

Copied and adapted from https://github.com/forslund/mycroft-media-skills/, for complete history check that out.


ethanaward:
Thanks for opening this up! I agree, I think that the media base skill is going to be very important, especially deciding on a good baseline for what all media skills should include.

mycroft/skills/media/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+        to speak and raise it again when (s)he stops.
+
+        But wait there is one more thing! A small event emitter and matching
+        handler to stop media currently playing when new media is started.
+    """
+    def __init__(self, name):
+        super(MediaSkill, self).__init__(name)
+        self.isPlaying = False
+        config = ConfigurationManager.get()
+        self.base_conf = config.get('Media')
```
ethanaward: If you look at the base `MycroftSkill`, you should be able to replace `self.base_conf` with just `self.config`, since that's defined in the base class as getting the section of the configuration file with the name of the current skill. Just replace the `Media` section of the config with `MediaSkill`.

mycroft/skills/media/vocab/en-us/NextKeyword.voc
```diff
@@ -0,0 +1,3 @@
+next
+skip
+play next
```
ethanaward: We shouldn't need play next, since it'll be covered by next.

mycroft/skills/media/vocab/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+        self.isPlaying = False
+        config = ConfigurationManager.get_config()
+        self.base_conf = config.get('Media')
+
+    def initialize(self):
+        logger.info('Initializing MediaSkill commons')
+        logger.info('loading vocab files from ' + join(dirname(__file__),
+                    'vocab', self.lang))
+        self.load_vocab_files(join(dirname(__file__), 'vocab', self.lang))
+
```
ethanaward: Is this intended to load common vocab files for all media skills that inherit from this base class, or just the vocab from the current media skill that's inheriting? If it's the latter, it might be better to just leave this to the inherited class.

mycroft/skills/media/vocab/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+        self.register_intent(intent, self.handle_prev)
+
+        intent = IntentBuilder('StopIntent').require('StopKeyword')
+        self.register_intent(intent, self.handle_stop)
+
+        intent = IntentBuilder('PauseIntent').require('PauseKeyword')
+        self.register_intent(intent, self.handle_pause)
+
+        intent = IntentBuilder('ResumeIntent').require('ResumeKeyword')
+        self.register_intent(intent, self.handle_resume)
```
ethanaward: Is there a reason to not make this a PlayIntent rather than Resume? It seems to me like they'll be the same, and `play` seems like a more intuitive name.

mycroft/skills/media/vocab/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+           except the intents to start playback.
+        """
+        intent = IntentBuilder('NextIntent').require('NextKeyword')
+        self.register_intent(intent, self.handle_next)
+
+        intent = IntentBuilder('PrevIntent').require('PrevKeyword')
+        self.register_intent(intent, self.handle_prev)
+
+        intent = IntentBuilder('StopIntent').require('StopKeyword')
+        self.register_intent(intent, self.handle_stop)
```
ethanaward: We might not need this, since we already have the `stop` function that triggers on `Hey Mycroft stop`.

mycroft/skills/media/vocab/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+            .require('CurrentlyPlayingKeyword')
+        self.register_intent(intent, self.handle_currently_playing)
+
+    def _register_event_handlers(self):
+        """
+           Register event handlers for stopping currently playing media
+           when new media is started and handlers for lowering media volume
+           while mycroft is speaking.
+        """
+        self.emitter.on('mycroft.media.stop', self.handle_stop)
```
ethanaward: Do we need this special stop functionality, since we already have the default mycroft stop?

mycroft/skills/media/vocab/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+        logger.info('handle_stop not implemented')
+
+    def stop(self):
+        logger.debug('No stop method implemented')
+
+    def lower_volume(self, message):
+        logger.debug('Lower volume not implemented')
+
+    def restore_volume(self, message):
+        logger.debug('Restore volume not implemented')
```
ethanaward: For all of these debug lines, change them to something like `"... not implemented in" + self.name` to give developers more useful information.

mycroft/skills/media/vocab/en-us/FromKeyword.voc
```diff
@@ -0,0 +1 @@
+from
```
ethanaward: Is this used in the base media skill?

mycroft/skills/media/vocab/en-us/PauseKeyword
```diff
@@ -0,0 +1 @@
+pause
```
ethanaward: Change this to be a .voc.

mycroft/skills/media/vocab/en-us/ResumeKeyword
```diff
@@ -0,0 +1,2 @@
+resume
+unpause
```
ethanaward: Change this to be a .voc.

mycroft/skills/media/__init__.py
ethanaward: Actually, never mind. I see that this is intended to provide a base config area for general media skills.

ethanaward: Actually, never mind. I see that this is intended to provide a base config area for general media skills.

mycroft/skills/media/vocab/__init__.py
forslund: I'm not sure it's needed. I did this originally before the stop functionality in mycroft was added. After it was added I felt it was nicer to the rest of the skills if I sent a stop request only to the skills I was interested in stopping.

Couldn't for example a timer skill be using the stop functionality and the user would be annoyed that the timer stops just because she switched from a local album to a spotify playlist. (this would cause the event to be emitted to ensure that two media sources aren't playing on top of each other)

mycroft/skills/media/vocab/__init__.py
forslund: There's a couple of reasons
I have intentionally left all playIntents outside the base class. They will all be different, there might be several for each skill so I've left it for each skill to implement. Basically anything with "play" in it I've left for the inherited class to handle.

Pause/Resume was the latest additions so I haven't quite landed in how I to best implement it. A separate intent for resume was the most straightforward way without adding lots of logic to the existing skills.

I'm willing to do merge this into a PlayIntent if you'd like that better.

mycroft/skills/media/vocab/__init__.py
forslund: It loads the common vocab files under `media/vocab/...`

The intention was that all skills that inherit from this skill should have access to the same keywords for normal operation to make operation easier for the user.

mycroft/skills/media/vocab/en-us/FromKeyword.voc
forslund: It was at one point, currently it isn't. It might not be necessary at all come to think of it...I used it when I want to specify a skill exclusively to handle a play request...
```py
intent = IntentBuilder('PlayFromIntent' + self.name)\
            .require('PlayKeyword')\
            .require('PlaylistKeyword')\
            .require('FromKeyword')\
            .require('NameKeyword')\
            .build()
```

but the from keyword isn't really necessary is it?

```py
intent = IntentBuilder('PlayFromIntent' + self.name)\
            .require('PlayKeyword')\
            .require('PlaylistKeyword')\
            .require('NameKeyword')\
            .build()
```
would work just as nicely right?

ethanaward: Yeah, and you can even register some regex if you don't want to go through the process of finding every single possible playlist or song name. Something like `from (?P<MediaName>.*)` would match any request to play from somewhere.

mycroft/skills/media/vocab/__init__.py
ethanaward: I think for reminders it has a different stop functionality of its own, but I get your point, yeah. I think that keeping that special `mycroft.media.stop` message is fine so that you can do things like stop one media source while playing anonother one. 

I think that you should remove the StopIntent though, and instead put an implementation in the `stop` function, since any stop vocab will always conflict with the default mycroft stop vocab.

mycroft/skills/media/vocab/__init__.py
ethanaward: I see, so you want to leave room for custom plays for different media types and sources. That does make sense, but I think that shouldn't conflict with a base `PlayIntent`, since just playing something paused is something that has meaning across all types of media. I do think it would work better worded as `PlayIntent` versus `ResumeIntent`.

mycroft/skills/media/vocab/__init__.py
forslund: Good point, I'll get right on that

mycroft/skills/media/vocab/__init__.py
ethanaward: Alright, I see. That definitely makes sense. Creating common sources of vocab for multiple skills to use is definitely something that I've been thinking about, as it removes redundancy as well as possible conflicts. 

The one thing I'm not sure about is whether this is the best way to do it. I might be wrong, but wouldn't this require loading the base MediaSkill as a skill on its own, and so registering each of its intents to stub handlers? @clusterfudge, what do you think?

mycroft/skills/media/vocab/en-us/FromKeyword.voc
forslund: I used that in the beginning but I went over to the from keyword to make it easier to translate the skill... I do a dirty hack in the spotify skill to allow some free searches...

Any way I've removed it from the media common vocabulary and just use it in the spotify skill. I'll see if I can figure out how to remove it from there as well...

forslund:
I think I've updated most of the concerns that has been discussed tonight. Thanks for the pointers so far. I'm in the middle of doing a refactoring that will make some of the media skills look nicer and allow using Kodi as an alternative to mopidy. (only for local music skill so far but they're adding spotify support so hopefully that too). Hopefully I can push that tomorrow morning.

Looking forward to more pointers and maybe a better way to handle the common vocabulary.

ethanaward:
No problem, thanks for the feedback. Do you think we would be able to get at least the media base skill into the release tomorrow, or do you think we should wait?

forslund:
I'm happy with how the base class work and I think it's good to add it as soon as possible so others can build on it and comment on flaws and irregularities.

If only I can find what I've done to break the test...

ethanaward:
It looks like two things: intent tests
```
======================================================================

ERROR [0.313s]: test_IntentValidation[npr_news:sample2] (skills.discover_tests.IntentTestSequence)

----------------------------------------------------------------------

  File "/home/travis/build/MycroftAI/mycroft-core/test/skills/discover_tests.py", line 34, in test

    SkillTest(a, b, self.emitter).run()

  File "/home/travis/build/MycroftAI/mycroft-core/test/skills/skill_tester.py", line 64, in run

    assert False

AssertionError

======================================================================

ERROR [0.316s]: test_IntentValidation[reminder:sample4] (skills.discover_tests.IntentTestSequence)

----------------------------------------------------------------------

  File "/home/travis/build/MycroftAI/mycroft-core/test/skills/discover_tests.py", line 34, in test

    SkillTest(a, b, self.emitter).run()

  File "/home/travis/build/MycroftAI/mycroft-core/test/skills/skill_tester.py", line 64, in run

    assert False

AssertionError

----------------------------------------------------------------------
```

and one pep8 error: 
```
1.28s$ pep8 mycroft test

mycroft/skills/media/__init__.py:58:21: E127 continuation line over-indented for visual indent

The command "pep8 mycroft test" exited with 1.
```

forslund:
Closing this since #267 implemented the core of this one. Might reopen a pull request with implementations using the base class.

PR #265, opened 2016-07-07
Issue-264 cairo url updated.

el-tocino:
This should fix the problem with the old url downloading the error page instead of the desired file. I hope.


jdorleans:
@ethanaward could you confirm this?

ethanaward:
As far as I can tell, both of these links download the file. I'm not sure that this is needed.

ethanaward:
Closing, since this is not needed.

PR #266, opened 2016-07-07
Add new listener config options- Fixes #260

### LINKED ISSUE #260: Extract listener values into configuration
A few options for the listener:

- [ ] Wake word
- [ ] Pronunciations
- [ ] Threshold
- [ ] 3 other listener parameters

Wolfgange3311999:
This brings in a set of new config vaues for adjusting the behavior of the listener along with the wake word. Here's the new values:
```ini
[listener]
wake_word = "hey-mycroft"
phonemes = "HH EY M AY K R AO F T"
threshold = "1e-90"

threshold_multiplier = "1.0"
dynamic_energy_ratio = "1.5"
```

Instead of keeping a dictionary with all used word pronunciations LocalRecognizer now creates a temporary .dict file named the wake word for every recognizer instantiated. This allows the .dicts to be generated from the config rather than having the user manually edit the embedded file.


Wolfgange3311999:
I've just verified the audio accuracy test returns the same value as before this change.

ethanaward:
Test are passing, LGTM, merging.

PR #267, opened 2016-07-07
Add media base class

forslund:
As per request by @ethanaward, a base class for media skills. The content has been previously discussed in PR #257


ethanaward:
@clusterfudge, please review

mycroft/skills/media/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+        to speak and raise it again when (s)he stops.
+
+        But wait there is one more thing! A small event emitter and matching
+        handler to stop media currently playing when new media is started.
+    """
+    def __init__(self, name):
+        super(MediaSkill, self).__init__(name)
+        self.isPlaying = False
+        config = ConfigurationManager.get()
+        self.base_conf = config.get('Media')
```
jdorleans: Let's keep our standard and use MediaSkill instead of Media: `self.base_conf = config.get('MediaSkill')`

mycroft/skills/media/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def initialize(self):
+        logger.info('Initializing MediaSkill commons')
+        logger.info('loading vocab files from ' + join(dirname(__file__),
+                    'vocab', self.lang))
+        self.load_vocab_files(join(dirname(__file__), 'vocab', self.lang))
+
+        self.register_vocabulary(self.name, 'NameKeyword')
```
jdorleans: Are we supposed to load `NameKeyword` file from any place? If so, why it is not inside the vocabulary folder? We should already pick this during the above call: `self.load_vocab_files(join(dirname(__file__), 'vocab', self.lang))`

mycroft/skills/media/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+           handle_currently_playing() should be implemented to tell the user
+           what is currently playing
+        """
+        logger.debug('handle_currently_playing not implemented in ' +
+                     self.name)
+
+    def play(self):
+        """ Stop currently playing media before starting the new. """
+        logger.info('Stopping currently playing media if any')
+        self.emitter.emit(Message('mycroft.media.stop'))
```
jdorleans: Since this MediaSkill represents an abstraction, we should just call the method `stop` directly instead of sending an event to be consumed by exactly this class.

mycroft/skills/media/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+        logger.debug('handle_play not implemented in ' + self.name)
+
+    def handle_stop(self, message):
+        """
+           handle_stop() should be implemented to stop currently playing media
+        """
+        logger.debug('handle_stop not implemented in ' + self.name)
+
+    def stop(self):
+        self.handle_stop(None)
```
jdorleans: I'd say, the call dependency should be the opposite here. `handle_stop` should call `stop` and not the contrary.
Basically, `handle_stop` will receive a message, extract data from it, then perform some validation and finally execute the real stop action. This action should be implemented inside the `stop` method.

mycroft/skills/media/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def lower_volume(self, message):
+        logger.debug('Lower volume not implemented in ' + self.name)
```
jdorleans: We do have VolumeSkill. Let's integrate with it by sending a message to reduce the volume instead of reimplementing the same behaviour here. We should be able to do this by calling: 
`self.client.emit(Message("DecreaseVolumeIntent"))`

mycroft/skills/media/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def restore_volume(self, message):
+        logger.debug('Restore volume not implemented in ' + self.name)
```
jdorleans: Same as above, we should be able to increase the volume by sending the message: 
`self.client.emit(Message("IncreaseVolumeIntent"))`

mycroft/skills/media/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def _set_sink(self, message):
```
jdorleans: Let's rename this method to something more meaningful. Just call it: `_set_output_device` or similar.

mycroft/skills/media/vocab/en-us/CurrentlyPlayingKeyword.voc
```diff
@@ -0,0 +1,4 @@
+what's currently playing
+what are you playing
+What are we listening to
+What's playing
```
jdorleans: Please, also add `what is ...` (without abbreviation) as possible utterances.

mycroft/skills/media/vocab/en-us/PrevKeyword.voc
```diff
@@ -0,0 +1 @@
+play previous
```
jdorleans: It's better to follow similar commands for Next and Prev in order to help the user to easily understand the feature.
So, why not add `previous track` too?

mycroft/skills/media/__init__.py
jdorleans: Ok, I've just understood what you wanna do here... control the media volume as opposite to Mycroft volume. Cool... so forget about my previous comment.

mycroft/skills/media/__init__.py
jdorleans: Forget this previous comment as well.

mycroft/skills/npr_news/test/intent/sample2.intent.json
```diff
@@ -1,7 +1,7 @@
 {
-  "utterance": "play the news",
+  "utterance": "is there any news",
```
jdorleans: Why you changed this?

jdorleans:
@forslund that's a very good model. Thanks for providing this. There are a few comments to be addressed before we merge.

mycroft/skills/media/__init__.py
forslund: This is used to stop any other currently playing media skill when starting playback. The current name comes from being used as super for the `play` implementation in the skills. I decided against sending `mycroft.stop` to not interfere with other skills and only target the media skills.

maybe `stop_currently_playing_media` would be a better name?

mycroft/skills/media/__init__.py
forslund: `handle_stop` is there for consistency with all the other `handle_` functions. The idea was that the skill developer should pretty much exclusively be implementing `handle_` functions.

There is no specific stop intent for the media skills instead the mycroft stop feature is used and that always uses the `stop` method.

I could remove the `handle_stop` method completely if you like...

mycroft/skills/npr_news/test/intent/sample2.intent.json
forslund: The play skill of implemented media skills in the last PR ate the "play the news" utterence. To make sure the test wasn't broken @ethanaward suggested that I'd change the test.

mycroft/skills/media/__init__.py
forslund: I'll remove it,The NameKeyword contains the current skills name in this case (retrieved from self.name). It shouldn't be here, it should be in the inheriting skill.

mycroft/skills/media/__init__.py
jdorleans: The name is not a concern. It's good enough. I was wondering if we wanna play multiple skills at the same time though. Does this make any sense? Do you have a practical example? If so, we can keep this behaviour by sending an event instead of calling the method itself.

mycroft/skills/media/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def handle_currently_playing(self, message):
+        """
+           handle_currently_playing() should be implemented to tell the user
+           what is currently playing
+        """
+        logger.debug('handle_currently_playing not implemented in ' +
+                     self.name)
+
+    def play(self):
```
jdorleans: What's the difference between this `play` and `handle_play`?

mycroft/skills/media/__init__.py
forslund: It's used all the time by me...

I start a local album
*Hey Mycroft play something by the Eels*
**Eels starts playing by the local_music skill**

Then I want to listen to the radio channel P3 (Eels are still playing)
*Hey mycroft play P3*

At this point there has been no stop command stopping the music instead the play handler of the swedish radio skill emits the `mycroft.media.stop` message.
All media skills receives this and their stop method is called. In this case the local music skill stops playing eels.
The swedish radio skill starts playing.

That is how I've been using it.

mycroft/skills/media/__init__.py
jdorleans: To be honest, I don't see a scenario where we want to stop the media only and keep something else from mycroft going. So, It seems to me we should not have a specific `handle_stop`, only `stop` though.
Do you have an example where makes sense to stop a media only?

mycroft/skills/media/__init__.py
forslund: handle_play is a mehod for generic playback/unpause resume that the skill creator should implement.

`play` is a function that the user should call to let the media handle internal stuff before new playback is started. The name might not be the best choice. So far all it does is to make sure other media is stopped.

mycroft/skills/npr_news/test/intent/sample2.intent.json
ethanaward: Yeah, there was a conflict between `play` and `news` as keywords in the `play the news` utterance. That's a larger problem with intent parsing in general, and can't really be solved here. I suggested that he just change that test for the moment, since the play part of it isn't an essential part of the test.

jdorleans: If the npr test broke, it means we gonna have conflicts and trouble between media and npr skills. Unless we enforce the utterance somehow on one of the sides that'll still happen. Have we tried a real media skill using the keyword `play` and verify if it conflicts with npr?

mycroft/skills/media/__init__.py
forslund: Only a theoretical one... from discussion in the previous thread:
"Couldn't for example a timer skill be using the stop functionality and the user would be annoyed that the timer stops just because she switched from a local album to a spotify playlist. (this would cause the event to be emitted to ensure that two media sources aren't playing on top of each other)"

The event will be sent each time the play is triggered

mycroft/skills/npr_news/test/intent/sample2.intent.json
forslund: Yes in PR #257 there was a conflict here and the test case broke.

ethanaward: No, there's no conflict between the actual vocab in the NPR news skill and these play skills. The only conflict was in a hypothetical `play the news

mycroft/skills/media/__init__.py
jdorleans: Very cool and interesting behaviour. It seems very practical. However, I got another question. 

Since sending a message is an asynchronous process, how can we guarantee the RadioSkill will be first stopped and then start playing the radio after that?

It seems to me we may also end-up in the following flow:
0. MusicSkill is playing some music...
1. User triggers the RadioSkill by saying: `play P3`
2. RadioSkill sends the `media.stop` --> async process
3. RadioSkill starts playing the radio
4. RadioSkill is notified of `media.stop` and stops playing the radio - KO

Any idea how to avoid that?

jdorleans: Maybe we could add some data along `media.stop` to identify the sender. Something like:
`self.emitter.emit(Message('mycroft.media.stop', metadata={sender=self.name}))`

That way, all media skills could read the `sender` and verify whether the message was sent by itself or not. It can then decide to process it or not.

mycroft/skills/media/__init__.py
jdorleans: Ok, the example above which integrates `play` and `stop` answers this question. As a result, it'll make sense to have `handle_stop` for media only if we get the `play` to work properly.

mycroft/skills/media/__init__.py
forslund: I've never run into it but there might be an issue.

It might have to do with the fact that I always have a delay allowing mycroft to present what he's starting.

Another way is to add a flag a lock mechanism.
```python

def stop_currently_playing(self):
        """ Stop currently playing media before starting the new. """
        logger.info('Stopping currently playing media if any')
        self.wait_for_stop = True
        self.emitter.emit(Message('mycroft.stop'))
        while(self.wait_for_stop:
              time.sleep(0.5)
    def stop(self):
        self.wait_for_stop = False
        handle_stop(None)
```

Or something more fancy perhaps

mycroft/skills/media/__init__.py
jdorleans: Ok, it seems then `play` should be renamed to `before_play` and then be called inside `handle_play`. I'll add a note on `handle_play` method then.

mycroft/skills/media/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+        self.emitter.emit(Message('mycroft.media.stop'))
+
+    def handle_pause(self, message):
+        """ handle_pause() should pause currently playing media """
+        logger.debug('handle_pause not implemented in ' + self.name)
+
+    def handle_play(self, message):
+        """ handle_play() generic play handler. Should resume paused media
+            and/or implement generic playback functionality."""
+        logger.debug('handle_play not implemented in ' + self.name)
```
jdorleans: As mentioned on the `play` comment, we should call `before_play` here.

forslund:
I think I've taken care of most of the comments above. I might be a bit slow in replying from here on in, since I'm leaving for vacation this afternoon, but I'll try to check in regularly and continue to address issues and suggestions.

PR #269, opened 2016-07-07
Feature/use compiled ap voice

aatchison:
This PR should removes the flitevox file, as mimic now has it compiled in. It will require you to run dev_detup.sh again to compile the AP voice into mimic.


ethanaward:
LGTM, merging

PR #271, opened 2016-07-08
Update load_data_files to load regex files (Fixes #256)

### LINKED ISSUE #256: load_data_files should call load_regex_files
I see that not having `load_data_files` call `load_regex_files` was done on purpose, according to the first comment on https://github.com/MycroftAI/mycroft-core/pull/222, however, it seems strange to have `load_data_files` not load all the data files, so I wanted to open this up for discussion.

The function is currently written as:
```
    def load_data_files(self, root_directory):
        self.init_dialog(root_directory)
        self.load_vocab_files(join(root_directory, 'vocab', self.lang))
```
It seems like it could be:
```
    def load_data_files(self, root_directory):
        self.init_dialog(root_directory)
        self.load_vocab_files(join(root_directory, 'vocab', self.lang))
        regex_path = join(root_directory, 'regex', self.lang)
        if os.path.exists(regex_path):
            self.load_regex_files(regex_path)
```
This should safely load it for the skills that rely on it, without causing harm to the others.

Is there any strong argument for not having `load_data_files` load the regex files in this, or a similar manner?

ethanaward:
This changes load_data_files to also load regex files if they are available. This addresses #256.

### DIFF
--- a/mycroft/skills/send_sms/__init__.py
+++ b/mycroft/skills/send_sms/__init__.py
@@ -42,7 +42,6 @@ class SendSMSSkill(MycroftSkill):
 
     def initialize(self):
         self.load_vocab_files(join(dirname(__file__), 'vocab', self.lang))
-        self.load_regex_files(join(dirname(__file__), 'regex', self.lang))
 
         intent = IntentBuilder("SendSMSIntent").require(
             "SendSMSKeyword").require("Contact").require("Message").build()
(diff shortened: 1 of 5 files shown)


mycroft/skills/send_sms/__init__.py
```diff
@@ -42,7 +42,6 @@ def __init__(self):
 
     def initialize(self):
         self.load_vocab_files(join(dirname(__file__), 'vocab', self.lang))
-        self.load_regex_files(join(dirname(__file__), 'regex', self.lang))
```
jdorleans: It seems we should keep this one, since we load a vocab instead of all data.

ethanaward:
Good point, yeah. I readded that.

jdorleans:
Great, merge it!

PR #280, opened 2016-07-13
Issues-279 - Added 'extras' system so skills can send more information.

joshymcd:
Also added default of the skills name to the extras. skills should now be able to call the inherited 'add_extra' method to add extras to its responses


joshymcd:
re pulled the repo before making this PR.
not sure why the build is now failing. will close this PR and make a new one once ive figured it out

Issue #279, opened 2016-07-13
Allow skills to send more information than the utterance to the messagebus


joshymcd:
Currently skills send their 'utterance' to the message bus but with no extra information.
there should be a way to allow skills to send extra information so custom UI's can display the data whichever way it feels fit.
e.g the wiki skill should/could be able to send the URL to the wiki page it is getting its information from:
`{'Title': 'Ubuntu (operating system)', 'URL': 'https://en.wikipedia.org/wiki/Ubuntu_(operating_system)'}`
or the weather skill should be able to send structured information e.g 
`{'High': 30, 'Low': '20', 'Unit': 'Celsius' }`

joshymcd:
Created PR #280 for this issue

krisgesling:
Closing very old issue.

Worth noting, we now have a defined range of [Message types](https://mycroft-ai.gitbook.io/docs/mycroft-technologies/mycroft-core/message-types#) along with the ability for custom messages namespaced to the Skill. 

This also ties into the new GUI framework.

PR #281, opened 2016-07-13
Issue #279 - Added results property to skills

joshymcd:
Also added default of the skills name to the extras. 
Skills should now be able to call the inherited 'add_extra' method to add extras to its responses
ref issue #279

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -207,8 +207,8 @@ class MycroftSkill(object):
                 handler(message)
                 if len(self.results) > 0:
                     self.emitter.emit(
-                        Message("results", metadata=self.results))
-                    self.results.clear()
+                        Message("results",
+                                metadata=self.results))
             except:
                 # TODO: Localize
                 self.speak(
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -162,6 +162,7 @@ class MycroftSkill(object):
         self.dialog_renderer = None
         self.file_system = FileSystemAccess(join('skills', name))
         self.registered_intents = []
+        self.results = {}
 
     @property
     def location(self):
@@ -204,6 +205,10 @@ class MycroftSkill(object):
         def receive_handler(message):
             try:
                 handler(message)
+                if len(self.results) > 0:
+                    self.emitter.emit(
+                        Message("results", metadata={"data": self.results}))
+                    self.results.clear()
             except:
                 # TODO: Localize
                 self.speak(
@@ -224,6 +229,10 @@ class MycroftSkill(object):
         re.compile(regex_str)  # validate regex
         self.emitter.emit(
             Message('register_vocab', metadata={'regex': regex_str}))
+
+    def add_result(self, key, value):
+        self.results['name'] = self.name
+        self.results[str(key)] = str(value)
 
     def speak(self, utterance):
         self.emitter.emit(Message("speak", metadata={'utterance': utterance}))
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -207,7 +207,7 @@ class MycroftSkill(object):
                 handler(message)
                 if len(self.results) > 0:
                     self.emitter.emit(
-                        Message("results", metadata={"data": self.results}))
+                        Message("results", metadata=self.results))
                     self.results.clear()
             except:
                 # TODO: Localize
@@ -231,8 +231,8 @@ class MycroftSkill(object):
             Message('register_vocab', metadata={'regex': regex_str}))
 
     def add_result(self, key, value):
-        self.results['name'] = self.name
-        self.results[str(key)] = str(value)
+        self.results['skill_name'] = self.name
+        self.results[str(key)] = value
 
     def speak(self, utterance):
         self.emitter.emit(Message("speak", metadata={'utterance': utterance}))
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -209,6 +209,7 @@ class MycroftSkill(object):
                     self.emitter.emit(
                         Message("results",
                                 metadata=self.results))
+                    self.results.clear()
             except:
                 # TODO: Localize
                 self.speak(
--- a/mycroft/skills/date_time/__init__.py
+++ b/mycroft/skills/date_time/__init__.py
@@ -76,6 +76,10 @@ class TimeSkill(MycroftSkill):
                 return
 
         time_value = now.astimezone(tz)
+        add_result("location" : location)
+        add_result("time": time_value)
+        add_result("timezone": tz)
+        
         self.speak("It is currently " + self.get_time_format(time_value))
 
     def stop(self):
--- a/mycroft/skills/date_time/__init__.py
+++ b/mycroft/skills/date_time/__init__.py
@@ -76,10 +76,10 @@ class TimeSkill(MycroftSkill):
                 return
 
         time_value = now.astimezone(tz)
-        add_result("location" : location)
-        add_result("time": time_value)
-        add_result("timezone": tz)
-        
+
+        self.add_result("location", location)
+        self.add_result("time", self.get_time_format(time_value))
+
         self.speak("It is currently " + self.get_time_format(time_value))
 
     def stop(self):


Wolfgange3311999:
@joshymcd Thanks for submitting this! I like the idea however there's a few things I'd like to point out:
 - I don't know if Extra is the right word to describe this. How about using the word `Result` instead?
 - Currently this would output the extra info every time a skill speaks. When a skill speaks multiple times such as in the weather skill it will send the info twice which doesn't make much sense. Along with this the developer must make sure to add results before calling speak which might not be that obvious. Instead can you make something like a `self.output_results` method that runs once right after the skill's intent is handled.
 - I think `self.extras` should only be printed if it has anything beyond `'skill_name'`.
 - Lastly, the `self.extras` should be cleared when outputting to prevent sending old information that wasn't replaced in the current intent.

joshymcd:
@Wolfgange3311999  re point 2 - would this be a separate method/function the skill runs to submit results in the same way it implements self.speak when it needs to ?

also to be clear would you add ~~extras~~ results using the same ~~add_extra~~ add_results function then send them to the messagebus using self.output_results?

Wolfgange3311999:
@joshymcd Yes, that's along the lines of what I was thinking. Although perhaps the `output_results` method could be automatically called at the end of every intent handler (by placing it [right after the call here][1]).

[1]:https://github.com/joshymcd/mycroft-core/blob/7e91f8e2291a4871d65946ddc317bc3bafc03977/mycroft/skills/core.py#L207

ethanaward:
Any update on this?

joshymcd:
still in progress, will push my update later today/tonight :)

ethanaward:
Sounds good, thanks!

joshymcd:
how about now?
should be ready to go

Also i have another branch im going to PR with a modded time skill which uses this functionality to give: 
```javascript
{
  message:{
    context: null,
    id: 1469742948393,
    message_type: "results",
    metadata:{
      data:{
        name: "TimeSkill",
        time: "2016-07-22 00:15:48.337602+01:00",
        timezone: "Europe/London"
      }
    }
  }
}
```

Wolfgange3311999:
@joshymcd I'll have to test this out tomorrow with something like the weather skill, but overall it looks great!

joshymcd:
@Wolfgange3311999 you should just be able to add
```python
add_result("location" : location)
add_result("time": time_value)
add_result("timezone": tz)
```
to the bottom of the date_time skills `handle_intent()` method (just before the `speak()` command)

Wolfgange3311999:
@joshymcd I decided to begin writing a standard system for the enclosure client to interact with Mycroft using this change. After doing so I noticed a few things:
 - If there aren't plans to put anything else in the message `metadata`, the keys can all be put directly into there rather than being inside `data`.
 - `name` should probably be renamed to `skill_name` to prevent any conflicts.
 - The keys and values should be able to be anything rather than getting casted to a string.

So I was thinking that the results message should probably look like this:
```
{
  "message_type": "results",
  "context": null,
  "metadata": {
    "data": {
      "skill_name": "WeatherSkill"
      "temp_current": 92,
      "scale": "fahrenheit",
      "temp_max": 98,
      "location": "Lawrence, KS",
      "temp_min": 87,
      "condition": "clear sky"
    }
  }
}
```
After that change I think this is merge ready :).

joshymcd:
Should have this done by end of the day (was on holiday)

aatchison:
Ahh, cool. How are we doing?

aatchison:
*bump*

joshymcd:
should be good to go i think
i also updated the date/time skill with some 'results'

joshymcd:
also thanks for the bump @aatchison :D

Wolfgange3311999:
@joshymcd Thanks for those last touches. There's still a `str()` cast around `key` although I think that might be better; it wouldn't make too much sense to ever have a result key be an integer.

So, LGTM.

clusterfudge:
If results is not a property on the skill itself (via the MycroftSkill base class), the skill remains stateless (which is good!). Concurrent calls to handle_request (which doesn't apply to this skill, but may to others) could result in the results dict containing properties from both calls. The most accurate description I could use to describe this is 'not thread-safe'. 

What you're trying to provide here is a result object that's scoped to a call of the handler, and manage the state for the developer. I think the former is a good call, but the latter may not be entirely practical. Aside from being slightly more verbose, I don't think there's any problem with the developer of a skill managing the result object in the scope of the handler, and then either emitting a 'result' Message themselves, or providing a send_result(Message) convenience function as part of the MycroftSkill class. 

Lastly, time is a pretty simple skill, and the results concept doesn't account for conditional logic that may exist in other skills (including error handling). Allowing users to construct their result object and only emit it when appropriate (i.e., themselves, as opposed to as part of the framework) is probably a good idea.

clusterfudge:
I would also recommend including some kind of type information in the result message. Something like "TimeResult", or "WeatherResult" would be really helpful when something actually starts consuming these messages (instead of trying to infer the type from the contents of the message), and you should also keep in mind that skills may have multiple types of results, so tying to the skill name may not be sufficient either.

mycroft/skills/core.py
```diff
@@ -204,6 +205,11 @@ def register_intent(self, intent_parser, handler):
         def receive_handler(message):
             try:
                 handler(message)
+                if len(self.results) > 0:
+                    self.emitter.emit(
+                        Message("results",
+                                metadata=self.results))
+                    self.results.clear()
```
clusterfudge: I think you would want this to be in a finally block; if handler throws an exception, it would leave the results dict in an unknown state, to be flushed by the next caller.

joshymcd:
@clusterfudge so would you want a 'emmit_results' method instead of being done automagically?
python isnt my native language so sorry if i'm being a bit slow!

main question: What would be the user process you having in mind (user being the person writing the skill)?

also i was moving away from having "TimeResult", or "WeatherResult" to make it less skill/coder dependent. currently no matter what is put into the result object there is a property in the result object called skill_name which automatically takes the names of the skill. that way it should be easy to parse and always unique (maybe)?

clusterfudge:
@joshymcd I think the emit_results method will allow for more explicit error handling, as well as more explicit typing of the results themselves. 

> that way it should be easy to parse and always unique (maybe)?

I think you're making assumptions about skills that can't/shouldn't be made. Any given intent could have multiple result types for success as well as for errors. I can also imagine a world where these result objects are emitted asynchronously (think progress or next track info for a media player skill), which doesn't exist in this world. Something that's more explicit in this case can be reused elsewhere in ways we don't yet understand :). 

Lastly, since most event handlers within mycroft (not thinking about the clients consuming these result messages now) key off the message_name property of Messages, it would probably be better to have a more explicitly named message than "results". This would allow for simpler logic in clients consuming these messages, as well as potential optimizations later (where messages of certain types are only routed to the appropriate clients). Example

```
Message(message_type="WeatherForecast",
metadata={"source": "OpenWeatherMapSkill", "temperature".....})
```

I think you're on the right track with separating the result from the skill implementation (where appropriate), for example in the above we have a WeatherForecast result, but it could come from OWM or weather.com or weather underground or any other source, as long as we can get the data into the right format. I'm a huge fan of having multiple providers of the same functionality, but it will require a definition of some core result types over time.

so to come back and answer your original question, "what would the user process look like", I would imagine something like this

```
forecast = self.OWM.getForecast("Seattle, WA")
self.emit_results("WeatherForecast", forecast)
self.speak_dialog("weather.forecast", forecast)
```

joshymcd:
back off holiday :) 
Ill have a play and see what i come up with. thanks for the assistance :)

aatchison:
We have made a major update to master and plan on getting to this, thanks!

aatchison:
Ok, we are officially released in master for our new back end at home.mycroft.ai. If you wouldn't mind re basing and making a PR against the dev branch, we can continue the discussion there. Thanks!

JarbasAI:
im using this in jarbas, added emit_results and automagicaly added speak/dialog as a result also

for reference:  https://github.com/JarbasAI/jarbas-core/blob/dev/mycroft/skills/core.py

implemented results in all my skills

https://github.com/JarbasAI/jarbas-core/tree/dev/mycroft/skills

clusterfudge:
How do the consumers of the results message disambiguate between result types? It appears all of your results messages have the type "results".

JarbasAI:
one of the results is skill_name , should be enough to id the source

clusterfudge:
So the point isn't to id the source, the point should be to identify the structure of the data in the result. We talked about this earlier up in the PR, but I'll summarize again here:

1) By tying result type to the skill name, you're unnecessarily restricting a skill to a single type of result. A weather skill, for example, could have both a "CurrentWeather" and "NDayForecast" result type, but this route would prevent that, and would result in breaking changes to any consumers of results when you finally got back to fixing it.

2) By tying result type to skill name, you 're unnecessarily bleeding/taking dependency on the implementation of the skill, when multiple skills could (and should!) implement the same message interface or format, without the consumer of results needing to care which one it's looking at. For example, there could be two implementations of a weather skill, OWMWeatherSkill and WeatherUndergroundWeatherSkill, each simply yielding "WeatherForcast" results. Consumers of results would not need to know which implementation was in place, as long as they could trust the format of WeatherForecast.

My impression is that this pull has been abandoned, and people shouldn't be building off of it.

joshymcd:
@clusterfudge - agreed. i think i need to have a think about a more robust solution to getting structured data from a skill. i'll close this PR and when i come up with a better solution i'll make anew one 

Josh

clusterfudge:
thanks @joshymcd , looking forward to the new work!

JarbasAI:
@clusterfudge CurrentWeather would be a key of the result, just by listening to results message you can check if that key is in it without caring which skill it came from, i dont get your point...

AIIX:
I'd just like to add my view from a GUI perspective, taking the Current Weather skill example that displays three different variables i.e. Current Weather, High and Low temps, Currently it is really difficult to capture these variables individually from a single web socket message where type = speak to display them within a custom GUI as I have to manually filter each string received from skill dialog to search for certain keywords in the string index to match a skill type (in this case weather) and then further filter this to get variables again by hacking away at the fixed index locations of each variable. This also limits the ability to capture data from a dynamic skill dialog.

Example in JS:

Current Temp Dialog:  
Right now, it's {{condition}} and {{temp_current}} degrees, for a high of {{temp_max}} and a low of {{temp_min}} in {{location}}.

Filtering Dialog of Each Skill to Match Skill Type and Filter for variables: 
// If state first matches string contents for keywords of the above dialog to recognize this is from current weather. 
if (websocket.msgtext.indexOf("Right") != -1 && websocket.msgtext.indexOf("now") != -1 && websocket.msgtext.indexOf("and") != -1 && websocket.msgtext.indexOf("degrees") != -1 && websocket.msgtext.indexOf("for") != -1 && websocket.msgtext.indexOf("a") != -1 && websocket.msgtext.indexOf("high") != -1 && websocket.msgtext.indexOf("low") != -1) {

var totalclimate = websocket.msgtext.match(/\d/g)
var hightempclimate = totalclimate.toString().substring(4,7) // <--- Finding Index for High Temp Value.
console.log(hightempclimate)
}

What i believe would really help would be the ability of getting the above data in separate websocket messages where "result=weatherskillresult type=hightemp data=50 unit=imperial", "result=weatherskillresult type=lowtemp data=30 unit=imperial"  and "result=weatherskillresult type=currenttemp data=50 unit=imperial"

clusterfudge:
@joshymcd I think the above speaks to what I was describing; having a strong format for each type of response, and not having to inspect each message (instead routing off of the message type) is easier for consumers, and in line with how messages are used in the rest of mycroft.

JarbasAI:
so instead of a general "results" message that we could check for the keys of "desired data"  without caring for source skill

results  {"skill_name":weatherskill, "min temp":x , "max temp":y, "forecast":w, unit":imperial}
results  {"skill_name":wikipediaskill, "link":x , "text":y, }

- parse every results message for key "forecast", if key forecast in message display in gui else ignore
- for failsafe parsing may be applied in dev side by checking skill_name, so if skill_name = troll (not in whitelist / in blacklist) ignore message

every skill would have a "skill name result type" message emitted with data for every individual result?

weatherskillresult  {"type":"min temp" , "value":x}
weatherskillresult  {"type":"max temp" , "value":x}
wikipediaskillresult  {"type":"link" , "value":x}
wikipediaskillresult  {"type":"text" , "value":x}

- on weatherskillresult do something with gui
- no parsing needed on listener side

not sure im understanding this correctly

JarbasAI:
i did the following:
- add_result sends a register_result message with result message type (not really needed for core, just for me)
- emit_results send a message with "key"_result, {data}

self.add_result("WeatherForecast", {weather_data}) 
self.add_result("Time",{time_data})
self.emit_results()

sends the messages:
WeatherForecast_result , weather_data
Time_result, time_data

is this satisfactory?

Issue #273, opened 2016-07-08
Wolfram problem with contractions


Wolfgange3311999:
`what's 2 + 2` gives something about foot seconds since it searches for `'s 2 + 2`. This relates to the verb now being `'s` for `what's`.

PR #283, opened 2016-07-14
Issues #273 - Add case for when verb is 's

ethanaward:
This adds code to handle the case for when the parsed question verb is `'s` so that it doesn't send `'s` to be parsed as `foot seconds`.

### DIFF
--- a/mycroft/skills/wolfram_alpha/__init__.py
+++ b/mycroft/skills/wolfram_alpha/__init__.py
@@ -91,7 +91,7 @@ class CerberusWolframAlphaClient(object):
         response = requests.get(url, headers=headers)
         if response.status_code == 401:
             raise CerberusAccessDenied()
-        logger.debug(response.content)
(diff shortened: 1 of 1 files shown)


mycroft/skills/wolfram_alpha/__init__.py
```diff
@@ -153,7 +153,11 @@ def handle_fallback(self, message):
             utt_word = parsed_question.get('QuestionWord')
             utt_verb = parsed_question.get('QuestionVerb')
             utt_query = parsed_question.get('Query')
+            if utt_verb == "'s":
+                utt_verb = 'is'
+                parsed_question['QuestionVerb'] = 'is'
             query = "%s %s %s" % (utt_word, utt_verb, utt_query)
+            logger.debug(query)
```
isaacnward: Please, remove this line.

isaacnward:
Okay, LGTM. Merging.

PR #284, opened 2016-07-14
Temporarily removing  from install-mimic.sh to avoid overwhelming bui…

aatchison:
…ld machines and failing out.


aatchison:
opsie

Issue #259, opened 2016-07-03
cleanup root directory


clusterfudge:
things have gotten cluttered with new scripts and config for those scripts.

ethanaward:
Addressed by #301.

PR #285, opened 2016-07-14
Issues #259 - Move scripts and config out of root directory

ethanaward:
This moves a few scripts and some config file into a separate `scripts` directory to clean up the root directory. It also modifies scripts to use the new locations.


jdorleans:
It seems all of them are dev related. If so, ship it!

PR #286, opened 2016-07-14
Removed  -j$CORES from install-mimic.sh

aatchison:
This is temporary relief for those with mimic builds failing.


ethanaward:
LGTM, merging

PR #289, opened 2016-07-15
Make Mimic interface send phonemes on message bus

forslund:
This is a take on providing information to allow mycroft to "lip-sync" as requested by "Matthew" on slack. It's mainly for his testing.

Using the -psdur flag mimic will write phonemes and their duration to
standard out. These are captured and sent over the message bus as a list
of phoneme-duration pairs.

It'll look like this:
```
2016-07-14 13:37:16,174 - Skills - DEBUG - {"message_type": "mycroft.tts", "context": null, "metadata": {"phonemes": [["pau", "0.152"], ["w", "0.244"], ["ay", "0.389"], ["l", "0.496"], ["d", "0.554"], ["ih", "0.593"], ["b", "0.670"], ["iy", "0.822"], ["s", "0.950"], ["t", "0.975"], ["s", "1.077"], ["ae", "1.289"], ["n", "1.349"], ["ax", "1.383"], ["m", "1.475"], ["ax", "1.521"], ["l", "1.656"], ["z", "1.743"], ["pau", "1.925"], ["ih", "2.003"], ["z", "2.087"], ["k", "2.214"], ["aa", "2.330"], ["n", "2.376"], ["ow", "2.519"], ["k", "2.624"], ["aa", "2.747"], ["iy", "2.895"], ["t", "2.977"], ["s", "3.170"], ["pau", "3.336"]]}}
```
"duration" is the mimic term for it but it's not really a duration, more like stop time for the phoneme.

Currently this uses a workaround to force mimic to linebuffer the stdout stream. Mimic is called via `stdbuf -oL` so please verify that the `stdbuf` command exists on the test system. I'll fix the buffering in mimic so next version won't require this hack.

Also, this is using simple pipes to communicate with mimic. It might be too slow and in that case I have to think of something new or get *pymimic* ready.

### DIFF
--- a/mycroft/client/speech/main.py
+++ b/mycroft/client/speech/main.py
@@ -62,7 +62,7 @@ def mute_and_speak(utterance):
     try:
         logger.info("Speak: " + utterance)
         loop.mute()
-        tts.execute(utterance)
+        tts.execute(utterance, client)
     finally:
         loop.unmute()
         mutex.release()
--- a/mycroft/client/text/cli.py
+++ b/mycroft/client/text/cli.py
@@ -36,10 +36,11 @@ def handle_speak(event):
     try:
         utterance = event.metadata.get('utterance')
         logger.info("Speak: " + utterance)
-        tts.execute(utterance)
+        tts.execute(utterance, client)
     finally:
         mutex.release()
         client.emit(Message("recognizer_loop:audio_output_end"))
+        logger.info("Speak complete")
 
 
 def connect():
--- a/mycroft/configuration/mycroft.ini
+++ b/mycroft/configuration/mycroft.ini
@@ -60,6 +60,7 @@ session_ttl_seconds = 180
 [tts]
 module = "mimic"
 mimic.voice = "ap"
+mimic.phonemes = "startstop"
 espeak.lang = "english-us"
 espeak.voice = "m1"
 
--- a/mycroft/tts/__init__.py
+++ b/mycroft/tts/__init__.py
@@ -31,7 +31,7 @@ class TTS(object):
     TTS abstract class to be implemented by all TTS engines.
 
     It aggregates the minimum required parameters and exposes
-    ``execute(sentence)`` function.
+    ``execute(sentence, client)`` function.
     """
 
     def __init__(self, lang, voice, filename='/tmp/tts.wav'):
@@ -41,7 +41,7 @@ class TTS(object):
         self.filename = filename
 
     @abc.abstractmethod
-    def execute(self, sentence):
+    def execute(self, sentence, client):
         pass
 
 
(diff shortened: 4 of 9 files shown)


aatchison:
This is very nice! We have needed something like this all along. Thanks.

Wolfgange3311999:
I just wanted to mention I've tested this out and it works great! With a small change the talk animation can actually start only when mimic actually begins to speak. My only concern would be that with especially long responses it will clutter up the message bus.

forslund:
Could we reduce the messagebus utilization if I convert the phoneme to a
viseme set suitable for mycroft before sending? (Could be configurable).

Or can the information be sent by other means than the message bus?

Den 22 jul 2016 10:07 em skrev "Matthew D. Scholefield" <
notifications@github.com>:

Wolfgange3311999:
@forslund We talked and decided the ability to know the true start and end of mimic's speaking is crucial. The actual annotated phonemes will hopefully be necessary in the future when when someone gets around to adding that support in the client. So do you think you could keep the `-psdur` flag in but only output a message when it receives the first thing from stdout? And then in the future we can decide about how to output the actual phonemes.

forslund:
Ok, this might be a bit of over engineering things...

Now `mimic.phonemes` can be set in *mycroft.ini* to select a couple of representation.
- `all`: Same behaviour as previously, all phonemes are sent.
- `startstop`: Will only send times for start of speech and stop of speech.
- `speaking`: Will send time info for speech and pauses (Basically pauses will be produced by ',' and '.' in strings sent to mimic.)
- `none`: Will not send phoneme information

aatchison:
@Wolfgange3311999 @forslund 
Do you think there is any chance that we can get this PR merged this week, including the enclosure changes? It would really help me out.

Wolfgange3311999:
@aatchison This PR looks merge ready to me. Once this is merged I can surely make a new PR for getting the mouth the sync up. It wouldn't even require enclosure changes; just changing where it sends the `audio_output_begin` and `end` messages.

aatchison:
Ok. LGTM. What say you, team?

forslund:
#334 will conflict with this one. Depending on merge order I'll update the appropriate PR.

forslund:
Do you want me to do any more work on this one to get it merged?

aatchison:
I'm not really sure what the sate is currently? I would really like to be able to merge this.

forslund:
I've retested against latest master and it's still working as intended. If you find any issues let me know and I'll take care of them.

forslund:
Should I close this now that another way has been chosen or do you want to have it updated to be mergable?

aatchison:
@penrods has been working on this. What do you think? It's one less file to write to on the file system.

forslund:
You know what, I'll Close this one and make a new PR with the actual changes required.

PR #292, opened 2016-07-15
Temp dict files now have generated names in /tmp/

Wolfgange3311999:
This prevent permission errors associated with storing things in the
/tmp/mycroft directory and with reusing files. When mycroft was first run as a different user this could cause the voice thread to fail when it attempts to create or open the already-existing folder or file. Now, instead of being put in something like `/tmp/mycroft/hey mycroft.dict` the generated dicts are placed into a file such as `/tmp/tmpxl6Ysh`.


aatchison:
lgtm

PR #294, opened 2016-07-15
Update test coverage to include loading vocab

ethanaward:
This updates the test coverage of `core.py` to include the vocab loading functions. I was going to also include tests for skill loading functions, but those have proven to be harder to write.


jdorleans:
Seems good to me. Merge it.

PR #296, opened 2016-07-18
Pyflakes - Fix unused imports

ethanaward:
This removes unused imports from several files in the repo, as shown by Pyflakes.


jdorleans:
LGTM

PR #298, opened 2016-07-20
Feature/remove build scripts

aatchison:
I am nearly ready to replace the core repo build scripts with an outside repository.


aatchison:
The external build process is working, merging.

PR #300, opened 2016-07-21
The volume knob now plays a noise for feedback

### LINKED ISSUE #299: Volume change should not be spoken every time
When changing the volume with hardware it should not speak the result but rather play a sound.

Wolfgange3311999:
This fixes #299 .

Before it would speak the response for every volume change and take a long time to finish speaking all the redundant volume changes. This is a simple fix to the problem and I think the right long term solution because IMO it fells more responsive and natural than having him respond with his voice.

However, when you ask him to change his volume that has remained unchanged as I feel the above does not apply.


aatchison:
Nice, reviewing.

aatchison:
We are going to need an entry in the mycroft-base-manifest for the new .wav file.

aatchison:
LGTM

PR #301, opened 2016-07-22
Issues #259 - Move scripts out of root directory

ethanaward:
Similar to #285, this cleans up the root directory by moving several scripts into a new scripts directory. #285 had to be reverted because of an accidental missing line.


Wolfgange3311999:
LGTM but I only ran `mycroft.sh` and `start.sh`. I did not test `dev_setup.sh` although I don't have any reason to believe it wouldn't work.

Issue #194, opened 2016-06-20
Wolfram should be more concise


Wolfgange3311999:
For instance, when asking "What is a power", it should read only the first definition rather than all 10.

ethanaward:
This is related to a larger issues with parsing Wolfram output as well, since Mycroft frequently gives back responses that don't sound very natural.

Wolfgange3311999:
I don't know if there's a unified solution to them all. However, would it be effective to read a minimum of of certain amount of characters and then stop on the next newline?

ethanaward:
Stopping on a newline would likely help somewhat, especially with the definitions.

ethanaward:
We'll likely want to have some kind of discussion about this to determine how we want Wolfram output formatted.

marksev1:
An idea would be too that one could then trigger to read further with: "Read on."

isaacnward:
Solved by #302.

PR #302, opened 2016-07-22
Issues #194 - Clean up Wolfram syntax

ethanaward:
This partially addresses #194. The WA skill will now no longer search again for things it thought you meant, and instead will just say a possible alternative. Also, I've created some regex to parse definitions so that it will say only one definition of a word rather than all of the definitions it can find, as this often takes far longer than needed.


Wolfgange3311999:
I think that `Did you mean` insinuates a response of which is not yet support. Could we change that to something like `Perhaps you meant`?

mycroft/skills/wolfram_alpha/__init__.py
```diff
@@ -91,6 +91,7 @@ def query(self, query):
         response = requests.get(url, headers=headers)
         if response.status_code == 401:
             raise CerberusAccessDenied()
+        # logger.debug(response.content)
```
Wolfgange3311999: Debug line accidentally added.

Wolfgange3311999:
LGTM.

isaacnward:
LGTM too, merging.

PR #306, opened 2016-07-27
Fix install-pygtk to use new version of cairo

ethanaward:
In the install-pygtk script, the installation of cairo was failing because it was using an outdated URL. I've updated the URL, and cairo now installs correctly.


isaacnward:
LGTM, merging.

PR #307, opened 2016-07-27
Replace outdated mustache library

ethanaward:
This replaces the outdated [mustache Python liibray](https://github.com/peterldowns/python-mustache), as it is no longer supported. It replaces it with [pystache](https://github.com/defunkt/pystache), which provides the exact same functionality but is still being maintained.


isaacnward:
LGTM, merging.

Issue #56, opened 2016-05-25
Mycroft should stop TTS


jdorleans:
Mycroft should be able to stop any TTS engine by voice or button pressing. This task depends on #55.

aatchison:
This issue is particularly encumbered by
https://github.com/MycroftAI/mycroft-core/issues/57

ryanleesipes:
We should get the button press working again.

jdorleans:
@ryanleesipes the current knowing problem we have with the button is related to the hardware. @aatchison thinks it might be caused by a wrong position of the cables that connects the encoder to the arduino. My unit has this problem and I'll try to fix it by switching the cable position during this weekend.

ryanleesipes:
Until you get that fixed, perhaps you could coordinate with @isaacnward to test it on a "good" unit.

ryanleesipes:
So close! It looks like @isaacnward might have a solution started this week. **fingers crossed**

ryanleesipes:
Looks like at this point, if I'm not mistaken, we need to build this into each of the existing skills (to kill Mimic), I guess we can do that via the "stop" function?

ryanleesipes:
Also @jdorleans - @isaacnward fixed our encoder issue.

isaacnward:
The encoder's working now, but there are still two things standing in the way of this issue getting fixed: the lack of stop functionality in existing skills (which I can work on if needed), and the slow response time described in MycroftAI/enclosure#7 (which is a pretty large problem, but I'll be prioritizing it this week).

ryanleesipes:
Alright, let's close this and get a different issue popped up to implement stop functionality into existing skills.

jdorleans:
Guys, this task is about dealing with TTS stop feature. It requires migrating the kill function from inside the core Enclosure to the TTS engines and then stopping any related TTS process properly. 
This task is not about fixing the hardware encoder issue nor stopping skills.

isaacnward:
Solved by #308.

PR #308, opened 2016-07-27
Issues #56 - Move TTS process killing to speech client

ethanaward:
This removes `kill(['mimic'])` from the enclosure client and moves it to the speech client. It also changes it to use the configured TTS client rather than assuming the user is using mimic.

Note that saying `Mycroft stop` still will not work most of the time, as the speech client mutes the microphone when saying something. One possible fix would be to add a config option that the user can set if they are using headphones or a system with echo cancellation, as the main reason that the microphone is muted is to prevent Mycroft from hearing himself while speaking.


isaacnward:
LGTM, merging

PR #312, opened 2016-07-29
Added white noise and dialog to hardware test

Wolfgange3311999:
This adds dialog when uploading a hex, as well.

In order for Mycroft to speak, another thread must be executed to run the client since setup now uses it.


Wolfgange3311999:
LGTM.

PR #313, opened 2016-07-29
Refactored VolumeSkill to use configurable min and max

### LINKED ISSUE #303: Volume range should be configurable
The minimum and maximum volume should be configurable so that the max isn't excessively loud on the unit and the minimum is something more reasonable like 20.

Wolfgange3311999:
This fixes #303 by making the maximum and minimum configurable.


aatchison:
LGTM

aatchison:
I meant to say, this needs a refactor ;)

PR #314, opened 2016-07-29
Issue/dont compile pocketsphinx

aatchison:
I can't understand what is wrong with using pip to install pocketsphinx with python bindings. Unlike the debian packages python-pocketsphinx, the pip method installes a newer version '0.1.0'. I have run dev_setup and my packaging scripts to create debs by adding pocketsphinx to `requirements.txt` and it seems to work just fine. 

it would really help the packaging efforts as well as the development setup if we didn't have to compile  pocketsphinx each time.

### DIFF
--- a/build_host_setup.sh
+++ b/build_host_setup.sh
@@ -20,7 +20,8 @@ sudo apt-get install -y \
     portaudio19-dev \
     mpg123 \
     screen \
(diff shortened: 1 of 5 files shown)


aatchison:
LGTM, revert if need be.

PR #324, opened 2016-08-12
Added support for power control through unit menu mode

isaacnward:
This PR allows users to reboot or shut down the Raspberry Pi on the hardware unit through its menu mode.

It relies on MycroftAI/enclosure#55, which implements menu mode itself.


isaacnward:
LGTM, merging

PR #325, opened 2016-08-15
Feature/break tts into smaller chunks

aatchison:
This is a temporary solution to break utterances into smaller chunks. When pymimic arrives, we should be able to handle this better.


aatchison:
This helps a bit to break up the TTS. Alan's voice still takes a long time to load, even for just a sentence.

PR #326, opened 2016-08-16
Feature/rename build host setup

aatchison:
After merging the Fedora build script, I thought I would be good to change the name of the default one to indicate that it is for Debian. Perhaps we should look into separating the build host setup into another sub directory


aatchison:
LGTM

PR #327, opened 2016-08-16
Add checks for wifi connection (Fixes #321)

### LINKED ISSUE #321: Mycroft should indicate when there is no internet connection.,
Since an internet connection is required, we need to know if there is one at any given time.

ethanaward:
This adds checks for an internet connection when the wakeword is detected as well as when the speech client is started to address #321.


aatchison:
Blind merge with a flaming fireball vengence

PR #328, opened 2016-08-16
Add configurable volume levels (Fixes #303)

### LINKED ISSUE #303: Volume range should be configurable

ethanaward:
This is a rebased version of #313 to fix merge conflicts.


aatchison:
Super premature auto fast hand ultimate risk-taking behavior merge

PR #330, opened 2016-08-17
Try setting PATH variable before sudo commands...

aatchison:
in enclosure.py.
This seems to work, I need to build a package to make sure.


aatchison:
I'm thinking that I might need to export to child processes instead, but I'm not sure just yet.

PR #335, opened 2016-08-18
Eye shutdown

isaacnward:
Added functionality for the eyes to spin while the unit shuts down, then turn off when it finishes. This corresponds with an enclosure PR.


aatchison:
lgtm

PR #338, opened 2016-08-18
Factory reset

isaacnward:
Added a "factory reset" (deleting identity.json, then rebooting), which is accessed through the unit menu.


aatchison:
lgtm

PR #341, opened 2016-08-18
Fix wifi indication in speech client

ethanaward:
The check for internet connection in the speech client wasn't working, as it started too early for it to be useful.


aatchison:
lgtm

PR #357, opened 2016-09-05
Added viseme support for TTS, allowing enclosure to display visemes

penrods:
This PR allows pulling the phonemes from Mimic and converting those to visemes, then sending the visemes to the Enclosure for a visualization.  This is available for all TTS clients, but currently only Mimic performs it.


aatchison:
Lgtm

PR #365, opened 2016-09-22
Feature button behavior

penrods:
Several changes related to button pressing on the Mycroft unit:
- Pressing the button when it isn't listening starts it listening
- Pressing the button when listening will stop the listen
- Added a mycroft.util.signal() mechanism for out-of-thread communication
- Pressing the button now creates an "buttonPress" signal from the Enclosure
- The viseme playback and aplay check for the 'buttonPress' signal to abort
- Removed "Sorry I didn't catch that", irritating during false activations


aatchison:
LGTM

mycroft/util/__init__.py
```diff
@@ -109,5 +111,21 @@ def connected(host="8.8.8.8", port=53, timeout=3):
             return False
 
 
+def create_signal(signalName):
```
clusterfudge: Sorry for the 9-months late review. I have concerns about this on picroft (and mark 1s) as it will unnecessarily abuse the file system. SD cards have a finite number of writes, so I had been purposely avoiding things like this. 

A better practice would probably be something like emitter.once , but requires that you put more thought into an order of operations of sorts (i.e. I must set my listener before the event occurs, likely by setting the listener and then beginning a chain of events). An example from the pyeventemitter docs:

```
import event_emitter as events

em = events.EventEmitter()

def hello(who):
    print('Hello {}'.format(who))

em.once('hello', hello)
em.emit('hello', who='World')  # prints Hello World
em.emit('hello', who='World')  # nothing happens
```

forslund: Will that work across processes?

An alternative might be to reimplement the current signaling system using a dict and the messagebus.

A module along the lines of:
```python
from mycroft.messagebus.client.ws import WebSocketClient
from mycroft.messagebus.message import Message
from time import time

_signal_dict = {}

ws = WebSocketClient()

def _set_signal_handler(message):
    sig = message.data['name']
    _signal_dict[sig] = time()

def _clear_signal_handler(message):
    sig = message.data['name']
    _signal_dict[sig] = False


def create_signal(signal_name):
    ws.emit(Message('mycroft.signal.set', data={'name': signal_name}))

def check_for_signal(signal_name, sec_lifetime=0):
    signal = _signal_dict[sig].get(signal_name, False)
    if signal == False:
        return False
    elif sec_lifetime != -1 and signal + sec_lifetime < time.time():
        ws.emit(Message('mycroft.signal.clear', data={'name': signal_name}))
    return True

ws.on('mycroft.signal.set', _set_signal_handler)
ws.on('mycroft.signal.clear', _clear_signal_handler)
```

would be compatible with the current signaling system.

clusterfudge: Sorry I wasn't more explicit. The existing messagebus clients [all use an instance of event emitter](https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/messagebus/client/ws.py#L46) under the covers, and it should be pretty easy to expose the 'once' functionality.

I'm curious, why do you think the signal state needs to be persisted like this? Could you instead have something like this, assuming there's no naming collisions?
```
ws.emit(Message(signal_name))
```

forslund: That is definitely possible, many places will however need to be looked over and maybe redesigned to work nicely. I mainly suggested the above as a quick drop in replacement keeping the same api and functionality without touching the disk.

PR #370, opened 2016-10-06
Wifi setup usability

penrods:
Enhancing the wifi setup system:
* When no internet is detected, Mycroft now instructs user
  to connect cable or tells how to start wifi setup
* Wifi setup talks user through connection process
* Setup will auto-shutdown after 5 minutes
* Completion screen now goes to cerberus.mycroft.ai to
  allow registration immediately
  (TODO: custom url/landing page to that will already have the pairing code)


mycroft/client/wifisetup/main.py
```diff
@@ -135,38 +208,145 @@ def save(self):
...
+                            else:
+                                bConnected = True
+                if not bConnected:
+                    cARPFailures += 1
+                    if cARPFailures > 3:
+                        self.connectionPrompt("Connection lost,")
+                        bHasConnected = False
+                else:
+                    cARPFailures = 0
+            sleep(5)  # wait a bit before continuing
```
jdorleans: Unnecessary comment

mycroft/client/wifisetup/main.py
```diff
@@ -135,38 +208,145 @@ def save(self):
...
+                    sleep(5)
+
+                # now look at the hardware that has responded, if no entry
+                # shows up on our access point after 3*10=30 seconds, the user
+                # has disconnected
+                bConnected = False
+                res = cli_no_output('/usr/sbin/arp', '-n')
+                out = str(res.get("stdout"))
+                if out:
+                    # Parse output, skipping header
```
jdorleans: Make this parsing in a new method.

mycroft/client/wifisetup/main.py
```diff
@@ -218,12 +402,20 @@ def connect(self, event=None):
                 connected = self.get_connected(ssid)
                 if connected:
                     wpa(self.iface, 'save_config')
-                    # ConfigurationManager.set(self.NAME, 'setup', False, True)
 
             self.client.emit(Message("mycroft.wifi.connected",
                                      {'connected': connected}))
             LOG.info("Connection status for %s = %s" % (ssid, connected))
 
+            if connected:
```
jdorleans: This block is weird. This feedback should either be moved to the top connected check or the inside one.

mycroft/client/wifisetup/main.py
```diff
@@ -218,12 +402,20 @@ def connect(self, event=None):
...
+                if True:
```
jdorleans: What's that?

mycroft/client/wifisetup/main.py
```diff
@@ -218,12 +402,20 @@ def connect(self, event=None):
...
+                    self.client.emit(Message("speak", metadata={
+                        'utterance': "Thank you, I'm now connected to the "
+                                     "internet and ready for use"}))
+                else:
```
jdorleans: This If True / else does not make sense

mycroft/client/wifisetup/main.py
```diff
@@ -38,6 +42,29 @@
...
+        if type(i) in (list, tuple, dict, set):
+            var_dump(i, prefix)
+        else:
+            if isinstance(var, dict):
+                print(prefix, i, ': (', var[i].__class__.__name__, ')', var[i])
+            else:
+                print(prefix, '(', i.__class__.__name__, ')', i)
+
+
+def cli_no_output(*args):
```
penrods: That was an explicit dump being used to debug an odd network name we were seeing around the office.  I've removed it.

mycroft/client/wifisetup/main.py
```diff
@@ -61,17 +88,62 @@ def sysctrl(*args):
     return cli('systemctl', *args)
 
 
-class WebServer(Thread):
-    DIR = dirname(realpath(__file__))
+class MycroftHTTPRequestHandler(SimpleHTTPRequestHandler):
```
penrods: I had it commented out until I was able to test it.  With an iOS device, this makes the portal appear automatically when the device connects to the MYCROFT network.

mycroft/client/wifisetup/main.py
```diff
@@ -135,38 +208,145 @@ def save(self):
...
+        # Fire up the MYCROFT access point for the user to connect to
+        # with a phone or computer.
         LOG.info("Starting access point...")
-        self.client.emit(Message("speak", metadata={
-            'utterance': "Initializing wireless setup mode."}))
+
+        # zap any existing leases file because:
+        # a) we don't care about previous leases
+        # b) we will monitor it for connections
+        # os.remove('/var/lib/misc/dnsmasq.leases')
```
penrods: Yeah, I'm pretty sure we won't need this.  Zapped.

mycroft/client/wifisetup/main.py
```diff
@@ -135,38 +208,145 @@ def save(self):
...
+
+        # Fire up our access point
         self.ap.up()
         if not self.server:
+            LOG.info("Creating web server...")
```
penrods: Those were needed to debug a situation where the self.server.start() was not firing.  Given how rarely this process will be invoked by a user, having extra logs to debug situations where it isn't working is worth it.  This doesn't run constantly and the logs aren't going to overflow from a couple extra messages here.

mycroft/client/wifisetup/main.py
penrods: This is a placeholder for code that will determine if pairing of the unit has already occurred.  When it hasn't, the pairing code will be displayed at this point to continue walking the user through setting up their device.

mycroft/client/wifisetup/main.py
```diff
@@ -61,17 +88,62 @@ def sysctrl(*args):
...
+            LOG.info(tb)
+            return False
 
+
+class WebServer(Thread):
     def __init__(self, host, port):
         super(WebServer, self).__init__()
         self.daemon = True
+        LOG.info("Creating TCPServer...")
+        # self.server = TCPServer((host, port), MycroftHTTPRequestHandler)
```
penrods: There was some stuff left behind while this was still in development.  It wasn't being kept around for legacy reasons, it was potentially going to be brought into play.  (In fact, MycroftHTTPRequestHandler is being used in my latest commits to enable automatic display of the portal on an iOS device.)

mycroft/client/wifisetup/main.py
penrods: The state of 'connected' changes in the body of the else, this code can't be moved to the top check.  This catches when you are connected after both the "then" and the "else".

mycroft/client/wifisetup/main.py
```diff
@@ -135,38 +208,145 @@ def save(self):
...
         self.client.on('mycroft.wifi.scan', self.scan)
         self.client.on('mycroft.wifi.connect', self.connect)
 
-    def first_setup(self):
-        if str2bool(self.config.get('setup')):
-            self.start()
-
     def start(self, event=None):
+        # Fire up the MYCROFT access point for the user to connect to
+        # with a phone or computer.
```
penrods: Not a class-level comment, but I made it into a doc string.

mycroft/client/wifisetup/main.py
penrods: Since I don't see where I can respond to the rebase commend above, I'll do it here:

This code needs to be applied to Master as soon as possible.  I'm afraid we cannot wait until we are ready to release the Dev branch on everyone because:
1) This process is particularly needed now after the package change that wiped out the wpa_supplicant file.
2) This functionality makes doing demos WAY easier
3) This process is the most critical thing that has to be in place on the initial device image, and the day we need to make that is rapidly approaching.  Without it the unit is useless to users who don't use Ethernet cables.  Everything else can be upgraded once the unit is on the network.

penrods: I respectfully disagree about the comments.  The mycroft-core code has virtually no comments besides the boilerplate copyright at the top, and that is not a good thing.  Comments are important to convey intent so subsequent programmers can tell what the code is *supposed* to be doing when there is a bug and it doesn't actually do something correctly.  Check out the source code for Linux or any other major project sometime, it is chock full of comments.

penrods: As for the spoken sentences, the phrases are being specifically designed to walk a novice user through the process.  I may refine them slightly beyond this, but this is a rare-enough event that I have no significant concern that expert users will be irritated by the wordiness.  Before these changes there was no way a user would have been able to perform the wifi setup without reading the source code.

penrods: P.S.  I hope none of the responses here seem snippy.  I was writing them around 5am.  :)

aatchison:
LGTM.
Sorry Jonathan, after we push this into the public release, we can focus on refinement in the dev branch.

PR #374, opened 2016-10-13
Issue 372

### LINKED ISSUE #372: dev_setup.sh ignores WORKON_HOME when VIRTUALENV_ROOT is unset
Working with a shell configured to use virtualenv via virtualenvwrapper may result in the creation of a new virtualenv root directory.

For example if WORKON_HOME is set to $HOME/projects, executing dev_setup.sh will create the $HOME/.virtualenv and use it as the location for the "mycroft" virtual environment.

Since virtualenv's documentation does not describe the use of the VIRTUALENV_ROOT variable, I would like to suggest evaluating if WORKON_HOME is not empty before using VIRTUALENV_ROOT, in which case its value becomes used over VIRTUALENV_ROOT's default.

ncouture:
- favor using the WORKON_HOME environment variable over
        VIRTUALENV_ROOT
      - closes #372


aatchison:
Hey, thank you. Would you mind reverting the requirements.txt? You can keep the new gtts version.
I'm sure that all the of requirements could be updated, but as we tested them, we don't want to introduce too many variables at once.

aatchison:
Oh, I see that you added them again...
Did you only chagne the gtts version?

ncouture:
I wasn't aware that PRs were branch-specific and not commit-specific.

I'll submit a new PR.

PR #376, opened 2016-10-14
#372 - dev_setup.sh ignores WORKON_HOME when VIRTUALENV_ROOT is unset

### LINKED ISSUE #372: dev_setup.sh ignores WORKON_HOME when VIRTUALENV_ROOT is unset

ncouture:
```
  - favor using the WORKON_HOME environment variable over
    VIRTUALENV_ROOT
  - closes #372 
```

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -33,7 +33,12 @@ if [ $(id -u) -eq 0 ]; then
 fi
 
 TOP=$(cd $(dirname $0) && pwd -L)
-VIRTUALENV_ROOT=${VIRTUALENV_ROOT:-"${HOME}/.virtualenvs/mycroft"}
+
+if [ -z "$WORKON_HOME" ]; then
+    VIRTUALENV_ROOT=${VIRTUALENV_ROOT:-"${HOME}/.virtualenvs/mycroft"}
+else
+    VIRTUALENV_ROOT="$WORKON_HOME"
+fi
 
 # create virtualenv, consistent with virtualenv-wrapper conventions
 if [ ! -d ${VIRTUALENV_ROOT} ]; then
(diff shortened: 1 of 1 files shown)


ncouture:
These changes are included in PR #378.

dev_setup.sh
```diff
@@ -33,7 +33,12 @@ if [ $(id -u) -eq 0 ]; then
 fi
 
 TOP=$(cd $(dirname $0) && pwd -L)
-VIRTUALENV_ROOT=${VIRTUALENV_ROOT:-"${HOME}/.virtualenvs/mycroft"}
+
+if [ -z "$WORKON_HOME" ]; then
+    VIRTUALENV_ROOT=${VIRTUALENV_ROOT:-"${HOME}/.virtualenvs/mycroft"}
+else
+    VIRTUALENV_ROOT="$WORKON_HOME"
```
jdorleans: It appears to me we should concatenate WORKON_HOME + /mycroft

`VIRTUALENV_ROOT="$WORKON_HOME/mycroft"`

ncouture:
@jdorleans see last commit

aatchison:
We have made a major update to master and plan on getting to this, thanks!

Issue #380, opened 2016-10-18
Implement refresh token


augustnmonteiro:
The new API sends an access token that expires after 24 hours and a refresh token is used to update this token. This task is intended to ensure all requests are authorized with a valid access token once user has paired a device. To do so, we have to check if the access token is still valid before performing a request. If not, a new access token is requested to the server using the refresh token and, as a result, updating the mycroft identity (login session).

PR #381, opened 2016-10-18
Issues 380 - Implement refresh token

augustnmonteiro:
Issues #380 
- Checking if access token is up to date before API requests
- When access token is expired, update it using refresh token

### DIFF
--- a/mycroft/api/__init__.py
+++ b/mycroft/api/__init__.py
@@ -1,5 +1,6 @@
 import requests
 from requests import HTTPError
+import time
 
 from mycroft.configuration import ConfigurationManager
 from mycroft.identity import IdentityManager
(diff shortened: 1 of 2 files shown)


mycroft/api/__init__.py
```diff
@@ -26,6 +37,11 @@ def request(self, params):
                                     data=params.get("data"), json=json)
         return self.get_response(response)
 
+    def request(self, params):
+        self.check_token()
+        params["path"] = self.build_path(params)
```
jdorleans: Move the initialization to inside the build process and return the final result instead

mycroft/api/__init__.py
```diff
@@ -16,7 +17,17 @@ def __init__(self, path):
         self.version = config_server.get("version")
         self.identity = IdentityManager.get()
 
-    def request(self, params):
+    def check_token(self):
+        if (self.identity.refresh and self.identity.expires_at <= time.time()):
```
jdorleans: We have to load the identity to ensure all processes will have the latest state of it. This solution will be temporary until we fix the way we manage the identity and configuration.

PR #387, opened 2016-10-24
Fix NPR News Skill by regexing https to http

aatchison:
This is PR implements the same changes I made to master with https://github.com/MycroftAI/mycroft-core/tree/bugfix/issues-385


jdorleans:
We did this in #397 because mpg123 is not capable of playing https. So we have to enforce http.

PR #388, opened 2016-11-04
Bugfix/issues 343

penrods:
Correct several issues that broke Mycroft on a generic Raspberry Pi:
* Removed a direct write to the serial port in the voice loop, now using the EnclosureAPI
* Added a system_reset() to the EnclosureAPI.  On the Mycroft Mark 1 this now resets both the mouth and eyes.
* The enclosure.py auto-detects if running on a Mark 1 or a generic Pi.  It kills itself if not on a Mark 1 (which prevents writing out the serial port connected to who-knows-what)

### DIFF
--- a/mycroft/client/enclosure/enclosure.py
+++ b/mycroft/client/enclosure/enclosure.py
@@ -191,6 +191,10 @@ class Enclosure:
     """
     Serves as a communication interface between Arduino and Mycroft Core.
 
+    This interface is currently designed to shut itself down when run on
+    a Raspberry Pi that is not connected to a Mycroft Mark 1 board on
+    the serial port.  So it can safely be started on any Raspberry Pi.
+
     ``Enclosure`` initializes and aggregates all enclosures implementation.
 
     E.g. ``EnclosureEyes``, ``EnclosureMouth`` and ``EnclosureArduino``
@@ -202,6 +206,7 @@ class Enclosure:
     """
 
     def __init__(self):
+        resetSelf = False
         try:
             self.config = ConfigurationManager.get().get("enclosure")
 
@@ -217,8 +222,12 @@ class Enclosure:
                     # Mycroft Mark 1
                     platform = self.detect_platform()
 
-                ConfigurationManager.set('enclosure', 'platform',
-                                         platform)
+                ConfigurationManager.set('enclosure', 'platform', platform)
+
+                # After a platform detection, the system is usually
+                # already active, so the message from the loop
+                # has already gone by on the messagebus.  So self reset.
+                resetSelf = True
         except Exception as e:
             self.disconnect()
             raise Exception("Exception: Unable to determine platform\n" +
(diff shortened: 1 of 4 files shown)


mycroft/client/enclosure/eyes.py
```diff
@@ -92,6 +94,13 @@ def volume(self, event=None):
     def reset(self, event=None):
         self.writer.write("eyes.reset")
 
+    def system_reset(self, event=None):
```
jdorleans: @penrods we have no guarantee eyes and mouth will be reset at this point because system.reset may restart arduino first. Then, we probably should move all logic regarding the reset of mycroft components to inside arduino and reset them before arduino itself.

mycroft/client/enclosure/enclosure.py
```diff
@@ -202,16 +202,91 @@ class Enclosure:
...
         self.eyes = EnclosureEyes(self.client, self.writer)
         self.mouth = EnclosureMouth(self.client, self.writer)
         self.system = EnclosureArduino(self.client, self.writer)
+
+        # TODO: Remove EnclosureWeather once the Skill can send images
+        #       directly to the EnclosureAPI.
         self.weather = EnclosureWeather(self.client, self.writer)
         self.__register_events()
 
+    def detect_platform(self):
```
jdorleans: We should use the reader and writer to do this sort of negotiation. Also, we can just get enclosure platform as a ping operation. So, python sends "system.platform", then arduino answers it as its current platform == "Mark 1" for instance.
Then, python will read it and confirm. I'll right the flow we should do instead of all this.

mycroft/client/enclosure/enclosure.py
```diff
@@ -202,16 +202,91 @@ class Enclosure:
     """
 
     def __init__(self):
-        self.__init_serial()
+        try:
```
jdorleans: This logic together with the platform detection method should use our enclosure reader and writer instead. I'll write the flow in the review comments.

jdorleans:
Ultimately my suggestion would be to:

In Enclosure class
```
def __init__(self):
        self.config = ConfigurationManager.get().get("enclosure")
        self.__init_serial()
        self.client = WebsocketClient()
        self.reader = EnclosureReader(self.serial, self.client)
        self.writer = EnclosureWriter(self.serial, self.client)
        self.client.on('enclosure.start', self.start)
        self.detect_platform()

    def detect_platform(self):
        platform = self.config.get("platform")

        if not platform:
            trials = 5

            while trials > 0:
                self.serial.flush()
                self.writer.write("arduino.started")
                time.sleep(2)
            self.stop()
        return platform


    def start(self):
        self.eyes = EnclosureEyes(self.client, self.writer)
        self.mouth = EnclosureMouth(self.client, self.writer)
        self.system = EnclosureArduino(self.client, self.writer)
        self.weather = EnclosureWeather(self.client, self.writer)
        self.__register_events()
        self.update()
        self.test()
```


In EnclosureReader you add this line to the process method:
```
        if "Command: arduino.started" in data:
            self.client.emit(Message("enclosure.start"))
 ```

penrods:
Finished changing EnclosureAPI.system_reset() and "enclosure.system.reset" to simply EnclosureAPI.reset() and "enclosure.reset".  Tested all changes under Mycroft Arduino and generic Raspberry Pi (no serial connection) environments.  Confirmed that there is garbage in the middle of a Mycroft Mark 1 unit's serial stream even immediately after a self.serial.flushInput()/flushOutput() -- presumably due to electrical line noise on the serial cable while the system is starting up.

PR #390, opened 2016-12-04
Fix arch build.

akegalj:
python2-virtualenvwrapper -> python-virtualenvwrapper
It is in Community repo:
https://www.archlinux.org/packages/?q=python-virtualenvwrapper


aatchison:
Thanks

PR #400, opened 2016-12-15
Update LICENSE.md

joshua-montgomery:
Proper license for this repo at this time.  Was orignially GPL which is too restrictive.


aatchison:
We need to adjust licence headers as well. Reverting.

PR #404, opened 2016-12-16
Issues 403 -- Changed License.md to GNU Lesser Public liscense; Replaced GPL v3 in …

aatchison:
…file headers

This is in reference to this reverted PR:
https://github.com/MycroftAI/mycroft-core/pull/400

PR #405, opened 2016-12-16
issues 403 -- Changing to GPL Lesser Public License

aatchison:
This is a replication of this PR for the master branch.
https://github.com/MycroftAI/mycroft-core/pull/404

@joshua-montgomery @penrods @jdorleans, please review


aatchison:
Looks like travis is halted on this one. Seems to be some pep8 issues

penrods:
For mycroft-core, GPLv3 is actually fine.  We just need to clean up the pieces that are the "Skill API" and add the GPLv3 with Classpath Exception to them.  That will retain the GPL on all of the core Mycroft technology (so somebody who creates a new device and that improves the core will have to give that part back to the community) but allow people to build their own Skills under whatever license they'd like -- free, GPL, Apache, MIT, even proprietary.

aatchison:
This seems no longer relevant.

PR #422, opened 2016-12-20
Update skills system to load from subfolders

ethanaward:
This updates the skill loading to be able to look one subfolder deep for skills. It also adds back `/opt/mycroft/third_party` as a possible place to look for skills for backwards compatibility purposes.

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -37,7 +37,7 @@ __author__ = 'seanfitz'
 PRIMARY_SKILLS = ['intent', 'wake']
 BLACKLISTED_SKILLS = ["send_sms", "media"]
 SKILLS_BASEDIR = dirname(__file__)
-THIRD_PARTY_SKILLS_DIR = "/opt/mycroft/skills"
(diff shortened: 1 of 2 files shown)


jdorleans:
The path should follow our standard: `/opt/mycroft/skills`

mycroft/skills/core.py
```diff
@@ -117,6 +117,14 @@ def get_skills(skills_folder):
     possible_skills = os.listdir(skills_folder)
     for i in possible_skills:
         location = join(skills_folder, i)
+        if (isdir(location) and
```
jdorleans: Remember we don't use parantheses in python

ethanaward:
Right, `/opt/mycroft/third_party` is only for backwards compatibility.

mycroft/skills/core.py
ethanaward: The parentheses are needed for pep8, as it gives an error without them.

aatchison:
Are we GTG on this?

PR #433, opened 2016-12-25
Audio service

forslund:
In late November @penrods led a discussion about how an audio service in Mycroft could work. This is a simple implementation of what was discussed. This is a quick implementation of the system as I understood it and I think this can be used to continue the discussion. Ideas and suggestions on improvements are very welcome.

The PR adds support for commands like

"Hey Mycroft, play the news in the livingroom"

# Backends

The current system includes support for the following "backends"

- Mpg123, very limited implementation
- VLC
- Mopidy

"backends" currently in the works:

- Chromecast
- Kodi

Backends are quite easy to add and an existing baseclass can be extended to add new ones.

# AudioService
Skill creators will use an audio service interface to communicate with the audio service / playback control. the AudioService interface uses the standard message bus.

```python
[...]
  as = AudioService()
[...]
  as.play(file_or_uri) # handles both lists and single uris
```

See the npr_news skill for basic usage.

The AudioService currently only support play and track_info requests but this will probably be expanded for increased usefulness.

# Playback control skill

Features:
* backend/location selection from utterance ("...in kitchen" etc.)
* lowering volume while Mycroft speaks
* playback control (Pausing/Resuming, skipping, Stopping)

The backend/location selection is currently done using basic string parsing and should probably be done using adapt instead.

This skill removes much of the basic functions that needs to be implemented for an audio playback skill and the skill creator can focus on the trickier parts.

Currently most of the audio logic is implemented in the playback control skill and should maybe be moved but this was a good starting point.

# Configuration

A Configuration the audio services:
```json
{
  "Audio": {
    "backends": {
      "local": {
        "type": "mpg123",
        "active": true
      },
      "kitchen": {
        "type": "mopidy",
        "url": "http://localhost:6680",
        "active": true
      },
      "livingroom": {
        "type": "mopidy",
        "url": "http://192.168.10.232:6680",
        "active": false
      },
      "vlc": {
        "type": "vlc",
        "active": true
      }
    },
    "default-backend": "local"
  }
}
```

If you try it I suggest you install vlc and pyvlc and use the vlc backend since mpg123 only has basic functionallity.

I've tested this mainly with the mopidy skill and the npr_news skill.

# Todo

- more backends, Kodi and chromecast is a minimum
- possibly separate the backends from the playback control skill
- populate the AudioService interface with more functions so skill creators can do more interesting things with the audio (A music quiz-skill should be able to skip ahead in the playlist when a correct answer is provided)
- More configuration options
- Maybe some way to send credentials from a skill to the backend to access some music services
- Better utterance parsing when starting playback, along with more keywords (like half volume)


forslund:
Major refactoring done after discussion with @ethanaward. The audio backends now runs separately from the skills. The loading of backends is now done using a method similar to how the skills work and should be easy to extend to include thirdparty backends.

forslund:
Skills using the audio service interface
NPR news:
https://github.com/forslund/skill-npr-news/tree/audio-service
Ethan's demo skill:
https://github.com/forslund/demo_skill/tree/audio_service/
`play all using vlc` to enque all tracks and play using vlc for example

most important, the playback control skill:
https://github.com/forslund/playback_control_skill

The playback control skills implement the common commands for skills using the audio service. like skipping tracks, pausing and resuming. The idea is that the skill creator only needs to implement the finding of audio files and the command that starts playback. (The other commands are of course available to the skill creator if they wish to use them)

forslund:
Finally rebased, *phew*

augustnmonteiro:
Let's rebase and wait finish the new release to merge it

penrods:
Let's get this merged after the next (July 6th) release.

augustnmonteiro:
@forslund can we merge this one?

forslund:
Fixed the conflict it should be good to merge now.

penrods:
So why isn't this merged yet!  Reviewed by Augusto, merge away!

augustnmonteiro:
@penrods I reviewed before the last release and didn't merged to that release and forgot to merge after the release is done, Sorry @forslund

PR #439, opened 2016-12-30
Add pymimic tts backend

forslund:
The development of pymimic has been moving along and is pretty functional. It currently relies on the development branch of mimic so it's not mergable yet. Currently I'd like to get some feedback on the pymimic backend for mycroft if there's any big mistakes or if there is something you'd like to have implemented.

pymimic wraps the mimic libraries and let's functions be called directly from python. The big structs are wrapped in python classes making it (in my opinion) quite easy to work with.

The main advantages over standard mimic:
- Loads voice from file once at startup, hence voices doesn't need to be compiled into mimic for speed
- No need to first write to file to get phonemes and then play the file
- Future releases will have possibility to abort playback in a neat way

Installation:
Install dev branch of mimic

```
  ./scrips/install-mimic-dev.sh
```

Install pymimic

```
  pip install git+git://github.com/forslund/pymimic.git@0.2.1
```

Edit mycroft.conf to something like

```json 
    "tts": {
    "module": "pymimic",
    "pymimic": {
      "voice": "/home/ake/projects/python/mycroft-core/mimic/voices/cmu_us_awb.flitevox"
    }
  }

```

This is very much a work in progress but is working decently. As soon as I can figure out the intended way to use PyAudio in a non-blocking way I will have visimes working so it can replace 100% of original mimic backend's functionality.

Any feedback is welcome.


augustnmonteiro:
That's awesome!

forslund:
Should probably rebase this...

forslund:
This has been rebased and updated to use the same caching method as normal mimic

labiadh:
@forslund the attached files contain the integration of pyaudio within pymimic :
- utterance split is moved to tts side
- pymimic splits the utterance into sentences, converts each sentence into speech, pushes the synthesized wave to pyaudio via a queue
- pyaudio plays back the received wave, using stream.write(), then notifies pymimic; runing as a thread
- in parallel pymimic continues converting and pushing speech waves, then wait until all sentences are played back

[playback.patch.txt](https://github.com/MycroftAI/mycroft-core/files/1044698/playback.patch.txt)
[pymimic.patch.txt](https://github.com/MycroftAI/mycroft-core/files/1044699/pymimic.patch.txt)

zeehio:
I believe the `VISIMES` variable should be localized according to the language used, as the phoneset will be different for each language.

penrods:
I believe the phoneme -> viseme translation is language independent as it is more a question of human physiology to produce the sound of the phoneme, not the language being spoken.  However there will be a few additional phonemes that need to be added to the map since English doesn't include all phonemes that humans can produce.

Correct me if I'm wrong on my assumption that the various mimic language implementations all make use of the same phonemic alphabet.

zeehio:
Each mimic lexicon defines its own set of phonemes, so I did not think that they were shared. You are right, all the supported languages use the same [SAMPA](https://en.m.wikipedia.org/wiki/Speech_Assessment_Methods_Phonetic_Alphabet) alphabet (each language uses a subset of SAMPA). It's nice to know they are compatible, I will document that to make sure the compatibility is preserved. Thanks!

PR #378, opened 2016-10-14
Bugfix/gtts version update speech impl fix

ncouture:
Adds newer versions of dependencies in requirements.txt and modifies Google TTS implementation to address issues listed in #377.

This should close #377.

PR #454, opened 2017-01-15
gTTS version bump

forslund:
This is a rebase of @ncouture's PR #378 against *dev* and fixes the `GoogleTTS` backend.

A couple of minor changes to make the TTS work reliably was made, such as waiting for playback completion and making sure the a valid `lang` value is set.


forslund:
There, rebased and clean.

PR #458, opened 2017-01-19
I would like to re base and merge against master. 

aatchison:
This would allow us to update to the hotfixes.


aatchison:
Sorry, I messed this up
master -> dev is what I need

PR #472, opened 2017-01-21
Listen Confirmation Ding

tmajibon:
Apologies in advance for anything I did wrong in my submission.

Adds an option (default off) to play a wave file when the listener starts recording.

Basically:

"Hey Mycroft" ding

Options are "confirm_listening" and "confirm_ding". No wave is included, but I used the following file which SoundBible lists as public domain.

http://soundbible.com/1815-A-Tone.html
direct link: http://soundbible.com/grab.php?id=1815&type=wav

### DIFF
--- a/mycroft/client/speech/main.py
+++ b/mycroft/client/speech/main.py
@@ -30,7 +30,7 @@ from mycroft.tts import TTSFactory
 from mycroft.util import kill, play_wav
 from mycroft.util.log import getLogger
 
-from os.path import join, dirname, expanduser
+from os.path import expanduser
 
 logger = getLogger("SpeechClient")
 ws = None
--- a/mycroft/client/speech/main.py
+++ b/mycroft/client/speech/main.py
@@ -44,7 +44,7 @@ def handle_record_begin():
 
     # If enabled, play a wave file with a short sound to indicate recording has
     # begun.
-    if config.get('confirm_keyword'):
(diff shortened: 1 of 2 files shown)


jhenriquez:
I couldn't wait and had to patch this into my code :) I love it! Personally, I like the idea of specifying the wav file. I'm using a computer beep from StarTrek TNG (http://www.trekcore.com/audio/). 

I had to do a minor change, I'll leave a comment on the PR for reference. That aside, works perfectly! Really appreciated it.

mycroft/client/speech/main.py
```diff
@@ -41,6 +41,12 @@
 
 def handle_record_begin():
     logger.info("Begin Recording...")
+
+    # If enabled, play a wave file with a short sound to indicate recording has
+    # begun.
+    if config.get('confirm_listening'):
+        play_wav(expanduser(config.get('confirm_ding')))
```
jhenriquez: `expanduser` has not been imported into the file. Locally, I just removed the function call. But probably, should consider adding the import if this is a best practice or style guideline.

tmajibon:
Thanks for catching that, I forgot the import line when putting it in. (Didn't have my dev environment set up at the time)

Was just using the function from os.path so that it can expand "~" in the config.

penrods:
This is something I've definitely wanted to get implemented, thanks for getting it going!

LGTM

PR #474, opened 2017-01-23
Fix silent mycroft when config lacks enclosure/platform

### LINKED ISSUE #473: Mycroft is silent if enclosure/platform isn't configured
The platform check in `handle_speak` will fail if the `platform` or `enclosure` is missing from the config. For me it seems like the platform value isn't set.

forslund:
Resolves #473


penrods:
LGTM

Issue #460, opened 2017-01-19
Phonetic Alphabet for Code Speach


kfezer:
Update the Pairing Skill to Speak the Phonetic Alphabet instead of letters.

Avoids confusing homoglyphs.

PR #476, opened 2017-01-24
Issue #460- Adds Phonetic Spelling to Pairing Skill

kfezer:
Adds NATO Phonetic alphabet to pairing skill, so reads codes as:
"A as in Alpaha, Three, B as in Bravo...."


jdorleans:
Thank you for the PR @kfezer ! Quesitons... 
Is that something we really want to behave like this? 
Could you please run some UX tests with @akuAtaja and @dschweppe so we get feedbacks?

kfezer:
Sure, I'll ping them and see what they think and work out some tests. Steve suggested it and it does make understanding the code easier.

kfezer:
Needs PEP8 testing

Issue #479, opened 2017-01-27
Shutdown all skills on process termination


augustnmonteiro:
- When mycroft-core is stopped it won't stop because have some timer threads running

PR #480, opened 2017-01-27
#479 - Shutdown all skills on process termination

augustnmonteiro:
#479 
- On stop mycroft-core executing cleanup of all skills


forslund:
Should this be done for `container.py` as well?

jdorleans:
@forslund thank you for pointing this out. I'm reviewing this PR and we'll address the container too.

PR #486, opened 2017-01-29
468 - Fix tests

ethanaward:
This fixes some of the tests that are currently broken. Most only required minor changes. Some of the tests seem to require pairing to work properly, and a few of the audio consumer tests are still not working.

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -33,7 +33,12 @@ if [ $(id -u) -eq 0 ]; then
 fi
 
 TOP=$(cd $(dirname $0) && pwd -L)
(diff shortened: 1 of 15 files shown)


jdorleans:
Ok, here is a first suggestion... let's break this into small chuncks. Don't take all tests to be fixed into one task. Let's concentrate on fixing a group of them properly then we can already ship them into the next release. So we tackle the other ones. How that sounds? I'll check on what you've done so far. Thx for that again!

jdorleans:
Let's ship this now, then we create specific tasks for further fixes.

aatchison:
Hmm, we seem to have a few conflicts here now.

forslund:
This one can be closed right?

ethanaward:
This has been closed, the bugfixes have already been addressed.

Issue #487, opened 2017-01-31
Integrate location with backend


augustnmonteiro:
- Configuration should get the location of device and save on mycroft.conf

PR #488, opened 2017-01-31
#487 - Integrating location microservice with mycroft core

augustnmonteiro:
- Adding find_location on device api
- Getting location and adding on settings

### DIFF
--- a/mycroft/configuration/__init__.py
+++ b/mycroft/configuration/__init__.py
@@ -110,11 +110,11 @@ class RemoteConfiguration(object):
         if update:
             try:
                 from mycroft.api import DeviceApi
-                setting = DeviceApi().find_setting()
-                location = DeviceApi().find_location()
+                api = DeviceApi()
+                setting = api.find_setting()
+                location = api.find_location()
                 if location:
-                    city = location.get("city")
-                    config["location"] = city["name"] + ", " + city["state"]["name"] + ", " + city["state"]["country"]["name"]
+                    setting["location"] = location
                 RemoteConfiguration.__load(config, setting)
             except Exception as e:
                 LOG.warn("Failed to fetch remote configuration: %s" % repr(e))
--- a/mycroft/configuration/__init__.py
+++ b/mycroft/configuration/__init__.py
@@ -111,6 +111,10 @@ class RemoteConfiguration(object):
             try:
                 from mycroft.api import DeviceApi
                 setting = DeviceApi().find_setting()
+                location = DeviceApi().find_location()
+                if location:
+                    city = location.get("city")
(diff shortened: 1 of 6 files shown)


mycroft/configuration/__init__.py
```diff
@@ -111,6 +111,10 @@ def load(config=None):
             try:
                 from mycroft.api import DeviceApi
                 setting = DeviceApi().find_setting()
+                location = DeviceApi().find_location()
```
jdorleans: Reuse DeviceApi into a var.

mycroft/configuration/__init__.py
```diff
@@ -111,6 +111,10 @@ def load(config=None):
...
+                if location:
+                    city = location.get("city")
+                    config["location"] = city["name"] + ", " + city["state"]["name"] + ", " + city["state"]["country"]["name"]
```
jdorleans: We should keep the location in the config as it is, an obj. So, no need for this concat logic. Each skill will deal with the data as they need. Some willnuse only the city name, others some other data, others will do a concat etc.

mycroft/skills/configuration/__init__.py
```diff
@@ -60,6 +60,10 @@ def notify(self, timestamp):
 
     def update(self):
         config = self.api.find_setting()
+        location = DeviceApi().find_location()
+        if location:
+            city = location.get("city")
+            config["location"] = city["name"] + ", " + city["state"]["name"] + ", " + city["state"]["country"]["name"]
```
jdorleans: Same as above.

mycroft/skills/weather/__init__.py
```diff
@@ -46,25 +46,29 @@ def build_query(self, params):
         params.get("query").update({"lang": self.lang})
         return params.get("query")
 
+    def build_location(self, location):
+        return location["city"]["name"] + ", " + location["city"]["state"]["name"] + ", " + \
```
jdorleans: This will break if any of those nested data are None. We have to check on them.

mycroft/skills/weather/__init__.py
```diff
@@ -46,25 +46,29 @@ def build_query(self, params):
...
+               location["city"]["state"]["country"]["name"]
```
jdorleans: Same as above

mycroft/skills/weather/__init__.py
```diff
@@ -189,7 +193,7 @@ def __build_data_condition(
             self, location, weather, temp='temp', temp_min='temp_min',
             temp_max='temp_max'):
         data = {
-            'location': location,
+            'location': location["city"]["name"],
```
jdorleans: Same as above

PR #499, opened 2017-02-07
Feature/issue #498 Automatically load data files

### LINKED ISSUE #498: Automatically load skill data files
In most skills the following is seen:

```python
    def initialize(self):
        self.load_data_files(dirname(__file__))
        [...]
```

The loading could be done automatically when loading skills instead of requiring the skill developer to do it manually. It's a small thing but in my opinion as much in the skill creation as possible should be automated.

This will require the `load_data_files()` method to handle missing directories more gracefully by loggin an error instead of throwing an exception.

forslund:
Load data files for skills automatically when loading the skill. Resolves #498

### DIFF
--- a/mycroft/skills/alarm/__init__.py
+++ b/mycroft/skills/alarm/__init__.py
@@ -33,12 +33,12 @@ __author__ = 'jdorleans'
 # TODO - Localization
 class AlarmSkill(ScheduledCRUDSkill):
     def __init__(self):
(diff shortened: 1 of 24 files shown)


forslund:
Sure, I'll remove it from all skills in the skills directory and update the PR

forslund:
All loading of data files in the skills have been removed. I've tested as many skills as I've been able to and everything seems fine. I haven't been able to test the *dial_call* and *send_sms* skills but they should be fine.

PR #506, opened 2017-02-09
Refining new "listening" sound mechanism

aatchison:
This adds several refinements to the listening sound mechanism added by:
* Added a default sound file
* Allowing various ways to override "resource files" for customization
* Moved the sound configuration path from "confirm_ding" to
  "sounds" > "start_listening"
* Also added "sounds" > "end_listening" configuration for the future

This submission adds the new mycroft.util.resolve_resource_file(res_name)
method.  This method takes a name such as "snd/start_listening.wav" and
looks (in order):
* For an absolute path <res_name>
* For ~/.mycroft/<res_name>
* For /opt/mycroft/<res_name>
* For mycroft/res/<res_name> within the source package


mycroft/util/__init__.py
```diff
@@ -28,6 +28,52 @@
...
+    if os.path.isfile(filename):
+        return filename
+
+    # Next look for /opt/mycroft/res/res_name
+    filename = os.path.expanduser("/opt/mycroft/"+res_name)
+    if os.path.isfile(filename):
+        return filename
+
+    # Finally look for it in the source package
+    filename = dirname(os.path.abspath(__file__))+"../res/"+res_name
```
augustnmonteiro: this code isn't upping the directory level and because that the isn't founding the file
a solution could be `os.path.abspath(os.path.join(os.path.dirname( __file__ ), '..', 'res', res_name))`

aatchison:
Thanks @augustnmonteiro, it's working great now.

Issue #527, opened 2017-02-23
Skill to install a new skill


augustnmonteiro:
mycroft should be able to install a skill from our repository
`hey mycroft install skill <skill_name>`

for do this mycroft should use msm

PR #524, opened 2017-02-21
#521 - Skill auto reload

augustnmonteiro:
- Checking by modification in skills and reloading

PR #533, opened 2017-02-24
#527 - skill to skill installation

augustnmonteiro:
For this skill work well the [auto reload skill](https://github.com/MycroftAI/mycroft-core/pull/524) should be merged.

### how it works:
say `hey mycroft install bitcoin` and [msm](https://github.com/MycroftAI/msm) will install the skill


aatchison:
So, here is what I'm thinking.
I have an old version of msm hanging around on my computer.
Let's add a script like scripts/install-msm.sh.
This script can then clone the current version of msm that is working with the core (similar to mimic)
Add a configuration option to define the msm script location, with default in the mycroft-core/msm directory.
Then, when packaging we can define another location for installation by package (like mimic.)
Does this sound okay?

augustnmonteiro:
yeah, it sounds good.
what you think about we install in /usr/bin ?

aatchison:
This is where I install mimic.
mimic: /usr/local/bin/mimic
I can take care of the path in packaging as long as you have a line like this in config:

```
{
  "tts": {
    "module": "mimic",
    "mimic": {
      "voice": "ap",
      "path":"/usr/local/bin/mimic"
    }
  }
}
```

aatchison:
Sort of like this:
https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/tts/mimic_tts.py#L32
The mimic path defaults to `./mimic/bin/mimic` in a git checkout environment, you could do the same for msm.
Checkout `./scripts/install-mimic.sh`

PR #538, opened 2017-03-02
Changes to record mycroft wake word when configured to do so.

amcgee7:
NOTE: this doesn't check for disk space and doesn't clean up
file stored in /tmp/ called mycroft_wake_sucess<time>.wav

TODO: some cleanup should be considered.

Signed-off-by: Art McGee <amcgee7@jaguarlandrover.com>


disheet:
how can i change wake word from HEY MICROFT to some other word?

PR #539, opened 2017-03-04
add converse method to allow skills to handle utterance

JarbasAI:
1) In the base MycroftSkill class, add a method like this:
```
def Converse(self, transcript, lang):
    return False
```
2) In the IntentSkill add a few things to handle_utterance():

a) Track recently invoked skills.  This would be a list of the Skills associated with the best_intent that comes out of Adapt.  Whenever a Skill's intent is invoked, it gets moved/placed at the top of the list.  This list also gets curated to remove Skills that haven't been invoked in longer than, say 5 minutes.

b) Before going through the current code that figures out best_intent, loop through the Skills in this list and call skill.Converse(utterance).  If one returns True, then they have handled the utterance and there is no need to do further intent processing or fallback.

3) in register_skill map intents to skills 

4) in load skills add a unique id for each skill instead of using skill_name (because its fragile)


penrods:
Couple comments:
* It is important to sort the skills_5min.  We should always give the most recently invoked skill the first chance to handle the utterance.

I think the easiest way to handle this would be to use a list instead of a dictionary.  I also don't like the name "skills_5min" because it won't make sense if we go to, say, a 7 minute window.  So I'd rather use the name "active_skills".

So in init you'd do:

   self.active_skills = []

When you invoke a skill you'd:

   self.active_skills.insert(0, [skill, time])

When you "touch" a Skill you can just remove it from this list and prepend it with skills_5min.insert(0, skill) to keep the list sorted.  I realize this causes a little headache whenever you add a skill to the list, since you have to search the list for and existing entry that already contains it and remove that reference.  You can minimize the effort by creating helpers like:  remove_active_skill(skill) and add_active_skill(skill).  

There are probably a few other implementations.  But having this sorted is a pretty big deal that I'd like in place from the beginning.

* Nitpick:  "whether" instead of "wether"  :)
* Nitpick:  let's avoid mixing in camelCase.  So use "converse()" instead of "Converse()" and "do_conversation()" instead of "doConversation()".  My apologies if I used camelCase when we were talking about this in Slack -- bad habit of mine.

penrods:
I will also look into the problem you were talking about with the empty list when attempting to invoking the conversation skill.  More on that tomorrow.

JarbasAI:
better now?

mycroft/skills/intent/__init__.py
```diff
@@ -31,17 +34,61 @@ class IntentSkill(MycroftSkill):
...
+        self.active_skills.insert(0, [skill_name, time.time()])
 
     def handle_utterance(self, message):
+
         utterances = message.data.get('utterances', '')
 
+        for skill in self.active_skills:
+            # check for conversation time-out
+            if time.time() - skill[1] >= self.converse_timeout * 60:
+                self.remove_active_skill(skill[0])
```
penrods: This is going to be a problem.  You never want to alter a list while you are running through it.  You end up changing it "out from under" the loop index and not getting what you expect.  Here is an example that mirrors what you are doing here:

>>> mylist = [1,2,3,4,5,6,7,8]
>>> for x in mylist:
...     if x % 2 == 0:
...         mylist.remove(x)
...     print x
...
1
2
4
6
8
>>> mylist
[1, 3, 5, 7]

Instead, you have to edit the list first, then loop through the edited list.  Something like this:

>>> mylist = [1,2,3,4,5,6,7,8]
>>> mylist = [x for x in mylist if x % 2 == 0]
>>> mylist
[2, 4, 6, 8]

Now you can safely loop through the edited list:

for x in mylist:
    print x

mycroft/skills/intent/__init__.py
```diff
@@ -57,6 +104,9 @@ def handle_utterance(self, message):
             reply = message.reply(
                 best_intent.get('intent_type'), best_intent)
             self.emitter.emit(reply)
+            # best intent detected -> update called skills dict
+            name = self.intent_to_skill[best_intent['intent_type']]
```
penrods: Your variable names are misleading.  The intent_to_skill is more like "intent_to_skill_id"; and name is "skill_id".  It _works_, but will confuse those trying to figure out the code later.  Be kind to future generations.  :)

JarbasAI: name is an artifact from before when i used the skill_name instead of skill_id, changing now : P

totally missed that list thing, will also look into it

JarbasAI:
made some more changes:
- with PR#558 intent names become skill_name:intent_name changed this to skill_id:intent_name
- parse source skill_id from this 
- removed the intent to id dict and changes to register_intent message mechanism
- changed self.id to self.skill_id inside skill base class
- if we requested a converse for an invalid id there would be an infinite loop, added a default response if not skill exists and made int(skill_id) in case someone passes a string by mistake

JarbasAI:
resubmitted has #756

PR #541, opened 2017-03-05
allow skills not to reload

JarbasAI:
when changing intent skill it would get auto-reloaded and miss register_intent messages causing weird behaviour

in my dictation skill, writign dictated file to skill folder caused skill to reload and reset some variables

skills should be allowed to ask not to be reloaded


penrods:
Makes sense to allow the author control, as this is really largely a convenience for developers.  I can also imagine situations where a skill is storing info in a local file of some sort and wants to prevent the reloads of the current implementation.

LGTM.

Issue #361, opened 2016-09-17
Allow for non-speach events


chrisl8:
I would like to be able to put messages onto the bus and "flag" them in some way that will prevent the TTS engine from saying them when the skill emits the "speak" message.

I have in mind that the message could include a "source" key that tracks where the message came from, and the TTS engine will ignore messages with this flag set.

This would allow sending queries to MyCroft via SMS, web, or any other source and picking the result up off of the message bus without MyCroft speaking the answers out loud into the room when the source was not the microphone.

Unfortunately, it seems like none of the data from the original message is preserved through the skill process. The skill simply emits a "speak" message without copying any of the data from the original message into it. It appears to be non-trivial to change this.

forslund:
This can be done now (thanks to @JarbasAI) by using setting the target of an interaction to something other than "audio" allowing for precisely this sort of thing. Closing this issue.

PR #556, opened 2017-03-13
Enable/Disable speak

JarbasAI:
sometimes it is useful to have mycroft do everything except speak, for example if you want it to play music while using the cli without getting interrupted by speech

we mustn't assume the user will always start things in quiet mode and it is good to be able to control this trough intents

a generic mute skill will give user control to disable speech

other use-cases can send a bus message to control this

issue#https://github.com/MycroftAI/mycroft-core/issues/361

### DIFF
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -23,6 +23,7 @@ from mycroft.util import create_signal, check_for_signal
 from mycroft.util.log import LOG
 from mycroft.messagebus.message import Message
 
+speak_muted = False
 bus = None  # Mycroft messagebus connection
 config = None
 tts = None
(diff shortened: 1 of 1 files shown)


augustnmonteiro:
Please fix conflicts and rebase : D

ethanaward:
This seems good, although we don't have skills in the core anymore. Could you remove the skill and rebase?

augustnmonteiro:
@JarbasAI can you rebase this?, thank you

JarbasAI:
rebased and updated this

AIIX:
+1 vote for this feature another use case: office working environments or public places where one would like to use mycroft using text instead of speech on enclosures that support text

JarbasAI:
we now can not execute TTS with #1408 , but i would still like this feature as a global switch

i like telling mycroft to mute and to start playing music, then i can still trigger skills because i always have a ssh terminal open to my mark1, i can still type in questions without having the speech

if this is desired i will rebase again

Issue #540, opened 2017-03-05
APi de-cluttering


JarbasAI:
at a certain point there will sure be lots of api keys, sometimes shared between skills (mashape key used in a bunch of my skills), having the same api in a section for each skill in .config is ugly,  we should have a section for apis only maybe?

in my build i added a self.APIconfig to base MycroftSkill , so i just do self.APIconfig.get(api), but not sure this is the best approach

forslund:
Ok, closing this one. Feel free to reopen if you think it should remain.

JarbasAl:
this served 2 purposes, ease of use which is no longer an issue because of skill settings and privacy reasons

i still like this approach for privacy reasons, keep your api keys in a single place and out of mycroft-home

it's less important now that we have skill settings, i will give it some more thought and open a PR if i come up with some cool solution

devs-mycroft:
If it's a convention how to organize it may just be documentation needed. Since we're removing the Per skill config-blocks in 19.08 this might be something...

forslund:
If it's a convention how to organize it may just be documentation needed. Since we're removing the Per skill config-blocks in 19.08 this might be something...

PR #557, opened 2017-03-13
Centralizing APIs in config file#540

JarbasAI:
API keys can often used in more than one skill, to use same key in more than one skill requires developer to know in which part of config file it is stored and manually accessing it, or to duplicate the key entry

in long term this makes it messy for user-setup and dev-usage

issue #540 

my use-case:

´´´
  "APIS": {
    "WolframAlphaAPI": "key",
    "WeatherAPI": "key",
    "GraphAPI": "key",
    "MashapeAPI": "key",
    "WikimapiaAPI": "key",
    "CloudsightAPI": "key",
    "RedditAPI": "key",
    "CloudsightSecret": "secret",
    "RedditSecret": "secret"
  },
´´´


ethanaward:
This needs to be rebased to remove references to skills being in the core, and then PRs should be sent to the core skill repos instead. I'm also not sure whether this will affect changes on the backend, @augustnmonteiro?

Issue #553, opened 2017-03-12
way to de-register intents


JarbasAI:
in some cases it would be helpful to be able to de-register intents

use case example:

this would also allow intents to only execute if you give a sequence of orders, intent 1 registers intent 2 
intent 2 registers intent 3, if intent 2 not executed intent 2 is de-registered after a timer

continuous conversation use-case https://community.mycroft.ai/t/keep-skill-active-and-store-input-variables/1447/3?u=jarbas_ai

JarbasAI:
seems to work: 

https://github.com/JarbasAI/mycroft---konami-code

personal taste changes, return instead of break, more logs

    def enable_intent(self, intent_name):
        """Reenable a registered intent"""
        for (name, intent, handler) in self.registered_intents:
            if name == intent_name:
                self.registered_intents.remove((name, intent, handler))
                intent.name = name
                self.register_intent(intent, None)
                self.log.info("Enabling Intent " + name)
                return
        self.log.error("Could not Re-enable Intent " + intent_name)

augustnmonteiro:
Hi,

you can use self.disable_intent(name)

thank you!

PR #558, opened 2017-03-13
Add functionality to remove and restore intents.

forslund:
This is a solution for #553 which adds two new methods to the `MycroftSkill` `disable_intent()` and `enable_intent()`

The skills takes an intent name and enables/disables it.

A test skill that uses this can be found here: https://drive.google.com/open?id=0B2_BylN8IxZUbmM1eWE2OHBEUWs

`run test` will run the test intent once and then disable itself, `enable test` will re-enable the test intent.

This PR also changes the internal name that's associated with the intent, it now includes the skill name as well to make the intents uniquely identifiable.


JarbasAI:
seems to work:

https://github.com/JarbasAI/mycroft---konami-code

personal taste changes, return instead of break, more logs

def enable_intent(self, intent_name):
    """Reenable a registered intent"""
    for (name, intent, handler) in self.registered_intents:
        if name == intent_name:
            self.registered_intents.remove((name, intent, handler))
            intent.name = name
            self.register_intent(intent, None)
            self.log.info("Enabling Intent " + name)
            return
    self.log.error("Could not Re-enable Intent " + intent_name)

JarbasAI:
I don't really like the changing of intent names to skill_name:intent_name, tracking wish skill owns each intent seems a separate issue

forslund:
I think it (or a similar mechanism) is needed, adding this feature without making sure it doesn't mess up for other skill creators would be irresponsible. I could separate it into two pull requests if that would be preferable.

JarbasAI:
if PR #539 goes forward mapping skills to intents is tracked there, your approach could make more sense if it is desired to keep track of skill_source outside intent skill

forslund:
Hmm, but that doesn't register/detach unique intent id's into adapt (as far as I can determine) I could move the name modification into the intent skill (register and detach modifies the name when sending them into adapt) but that would require that I duplicate some of the work in #539.

It might be best to keep it like this for the time being and having it working stand alone and then update it when #539 gets merged.

JarbasAI:
i would have enable_intent emit a message  self.emitter.emit(Message("enable_intent", {"intent_name": name}))   and have a handle_enable_intent function performing the enabling itself, it can be useful to enable intents from outside the skill itself

(i'm messing with a intent tree kind of structure that needs this for example- https://github.com/JarbasAI/jarbas-core/blob/dev/mycroft/skills/intent_parser.py )

Issue #560, opened 2017-03-14
Fix skills loading


augustnmonteiro:
Some skills are not working

kfezer:
This fixes the order. 

Apparently WA needs to be listed last. I created an issue so we can fix this as right now I'd consider this a workaround.

PR #563, opened 2017-03-14
#560 - loading skills sorted

augustnmonteiro:


kfezer:
If wolfram alpha skill isn't loaded last, it doesn't work correctly. Making an issue.

kfezer:
This fixes the order. 

Apparently WA needs to be listed last. I created an issue so we can fix this as right now I'd consider this a workaround.

PR #567, opened 2017-03-15
#566 - Adding normalize import on wolfranalpha

augustnmonteiro:


kfezer:
Fixed issue with WA and normalize

PR #576, opened 2017-03-17
Add expect_response option to speak_methods

forslund:
If `expect_response` flag is set to `True` the STT will be triggered just as if the wakeword has been received or the button on the Mycroft enclosure has been pressed.

The `speak` and `speak_dialog` methods will continue to work as before

```python
self.speak('Hello there')
```

But if Mycroft asks a user a question and would like a response from the user the STT can be triggered directly after mycroft finish speaking by setting the `expect_result` parameter to `True`

```python
self.speak('Would you like a cookie?', expect_response=True)
```

Can be useful to create some more natural interaction between the user and Mycroft.

A test skill using this can be downloaded here:
https://drive.google.com/open?id=0B2_BylN8IxZUal9VWFoyNTIyS3c
`Hey mycroft I'd like a cup of tea`

### DIFF
--- a/mycroft/client/speech/main.py
+++ b/mycroft/client/speech/main.py
@@ -27,7 +27,7 @@ from mycroft.identity import IdentityManager
 from mycroft.messagebus.client.ws import WebsocketClient
 from mycroft.messagebus.message import Message
 from mycroft.tts import TTSFactory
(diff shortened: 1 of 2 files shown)


forslund:
A possible improvement is to check for a `?` at the end of the utterance given to `speak`

PR #591, opened 2017-03-27
Fixes Issue 588

kfezer:
Fixed typo

	modified:   mycroft/skills/core.py


SoloVeniaASaludar:
I'm wondering if another programming language is not a better alternative for mycroft-core. Python gives a false impression of fast development, in addition to other drawbacks (low performance, ...). Any other language as c/c++, java, **scala**, ... can detect immediately an issue like this one.

kfezer:
I would say that those points are valid, but we're not changing the language for mycroft-core anytime in the near future. I prefer python; it just requires a little more testing procedure and vigilance since those flags don't show.

Issue #595, opened 2017-03-29
error in mic.py causes no phrase longer to 3 seconds can be used


SoloVeniaASaludar:
This is a trace of an execution of method "record_phrase" in mycroft/client/speech/mic.py, where the user has been speeking in loud voice continuously:

```
2017-03-29 10:31:51,573 - mycroft.client.speech.mic - DEBUG - energy=3234 thr=4961.39655261 noise=0
2017-03-29 10:31:51,634 - mycroft.client.speech.mic - DEBUG - energy=3391 thr=4948.7744434 noise=0
2017-03-29 10:31:51,694 - mycroft.client.speech.mic - DEBUG - energy=3242 thr=4964.52119432 noise=0
2017-03-29 10:31:51,765 - mycroft.client.speech.mic - DEBUG - energy=3364 thr=4952.91384272 noise=0
2017-03-29 10:31:51,826 - mycroft.client.speech.mic - DEBUG - energy=3238 thr=4963.55678052 noise=0
2017-03-29 10:31:51,886 - mycroft.client.speech.mic - DEBUG - energy=3205 thr=4951.37368885 noise=0
2017-03-29 10:31:51,957 - mycroft.client.speech.mic - DEBUG - energy=3370 thr=4934.92399572 noise=0
2017-03-29 10:31:52,017 - mycroft.client.speech.mic - DEBUG - energy=3216 thr=4948.65279797 noise=0
2017-03-29 10:31:52,078 - mycroft.client.speech.mic - DEBUG - energy=3354 thr=4934.40071136 noise=0
2017-03-29 10:31:52,148 - mycroft.client.speech.mic - DEBUG - energy=3224 thr=4945.44532047 noise=0
2017-03-29 10:31:52,209 - mycroft.client.speech.mic - DEBUG - energy=3181 thr=4932.93196969 noise=0
2017-03-29 10:31:52,270 - mycroft.client.speech.mic - DEBUG - energy=3353 thr=4914.47476334 noise=0
2017-03-29 10:31:52,340 - mycroft.client.speech.mic - DEBUG - energy=3218 thr=4927.62608976 noise=0
2017-03-29 10:31:52,401 - mycroft.client.speech.mic - DEBUG - energy=3346 thr=4916.12107929 noise=0
2017-03-29 10:31:52,462 - mycroft.client.speech.mic - DEBUG - energy=3218 thr=4927.88366554 noise=0
[...]

PR #596, opened 2017-03-29
error in mic.py causes no phrase longer to 3 seconds can be used

SoloVeniaASaludar:
Error in mic.py causes no phrase longer to 3 seconds can be recorded.
See issue #595 for details.

### DIFF
--- a/mycroft/client/speech/mic.py
+++ b/mycroft/client/speech/mic.py
@@ -179,15 +179,15 @@ class ResponsiveRecognizer(speech_recognition.Recognizer):
         max_noise = 25
         min_noise = 0
 
-#        def increase_noise(level):
-#            if level < max_noise:
-#                return level + 200 * sec_per_buffer
-#            return level
-#
-#        def decrease_noise(level):
-#            if level > min_noise:
-#                return level - 100 * sec_per_buffer
-#            return level
+        # def increase_noise(level):
+        #   if level < max_noise:
+        #     return level + 200 * sec_per_buffer
+        #   return level
+        #
+        # def decrease_noise(level):
+        #   if level > min_noise:
+        #     return level - 100 * sec_per_buffer
+        # return level
 
         # Smallest number of loud chunks required to return
         min_loud_chunks = int(self.MIN_LOUD_SEC_PER_PHRASE / sec_per_buffer)
@@ -214,11 +214,11 @@ class ResponsiveRecognizer(speech_recognition.Recognizer):
             test_threshold = self.energy_threshold * self.multiplier
             is_loud = energy > test_threshold
             if is_loud:
-#                noise = increase_noise(noise)
+                # noise = increase_noise(noise)
                 num_loud_chunks += 1
                 num_silence_chunks = 0
             else:
-#                noise = decrease_noise(noise)
+                # noise = decrease_noise(noise)
                 num_silence_chunks += 1
                 self.adjust_threshold(energy, sec_per_buffer)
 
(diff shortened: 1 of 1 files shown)


MatthewScholefield:
I should have probably put this in the code beforehand to explain it, but that line, `recorded_too_much_silence = num_chunks > max_chunks_of_silence` is checking whether or not it should continue waiting for more audio even though it hears silence. It should act like a timer since it essentially prevents quiet noise less than 3 seconds from being recorded. However, the moment the user speaks, the noise will go above 0 and thus will indeed record more than 3 seconds. [Here's a better explanation](https://gist.github.com/MatthewScholefield/fb66d02e315d27a08094e1162dc31ba1#recording-the-query). (Specifically, the last part of `Recording the Query` talks about this).

However, if you were having issues with that part, it means there's something else going wrong. Either way, we should still test your changes and see if they work better on different devices than the original.

SoloVeniaASaludar:
Hi,

Thanks for your answer.

After 3 seconds from method start, the condition "recorded_too_much_silence" pass from false to true. For this reason, the if condition:

```
if quiet_enough and (was_loud_enough or recorded_too_much_silence):
                phrase_complete = True

```

is, after 3 seconds, equivalent to:
```

if quiet_enough:
                phrase_complete = True
```

In this situation, after these initial three seconds, the record finish any time that "quiet_enough" becomes true. And "quiet_enough" is related only to the characteristics of current chunk and noise threshold. In my test, near than in all chunks "quiet_enough" is true.

MatthewScholefield:
Well, the problem is that in the beginning, we are certain the user has not spoken. Because of this, we can wait with a fair amount of certainty for the user to speak. However, the longer we wait, there's more of a chance that the algorithm is incorrect and that the user has indeed spoken but possibly did not trigger enough noise. For this reason, it would be annoying to sit there waiting in the case that the user has actually already said their sentence. So, we can only use this little trick in the first three seconds. This is what I meant by the last set of parenthesis: `(however, after a reasonable time limit, the listener will timeout and return even if still waiting for some noise and was_loud_enough is still False).`

As for your problem with `quiet_enough` should not always be `True`. If this is the case, it probably means there's a problem with the way we track the threshold. Can you try getting the original code to work by only changing `self.multiplier`? Otherwise, I think I have another idea to improve the threshold. I can give you a version later today to test.

SoloVeniaASaludar:
In the issue you can see a full trace of the energy, noise and threshold levels during the test.

penrods:
I was also working on several things around the wake-up word in my PR #647.  I believe those changes make this unnecessary.

SoloVeniaASaludar:
yes, changes in #647 are very related. Can I suggest a "cosmetic" change, that made things more easy to follow ? Instead of:

```
            if quiet_enough:
                silence_duration += sec_per_buffer
                if silence_duration < self.MIN_SILENCE_AT_END:
                    quiet_enough = False  # gotta be silent for min of 1/4 sec
            else:
                silence_duration = 0
             recorded_too_much_silence = num_chunks > max_chunks_of_silence
             if quiet_enough and (was_loud_enough or recorded_too_much_silence):
                 phrase_complete = True

``` 
use:

```
            if quiet_enough:
                silence_duration += sec_per_buffer
            else:
                silence_duration = 0
             recorded_too_much_silence = num_chunks > max_chunks_of_silence
             recorded_enough_silence =  silence_duration >= self.MIN_SILENCE_AT_END:
             if recorded_enough_silence and (was_loud_enough or recorded_too_much_silence):
                 phrase_complete = True

```

PR #446, opened 2017-01-09
Adding support for Python3

the-kid89:
This is the first pull request to start adding support for Python3.

My first step was to make sure that all the unit tests would run without throwing errors. The next step is to get all the unittests passing and find any further Python2-3 errors.

This is for ticket https://github.com/MycroftAI/mycroft-core/issues/366

PR #604, opened 2017-03-30
fix syntax error for Python 3

stardiviner:


stardiviner:
@augustnmonteiro I updated `dev_setup.sh` to use Python 3 now. check out. Is there other places to modify?

stardiviner:
@augustnmonteiro I updated `dev_setup.sh` which is from your email attachment. Thanks. Please check it out the updated branch.

ethanaward:
This seems like it changes the whole git install to use python 3 rather than just fixing syntax errors in preparation for a change, which is a very substantial change. Do we know that all of our dependencies and the codebase work with python 3?

kfezer:
This is similar to the newer PR https://github.com/MycroftAI/mycroft-core/pull/446#issuecomment-311780314

Issue #612, opened 2017-03-31
improvement proposal: allow custom STT engines as plug-ins


SoloVeniaASaludar:
Custom STT engines are an interesting possibility. By example, to allow testing of a new engine, internalization, or any other activity that requires very different configuration, input or output than the existing one (i.e. use jsgf instead of ngram based stt).

With the current STT Factory class, addition of a new SST implies modify the core code: import the class with the new STT and add it to the CLASSES map in file stt/__init__,py

An small change (only addition of 5 statements in a single point of the current code) will allow to add custom STTs without impacts on core. 

To configure these new stt engines, the full path class path must be wrriten in config file. Existing engines can be configured using current single word method or full path. An example of new STT is:

```
  "stt": {
    "module": "mycroft.stt.LocalSTT.LocalSTT"
  },

```

PR #614, opened 2017-03-31
improvement: allow custom STT engines as plug-ins #612

SoloVeniaASaludar:
Allow custom STT engines as plug-ins.

See #612 for details

### DIFF
--- a/mycroft/stt/__init__.py
+++ b/mycroft/stt/__init__.py
@@ -14,6 +14,9 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with Mycroft Core.  If not, see <http://www.gnu.org/licenses/>.
+
+import importlib
+
 from abc import ABCMeta, abstractmethod
 
 from speech_recognition import Recognizer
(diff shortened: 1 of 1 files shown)


clusterfudge:
I might recommend something like the following:

```
class STTFactory(object):
    CLASSES = {
        "mycroft": MycroftSTT,
        "google": GoogleSTT,
        "wit": WITSTT,
        "ibm": IBMSTT
    }

    @staticmethod
    def register_stt(name, engine_class):
        # assert api signature
        # ...
        CLASSESS.put(name, engine_class)

    @staticmethod
    def create():
        config = ConfigurationManager.get().get("stt", {})
        module = config.get("module", "mycroft")
        clazz = STTFactory.CLASSES.get(module)
        return clazz()
```

And then we can avoid importlib, and the convention of splitting config strings to try to construct plugins.

SoloVeniaASaludar:
Thanks for the comment.

Who will call "register_stt" ?

clusterfudge:
I see the meat of your question :). I'm assuming you want something that's entirely config driven, as opposed to something that requires writing code to inject. I don't think there's a precedent for that in mycroft (aside from skills), so someone from the core team may want to chime in on what direction they see that going.

Issue #594, opened 2017-03-28
intent parsing inside skills


JarbasAI:
best intent is determined inside intent skill

sometimes it is useful to have intents registered inside a skill, for continuous conversation

# Intent Parsing inside skills

 with intent parser abstracted we can also do the parsing by registering intents inside the skill

            from mycroft.skills.intent_parser import IntentParser

            in initialize -> self.intent_parser = IntentParser(self.emitter)
            
            registering intents -> intent = IntentBuilder('NameofIntent')
                                -> self.register_intent(intent, self.handle_intent)
                                -> self.intent_parser.register_intent(intent.__dict__)
                          
            on converse method -> get intent instead of manually parsing utterance
                def converse(self, transcript, lang="en-us"):
                    determined, intent = self.intent_parser.determine_intent(transcript)
                    if determined:
                        # intent was registered and detected on own parser
                        # no need to parse utterance
                        # return false to handle in intent skill
                        return False
                    else:
                        handled = False
                        # Handle utterance manually
                        if handled:
                            return True
                    return False

this allows every skill to have a self-intent parser, builds on converse PR #539 

also simplifies intent skill

SoloVeniaASaludar:
I'm facing what I think is a similar problem, but solution points to a different mechanism. The scenario is:

1) skill handles an intent as "warning me in 10 minutes"
2) skill launches a thread
3) the thread of the skill requests client.speak to speak the phrase "what I must warning in 10 minutes?" and request also to record a phrase
4) client.speak plays the previous phrase "what I ... ?"
5) client.speak records a new phrase (without stt translation) and sends "record_end" message.
6) the thread in skill receives the "record_end" 
7) after  10 minutes, skill plays the recorded phrase.

note that in step 6 is one specific thread in a skill that must handle the message and continue with its logic. It seems difficult decide which skill and which thread of the skill must handle the answer from client.speak based only in the utterance content.

In order to solve this kind of dialog, recognition and decision of intent by content seems not a good approach. Probably, it is better provide client/speak with a "**answer identity**" in step 3. In step 5, the same "answer identity" is included in the utterance message or record_end message.

augustnmonteiro:
The IntentSkill is no longer a skill now is just class : D

JarbasAI:
how would you use the intent class for this purpose? import it from a skill?

PR #620, opened 2017-04-01
#594 - intent parsing inside skills

JarbasAI:
Solution for Issue #594 

this allows for skills to determine intents from utterances and execute intents without relying on intent skill

use case:
  - skills can parse intents from utterances to use in continuous dialog
  - skills can have "internal" intents and fully control them from registering to execution

made intent skill also use this


JarbasAI:
closing this, no longer applies, better alternative #756

Issue #621, opened 2017-04-02
i18n: skill supports language


SoloVeniaASaludar:
Only skills that supports current language must be loaded. That means loader of skills must verify if current language is supported by the skill. 

For backward compatibility, it is assumed "en-US" is the only language supported, except "intent" skill that is language independent.

PR #623, opened 2017-04-02
#621 - i18n: filter skill by supported language

SoloVeniaASaludar:
Load only skills that supports current language.

See issue #621 for details.


SoloVeniaASaludar:
Fixed conflict with changes in "intent".

PR #625, opened 2017-04-03
Bugfix/issue 624 - fix mimic_tts for unicode

### LINKED ISSUE #624: mimic tts fails silently when utterance contains unicode
It seems like hashing the unicode string for caching purposes causes an exception when unicode characters are parsed.

forslund:
Convert input to the hash function to `bytes`/`str`.

This PR also add some basic feedback if something in the tts process fails. Resolves #624


penrods:
Looks good to me

PR #628, opened 2017-04-03
Stop multi-sentence speech

### LINKED ISSUE #627: Multi-sentence speak commands can't easily be stopped
If `self.speak()` is called from a skill with multiple sentences the stop functionality doesn't quite work. Only the current sentence is stopped and the following sentence will immediately be started in it's place.

Example:
```python
self.speak('First sentence that is quite long. This is the second sentence that is also of some length. Finally a third sentence.
```

forslund:
This PR resolves #627 by storing the time for each stop command received in the voice client. The sentence loop is broken if the start time of the `handle_speak()` instance is older than the last stop request.

This approach should be fairly robust despite the global variable used for the time of the last received stop command since the variable is written at exactly one place and only read in the `handle_speak()` function.

### DIFF
--- a/mycroft/client/speech/main.py
+++ b/mycroft/client/speech/main.py
@@ -19,6 +19,7 @@
 import re
 import sys
 from threading import Thread, Lock
+import time
 
 from mycroft.client.enclosure.api import EnclosureAPI
 from mycroft.client.speech.listener import RecognizerLoop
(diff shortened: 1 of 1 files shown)


forslund:
Rebased and Conflict resolved.

augustnmonteiro:
Thank you!

Issue #630, opened 2017-04-03
fix skillInstaller


augustnmonteiro:
the popen is not getting the exit code from msm

penrods:
Can this be closed now?

augustnmonteiro:
Yeah, was fixed!

PR #631, opened 2017-04-03
#630 - Fixing SkillInstaller, using text to match instead of exit code

augustnmonteiro:


aatchison:
It looks like travis-ci fails on the install-msm.sh, due to inability to create the /opt/mycroft/skills directory.
Obviously, this isn't an issue with the code but it is for our tests...

PR #638, opened 2017-04-06
Code to prevent multiple instances of a service

dmendyke:
**Purpose**: Code in .../mycroft/lock can be used to create a text based lock file that can be used to prevent more than one instance of a process form running.

Currently the serivces _voice, skills_ and _service_ use this code

**Limitations**:   If the lock file is deleted by another process the protection fails.  Also, if many instances of a protected process are started within a few seconds, such as from within a loop in a shell script, the lock file could fall out of sync and the protection fails.


augustnmonteiro:
we are using pep8 for code style could you fix those errors https://travis-ci.org/MycroftAI/mycroft-core/builds/219454992
if you want to run pep8 on your machine you can install pep8 using pip and run `pep8 mycroft test`
and could you remove the skill daily meditation from your pr.

thank you!

dmendyke:
Pep8 compliant.
'daily_meditation' skill removed.  (Didn't need it anyway)

dmendyke:
Perhaps you could add 'autopep8' to the pull script to automatically configure incoming files.

PR #643, opened 2017-04-07
Fix creation of /opt/mycroft/skills

cincodenada:
I got an error when trying to set up about the creation of `/opt/mycroft/skills`.

Looking into it, the commands here seemed to be backwards - it tested for directory existence and if not tried to set permissions, then tested for write permission and then tried to create.

Flipped those two around, also made indenting consistent, and the setup ran as expected.


forslund:
Noted the same issue. Tested the PR and it resolves the issue.

augustnmonteiro:
Thank you!

PR #645, opened 2017-04-08
Update skill container to use Intent class

### LINKED ISSUE #644: skill container is broken
When the intent skill became the intent class the skill container wasn't updated to match the new structure and is currently not working.

forslund:
The skill container now uses the Intent class like the normal skill process.

The command line to enable the intent parsing is now `--enable-intent`. Resolves #644


forslund:
The fails here are completely unrelated to the changes in this PR...

Issue #609, opened 2017-03-31
improvement proposal: allow to save records in file and no stt 


SoloVeniaASaludar:
Current implementation of skill "audio_record" is based on tool "arecord". This approach has a problem, arecord tool fails because it can not access to system audio-in (the mycrophone, that is currently used by mycroft-core).

An improvement in mycroft-core could solve previous scenario and add some reusable new features: the skill can request next phrase to be saved in file; and/or that next phrase is not passed to the stt. 

In addition to the feature of disable wakeup word (added a few days ago), a new "audio_record" implementationl can solve previous problem following these steps:
1) speak something like "ready to record..." or "start of record..."
2) use "speak" in core.py to request to "client/speech" that next phrase must a) be saved in file; b) not translated by sst. 
3) Finally, request forced wakeup (do not wait for "hello mycroft" before to start recording).

In this way, the skill will have a .wav file with the voice recording.

Feature implementation has been tested using a new skill, but this one is in Spanish language, I can not share yet a functional copy. However, the relevant part of the skill is the call:

```
            self.speak(u'que debo avisar?', 
                expect_response=True, 
                record_characteristics =
                    { 'no_stt' : True, 
                      'record_filename' : '/tmp/prueba.wav' } )
```

that request play "¿que debo avisar?" and records a wav file '/tmp/prueba.wav'. 

The end of the record is know because message "recognizer_loop:record_end" is received in the skill.

PR #647, opened 2017-04-09
Issues related to wake-up words and audio recording, including #595

penrods:
Changes include:

* The wake up sound is now played synchronously, thus not included in the recorded audio
* The minimum recorded phrase must be at least 0.5 seconds (instead of 0.1), and must be continuously quiet for that duration
* The silence threshold is reset every time we begin listening for a wakeword
* The silence threshold adjusts upward to just above the ambient sound while waiting for the wake-word
* Reformatted some comments to use Google-style docstrings, and added more comments

### DIFF
--- a/mycroft/client/speech/main.py
+++ b/mycroft/client/speech/main.py
@@ -27,7 +27,7 @@ from mycroft.identity import IdentityManager
 from mycroft.messagebus.client.ws import WebsocketClient
 from mycroft.messagebus.message import Message
 from mycroft.tts import TTSFactory
-from mycroft.util import kill, play_wav, resolve_resource_file, create_signal
+from mycroft.util import kill, create_signal
 from mycroft.util.log import getLogger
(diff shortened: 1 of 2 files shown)


SoloVeniaASaludar:
Please, allow me to made a suggestion:

Recording options (as this interesting new flag "confirm_listening") are more practical if they are record specific (with default value from config) instead of globally (all record) flags.

In issue #609 you can see an example of two new record options ("save in file" and "no stt") and how they are implemented in a record specific way, not as a global (all record) configuration. "confirm_listening" could be next one.

Thanks.

penrods:
Using a single underscore is a general guideline in the Python world to mark a name as 'private'.  There is no enforced 'private' in Python like there is in C++, rather the single underscore is a way of telling other programmers that you really don't intend for this thing to be used outside of this module.  So if you muck with it, things may break later.

Issue #653, opened 2017-04-13
Conf options need to be documented


penrods:
The options in mycroft.conf are poorly understood and not documented.  What are the legal values?  What values get overridden locally and/or by home.mycroft.ai?

PR #654, opened 2017-04-13
The configuration values were poorly documented.  See issue #653

penrods:
This addresses this in several ways:
* Created mechanism to load 'commented' JSON (using '//' or '#' comments on a single line)
* Embedded comments into the mycroft.conf, indicating use, legal values, and where they get overridden
* Create ConfigurationManager.instance() static method to replace ConfigurationManager.get().  This produces more readable code like:
  ConfigurationManager.instance().get("value") instead of ConfigurationManager.get().get("value")
* Made _ConfigurationListener 'private'
* docstring'ed things
* Fixed "open(location, 'rw')".  "rw" is not documented as legal, although it wasn't crashing.  Now just using "w".

### DIFF
--- a/mycroft/configuration/__init__.py
+++ b/mycroft/configuration/__init__.py
@@ -39,7 +39,7 @@ load_order = [DEFAULT_CONFIG, REMOTE_CONFIG, SYSTEM_CONFIG, USER_CONFIG]
 class ConfigurationLoader(object):
     """
     A utility for loading Mycroft configuration files.
-    
+
     Mycroft configuration comes from four potential locations:
      * Defaults found in 'mycroft.conf' in the code
      * Remote settings (coming from home.mycroft.ai)
@@ -49,10 +49,10 @@ class ConfigurationLoader(object):
     in the Default would be overridden by a value with the same name found
     in the Remote.  And a value in the Remote would be overridden by a value
     set in the User settings.  Not all values exist at all levels.
-    
+
     See comments in the 'mycroft.conf' for more information about specific
     settings and where they reside.
-    
+
     Note:
         Values are overridden by name.  This includes all data under that name,
         so you if a value contains a complex structure, you cannot specify
@@ -173,7 +173,7 @@ class ConfigurationManager(object):
     """
     Static management utility for accessing the cached configuration.
     This configuration is periodically updated from the remote server
-    to keep in sync.    
+    to keep in sync.
     """
 
     __config = None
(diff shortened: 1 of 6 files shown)


SoloVeniaASaludar:
Please, one question: how is expected handle i18n in config file?

Will a single config file contain different data blocks for each language; or a set of different config files will be distributed  (mycroft-config.en-US, mycroft-config.es, ...) and user will overwrite default one with its language choice ?

penrods:
This PR is really just about getting things as they stand today documented, but I don't mind discussing localization/internationalization a bit.

What do you see in the config files that is troubling to you?  In my mind things aren't too bad.  In the long term the localized interface that users will see happens on the website (home.mycroft.ai) -- obviously that is not localized yet, but that is independent of these config files.

So, for example, a Spanish speaking user would go to home.mycroft.ai.  They would click a flag icon to get to the Spanish version of the website.  Once there, we can default the configuration of all provisioned devices to use "es-US" and the user can also choose a different varient ("es-MX", "es-ES", etc.).  Of course we'd present this in a user-friendly way not by those codes.  :)

Once this is set, Skills should start using the appropriate "es-es" folder for vocabulary and dialog.  I've been talking about the potential of auto-translation of stills that do not have vocabulary for a specific language already.  I think that is very do-able... it won't be perfect, but much better than not having access to "turn on the lights" just because the Skill author didn't speak Russian and provide "ru-ru" folders.

Make sense?  Please point out any specific concerns, recognizing problems sooner than later is a good thing!

SoloVeniaASaludar:
The origin of the question is that currently there are some hardcoded english messages in mycroft core. i18n needs move them from code to somewhere configurable. 

1) An easy option is to add them to mycroft.conf in a configuration block like:
```
"msg" {
   "failed_stt" : "Sorry, I can't catch that"
   ...
}

```
A different file mycroft.conf.es will contain all configuration entries, with the corresponding messages translations. In this way, mycroft.conf.es file can be copied over mycroft.conf to change the language.

2) Another option, using a single file is:

"msg" {
   "en-US" {
   "failed_stt" : "Sorry, I can't catch that"
   ...
  }
  "es" = {
   "failed_stt" : "Lo siento, no he entendido esa frase"
   ...
  }
}

3) A final option is to have the mycroft.conf file as in option 1, and a **delta** file mycroft.conf.es that will contain only the entries in mycroft.conf that changes due to i18n. 

This option needs a change in the configuration subsystem, first mycroft.conf must be loaded, next (optionally) mycroft.con.es will override the entries that changes due to i18n.

penrods:
Yep, I completely understand what you are saying .  I've been adding TODO comments near hard-coded strings that need localized that are inside mycroft-core code.  But those will live in a different place that mycroft.conf.  That is for configuration settings, not translations.

I will get this added in the next week or two.  It will follow a scheme similar to the Skill "dialog" ... subdirectories for "en-us", "ru-RU", etc.  I will probably place it under the mycroft\res directory, but let me think about it a litte more.

Issue #659, opened 2017-04-14
Git install broken after pull


penrods:
The recent change breaking out Skills from mycroft-core has broken installations in several ways:
* Naming the file Intent.py doesn't work on a pre-existing install, the folder Intent exists already and contains __init__.pyc.  So "import mycroft.skill.intent brings in the folder's __init__.py, not the intent.py file.
* The msm.sh script isn't necessarilly on the system, thus skill installation breaks.

augustnmonteiro:
Fixed!

PR #660, opened 2017-04-14
Fix for issue #659, broken git install after recent changes

penrods:
There are several issues with the changes recently made to break out
Skills from mycroft-core.
* The subdirectory mycroft/skills/intent/ is left behind when you do
  a pull, so the change to core.py that tries to
  "import mycroft.skills.intent" gets confused.  I renamed the file
  intent.py to intent_service.py
* Added a bunch of comments
* Made a bunch of functions "private", using the leading _
* The MSM install was failing.  Still working on that, but we should
  probably make msm.sh a part of mycroft-core instead of its own
  repo.
* Restored some mycroft.conf values that I noticed got lost during
  an automatic merge.

### DIFF
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -119,7 +119,7 @@
   // Override: SYSTEM (e.g. Picroft)
   "enclosure": {
     // Platform name (e.g. 'Picroft', 'Mark_1'
-    // Override: SYSTEM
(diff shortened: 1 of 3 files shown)


penrods:
@augustnmonteiro , let's add checks for an installed 'msm' to the start.sh and mycroft.sh scripts, prompting the dev to re-run the install script if they don't have them already.

PR #662, opened 2017-04-14
#661 - Auto installing msm on start.sh

augustnmonteiro:


penrods:
Please add a notification to the user.  e.g.

if [ ! -d "${TOP}/msm" ]; then
	echo "Installing Mycroft Skill Manager tool..."
	${TOP}/scripts/install-msm.sh &> /dev/null
else
...

Then we are good to go...

start.sh
```diff
@@ -2,6 +2,14 @@
 TOP=$(cd $(dirname $0) && pwd -L)
 VIRTUALENV_ROOT=${VIRTUALENV_ROOT:-"${HOME}/.virtualenvs/mycroft"}
 
+if [ ! -d "${TOP}/msm" ]; then
+	${TOP}/scripts/install-msm.sh &> /dev/null
```
penrods: Insert:  echo "Installing Mycroft Skill Manager tool..."

augustnmonteiro: Added! : D

PR #669, opened 2017-04-16
Drop rpmfusion from Fedora install script

nullr0ute:
From Fedora 25 mpg123 is now included in the main Fedora repositories
so we don't need to add the rpmfusion repos for the build dependencies.

Signed-off-by: Peter Robinson <pbrobinson@gmail.com>


aatchison:
Very cool. Merging.

PR #670, opened 2017-04-16
Attempt to reenable (most) of the unittests

forslund:
This is an attempt to get the unittests up and running with travis, based on previous work by @ethanaward. This PR will probably fail on travis as it is right now but will be updated disabling the failing parts until a passing unittest remains. This branch can then be merged and more unittests can be added and re-enabled.

Also I'm not sure how pairing will work for travis...

- The more complex failing tests have been disabled
- Pairing and wolfram alpha skill tests have been removed
- GoogleSTT has been replaced with MycroftSTT
- fixed local recognizer initialization


forslund:
Ok, this actually seems to be working... Even after a rebase!

Currently there are 8 disabled tests, some of these are disabled because I don't know enough about the underlying code to determine if the test is incorrect or if there is an actual issue.

A couple of tests are disabled since they call the mycroft API and will fail since travis isn't paired. In my opinion this is the big issue for future updates, the unittests should not rely on an external server at all. Instead the API calls should be mocked.

I do think this branch could be merged so we'd get a small base of unittests that we can expand on.

aatchison:
Hurray! Thanks to you!

Issue #674, opened 2017-04-17
Audio accuracy test is broken


ethanaward:
The audio accuracy test does not work right now, as it hasn't been updated for recent wakeword recognizer changes.

PR #675, opened 2017-04-17
#674 - Update audio accuracy test

ethanaward:
This fixes the audio accuracy test so that it can be used again, as well as moves it into the mycroft/ directory and makes sure any data files used to test it won't be included by git.


aatchison:
Thank you sir

Issue #676, opened 2017-04-19
Auto loading is watching .git folder modification


augustnmonteiro:
Auto load should not watch .git or any hide folder

PR #677, opened 2017-04-19
#671 - making auto load doesn't watch hidden folders changes

augustnmonteiro:
Fix for issue #676

### DIFF
--- a/mycroft/skills/main.py
+++ b/mycroft/skills/main.py
@@ -102,14 +102,15 @@ def _load_skills():
 
 
 def _get_last_modified_date(path):
-    dates = []
+    last_date = 0
     # getting all recursive paths
     for root, _, _ in os.walk(path):
         f = root.replace(path, "")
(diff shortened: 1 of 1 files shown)


mycroft/skills/main.py
```diff
@@ -101,6 +101,17 @@ def _load_skills():
     skill_reload_thread.start()
 
 
+def _get_last_modified_date(path):
```
penrods: No reason to over-optimize, but this can also be a lot easier and a better example.  The list of dates is unnecessary.  Do something like this:

def _get_last_modified_date(path):
    last_date = 0
    # getting all recursive paths
    for root, _, _ in os.walk(path):
        f = root.replace(path, "")
        # checking if is a hidden path
        if not f.startswith(".") and not f.startswith("/."):
            last_date = max(last_date, os.path.getmtime(path + f))
    return last_date

(The spacing is getting trashed in comments)

Issue #680, opened 2017-04-21
msm, wrong listing of skills and wrong copyrigth.


SoloVeniaASaludar:
1) One of the particularities of bash is that in a loop like:

```
    for f in *; do
      echo $f
    done
```

$f takes the value of all files in the current directory if there are any one, but it takes value '*' if no file exists.

The correct way to iterate over each subdirectory of current one, covering the case of none exists, is:

    for f in $(ls -d */); do
      ...

2) msm file starts with:

     Copyright (c) 2017 Mycroft AI, Inc.  All rights reserved.\n"

please, remove. Mycroft is open-source.

PR #682, opened 2017-04-21
#680 - error in msm script

SoloVeniaASaludar:
See issue #680

### DIFF
--- a/msm/msm
+++ b/msm/msm
@@ -30,7 +30,6 @@ echo "#######  Mycroft Skill Manager #######"
 
 function help() {
 	echo "msm: Mycroft Skill Manager"
-	echo -e "     Copyright (c) 2017 Mycroft AI, Inc.  All rights reserved.\n"
 	echo "usage: msm install <repository> or <name>"
 	echo "     Installs the given Skill into the /opt/mycroft/skills directory"
 	echo "     where <repository> is the address of the skill in Github."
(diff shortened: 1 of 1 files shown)


penrods:
Hey, thanks for this fix!  Did you intentionally delete the copyright line?

SoloVeniaASaludar:
Yes, I do. Is not mycroft open-source?

penrods:
It is!  But having a copyright does not stop it from being open source.  In fact, by granting the copyright to Mycroft you are insuring it remains open source.

Think of it this way:  If one doesn't explicitly grant the copyright to the project, then it is ambiguous.  That means if something occurs that requires a court to make a decision, then they have to reach out to *everyone* who has contributed to that non-copyrighted project, which could be hundreds or thousand of people.

See: http://www.infoworld.com/article/2839560/open-source-software/sticking-a-license-on-everything.html

SoloVeniaASaludar:
Of course they must have a license (or be public domain) but this license must respect the four freedoms. "All rights reserved", the supressed phrase, does not.
In fact, usual source files in mycroft starts with a header, copyleft, in perfect agreement with foss.

penrods:
I didn't notice that the original checkin held the "all rights reserved" phrase.  Please leave that for us to correct in an independent PR.  I'm reviewing all of the copyright and licensing (it's a big messy process), and we want to be clean when we change anything related to this to prevent legal confusion.  If Mycroft has reserved rights, you aren't able to legally make that change.  Make sense?

SoloVeniaASaludar:
Done.

(PS: it is perfectly legal I suggest any change.)

penrods:
Of course, no qualms about your suggestion -- it exposed a problem to me.  And while I don't honestly think it would cause any long term damage, I'd rather not have it look like you were the one who changed the license, even if it was just for a second.  I've been through a stupid lawsuit before and I don't want to repeat it.

Thanks again for the fix and for pointing this out!

Issue #688, opened 2017-04-21
merge of config files overrides previous data


SoloVeniaASaludar:
In addition to the default mycroft.conf with the official content,
let a config file /etc/mycroft/mycrof.conf with content:

```
{
  "listener": {
   "device_index": 1
  }
}

```
I expected (could be incorrectly) that "listener" configuration in default file was keep, changing **only** the item present on this file. However, all content of the block "listener" seems to be lost, overrriden by the content in /etc/mycroft/mycroft.conf.

Merge of configuration should be at entry level, not at block level.

PR #689, opened 2017-04-22
#688 - merge config files

SoloVeniaASaludar:
Allow coexistence of default, system and user config files.
See issue #688 for details.

### DIFF
--- a/mycroft/configuration/__init__.py
+++ b/mycroft/configuration/__init__.py
@@ -98,10 +98,20 @@ class ConfigurationLoader(object):
         return config
 
     @staticmethod
+    def merge_conf( base, delta ):
+        for k, dv in delta.iteritems():
+            bv = base.get(k)
+            if isinstance( dv, dict ) and isinstance( bv, dict):
+                ConfigurationLoader.merge_conf( bv, dv )
+            else:
+                base[k]=dv
+
+    @staticmethod
     def __load(config, location):
         if exists(location) and isfile(location):
             try:
-                config.update(load_commented_json(location))
+                ConfigurationLoader.merge_conf(
+                    config, load_commented_json(location))
                 LOG.debug("Configuration '%s' loaded" % location)
             except Exception, e:
                 LOG.error("Error loading configuration '%s'" % location)
--- a/mycroft/configuration/__init__.py
+++ b/mycroft/configuration/__init__.py
@@ -98,13 +98,13 @@ class ConfigurationLoader(object):
         return config
 
     @staticmethod
-    def merge_conf( base, delta ):
+    def merge_conf(base, delta):
         for k, dv in delta.iteritems():
             bv = base.get(k)
-            if isinstance( dv, dict ) and isinstance( bv, dict):
-                ConfigurationLoader.merge_conf( bv, dv )
+            if isinstance(dv, dict) and isinstance(bv, dict):
+                ConfigurationLoader.merge_conf(bv, dv)
             else:
-                base[k]=dv
+                base[k] = dv
 
     @staticmethod
     def __load(config, location):
--- a/mycroft/configuration/__init__.py
+++ b/mycroft/configuration/__init__.py
@@ -99,6 +99,14 @@ class ConfigurationLoader(object):
 
     @staticmethod
     def merge_conf(base, delta):
+        """
+            Recursively merging configuration dictionaries.
+
+            Args:
+                base:  Target for merge
+                delta: Dictionary to merge into base
+        """
+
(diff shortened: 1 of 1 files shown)


mycroft/configuration/__init__.py
```diff
@@ -98,10 +98,20 @@ def load(config=None, locations=None, keep_user_config=True):
         return config
 
     @staticmethod
+    def merge_conf(base, delta):
```
forslund: Would you mind adding a short docstring here explaining the function.

Something along the lines

```python
    """
        Class method recursively merging configuration dictionaries.

        Args:
            base:  Target for merge
            delta: Dictionary to merge into base
    """

forslund:
Other than the minor detail of the docstring this looks really good and works great! This will be really useful!

SoloVeniaASaludar:
Updated with the addition of the comment.
Nice it was useful.

SoloVeniaASaludar:
Do you agree to add two more config files to the loading process ?

1) .../mycroft/configuration/mycroft.conf 
2) .../mycroft/configuration/mycroft.conf.<lang> 
3) /etc/mycroft/mycroft.conf
4) /etc/mycroft/mycroft.conf.<lang>
5) .../.mycroft//mycroft.conf

- Files "2" and "4" are the new ones that are add to the configuration list of files. They will contain i18n deltas. 
- The official mycroft distribution will contain only file "2". Example: .../mycroft/configuration/mycroft.conf.en-US, .../mycroft/configuration/mycroft.conf.es, etc.
- It seems not necessary a .../.mycroft/mycroft.conf.<lang> because user config file is expected be in the native user language.

forslund:
I'm not entirely sure I get the reasoning here.

Why there should be language specific config files outside the normal ones. The config in `/etc` and `~/.mycroft` system/user specific. It seems unlikely that the user would use two config files in `~/.mycroft/` when one would do just fine.

What sort of configuration do you propose to put in the language specific config files?

forslund:
I'm not even sure the language is defined until after the `mycroft/configuration/mycroft.conf` is loaded. I think this would be fetched from one home.mycroft.ai or the user/system configs.

penrods:
Which elements do you envision translating?  I can see having different STT or TTS engine depending on the language being spoken.  I think I'd prefer things in one file with a language code spelled to the element, e.g. "tts-en-us".

SoloVeniaASaludar:
Hi,

1) About the necessity of the files, by example, nowadays in file skill/core.py there are this statement:

                # TODO: Localize
                self.speak(
                    "An error occurred while processing a request in " +
                    self.name)

If this hard-coded English phrase is moved to config file, a good one could be "mycroft/configuration/mycroft.conf.en-US". In this way, internalization needs only create a new file "mycroft/configuration/mycroft.conf.es" for Spanish.

2) About the availability of "lang", yes, language could be not defined in mycroft/configuration/mycroft.conf. In this case, a better order of loading of config file could be:

- .../mycroft/configuration/mycroft.conf
- /etc/mycroft/mycroft.conf
- .../.mycroft//mycroft.conf
- .../mycroft/configuration/mycroft.conf.<lang>
- /etc/mycroft/mycroft.conf.<lang

Kind regards.

SoloVeniaASaludar:
With previous, internalization is allowed by the configuration subsystem. Looking further, there are two more points to be considered: 

a) remote configuration.

Remote configuration could be considered a new configuration delta, to be loaded between files 3 and 4 of previous list, or after the last one.

b) language based on wake word.

Language could be selected taken into account he wake word: "Hey, mycroft" => en; "Hola, mycroft" => es. In this case, language dependent config files must be loaded after wake word recognition.

penrods:
I have created a localization mechanism for those embedded strings, I need to stay a PR so you can see it.  Basically, I am creating a structure under mycroft/res/text similar to the dialog folders in skills.  I don't consider these "configurations", but this will expose it for translation.

I'll update this with a link once I have the PR pushed.

SoloVeniaASaludar:
ok, thanks.

forslund:
I agree that the remote configuration should be considered just another delta. I think the current resolution order is
`.../mycroft/configuration/mycroft.conf`
Remote config
`/etc/mycroft/mycrof.conf`
`~/.mycroft/mycroft.conf`

All these could be considered deltas if starting with an empty dict as base.

I am still reluctant to adding more config files, the number of places to check if some config value gets overridden to an incorrect value should be kept to as few as possible.

forslund:
I think this PR looks good right now. Works great and is nicely documented

PR #703, opened 2017-04-27
Bugfix/issue 702

### LINKED ISSUE #702: ipc signals fails
interprocess communication signals may fail if the `$TMP/mycroft/ipc/signal` direcory is missing. The creation needs to make sure the directory exists before creating the signal file.

forslund:
Uses `create_file()` to create the signal file, ensuring the directory is created.

Also adds basic test cases for the function. Resolves #702


kfezer:
Looks good to me.

PR #712, opened 2017-04-28
	Adding SSH enable option

kfezer:
modified:   client/enclosure/__init__.py

Adds a listener for a button-pressed induced SSH enable.


kfezer:
This isn't working yet...

kfezer:
This still isn't right. Thinking about the best way to do this.

kfezer:
This should work as of now.

Issue #610, opened 2017-03-31
i18n: language codes are not always two words.


SoloVeniaASaludar:
Language codes are not always one pair of two letters with "-" as separator. ISO 639 / BCP 47 / IETF language tags (see [here](https://en.wikipedia.org/wiki/IETF_language_tag#ISO_639-3_and_ISO_639-1)) is a near than incomprehensible standard. 

In short, we can say that valid language codes are "es", "en-US", someone with three parts, ... .

Code in stt/__init__.py expects always two words. It could be modified to accept remainder formats.

eccco3:
Is this something that a relative beginner could tackle? I have about 6 months experience with using Python and 1 year of experience programming before that.

SoloVeniaASaludar:
Tests can be done changing in conf file "lang" from default "en-US" to "es".

PR #717, opened 2017-05-02
Adding localization mechanism for strings embedded in mycroft-core code

penrods:
Added mycroft.dialog.get() function.  This behaves much like the localization
mechanism for dialogs in Skills.  So you can do things like this:

   lang = "en-us"
   str = mycroft.dialog.get("how are you", lang)

Which will look in mycroft/res/text for the dialog file containing strings to
use as templates for the actual output.  This depends on the language being
currently used.  When operating in English this would be:

   mycroft/res/text/en-us/how are you.dialog

This function will pick a random line from that file to assign to str.

A more advanced use is to embed placeholders in the strings within the
template file.

   lang = "en-us"
   ctx = {"time" : "noon"}
   str = mycroft.dialog.get("current time", lang, ctx)

And the random template line picked was "the current time is {{time}} ", then
the output would be:

   "the current time is noon"

### DIFF
--- a/mycroft/client/speech/listener.py
+++ b/mycroft/client/speech/listener.py
@@ -25,6 +25,7 @@ from pyee import EventEmitter
 from requests import HTTPError
 from requests.exceptions import ConnectionError
 
+import mycroft.dialog
 from mycroft.client.speech.local_recognizer import LocalRecognizer
 from mycroft.client.speech.mic import MutableMicrophone, ResponsiveRecognizer
 from mycroft.configuration import ConfigurationManager
@@ -109,7 +110,7 @@ class AudioConsumer(Thread):
                                                 self.metrics):
             SessionManager.touch()
             self.state.sleeping = False
-            self.__speak("I'm awake.")
+            self.__speak(mycroft.dialog.get("i am awake", self.stt.lang))
             self.metrics.increment("mycroft.wakeup")
 
     @staticmethod
@@ -141,7 +142,8 @@ class AudioConsumer(Thread):
             LOG.error("Could not request Speech Recognition {0}".format(e))
         except ConnectionError as e:
             LOG.error("Connection Error: {0}".format(e))
-            self.__speak("Mycroft seems not to be connected to the Internet")
+            self.__speak(mycroft.dialog.get("not connected to the internet",
+                                            self.stt.lang))
         except HTTPError as e:
             if e.response.status_code == 401:
                 text = "pair my device"
@@ -149,7 +151,8 @@ class AudioConsumer(Thread):
         except Exception as e:
             LOG.error(e)
             LOG.error("Speech Recognition could not understand audio")
-            self.__speak("Sorry, I didn't catch that")
+            self.__speak(mycroft.dialog.get("i didn't catch that",
+                                            self.stt.lang))
         if text:
             # STT succeeded, send the transcribed speech on for processing
             payload = {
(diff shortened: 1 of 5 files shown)


SoloVeniaASaludar:
Thanks a lot for this feature, it will be very useful in i18n.

Can I suggest include in dialog.get or dialog.resolve_resource_file a replacement of spaces by underscores? Filenames with spaces are a source of troubles.

mycroft/dialog/__init__.py
```diff
@@ -108,3 +108,28 @@ def load(self, dialog_dir):
...
+        phrase (str): resource phrase to retrieve/translate
+        lang (str): the language to use
+        context (dict): values to be inserted into the string
+
+    Returns:
+        str: a randomized and/or translated version of the phrase
+    """
+
+    filename = resolve_resource_file("text/"+lang+"/"+phrase+".dialog")
+    print (filename)
```
forslund: This is a debug message and I don't think it should be included. at least not as a *print*

mycroft/dialog/__init__.py
```diff
@@ -108,3 +108,28 @@ def load(self, dialog_dir):
...
+    if not filename:
```
forslund: I'd like the failure to find a file to log an error or at least a debug message to tell developer that something is wrong.

mycroft/client/speech/listener.py
```diff
@@ -141,15 +142,17 @@ def transcribe(self, audio):
...
         except HTTPError as e:
             if e.response.status_code == 401:
                 text = "pair my device"
                 LOG.warn("Access Denied at mycroft.ai")
         except Exception as e:
             LOG.error(e)
             LOG.error("Speech Recognition could not understand audio")
-            self.__speak("Sorry, I didn't catch that")
+            self.__speak(mycroft.dialog.get("i didn't catch that",
+                                            self.stt.lang))
```
forslund: There's an issue with the `stt.lang` variable in this case. It forces the part after the `-` to be upper case, for example `en-us` becomes `en-US` see `mycroft/sst/__init__.py` L44, This will not match the directory structure.

Alternatives:
- `dialog.get()` should use `lang.lower()` or convert before sending
- use `config_core.get('lang', 'en-us')` where `self.stt.lang` is used
- change the way the STT stores the lang variable (and do the upper case conversion when the value is used if needed)

penrods: Yeah, I've been fighting this for a while... the official lang codes are mixed case.  But all the Skills in the world already use lowercase in the file structure.  I think using the lang.lower() within the dialog.get() is probably an acceptable convention.  I can't imagine anyone wanting to distinguish "en-US" from "en-us" or "en-Us".

mycroft/dialog/__init__.py
penrods: Yup, that was debugging.  Good catch.

mycroft/dialog/__init__.py
penrods: Will do

SoloVeniaASaludar:
Can I suggest replace space by underscore in "dialog.get" or "dialog.resolve_resource_file"? Filenames with spaces are a source of troubles.

About language codes, they can be one, two or more items with dash separator. See issue #610, pull requests #611. In addition, in a startup, it is interesting to have and allow a single "es", to cover all Spanish users, at least until more specifically regional translations exists. (Moreover, there are long discussions about if "es-XX" is correct from a point of view of grammatical normative).

forslund:
@SoloVeniaASaludar we'll add that to `find_resource_file()` (or similar) and treat it as a separate issue.

PR #720, opened 2017-05-02
Feature/wifi reset

kfezer:
Adds resetting of wifi networks to Reset

Also enables other options to call wifi reset easily using ws


aatchison:
nice

PR #725, opened 2017-05-03
	Adding an update time feature once Mycroft is connected (NTP)

kfezer:
modified:   mycroft/client/wifisetup/main.py
This forces mycroft Mark1 to update it's local time as soon as it connects.


kfezer:
Not working yet...

PR #730, opened 2017-05-05
Adding a script to generate a skill

augustnmonteiro:
* Creating a skill inside skill folders
* Creating the whole skill structure
* Creating a basic __init__


kfezer:
Do we want to do this as a script and not as python code? Like a method in the API?

PR #742, opened 2017-05-08
Add support for switching tts and tts-settings

### LINKED ISSUE #741: TTS doesn't update when websettings are changed
Currently to change voice, TTS type etc. mycroft need to be completely restarted. Like location this should be updated on the fly.

forslund:
- If tts config hash has changed re-initialize tts
- Cache is cleared on mimic initialization

This will allow the vocie system to be updated on the fly when web settings are changed. Partially resolves #741


forslund:
Replaced by #775

PR #743, opened 2017-05-08
Improved handling of button press

penrods:
Improved the way the button press is handled.  Now the listener gives the rest
of the system a moment to consume the button press before it interprets it
as a request to begin listening.

Also replaced the use of the "buttonPress" signal when "expect_response"
was indicated with an utterance.  Now the signal "startListening" is
created, differentiating it from a buttonPress.

### DIFF
--- a/mycroft/client/speech/mic.py
+++ b/mycroft/client/speech/mic.py
@@ -260,6 +260,8 @@ class ResponsiveRecognizer(speech_recognition.Recognizer):
             recorded_too_much_silence = num_chunks > max_chunks_of_silence
             if quiet_enough and (was_loud_enough or recorded_too_much_silence):
                 phrase_complete = True
+
+            # Pressing top-button will end recording immediately
             if check_for_signal('buttonPress'):
                 phrase_complete = True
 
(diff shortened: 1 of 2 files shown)


mycroft/client/speech/mic.py
```diff
@@ -303,11 +305,22 @@ def _wait_until_wake_word(self, source, sec_per_buffer):
...
+            # used to mean 'stop' instead)
+            if check_for_signal('buttonPress', 1):
+                # give others time to consume this as 'stop'
+                sleep(0.25)
+                if check_for_signal('buttonPress'):
+                    # signal is still here, assume it was to begin recording
+                    logger.debug("Button Pressed, wakeword not needed")
+                    said_wake_word = True
+                    continue
+
```
forslund: Is this done in preparation for future architecture change? To me it looks like the buttonPress can't be used in two places at once. The only other place is in series with this check. (while recording to send to STT)

In any case we might want to move this to a separate function `check_for_button()` (or similar) and have the if statement at L309 check both criterias
`if check_for_signal('startListening') or check_for_button():`

The `_wait_for_wakeword()` method is complex enough as it is without logic for whether the buttonPressed signal means stop or start recording.

penrods: There is an architectural change needed.  Basically, the way "stop" is implemented using the message bus, we don't now if a Skill consumed the button press.  The mechanism I tossed in there using the 1/4 second sleep allows me to detect if the mimic_tts process was active (i.e. if the unit was talking) and it consumes the buttonPress to stop speaking and viseme playback.  But there is no way to detect other consumptions, e.g. if NPR news is playing and the button press is just meant to stop that.

I might do something hacky for short-term such for anything that actually would need to "stop" independent of speaking -- such as NPR news or music playback.  I can have those call a function that consumes the signal.  I'll probably have them call a MycroftSkill.do_stop() helper function.

In the longer term, we need a SkillManager that is aware of the Skill processes that we have launched.  It will invoke the method on those Skills and that method will return either True or False depending on whether they stopped.  Then the SkillManager will be the one that consumes the buttonPress.

Make sense?

penrods: I added the "startListening" to make it clear that the request was from the framework and explicitly about listening, not to be confused with the consumption of the buttonPress.

penrods: I'll move both check in that hunk of code into a wrapper method, something like

```
   if _force_listen_without_wakeword():
        said_wake_word = True
        continue
```

penrods:
I'm going ahead and pulling this one in so we can build a pre-release

PR #744, opened 2017-05-08
Support for Mark 1 SSH > BLOCK menu item

penrods:
This implements the handler for the Mark 1 menu item SSH > BLOCK (the inverse of SSH > ALLOW)
* Added handler for "unit.disable-ssh" on the serial line. This turns around and emits "mycroft.disable.ssh" on the messagebus.
* Removed the automatic reboot, just let the user know it will be different after a restart
* Made the spoken message translatable
* Changed mycroft.dialog.get() to not require the "lang" parameter.  It will default to the mycroft.conf value.


penrods:
@kfezer This is the PR we were talking about

PR #752, opened 2017-05-11
This tests the messagebus for mycroft.

amcgee7:
This is just a basic test of sending between websocket clients.  It makes sure the message gets to the other side. It doesn't check for correct ness.  More testing sure could be done.

Signed-off-by: Art McGee <amcgee7@jaguarlandrover.com>


amcgee7:
@aatchison MyTesting Starts up the mycroft service.  Is that already started.  Maybe I should consider removing that action?

forslund:
This goes into the weird borderland between unit test and integration test. Technically launching an entire service for a unittest is not a good idea. In this case there may be complications since the travis instance isn't paired with mycroft home.

The `TestMessageMethods` unittests looks ok (as far as I can see).

I'm not sure why the `onHandle1` doesn't get called.

Looking at the log it may be an issue with loosing track of the threads created to run the websocket handlers. The framework may be waiting for those to shut down.

augustnmonteiro:
Please rebase : D

forslund:
Could you try doing a `git rebase dev` and cleaning up the commit history. It's kind of hard reviewing right now.

amcgee7:
@forslund Did we talk about you doing a pic on this?  I can rebase if needed.

forslund:
I'm going to cherrypick the relevant changes to another PR. Leave it, I'll
close it when I'm done.

Den 30 juni 2017 4:42 em skrev "Art McGee" <notifications@github.com>:

forslund:
I'm moving the relevant commit into a branch for adding integration tests. Closing this PR.

Issue #715, opened 2017-05-01
bug - call system.platform() fails


JarbasAI:
if i do 

```
import platform
print platform.system()
```

 works, but not if inside a skill, i get:

> Traceback (most recent call last):
>   File "/home/user/mycroft-core-base/mycroft/skills/core.py", line 106, in load_skill
>     skill = skill_module.create_skill()
>   File "/home/user/mycroft-skills/mycroft-facebook-skill/__init__.py", line 977, in create_skill
>     return FacebookSkill()
>   File "/home/user/mycroft-skills/mycroft-facebook-skill/__init__.py", line 760, in __init__
>     print platform.system()
>   File "/usr/lib/python2.7/platform.py", line 1333, in system
>      return uname()[0]
>    File "/usr/lib/python2.7/platform.py", line 1300, in uname
>      processor = _syscmd_uname('-p','')
>   File "/usr/lib/python2.7/platform.py", line 1035, in _syscmd_uname
>     rc = f.close()
> IOError: [Errno 10] No child processes

seems related to:
https://issues.apache.org/jira/browse/SPARK-1394?focusedCommentId=14046349&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14046349

this breaks a few packages, so far got problems on selenium

forslund:
If I recall correctly this is due to the handling of SIGCHLD, the current
handler interferes with the python internals.

I'll see if I can find a better way to clean up subprocesses.

Den 1 maj 2017 8:32 em skrev "JarbasAI" <notifications@github.com>:

SoloVeniaASaludar:
Doing some tests about this issue...

1) it is true that signal(SIGCHLD,SIG_IGN) breaks platform.system() call. Platform still uses deprecated "os.popen".

2) After an statement as "subprocess.Popen(["uname","-p"])", a defunct process appears. It still on this state until garbage collection does a cleanup (for testing, it can be forced with "gc.collect()").

3) After a call as "x=subprocess.Popen(["uname","-p"])" a defunct process appears. It still on this state until x is free and garbage collection does a cleanup. A way to force it is:

```
x=subprocess.Popen(["uname","-p"]
del(x)

```
this one seems the better way to force the dead of the zombie child.


In conclusion: call to "signal" can be suppressed. If the skill needs immediate finish of dead child, use the code said in point 3.

My apologizes for suggesting a bad change.

forslund:
[...]

PR #754, opened 2017-05-13
Remove ignoring SIG_CHLD

forslund:
Ignoring `SIG_CHLD` can be convenient but parts of the python standard
library relies on the behavior of `SIG_CHLD` and will not work correctly
in this configuration.

The best way to handle defunct processes is for the skill creator to handle them (by removing their reference or waiting for completion). This has been discussed somewhat in #715, there may be a possibility to create a watchdog loop to shut down defunct processes but would be kind of a blunt sword and would probably cause other more sneaky issues.

### DIFF
--- a/mycroft/skills/container.py
+++ b/mycroft/skills/container.py
@@ -25,10 +25,6 @@ from mycroft.messagebus.client.ws import WebsocketClient
 from mycroft.skills.core import create_skill_descriptor, load_skill
 from mycroft.skills.intent_service import IntentService
 from mycroft.util.log import getLogger
-import signal
-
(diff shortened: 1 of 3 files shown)


penrods:
@SoloVeniaASaludar I'd like your input on this before I merge it, as you submitted the original "ignore" code.

augustnmonteiro:
Please rebase! : D

forslund:
Rebased!

SoloVeniaASaludar:
I agree is best to remove the original change. As it causes conflicts, it is best to add this signal handling only in specific skills, when there no other alternative, and with effect local to the skill.

PR #756, opened 2017-05-14
converse, intent_parser, external shutdown/reload, priority and blacklist skills from config

JarbasAI:
- blacklisted skills now read from config
- added priority skills to be loaded first (any skill that needs to listen to register messages should load first) read from config
- skills now have unique ID
- converse method
- intent service now listens for external messages
- intent parser helper class, get intent from utterance, source skill from intent
- skills main now listens for external messages for skill_manifest, skill_shutdown and skill_reload
- enable and disable intent can be requested from outside

allows continuous dialog by using converse #539 method and giving intent parsing tools inside any skill

enable and disable intent can be requested from outside skill, allows for more tools like [intent layers](https://github.com/JarbasAI/service_intent_layer) class

external shutdown and skill reload allow for run levels, i am implementing a [control center](https://github.com/JarbasAI/JarbasAI/tree/dev/mycroft/jarbas-skills/skill_control_center) skill where i already do this, i have a "core" run level, a "offline" run level and a "dev" run level

priority skills is needed for example in my [objectives](https://github.com/JarbasAI/service_objectives) skills that needs to listen for register_objective messages from other skills, any "manifest" or "context" skill would also need this


forslund:
Hi @JarbasAI great work and a huge PR. It'll take me some time to wrap my head around all this.

I'm not sure about the approach to the priority skills. Did you consider to look for a `requirements` member in the skill module.

It would look something like this:

```python

import stuff

requires = ['skill_a']

class skill_b(MycroftSkill)
    [...]

def create_skill()
    return skill_b()
```

If a `requires` member is found the list of already loaded skills are checked and if the required skills are already loaded the skill is loaded. Otherwise the skill is put aside for later. After the normal loading procedure is finished the list of skills that couldn't be loaded due to requirements issues are re-evaluated in a loop until a pass fails to load a new skill.

(I'm sure even smarter approaches are possible)

This will remove the need for a configuration parameter telling which skills are needed by others, multiple levels of dependencies would work out of the box and if skills need to be reloaded after a priority skill is reloaded this can be handled too.

If you like I can whip up a proof of concept to show you what I mean in more details.

mycroft/skills/intent_service.py
```diff
@@ -37,6 +37,79 @@ def __init__(self, emitter):
...
+        self.skill_ids = {} # {skill_id: [intents]}
+        self.converse_timeout = 5  # minutes to prune active_skills
+
+    def do_conversation(self, utterances, skill_id, lang):
+        self.emitter.emit(Message("converse_status_request", {
+            "skill_id": skill_id, "utterances": utterances, "lang": lang}))
+        self.waiting = True
+        self.result = False
+        while self.waiting:
+            pass
```
forslund: Won't this consume a whole lot of cpu? or is there some async thing that protects?
In other cases a `time.sleep(0.1) might be a good idea

Maybe a timeout if the signal never arrives.

forslund:
I realize now that what I might have misunderstood your intention with this.

I still think a configuration parameter for this will create issues. Have you considered to use a separate directory for priority skills instead of the normal `/opt/mycroft/skills/` the priority skills are housed in `/opt/mycroft/priority-skills/` and that directory is loaded before the normal skills...

What register messages are we talking about here like intent registration or something different?

mycroft/skills/intent_service.py
JarbasAI: hum, must have missed this ot copy pasted wrong code, i usually put a timer in this case, what would a good time_out be? 5 seconds? ive been playing with 20 but thats way too much

will implement those changes

JarbasAI:
Its just like intent registration, in the objectives skill where i need this case, its something akin to the intent service skill, but chooses one randomly to execute, there's an helper class to use "objectives" inside any skill, and the execution is handled inside this skill. so this must be loaded before any skill that registers objectives. the requirements idea could work in this case

but what about a manifest or context skill that just wants to consume any kind of messages from all skills? that would need a priority load also, and it wouldn't be a requirement

JarbasAI:
added a handler in intent service to set any skill in active list, allows passive listeners (do stuff, return False) to ensure converse method is called, i was emitting utterances that triggered dummy intents for this periodically, this way is better

some more changes and code cleanup in shutdown/reload, its now possible to better manage what is loaded and shutdown

augustnmonteiro:
Please rebase : D

forslund:
@JarbasAI The branch is mergeable but it looks like the history is still a bit messed up. A proper `git rebase dev` might be difficult (sorry about this) instead what I've been doing with my PR's is to 
1. Make a backup of the branch `git checkout -b [BRANCH]-bak`
2. Switch to dev and pull latest from the repo
3. Delete the original branch
4. create a new branch with the same name from latest dev
5. use `git cherry-pick [COMMIT]` to copy the commits from the backup branch

It's a bit heavy on manual input but works nicely. There is probably some even better way but this is what I've used.

PR #760, opened 2017-05-16
Wifi setup changes

penrods:
Several changes surrounding the wifi setup process:
* The prompt to run wifi-setup is automatic on at startup if no connection detected
* Mycroft-core can initiate the same prompt if internet is lost later
* Skills manager doesn't attempt update unless connection is detected
* Added translate-able dialog for hard-coded strings
* Minor typo:  it had "diferent" instead of "different"


penrods:
Updated the description, expanding it to general "wifi setup" changes.

PR #775, opened 2017-05-21
adding in docker setup script

btotharye:
fixing rebase and adding in docker setup file.


augustnmonteiro:
In this case wasn't better add the dependencies on the Dockerfile?

btotharye:
Yea I just didn't know if you wanted it in core or on the docker repo either works

PR #789, opened 2017-05-23
Issues 788 - Adding remove_all_listeners

gregmccoy:
Adding method to EventEmitter that was missing


ethanaward:
@forslund Are you still concerned with this? It seems like it would be a pretty trivial change to make

forslund:
I think it would be good to have and pretty simple to do. I can do a quick update based on this PR tomorrow morning.

forslund:
@gregmccoy Thank you for your contribution, we merged it into #854 and it's now included in the dev branch.

PR #794, opened 2017-05-26
Make wake word listener have length depending on number of phonemes

### LINKED ISSUE #778: the WuW is not detected in slow speech or long WuW case
The hard coded duration of the wake up word, set to 1 sc,  is not suitable for slow speech or when the configured WuW is long. To handle such cases, the duration may be estimated from the phonological representation of the WuW.

As the phoneme duration highly depends on the speaking rate, the category of the phoneme and its context, a precise estimation will be hard, furthermore it is not required for the given purpose. So an overall estimation can be accepted.

From a fast review of some phonetical measures, the average duration of a vowel/consonant in different speaking rates is :
- fast speech :  vowel=60ms, consonant=30ms; syllabes=90ms
- normal speech : vowel=100ms, consonant=50ms; syllabes=150ms
- slow speech : vowel=250ms, consonant=75ms; syllabes=340ms

I think 100ms per phoneme will be enough, may be +10-20% to handle so slow speech.

possible solution :
- in mycroft.client.speech.mic, ResponsiveRecognizer::__init__() :
self.SAVED_WW_SEC = (wake_word_recognizer.phonemes.count( ' ' ) + 1 ) * 0.12

forslund:
Implementation of @labiadh's suggestion.

The wake up word will scale depending on how many phonemes are in the wake up word and the duration for each phoneme can be configured.

Resolves #778

### DIFF
--- a/mycroft/client/speech/mic.py
+++ b/mycroft/client/speech/mic.py
@@ -130,10 +130,6 @@ class MutableMicrophone(Microphone):
 
 
 class ResponsiveRecognizer(speech_recognition.Recognizer):
-    # The maximum audio in seconds to keep for transcribing a phrase
-    # The wake word must fit in this time
-    SAVED_WW_SEC = 1.0
-
     # Padding of silence when feeding to pocketsphinx
     SILENCE_SEC = 0.01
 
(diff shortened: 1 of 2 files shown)


forslund:
There are models for this but it will always depend a lot on who is speaking. In #778 a general duration of 100 - 120 ms per phoneme was suggested as a decent approximation for normal speech. (Approximates the previous window for *Hey Mycroft* really well)

As the config is pulled from the web at each startup this will have effect right away and should not be dependent on any other PRs.

kfezer:
Sounds good to me!

Issue #790, opened 2017-05-24
utterance source, speech metadata, speech targetting


JarbasAI:
1 - utterances should have a "source" so we can know were they come from
2 - speech commands should have a "mute" so we know if we are supposed to speak it
3 - speech should have arbitrary metadata, like adding source of spoken info
4 - speech should have a "more" so consumers know more speech is coming
5 - clients should check if they are the target of a "speech"

1 - this allows to ignore some utterances, for example if from remote acess
2 - we can use remote acess to "speak, check if the door is locked" to family, or just to query answer silently
3 - wiki skill could append wikipedia link
4 - consumers will know to wait for more speech before processing it
5 - we dont need clients processing utterance at all if target is a skill

Proposal for discussion: https://github.com/JarbasAI/JarbasAI/tree/patch-12

targetting is done crudely:
- intent class makes the target the source of utterance
- on register intent add a handler just to set target on receiving intent message
- receive intent message (with target), store target in self.target and call intent handler
- on speak method add self.target to metadata if none was provided

krisgesling:
My understanding is this now exists thanks to your PR #2461 

If I'm mistaken please reopen

PR #796, opened 2017-05-27
#790  utterance source, speech metadata, speech targetting

JarbasAI:
Proposal for discussion, solves #790 
- adds metadata field for speak messages to send extra info along
- adds source to recognizer_loop utterances data
- on register intent add a handler to update message.context
- on intent_service make the target the destinatary source of utterance, pass message.context to intent execution and intent failure methods

using this with success on facebook chat client, server and webchat to only get relevant messages


JarbasAI:
absolute since dig for message method was added and PR #1408

PR #800, opened 2017-05-31
Feature/install padatious

kfezer:
This adds Matthew's Padatious to the dev_setup script. 

This is still under construction and needs development, but by default, the build instructions are commented out. This will remain that way until more testing and expansion are in place.

For now, nothing will happen as we don't yet have  a repo for this. My intent was to fork it from Mathew's. 

SEE: https://github.com/MatthewScholefield/padatious


scripts/install-pedatious.sh
```diff
@@ -0,0 +1,35 @@
...
+#STILL UNDER CONSTRUCTION
+set -Ee
+
+PEDATIOUS_DIR=pedatious
+CORES=$(nproc)
+PEDATIOUS_VERSION=0.1
+
+# for ubuntu precise in travis, that does not provide pkg-config:
+pkg-config --exists icu-i18n || export CFLAGS="$CFLAGS -I/usr/include/x86_64-linux-gnu"
+pkg-config --exists icu-i18n || export LDFLAGS="$LDFLAGS -licui18n -licuuc -licudata"
```
MatthewScholefield: Once we set up padatious under MycroftAI, we should check if this is necessary.

scripts/install-pedatious.sh
```diff
@@ -0,0 +1,35 @@
...
+
+#install pre-reqs for building
+sudo apt-get install python3 ninja-build build-essential \
+    && pip3 install --user meson
+
+# download and install pedatious
+if [ ! -d ${PEDATIOUS_DIR} ]; then
+    git clone --branch ${PEDATIOUS_VERSION} https://github.com/MycroftAI/pedatious.git
+    cd ${PEDATIOUS_DIR}
+    make -j$CORES
```
MatthewScholefield: Here, and below, instead of using make, it uses:
```
meson build
cd build
ninja clean # Optional; same as make clean
ninja
```

As a result, when using `ninja`, the correct amount of cores are used automatically.

MatthewScholefield:
For the most part, it looks good to me. Only thing is make sure to change `pedatious` to `padatious` in the script (unless we want to rename it which would also be fine).

scripts/install-pedatious.sh
kfezer: That's what I thought and wanted to ask you about.

kfezer:
Should be updated as of now.

scripts/install-pedatious.sh
```diff
@@ -0,0 +1,40 @@
...
+
+
+#install pre-reqs for building
+sudo apt-get install python3 ninja-build build-essential \
+    && pip3 install --user meson
+
+# download and install pedatious
+if [ ! -d ${PADATIOUS_DIR} ]; then
+    #TODO: add versioning later
+    git clone https://github.com/MycroftAI/pedatious.git
```
MatthewScholefield: Change the `e` to `a` and, perhaps the url should be `padatious-mycroft` since that's actually [the component](https://github.com/MatthewScholefield/padatious-mycroft) that will run alongside Mycroft; the other is just a plain library.

scripts/install-padatious.sh
```diff
@@ -0,0 +1,40 @@
...
+    git clone https://github.com/MycroftAI/padatious.git
+    cd ${PADATIOUS_DIR}
+    meson build
+    cd build
+    ninja clean # Optional; same as make clean
+    ninja
+else
+    # ensure pedatious is up to date
+    cd ${PADATIOUS_DIR}
+    make clean 2> /dev/null || true
```
MatthewScholefield: Can be removed

kfezer:
Cool. So why padatious-mycroft and not just padatious? I wasn't going to fork yet just to verify licensing stuff. @penrods this uses a different license.

MatthewScholefield:
We should just create a new repo, pushing the sources directly. Also, I can relicense it to whatever, just let me know what works best. And padatious is a plain C++ library used for intent parsing. It doesn't, however, know where to load intent files from or reply to json queries on the messagebus. This is what padatious-mycroft does. It builds a standalone executable that can be run alongside mycroft to respond to utterances. With Adapt, this is done internally in `mycroft/skills/intent_service.py`. Let me know if there's a better name for it. (Maybe `padatious-service`?).

kfezer:
This was done in: Add Padatious install scripts #830

PR #804, opened 2017-06-05
TTS playback thread

forslund:
Using a thread to play audio allows TTS to generate new speech in the background. This should give muliti-sentence speech a more natural speed for slower devices.

There is also quite a bit of refactoring allowing the other TTS backends to use the caching mechanism and possibly Visimes in the future.


forslund:
Conflicts resolved, merging.

Issue #805, opened 2017-06-05
Requirements.txt has unused dependencies


ethanaward:
Currently, our python requirements.txt file has several dependencies that are no longer used. From what I can tell, these are: certifi, configobj, ndg-httpsclient, pdoc, pyasn1, pyOpenSSL, pyroute2, shortuuid, and urllib5.

MatthewScholefield:
Solved by #806

Issue #705, opened 2017-04-27
AttributeError: 'module' object has no attribute 'SSL_ST_INIT'


BoBeR182:
When installing mycroft-core from the AUR I get the following error when attempting to start the service.
```
systemctl --user status mycroft-service.service mycroft-skills.service mycroft-voice.service
● mycroft-service.service - Mycroft main service
   Loaded: loaded (/usr/lib/systemd/user/mycroft-service.service; disabled; vendor preset: enabled)
   Active: active (running) since Wed 2017-04-26 03:05:26 EDT; 49s ago
 Main PID: 14053 (bash)
   CGroup: /user.slice/user-1000.slice/user@1000.service/mycroft-service.service
           ├─14053 bash /usr/share/mycroft-core/start.sh service
           └─14068 python /usr/share/mycroft-core/mycroft/messagebus/service/main.py

Apr 26 03:05:26 $HOSTNAME systemd[1193]: Started Mycroft main service.
Apr 26 03:05:26 $HOSTNAME bash[14053]: Starting service
Apr 26 03:05:26 $HOSTNAME bash[14053]: 2017-04-26 03:05:26,395 - mycroft.configuration - WARNING - Failed to fetch remote configuration: AttributeError("'module' object has no attribute 'SSL_ST_INIT'",)

● mycroft-skills.service - Mycroft skills provider
   Loaded: loaded (/usr/lib/systemd/user/mycroft-skills.service; disabled; vendor preset: enabled)
   Active: active (running) since Tue 2017-04-25 21:52:28 EDT; 5h 13min ago
 Main PID: 2292 (bash)
   CGroup: /user.slice/user-1000.slice/user@1000.service/mycroft-skills.service
           ├─2292 bash /usr/share/mycroft-core/start.sh skills
           └─2318 python /usr/share/mycroft-core/mycroft/skills/main.py

Apr 26 03:01:21 $HOSTNAME bash[2292]: 2017-04-26 03:01:21,762 - mycroft.messagebus.client.ws - WARNING - WS Client will reconnect in 60 seconds.
Apr 26 03:02:21 $HOSTNAME bash[2292]: 2017-04-26 03:02:21,813 - mycroft.messagebus.client.ws - ERROR - Exception("Uncaught 'error' event.",)
Apr 26 03:02:21 $HOSTNAME bash[2292]: 2017-04-26 03:02:21,814 - mycroft.messagebus.client.ws - WARNING - WS Client will reconnect in 60 seconds.
[...]

Issue #690, opened 2017-04-24
Use python included uuid rather than third party


nullr0ute:
Git commit cd1a36b added uuid (https://pypi.python.org/pypi/uuid/1.30) as a dependency that doesn't support python3 and is unmaintained (last release was 2007). The python standard library already includes a rfc4122 generator so that should be used instead (https://docs.python.org/2/library/uuid.html).

aatchison:
Thanks!

nullr0ute:
I don't believe this is fixed, it still lists the other uuid in the requirements.txt in master

forslund:
Hmmm, it's present in the dev branch too. The package doesn't seem to be neccessary. Mycroft seems to run fine without it installed (using the one from standard lib). @ethanaward is looking over the requirements.txt in PR #805. I'll ask him to look at this aswell.

MatthewScholefield:
Closing since uuid is no longer in the requirements.

PR #806, opened 2017-06-05
#805 Removed unused dependencies from requirements.txt

ethanaward:
This removes several dependencies from requirements.txt, as they are no longer used. I've tested Mycroft without these dependencies on desktop and everything seems to be working. I'd recommend testing on Picroft and the Mark 1 to make sure there aren't any issues there.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,15 +1,9 @@
-shortuuid==0.4.3
 pystache==0.5.4
-configobj==5.0.6
 wikipedia==1.4.0
 requests==2.13.0
-pyOpenSSL==16.2.0
-ndg-httpsclient==0.4.0
-pyasn1==0.1.9
 gTTS==1.1.7
 gTTS-token==1.1.1
 backports.ssl-match-hostname==3.4.0.2
-certifi==2016.2.28
 PyAudio==0.2.8
 pyee==1.0.1
 SpeechRecognition==3.1.3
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,15 +1,9 @@
-shortuuid==0.4.3
 pystache==0.5.4
-configobj==5.0.6
 wikipedia==1.4.0
 requests==2.13.0
-pyOpenSSL==16.2.0
```
forslund: pyOpenSSL is used, I think websockets depend on it (and/or requests). For certain OS's the 16.0.0 package will cause problems and 16.2.0 or 17.0.0 is needed to make mycroft work across platforms.

We can probably remove it and let it be installed implicitly when the package requiring it is installed but it might cause problems if an older OS is upgraded.

ethanaward: So I looked into that earlier, and requests depends on pyOpenSSL only if you use the `security` version of it, which would be requests@security. We don't use that, so we should be good there. What issues were you seeing where pyOpenSSL not being installed led to errors, and what platforms?

forslund: Here's the original issue leading to upgrading the version listed in requirements: https://github.com/MycroftAI/mycroft-core/issues/705

augustnmonteiro:
Can we merge this one?

ethanaward:
I want to do a little bit of looking into the PyOpenSSL issue first just to double check, and possibly change this to updating more requirements.

forslund:
@ethanaward can you look at #690 and if you think it's ok remove uuid from requirements.txt along with the rest of your updates.

I've uninstalled it locally and mycroft works just fine with the one included in the standard library

ethanaward:
@forslund is right, looks like uuid isn't needed. I'll rebase and remove it as well. shortuuid isn't needed either, it's not used anywhere in the code.

ethanaward:
Tests are passing, this is good to go!

PR #808, opened 2017-06-07
Changes to out-of-the-box experience

penrods:
Implementing changes to out-of-the-box experience, providing an "onboarding" process.

* Added a mycroft.api.is_paired() method
* Added mycroft.util.is_speaking and mycroft.util.wait_while_speaking() methods
* RESET now waits for the spoken notice to complete
* Stopped the "Checking for updates" and "Skills updated" prompts (commented out for now, probably will eliminate)
* Wifi setup filters out hidden ("x00") networks
* Visemes should keep up better if they get behind (will skip)
* Mimic is now searched for on the users path
* Onboarding process:
  - wifi setup starts automatically
  - User is walked through the process
  - wake word and button pressing are ignored
  - At end, a short tutorial is given


forslund:
Fix the pep8 isdues and we should be good to merge

penrods:
Merged!

PR #812, opened 2017-06-08
Bugfix 807

el-tocino:
Some significant alterations beyond the initial parameter issue. 
 - altered the default install to use install
 - install can handle a list of skills now
 - if you set the mycroft_skills_dir in the env, you can have this run on alternative locations.
 - search feature added
 - now handles picroft and mycroft, in theory.


el-tocino:
missing a rebase somewhere, closing.

PR #813, opened 2017-06-08
msm updated for issue 807

el-tocino:
Several updates.  
 - parameter issue also fixed now.
 - picroft/mycroft can both be handled
 - alternate directory can be used if exported in env for mycroft_skills_directory
 - default calls install with a list
 - install handles a list of parameters
 - very simple search function (name only) added

The pip and git parts of the install function will need to be worked over further at some point in the future.

### DIFF
--- a/msm/msm
+++ b/msm/msm
@@ -37,7 +37,7 @@ else
   if [[ -r /etc/bash_completion.d/virtualenvwrapper ]]; then
     source /etc/bash_completion.d/virtualenvwrapper
   else
-    if locate virtualenvwrapper ; then
+    if locate virtualenmvwrapper ; then
       if ! source $(locate virtualenvwrapper) ; then
         echo "Unable to locate virtualenvwrapper.sh, will not be able to install skills!"
         vwrap='false'
@@ -46,7 +46,7 @@ else
   fi
 fi
 
-default_skills="alarm audio-record configuration date-time desktop-launcher ip joke hello-world media npr-news naptime pairing personal reminder installer singing speak spelling stop stock volume weather wiki wolfram-alpha mark1-demo"
+default_skills="skill-alarm skill-audio-record skill-configuration skill-date-time skill-desktop-launcher skill-ip skill-joke skill-hello-world skill-media skill-npr-news skill-naptime skill-pairing skill-personal skill-reminder skill-installer skill-singing skill-speak skill-spelling skill-stop skill-stock skill-volume skill-weather skill-wiki skill-wolfram-alpha skill-mark1-demo"
 
 echo "#######  Mycroft Skill Manager #######"
 
@@ -193,7 +193,7 @@ case ${OPT} in
   "list") list | grep 'submodule "' | sed 's/\[submodule "//g'| sed 's/"\]//g' ;;
   "update") update ;;
   "default") install $(echo ${default_skills}) ;;
-  "search")  if [[ $# -gt 0 ]] ; then search "$*" ; else  help ; fi;;
+  "search")  if [[ $# -gt 0 ]] ; then search $(echo "$*") ; else  help ; fi;;
   *) help ;;
 esac
 
--- a/msm/msm
+++ b/msm/msm
@@ -113,7 +113,7 @@ while [[ $# -gt 0 ]] ; do
   name=$(echo "$git_name" | sed 's/.git//')
   if [[ -d "${mycroft_skill_folder}/${name}" ]] ; then
     echo "Skill appears to exist already. Perhaps you meant to use update?"
-    return 169
+    continue 169
   fi
   echo "Cloning repository"
   git clone "${repo}" >> /dev/null
--- a/msm/msm
+++ b/msm/msm
@@ -124,7 +124,7 @@ while [[ $# -gt 0 ]] ; do
                 return 120
             fi
         else 
-            if ! pip install -r requirements.txt ; then
+            if ! sudo pip install -r requirements.txt ; then
                 echo "Unable to install requirements for skill ${iskill}!"
                 return 121
             fi
--- a/msm/msm
+++ b/msm/msm
@@ -99,7 +99,7 @@ while [[ $# -gt 0 ]] ; do
             return 3
         else
             if [[ -z "${git_line}" ]]; then
-                echo "Skill not found"
+                echo "Skill ${iskill} not found"
                 return 3
             fi
         repo_line=$(($git_line + 2))
@@ -176,7 +176,7 @@ done
 OPT=$1
 shift
 case ${OPT} in
-    "install") if [[ $# -gt 0 ]] ; then install "$*" ; else  help ; fi;;
+    "install") if [[ $# -gt 0 ]] ; then install $(echo "$*") ; else  help ; fi;;
     "list") list | grep 'submodule "' | sed 's/\[submodule "//g'| sed 's/"\]//g' ;;
     "update") update ;;
     "default") install $(echo ${default_skills}) ;;
@@ -185,6 +185,6 @@ case ${OPT} in
 esac
 
 if  [[ $? -gt 0 ]] ; then
-    echo "Sorry I'm unable to complete the request! Check the error messages above for why."fs
+    echo "Sorry I'm unable to complete the request! Check the error messages above for why."
 fi
 exit 0
--- a/msm/msm
+++ b/msm/msm
@@ -111,6 +111,10 @@ while [[ $# -gt 0 ]] ; do
   fi
   git_name=$(echo "${repo}" | sed 's/.*\///')
   name=$(echo "$git_name" | sed 's/.git//')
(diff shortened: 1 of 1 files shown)


forslund:
I've run into some trouble with `msm default`

```sh
(mycroft) ake@roadrunner:~/projects/python/mycroft-core$ LANG=C ./msm/msm default
Uname to access /opt/mycroft/skills!
(mycroft) ake@roadrunner:~/projects/python/mycroft-core$ LANG=C ./msm/msm default
/usr/share/virtualenvwrapper/virtualenvwrapper.sh
/home/ake/.virtualenvs/mycroft/bin/python: No module named virtualenvwrapper
virtualenvwrapper.sh: There was a problem running the initialization hooks. 

If Python could not import the module virtualenvwrapper.hook_loader,
check that virtualenvwrapper has been installed for
VIRTUALENVWRAPPER_PYTHON=/home/ake/.virtualenvs/mycroft/bin/python and that PATH is
set properly.
#######  Mycroft Skill Manager #######
Attempting to install alarm...
Cloning repository
fatal: repository '' does not exist
The alarm skill has been installed!
Attempting to install audio-record...
Cloning repository
fatal: repository '' does not exist
The audio-record skill has been installed!
Attempting to install configuration...
A configuration skill was not found
Sorry I'm unable to complete the request! Check the error messages above for why.
(mycroft) ake@roadrunner:~/projects/python/mycroft-core$
```

Is my setup broken in some way? (The msm in the dev branch works correctly)

msm/msm
```diff
@@ -23,114 +23,166 @@
...
+    exit 101
+fi 
+
+# picroft?
+if [[ "$(hostname)" == 'picroft' ]] || [[ -x /home/pi/bin/cli ]] ; then
+    picroft=0
+else
+    picroft=1
+    if locate virtualenvwrapper.sh ; then
+        source $(locate virtualenvwrapper.sh)
```
forslund: This produces error messages if run from within the virtualenv.

msm/msm
```diff
@@ -23,114 +23,166 @@
...
+        if [[ "${skill}" ==  *$'\n'* ]]; then
+            echo -e "Your search has multiple choices\n\n$skill" | sed 's/.*://g'
+            return 3
+        else
+            if [[ -z "${git_line}" ]]; then
+                echo "A ${iskill} skill was not found"
+                return 3
+            fi
+        repo_line=$(($git_line + 2))
+        repo=$(echo "$skill_list" | sed -n $repo_line'{p;q;}' | sed 's/[[:space:]]//g' | sed 's/url=//g')
```
forslund: `skill_list` is missing, I think this is what makes the installation of named skills fail.

msm/msm
```diff
@@ -23,114 +23,166 @@
...
+cd "${mycroft_skill_folder}"
+for d in $(find "${mycroft_skill_folder}" -mindepth 1 -maxdepth 1 -type d |grep -v '.git'$ ); do
+    if git -C "$d" rev-parse --git-dir > /dev/null 2>&1; then
+        cd "${d}" 
+        if [[ -z $(git status --porcelain) ]]; then
+        git fetch
+        git reset --hard origin/master 
+    fi
+fi
+done
```
forslund: Is the indentation of functions correct? I'm a rookie at shell scripting so I'm not sure what the standard is.

forslund:
Installation from git urls works fine. search also seems to be working as expected.

msm/msm
```diff
@@ -23,114 +23,166 @@
 # This script assists in the installation and management of
 # skills loaded from Github.  
 
-set -e
+mycroft_skill_folder=${mycroft_skill_folder:-"/opt/mycroft/skills"}
+if [[ ! -d "${mycroft_skill_folder}" ]] ; then
+    echo "Uname to access ${mycroft_skill_folder}!"
```
forslund: Should "Uname" be "Unable"?

msm/msm
```diff
@@ -23,114 +23,166 @@
...
+    exit 101
+fi 
+
+# picroft?
+if [[ "$(hostname)" == 'picroft' ]] || [[ -x /home/pi/bin/cli ]] ; then
+    picroft=0
```
forslund: It would be clearer if true/false were used for the picroft status variable

msm/msm
el-tocino: Ah, never thought of that, will add a check.

msm/msm
el-tocino: 2am typing fixed.

msm/msm
el-tocino: fixing.

msm/msm
el-tocino: Doh.  shortened the steps to get the skills, will undo that.

msm/msm
el-tocino: I believe so from a read of the google style guide. I've also been using 4 instead of two spaces...oops.

msm/msm
el-tocino: Same shortening applied here: called list directly instead of separately.

msm/msm
forslund: Ok, then I trust you =)

el-tocino:
I believe this will handle the virtualenv issue.  Also, the skill_list bit was trimmer because we can call list, so just re-used that call again.  Uname, while a useful command, adjusted to the more appropriate unable.  LMK if anything else!

forslund:
I've noticed a thing...

Skill-configuration is missing from the mycroft-skill repo and this will cause `msm default` to fail. I'll go through the mycroft-skills repo and make sure all default skills are there so this will work.

el-tocino:
I have made one additional change, it now checks for an existing dir on install and prompts you to update if it exists.  This does not currently error out, it merely continue's so as to allow for any other skills to be installed that were also passed.  

This may not be the desired behavior; if so, I'd suggest reverting to commit 1e32728 or suggesting alternative.

el-tocino:
@forslund the homeassistant skill being listed in upper and lower case also breaks installing it via msm.

forslund:
@el-tocino, I'll look into that as well.

msm/msm
```diff
@@ -23,114 +23,181 @@
...
+        echo "Unable to locate virtualenvwrapper.sh, will not be able to install skills!"
+        vwrap='false'
+      fi
+    fi
+  fi
+fi
 
-mycroft_skill_folder="/opt/mycroft/skills"
-mycroft_virtualenv=~/.virtualenvs/mycroft/bin/activate
+default_skills="alarm audio-record configuration date-time desktop-launcher ip joke hello-world media npr-news naptime pairing personal reminder installer singing speak spelling stop stock volume weather wiki wolfram-alpha mark1-demo"
```
forslund: These should be `skill-...` so the match is exact otherwise the install will fail if multiple skills matches. (Right now media fails on this)

el-tocino:
Took out the items from the default list that don't appear in the skills repo.

forslund:
No don't do that, it won't be usable without those skills. I'm going to be sending PR's for the missing skills today.

ethanaward:
@el-tocino @forslund @augustnmonteiro Is this good to merge?

forslund:
The missing skills were added to the mycroft-skills repo yesterday. I still need to verify that it works now,

forslund:
@el-tocino I'm getting a weird issue when installing default skills, seems like it enters the directory of the skill and never leaves causing the loop to install skills within skills (within skills...) resulting in a paths like this:
`/opt/mycroft/skills/skill-audio-record/skill-configuration/skill-date-time/skill-desktop-launcher/skill-ip/skill-joke/skill-hello-world/...`

msm/msm
```diff
@@ -23,114 +23,187 @@
...
-	echo "Installed!"
-	exit 0
+function update() {
+cd "${mycroft_skill_folder}"
+for d in $(find "${mycroft_skill_folder}" -mindepth 1 -maxdepth 1 -type d |grep -v '.git'$ ); do
+  if git -C "$d" rev-parse --git-dir > /dev/null 2>&1; then
+    cd "${d}" 
+    if [[ -z $(git status --porcelain) ]]; then
+    git fetch
+    git reset --hard origin/master
```
forslund: `cd ..` here maybe?

msm/msm
```diff
@@ -23,114 +23,187 @@
...
-        echo "Skill installed!"
-}
+cd "${mycroft_skill_folder}"
+if [[ "${vwrap}" == 'false' ]] ; then
+    echo "Missing virtualwrapper, cowardly refusing to install skills."
+    return 5
+fi
+# loop through list of arguments
+while [[ $# -gt 0 ]] ; do
+  iskill="${1}";
```
forslund: `  cd "${mycroft_skill_folder}"` here maybe?

forslund:
This works good now. As soon as the (hopefully) last missing core skills are added to mycroft-skills (https://github.com/MycroftAI/mycroft-skills/pull/120) this one is good to go as far as I can see.

Maybe @augustnmonteiro want to have a final look before merging

PR #814, opened 2017-06-11
doing german normalise and translate some basics

hersche:
well, i'm not shure how to run the mycroft-tests, but i tried it verbal and i became the same output as in the test. seems so.

hope it's usefull.


hersche:
i'm not really shure why the test is failing.. i tried the unittest which seems to fail on other places than my touchings, but about the lang-variable u"de", which couldn't be compared with "en-us" (maybe because of the unicode-sign on de?).. however,
a) is de!=en-us, so that fact is true and it's intepreted as this.
b) i couldn't set that in config, so could not be a issue from there..

if someone could help there, that would be nice.. the code is executeable and works, when i say german nr's, they're replaced (eins zwei = 1 2)- when i write it via cli, it didn't replace (eins zwei = eins zwei). i assumed there, that cli may delivers a direct utterance and don't need the normalize-method or something, but when the test fails, this didn't seem to be the case..

only thing i could imagine: when i speak, replace nr's-method is active, and normalize simply did not work yet?

MatthewScholefield:
Thanks for taking the time to start translating into German, this looks very useful.

It looks like the code formatting is the only thing that fails (like spaces after commas and ensuring lines don't go on too long). [Here's where it mentions the issues.](https://travis-ci.org/MycroftAI/mycroft-core/builds/243543177#L5001)

hersche:
thanks for your hint :)

hersche:
as you told me, dict- and lm-files are basicly unneeded (not as descriped in my wiki-entry, have to update). anyway, the hmm-files seems still needed, also when i'm not fully shure if it would be better to use the en-hmm-folder cause the "hey mycroft" is really bad reconized right now..

anyway, question: should i add those files to mycroft/client/speech/model/de/hmm ?

atm i upgrade my base (new pi, exclusively for mycroft) and could not test out of this..

MatthewScholefield:
How large are the files collectively under de/hmm?

hersche:
near 27mb, two files (variances and means) are 11mb each. so it's about double of en-us..

hersche:
resp. the cont-package is 26mb, the ptm is about 16mb.. i'm not shure about else diffrences

but as you might can imagine, the sphinx-topic is relativ irrelevant for me atm (snowboy olé). So if you want them, just say, else i let it be..

penrods:
Hey, I want to revive this work, it kinda got stalled.  I'm trying to get a good handle on how to approach not only this translation for German (danke!) but also other languages.  Would you mind working with me on breaking this into a few pieces?  I'm thinking:
1) The mycroft-core text parsing code (mycroft\util\parse.py:  creating normalize_XX(), ...
2) The mycroft-core text formatting code (mycroft\util\format.py): creating nice_number_XX(), ...
3) The mycroft-core resources: creating translations of the en-us text for mycroft\res\text\XX\*
4) Default skill translations: skill-pairing\dialog\*, skill-pairing\vocab\*, skill-whatever\* ...

I'd like to give some thought to whether we keep trained models in this repo.  I know the English hmm stuff is in there, but I'm worried about the repos size exploding as we refine models and the old models become part of the history that everyone has to drag down.

Thoughts?

valdr:
@hersche shouldn't 'de' not be 'de-de', as there are also 'de-at', and 'de-ch'?
There's 'en-us', but also 'en-gb', and even 'en-au'... They may not differ much when it comes to naming time, but thus is the convention...

penrods:
This seems to be an abandoned PR.  I hate to throw away work, but the code implementation is only partial, and there is no CLA.  So I'm going to go ahead and close this for now, but I'd be happy to revisit this work if you'd like to resubmit.

PR #817, opened 2017-06-12
Minor onboarding refinements

penrods:
* The wifi setup no longer stops after 5 minutes unless already paired (i.e. still onboarding)
* The mouth resets on wifisetup stop (clearing the scrolling home.mycroft.ai)
* Changed several ConfigurationManager.get() calls to ConfigurationManager.instance().  Exactly the same, but .instance() is clearer/preferred.
* Added a mycroft.util.stop_speaking() method.  Not perfect, but works for now and can be replaced later when AudioManager is in place.

A forthcoming change to the pairing skill will utilize the stop_speaking method.

### DIFF
--- a/mycroft/util/__init__.py
+++ b/mycroft/util/__init__.py
@@ -323,6 +323,7 @@ def check_for_signal(signal_name, sec_lifetime=0):
             valid in filenames.
         sec_lifetime (int, optional): How many seconds the signal should
             remain valid.  If 0 or not specified, it is a single-use signal.
+            If -1, it never expires.
 
     Returns:
         bool: True if the signal is defined, False otherwise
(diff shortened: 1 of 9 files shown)


mycroft/util/__init__.py
```diff
@@ -364,3 +366,11 @@ def wait_while_speaking():
...
     while is_speaking():
         time.sleep(0.1)
+
+
+def stop_speaking():
+    # TODO: Less hacky approach to this once Audio Manager is implemented
+    # Skills should only be able to stop speech they've initiated
+    config = mycroft.configuration.ConfigurationManager.instance()
+    kill([config.get('tts').get('module')])
+    kill(["aplay"])
```
forslund: Won't this leave the visimes running on the display?

Couldn't this check `is_speaking()` and if so `create_signal('buttonPress')` to stop speech.

forslund: Also I this will only stop the current sentence. On Picroft this isn't an issue but on PC it get's a bit weird during pairing.

penrods: I redid the shutdown of speech in a new util.stop_speaking()

penrods:
I believe this is ready for merge after you review it.

forslund:
Looks good, fixed the pep8 issues and will Merge!

PR #818, opened 2017-06-12
Check pairing after possible wakeword is detected

forslund:
This reduces the number of calls to the server since they are made only after a chunk of audio is identified.

### DIFF
--- a/mycroft/client/speech/mic.py
+++ b/mycroft/client/speech/mic.py
@@ -179,10 +179,8 @@ class ResponsiveRecognizer(speech_recognition.Recognizer):
         return audioop.rms(sound_chunk, sample_width)
 
     def wake_word_in_audio(self, frame_data):
-        if not is_paired():
-            return False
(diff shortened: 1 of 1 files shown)


penrods:
I removed this "is_paired" from the listener altogether.  It was just a bad idea.  :)

BenjaminHCCarr:
was this supposed to get merged? it looks like it got rolled into https://github.com/MycroftAI/mycroft-core/releases/tag/release%2Fv0.8.16

forslund:
@BenjaminHCCarr No, this was a proposed quick fix. Steve handled it in a different an more robust way instead. will check that it's not included in the release

BenjaminHCCarr:
@forslund thanks, the mention in the release note, and some oddness / lagging on picroft (low power.... it takes 20 minutes to install just that one `.deb`) had us digging through the release on slack trying to figure out why we were having issues..... likely cosmic rays. we'll see if @btotharye gets superpowers.

Issue #824, opened 2017-06-13
Mark 1 button stops working after initial onboarding.


ethanaward:
The button on the Mark 1 will sometimes stop working after the onboarding process, leading to needing a reboot.

ethanaward:
Closed by #825

PR #825, opened 2017-06-13
#824 Change has_been_paired to load the current identity file

ethanaward:
This changes the has_been_paired function to load the current identity file in case the pairing state has recently changed. It also removes the doubling of `mycroft.stop` messages.

### DIFF
--- a/mycroft/client/enclosure/__init__.py
+++ b/mycroft/client/enclosure/__init__.py
@@ -77,7 +77,8 @@ class EnclosureReader(Thread):
                 LOG.error("Reading error: {0}".format(e))
 
     def process(self, data):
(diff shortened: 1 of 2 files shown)


mycroft/client/enclosure/__init__.py
```diff
@@ -77,7 +77,8 @@ def read(self):
                 LOG.error("Reading error: {0}".format(e))
 
     def process(self, data):
-        self.ws.emit(Message(data))
+        if not "mycroft.stop" in data:
```
penrods: Let's add a comment here, like:
# TODO: Look at removing this altogether.  Review if any Arduino serial messages are handled by other messagebus subscribers

PR #829, opened 2017-06-14
Closes #801 - Refactor setup_base to use CORE_VERSION_STR

### LINKED ISSUE #801: setup_base.py uses deprecated __version__.py calls.
Setup_base.py by default tries to get the current version by importing `mycroft.__version`. However, this is no longer used. It should be updated to use the new versioning system.

ethanaward:
This changes the get_version method to get the version from CORE_VERSION_STR instead of the deprecated __version__.  This also doesn't have the issue of the previous PR, where it would check for a file that doesn't exist during the build process, and so break packaging. @aatchison, it would be good if you could double check that this works for the current packaging process.

### DIFF
--- a/mycroft/util/setup_base.py
+++ b/mycroft/util/setup_base.py
@@ -36,8 +36,8 @@ def place_manifest(manifest_file):
 def get_version():
     version = None
     try:
-        import mycroft.__version__
-        version = mycroft.__version__.version
+        from mycroft.version import CORE_VERSION_STR
+        version = CORE_VERSION_STR
     except Exception as e:
         try:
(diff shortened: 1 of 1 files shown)


aatchison:
Okie, I'm on it

penrods:
@aatchison This good?  If so let's merge

aatchison:
Oops, I do believe this works but I had better check again because I don't remember.

ethanaward:
The travis build on this is failing, do we have issues with travis normally in our tests right now or do I need to rebase?

ethanaward:
Tests are passing, this is good to go!

PR #841, opened 2017-06-20
mycroft-use script to switch builds

ajuhl:
mycroft-use is a script that allows you to change builds between stable, unstable, and github checkout
mycroft-use [unstable | stable | github]

### DIFF
--- a/scripts/mycroft-use
+++ b/scripts/mycroft-use
@@ -1,0 +1,19 @@
+#!/bin/bash
+#mycroft-use
+
+if [ $1 = "unstable" ]; then
+        echo "Switching to unstable build..."
+        sudo sh -c 'echo "deb http://repo.mycroft.ai/repos/apt/debian debian-unstable main" > /etc/apt/sources.list.d/repo.mycroft.ai.list'
+        sudo apt-get update
+        sudo apt-get install mycroft-core -y
+elif [ $1 = "stable" ]; then
+        echo "Switching to stable build..."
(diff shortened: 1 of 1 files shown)


forslund:
Looks like a useful script but it feels a bit out of place here in core. I wonder if it should be in the enclosure-picroft repo or among the scripts used when building the Mark-1 Image?

What do you think?

penrods:
I debated where to recommend it go.  This will be useful in Picroft and on the Mark 1.  I also think under Ubuntu and other package installs.  I'm open for a better suggestion.

penrods:
It will soon allow switching to a Github install, too.  So this is valuable to most devs.

forslund:
I would suggest to add the script when building apt-packages so keeping it
together with the packaging scripts would make sense to me.

Den 20 juni 2017 23:48 skrev "Steve Penrod" <notifications@github.com>:

penrods:
Hmm, yeah that makes sense.  Gotta do a package install first, but that isn't horrible.

aatchison:
I kind of feel like this would be better on picroft. Since we don't have a similar thing for the mark 1 though, I'm not sure. Let me think about this. Also, a package installation includes init scripts, so it would run automatically and would cause trouble with a dev setup. I guess that it might be best in the core repo.

penrods:
I talked to Austin about making the changes to switch the init scripts from package to Github.  Should be doable without too much effort.

ajuhl:
The github fix works with the default github path, but I will need to test to make sure it works when given a custom path.

penrods:
This got pulled in with modifications:

https://github.com/MycroftAI/mycroft-core/pull/1073

Issue #820, opened 2017-06-13
Python2.7/JSON error


iqtedar:
I'm testing both the master and dev branches. Initially I was able to run <./mycroft.sh start> on both, including its alternatives, but there was no audio. The log files revealed missing configuration files at ~/.mycroft/mycroft.conf and /etc/mycroft/mycroft.conf, so I placed symbolic links at these two places to the main configuration file in ~/mycroft-core/scripts/logs. Now, <./mycroft.sh start> doesn't work. The logs show no problems anymore with finding any configuration problems, but another problem related to python2.7/json:

```
File "/usr/lib/python2.7/json/__init__.py", line 291, in load
    **kw)
  File "/usr/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
File "/usr/lib/python2.7/json/__init__.py", line 291, in load
    **kw)
  File "/usr/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 364, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python2.7/json/decoder.py", line 380, in raw_decode
    obj, end = self.scan_once(s, idx)
ValueError: Expecting property name: line 2 column 3 (char 4)
  File "/usr/lib/python2.7/json/decoder.py", line 364, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python2.7/json/decoder.py", line 380, in raw_decode
    obj, end = self.scan_once(s, idx)
ValueError: Expecting property name: line 2 column 3 (char 4)
```
Any suggestions?

Matlar:
This most likely has to do with #685. I think I fixed this in #845

PR #845, opened 2017-06-22
Issue-685 - Using load_commented_json in log.py

### LINKED ISSUE #685: Mycroft doesn't starts if /etc/mycroft/mycroft.conf exists and has comments
In log.py there are these statements:

if isfile(SYSTEM_CONFIG):
    with open(SYSTEM_CONFIG) as f:
        config = json.load(f)
        log_level = config.get("log_level", "DEBUG")

note no filter of comments has been done using "load_commented_json".

Note: no pull request will be done, because could be the best solution is do not load any config here but use the common configuration procedure.

Matlar:
Fixes #685 and probably also #820

### DIFF
--- a/mycroft/util/log.py
+++ b/mycroft/util/log.py
@@ -18,6 +18,7 @@ import json
 import logging
 
 from os.path import isfile
+from mycroft.util.json_helper import load_commented_json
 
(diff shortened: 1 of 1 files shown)


forslund:
Works like a charm, thank you very much. Merging.

PR #851, opened 2017-06-28
feature/issue-850

kuboosoft:
We have mimic in Fedora 26 and Rawhide

### DIFF
--- a/scripts/install-mimic.sh
+++ b/scripts/install-mimic.sh
@@ -10,15 +10,18 @@ MIMIC_VERSION=1.2.0.2
 pkg-config --exists icu-i18n || export CFLAGS="$CFLAGS -I/usr/include/x86_64-linux-gnu"
 pkg-config --exists icu-i18n || export LDFLAGS="$LDFLAGS -licui18n -licuuc -licudata"
 
-# download and install mimic
-if [ ! -d ${MIMIC_DIR} ]; then
+# Check if you have installed mimic (Fedora 26 and higher provides a rpm for mimic)
+if [ -f /usr/bin/mimic ]; then
+ln -sf /usr/bin/mimic $(pwd)/mimic/bin/mimic
+# download and install mimic if the rpm does not exist
+elif [ ! -d ${MIMIC_DIR} ]; then
     git clone --branch ${MIMIC_VERSION} https://github.com/MycroftAI/mimic.git
     cd ${MIMIC_DIR}
     ./autogen.sh
     ./configure --with-audio=alsa --enable-shared --prefix=$(pwd)
     make -j$CORES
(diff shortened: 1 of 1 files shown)


forslund:
Good idea!

Is the symbolic link needed?

mimic tts backend should find the installed mimic by itself. (See https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/tts/mimic_tts.py#L38)

kuboosoft:
@forslund yes, because mycroft-core is expecting "mimic" in the specific path...

forslund:
It should fall back to a system install of mimic if the `./mimic/bin/mimic` isn't found. (See link in my earlier comment for the mechanism for this)
On Debian this works (tested it today), but maybe fedora is different? In that case we have a bug in `mimic_tts`

kuboosoft:
@forslund in Fedora with mimic installed (/usr/bin/mimic), ever installs mimic locally... I am not sure the mechanism about the python script, searches the excutable of mimic. But with it minor change, everyone can benefit :)

forslund:
Yeah, This is definitely a good idea. I'll just have to check how the mimic install script behaves if there is a symbolic link there (if we should start to require a newer version of mimic or something)

forslund:
I did a test with Fedora 26 on Virtual Box and the symbolic link doesn't seem to be necessary. `distutils.spawn.find_executable("mimic")` finds the mimic executable (`/usr/bin/mimic`) the same way as on Debian.

Adding the symbolic link will stop any future install attempts of mimic using the `install_mimic.sh` script if it should be needed for some reason.

Would you mind removing the symbolic link?

scripts/install-mimic.sh
```diff
@@ -10,15 +10,18 @@ MIMIC_VERSION=1.2.0.2
 pkg-config --exists icu-i18n || export CFLAGS="$CFLAGS -I/usr/include/x86_64-linux-gnu"
 pkg-config --exists icu-i18n || export LDFLAGS="$LDFLAGS -licui18n -licuuc -licudata"
 
-# download and install mimic
-if [ ! -d ${MIMIC_DIR} ]; then
+# Check if you have installed mimic (Fedora 26 and higher provides a rpm for mimic)
+if [ -f /usr/bin/mimic ]; then
+ln -sf /usr/bin/mimic $(pwd)/mimic/bin/mimic
```
forslund: replace this line with
```bash
    echo "Using the system's copy of mimic"
```

ethanaward:
@kuboosoft Can you rebase and remove the symbolic link?

forslund:
Closing this since the functionality has been added in a different way. Many thanks for this contribution in any case.

PR #853, opened 2017-06-28
feature/issue-852

kuboosoft:
We can run dnf using polkit...

### DIFF
--- a/build_host_setup_fedora.sh
+++ b/build_host_setup_fedora.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-sudo dnf install -y \
+pkexec dnf -y install \
     git \
     python \
     python-devel \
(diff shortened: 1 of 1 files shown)


build_host_setup_fedora.sh
```diff
@@ -24,6 +24,7 @@ sudo dnf install -y \
     curl \
     pkgconfig \
     libicu-devel \
+    procps-ng \
```
forslund: Is this a required dependency? Seems like it's a tool for browsing the proc file system

forslund:
There's a `sudo` at the end of the file as well upgrading the virtualenvironment, shouldn't this also be using policykit?

ethanaward:
@kuboosoft Any update on this? You might need to rebase on dev to fix those failing tests.

penrods:
The build_host_setup_fedora.sh no longer exists, closing this PR unmerged.

Issue #855, opened 2017-06-29
Fix msm


augustnmonteiro:
MSM isn't returning when skill isn't founded

el-tocino:
Does https://github.com/MycroftAI/mycroft-core/pull/813 help?

PR #856, opened 2017-06-29
#855 - MSM isn't returning when skill isn't founded

augustnmonteiro:
MSM isn't returning when skill isn't founded


msm/msm
```diff
@@ -23,8 +23,6 @@
 # This script assists in the installation and management of
 # skills loaded from Github.  
 
-set -e
```
forslund: This needs to be remain to make sure download errors makes the script exit with error

PR #857, opened 2017-06-29
Read Blacklisted skills from config

JarbasAI:
Blacklisted skills were hard-coded in skills core.py, now read from config

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -36,7 +36,9 @@ from mycroft.util.log import getLogger
 from mycroft.skills.settings import SkillSettings
 __author__ = 'seanfitz'
 
-BLACKLISTED_SKILLS = ["send_sms", "media"]
+skills_config = ConfigurationManager.instance().get("skills")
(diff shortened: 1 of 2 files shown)


forslund:
Looks good! I notice a small issue (not yours, it was there before): the name of the `media` skill is actually `skill-media`. Could you update the config to fix this as well?

JarbasAI:
done

mycroft/skills/core.py
```diff
@@ -36,7 +36,9 @@
 from mycroft.skills.settings import SkillSettings
 __author__ = 'seanfitz'
 
-BLACKLISTED_SKILLS = ["send_sms", "media"]
+skills_config = ConfigurationManager.instance().get("skills")
+BLACKLISTED_SKILLS = skills_config.get("blacklisted_skills",[])
```
forslund: there's a pep8 issue here. There should be a space between `,` and `[]`.

forslund:
Thank you. Merging!

PR #858, opened 2017-06-29
Load Piority skills first

JarbasAI:
priority skills is needed for skills that needs to listen for register_whatever messages from other skills, any "manifest" or "context" skill would also need this, service skills also make sense to be started first

my objectives skill, needs to listen to messages that can be emitted by any skill

some manifest skill that wants to listen to register messages of anything needs this

control center skill to enforce run levels at start up

browser service to ensure webdriver is running when some skill requests it


forslund:
Merged together with #1084 

Thanks @JarbasAI!

PR #859, opened 2017-06-29
Intent Parser helper class

JarbasAI:
helper class to query intent service instance

splitting #783 in smaller PRs

helper class to be imported and used in other skills

```
from mycroft.skills.intent_service import IntentParser

Parser = IntentParser(self.emitter)

intent_name = "HelloWorldIntent"
list_of_skills_with_this_intent = Parser.get_skill_from_intent(intent_name)

utterance = "Hello world"
skill_utterance_will_trigger = Parser.get_skill_from_utterance(utterance)
intent_utterance_will_trigger = Parser.get_intent(utterance, lang="en-us")

```


mycroft/skills/intent_service.py
```diff
@@ -101,3 +147,54 @@ def handle_detach_skill(self, message):
             p for p in self.engine.intent_parsers if
             not p.name.startswith(skill_name)]
         self.engine.intent_parsers = new_parsers
```
forslund: Shouldn't this remove the intent from the dictionary of skills and intents?

forslund:
As far as I can understand this it looks quite good. One small question that I'd like answered or fixed if it's an actual issue (see comment in code) and fix the pep8 problems and I think this is good to merge.

mycroft/skills/intent_service.py
forslund: @JarbasAI any comment on this?

JarbasAI: i missed this comment, will add the remove functionality, perhaps i should wait on the converse PR since skill name will become skill_id ?

penrods:
Converse() PR #925 is in!  Using the skill ID is a much better approach, I agree.

JarbasAl:
this one kinda got lost, now that we have padatious should i make it also get intents from there? how do i differentiate in the returned data?

JarbasAI:
should i update this for #1351 ?

PR #748, opened 2017-05-09
This add comments to document the message class

amcgee7:
Signed-off-by: Art McGee <amcgee7@jaguarlandrover.com>

PR #862, opened 2017-06-29
Fix cd "gotoskillsfolder" swap in msm

kfezer:
Fixes a cd change from:
https://github.com/MycroftAI/mycroft-core/pull/748


kfezer:
Fixed through revert.

PR #865, opened 2017-06-30
Feature/decorate intenthandlers

forslund:
A new way to create intent handlers for skills by simply decorating the methods with `@intent_handler(...)`.

Example of what it would look like:
```python


class ASkill(MycroftSkill):
    def __init__(self):
        super(ASkill, self).__init__('ASkill')

    @intent_handler(IntentBuilder('AskillIntent').require('ASkill1').build())
    def handle(self, message):
        self.speak('A Skill handler')

    @intent_handler(IntentBuilder('AskillIntent2').require('ASkill2').build())
    def handle2(self, message):
        self.speak('Another Skill handler')

    @intent_handler(IntentBuilder('AskillIntent3').require('ASkill3').build())
    def handle3(self, message):
        self.speak('Third Skill handler')


def create_skill():
    return ASkill()
```

The above test case can be downloaded from [GoogleDrive](https://drive.google.com/file/d/0B6NfzYSsivqTM0R0Wl9lMjBYNEk/view?usp=sharing). Each of the intent handlers can be triggered with _first skill_,  _another skill_ and _third_skill_

Normal intent registration is still available (`def initialize(self)...`) so this is 100% backward compatible.

Future improvement: Adding possibility to read intent from file.

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -161,14 +161,15 @@ def unload_skills(skills):
         s.shutdown()
 
 
-intent_list = []
+_intent_list = []
 
 
 def intent_handler(intent_parser):
+    """ Decorator for adding a method as an intent handler. """
     def real_decorator(func):
         def handler_method(*args, **kwargs):
             return func(*args, **kwargs)
-        intent_list.append((intent_parser, func))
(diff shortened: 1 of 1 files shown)


augustnmonteiro:
That's awesome!

PR #866, opened 2017-06-30
Adds the ExtractDateTime parse function from Christopher (plus relevant tests)

connor-penrod:
When imported from mycroft/util/parse.py, it'll take a sentence like "What's the weather like 5 weeks from next Wednesday?" and will extract a python datetime object for that date.

### DIFF
--- a/mycroft/util/parse.py
+++ b/mycroft/util/parse.py
@@ -170,7 +170,17 @@ def extractnumber_en(text):
 #the 'date' is a python datetime object
 #the 'leftover_string' contains all keywords not related to the date or time
 
-def extractdatetime_en(str, currentDate = datetime.now()):
+from datetime import datetime, timedelta
+from dateutil.relativedelta import relativedelta
+
+'''authors: spenrod and connorpenrod
+this function takes a sentence and extracts date and time information from it
+returns a 2-element array [date, leftover_string]
+the 'date' is a python datetime object
+the 'leftover_string' contains all keywords not related to the date or time'''
+
+
+def extractdatetime_en(str, currentDate=datetime.now()):
     if str == "":
         return None
     found = False
@@ -182,35 +192,38 @@ def extractdatetime_en(str, currentDate = datetime.now()):
     today = dateNow.strftime("%w")
     currentYear = dateNow.strftime("%Y")
     fromFlag = False
-    
     datestr = ""
     hasYear = False
     timeQualifier = ""
 
-    str = str.lower().replace('?','').replace('.','').replace(',','').replace(' the ', ' ').replace(' a ', ' ').replace(' an ', ' ')
-    words = str.split();
-    
+    str = str.lower().replace('?', '').replace('.', '').replace(',', '')\
+        .replace(' the ', ' ').replace(' a ', ' ').replace(' an ', ' ')
+    words = str.split()
+
     timeQualifiersList = ['morning', 'afternoon', 'evening']
     markers = ['at', 'in', 'on', 'by', 'this', 'around', 'for', 'of']
-    days = ['monday', 'tuesday', 'wednesday', 'thursday', 'friday', 'saturday', 'sunday']
-    months = ['january', 'february', 'march', 'april', 'may', 'june', 'july', 'august', 'september', 'october', 'november', 'december']
-    monthsShort = ['jan', 'feb', 'mar', 'apr', 'may', 'june', 'july', 'aug', 'sept', 'oct', 'nov', 'dec']
-    
-    #strip out 's, plurals, and ordinals (3rd becomes 3)
+    days = ['monday', 'tuesday', 'wednesday',
+            'thursday', 'friday', 'saturday', 'sunday']
+    months = ['january', 'february', 'march', 'april', 'may', 'june',
+              'july', 'august', 'september', 'october', 'november', 'december']
+    monthsShort = ['jan', 'feb', 'mar', 'apr', 'may', 'june', 'july', 'aug',
+                   'sept', 'oct', 'nov', 'dec']
+
+    '''strip out 's, plurals, and ordinals (3rd becomes 3)'''
     for idx, word in enumerate(words):
         word = word.replace("'s", "")
-        
-        ordinals = ["rd","st","nd","th"]
+
+        ordinals = ["rd", "st", "nd", "th"]
         if word[0].isdigit():
             for ord in ordinals:
                 if ord in word:
                     word = word.replace(ord, "")
         words[idx] = word
-        
+
     for idx, word in enumerate(words):
         if word == "":
             continue
-        
+
         wordPrevPrev = words[idx-2] if idx > 1 else ""
         wordPrev = words[idx-1] if idx > 0 else ""
         wordNext = words[idx+1] if idx+1 < len(words) else ""
@@ -218,23 +231,24 @@ def extractdatetime_en(str, currentDate = datetime.now()):
         word = word.rstrip('s')
         start = idx
         used = 0
-        #save timequalifier for later
+        '''save timequalifier for later'''
         if word in timeQualifiersList:
             timeQualifier = word
-        #parse today, tomorrow, day after tomorrow
+            '''parse today, tomorrow, day after tomorrow'''
         elif word == "today" and not fromFlag:
             dayOffset = 0
             used += 1
         elif word == "tomorrow" and not fromFlag:
             dayOffset = 1
             used += 1
-        elif word == "day" and wordNext == "after" and wordNextNext == "tomorrow" and not fromFlag:
+        elif word == "day" and wordNext == "after" and \
+                wordNextNext == "tomorrow" and not fromFlag:
             dayOffset = 2
             used = 3
             if wordPrev == "the":
                 start -= 1
                 used += 1
-        #parse 5 days, 10 weeks, last week, next week
+            '''parse 5 days, 10 weeks, last week, next week'''
         elif word == "day":
             if wordPrev[0].isdigit():
                 dayOffset += int(wordPrev)
@@ -253,7 +267,7 @@ def extractdatetime_en(str, currentDate = datetime.now()):
                 dayOffset = -7
                 start -= 1
                 used = 2
-        #parse 10 months, next month, last month
+            '''parse 10 months, next month, last month'''
         elif word == "month" and not fromFlag:
             if wordPrev[0].isdigit():
                 monthOffset = int(wordPrev)
@@ -267,7 +281,7 @@ def extractdatetime_en(str, currentDate = datetime.now()):
                 monthOffset = -1
                 start -= 1
                 used = 2
-        #parse 5 years, next year, last year
+            '''parse 5 years, next year, last year'''
         elif word == "year" and not fromFlag:
             if wordPrev[0].isdigit():
                 yearOffset = int(wordPrev)
@@ -281,7 +295,7 @@ def extractdatetime_en(str, currentDate = datetime.now()):
                 yearOffset = -1
                 start -= 1
                 used = 2
-        #parse Monday, Tuesday, etc., and next Monday, last Tuesday, etc.
+            '''parse Monday, Tuesday, etc., and next Monday, last Tuesday, etc.'''
         elif word in days and not fromFlag:
             d = days.index(word)
             dayOffset = (d+1)-int(today)
@@ -289,14 +303,14 @@ def extractdatetime_en(str, currentDate = datetime.now()):
             if dayOffset < 0:
                 dayOffset += 7
             if wordPrev == "next":
-               dayOffset += 7
-               used += 1
-               start -= 1
+                dayOffset += 7
+                used += 1
+                start -= 1
             elif wordPrev == "last":
                 dayOffset -= 7
                 used += 1
                 start -= 1
-        #parse 15 of July, June 20th, Feb 18, 19 of February
+            '''parse 15 of July, June 20th, Feb 18, 19 of February'''
         elif word in months or word in monthsShort and not fromFlag:
             try:
                 m = months.index(word)
@@ -304,9 +318,10 @@ def extractdatetime_en(str, currentDate = datetime.now()):
                 m = monthsShort.index(word)
             used += 1
             datestr = months[m]
-            if wordPrev[0].isdigit() or (wordPrev == "of" and wordPrevPrev[0].isdigit()):
+            if wordPrev[0].isdigit() or \
+                    (wordPrev == "of" and wordPrevPrev[0].isdigit()):
                 if wordPrev == "of" and wordPrevPrev[0].isdigit():
-                    datestr +=  " " + words[idx-2]
+                    datestr += " " + words[idx-2]
                     used += 1
                     start -= 1
                 else:
@@ -319,7 +334,7 @@ def extractdatetime_en(str, currentDate = datetime.now()):
                     hasYear = True
                 else:
                     hasYear = False
-            
+
             elif wordNext and wordNext[0].isdigit():
                 datestr += " " + wordNext
                 used += 1
@@ -329,7 +344,8 @@ def extractdatetime_en(str, currentDate = datetime.now()):
                     hasYear = True
                 else:
                     hasYear = False
-        #parse 5 days from tomorrow, 10 weeks from next thursday, 2 months from July
+        '''parse 5 days from tomorrow, 10 weeks from next thursday,
+        2 months from July'''
         validFollowups = days + months + monthsShort
         validFollowups.append("today")
         validFollowups.append("tomorrow")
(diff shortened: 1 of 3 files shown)


forslund:
Think of us running in the terminal, limit the line length to 79 characters :)

connor-penrod:
So the issue here is that PEP8 considers the indents that lead up to the line as part of the 79 character limit. Some of the lines have indents that are almost already past that 79 character limit and it's really not feasible to shorten the line lengths any farther. Still, I broke apart every long line that I could as long as it didn't look awful.

mycroft/util/parse.py
```diff
@@ -164,6 +169,571 @@ def extractnumber_en(text):
     return val
 
 
+'''authors: spenrod and connorpenrod
```
penrods: Move this into the function using docstrings, like what you see in the extractnumber() method.  We use Google style docstrings (see:  http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)

You can leave the Authors: stuff off.  That's all tracked by git anyway :)  I'm going to be removing the various Author tags from files at some point and moving everything into a master Contributor list instead.

mycroft/util/parse.py
```diff
@@ -164,6 +169,571 @@ def extractnumber_en(text):
...
+this function takes a sentence and extracts date and time information from it
+returns a 2-element array [date, leftover_string]
+the 'date' is a python datetime object
+the 'leftover_string' contains all keywords not related to the date or time'''
+
+
+def extractdatetime_en(str, currentDate=datetime.now()):
```
penrods: A couple minor things:
* Let's be Pythonic and name the function extract_datetime_en().  And yeah, we need to rename the existing extractnumber() to match -- but don't worry about that in this commit.
* Add a wrapper extract_datetime() just like there is on extractnumber() that turns around and calls the appropriate version based on a lang parameter.

mycroft/util/parse.py
```diff
@@ -164,6 +169,571 @@ def extractnumber_en(text):
     return val
 
 
+'''authors: spenrod and connorpenrod
+this function takes a sentence and extracts date and time information from it
+returns a 2-element array [date, leftover_string]
+the 'date' is a python datetime object
+the 'leftover_string' contains all keywords not related to the date or time'''
+
+
```
penrods: Don't use currentDate=datetime.now().  Python is a little odd in the way it handles default values.  It only gets executed once in the environment.  Try this example in the interactive interpreter:

```
> import datetime

> def func(paramVal=datetime.datetime.now()):
...    print paramVal
```
Now run func() several times.  You'll be surprised.


Instead, do this:

```
   def extract_datetime_en(str, currentDate=None):
       if currentDate is None:
           currentDate = datetime.now()
       ....
```

mycroft/util/parse.py
```diff
@@ -164,6 +169,571 @@ def extractnumber_en(text):
...
+            continue
+
+        wordPrevPrev = words[idx-2] if idx > 1 else ""
+        wordPrev = words[idx-1] if idx > 0 else ""
+        wordNext = words[idx+1] if idx+1 < len(words) else ""
+        wordNextNext = words[idx+2] if idx+2 < len(words) else ""
+        word = word.rstrip('s')
+        start = idx
+        used = 0
+        '''save timequalifier for later'''
```
penrods: Not a huge deal, but use # for comments instead of strings.  It is ever-so-slightly more efficient, and the floating string is just kinda weird.

mycroft/util/parse.py
```diff
@@ -164,6 +169,571 @@ def extractnumber_en(text):
...
+
+            for i in range(0, used):
+                words[i+start] = ""
+
+            if (start-1 >= 0 and words[start-1] in markers):
+                words[start-1] = ""
+            found = True
+            daySpecified = True
+
+        '''words[idx] = word'''
```
penrods: Is this supposed to be commented out?

mycroft/util/parse.py
```diff
@@ -164,6 +169,571 @@ def extractnumber_en(text):
     return val
 
 
+'''authors: spenrod and connorpenrod
+this function takes a sentence and extracts date and time information from it
+returns a 2-element array [date, leftover_string]
+the 'date' is a python datetime object
+the 'leftover_string' contains all keywords not related to the date or time'''
```
penrods: I'd modify the 'leftover_string' description a smidgeon for clairity.  More like:
   leftover_string (str):  the original phrase minus all words related to the datetime

In the longer description of the function, include an example.  Something simple like:
     extract_date_time_en("turn on the light at 3pm") will returns "turn on the light" as the leftover string

mycroft/util/parse.py
```diff
@@ -164,6 +169,571 @@ def extractnumber_en(text):
     return val
 
 
+'''authors: spenrod and connorpenrod
+this function takes a sentence and extracts date and time information from it
```
penrods: Since this is the primary documentation (docstrings are used to auto-generate the docs), this is probably a good place to explain some of the non-obvious behaviors this uses.  For example, an unqualified "evening" gets returned as 7pm, "morning" as 8am, "afternoon" as 3pm.

test/util/test_parse.py
```diff
@@ -40,6 +43,30 @@ def test_extractnumber(self):
...
+        self.assertEqual(extractWithFormat("What's the weather next Tuesday?"), "2017-07-04 00:00:00")  # noqa
+        self.assertEqual(extractWithFormat("What's the weather on August 11"), "2017-08-11 00:00:00")  # noqa
+        self.assertEqual(extractWithFormat("What was the weather last Monday?"), "2017-06-26 00:00:00")  # noqa
+        self.assertEqual(extractWithFormat("What's the weather the day after tomorrow"), "2017-06-29 00:00:00")  # noqa
+        self.assertEqual(extractWithFormat("What's the weather like on February 19?"), "2018-02-19 00:00:00")  # noqa
+        self.assertEqual(extractWithFormat("What's the weather at 2 am tomorrow?"), "2017-06-28 02:00:00")  # noqa
+        self.assertEqual(extractWithFormat("remind me to call mom in 8 weeks and 2 days"), "2017-08-24 00:00:00")  # noqa
+        self.assertEqual(extractWithFormat("start the attack on Monday at o 600 hours"), "2017-07-03 06:00:00")  # noqa
+        self.assertEqual(extractWithFormat("start the attack on Monday at 5 o'clock in the evening"), "2017-07-03 17:00:00")  # noqa
+
```
penrods: I think there were quite a few more test in the original.  Let's go ahead and test all of those cases plus also the returned leftover strings.

test/util/test_parse.py
```diff
@@ -4,9 +4,12 @@
 import unittest
 from mycroft.util.parse import normalize
 from mycroft.util.parse import extractnumber
-
+from mycroft.util.parse import extractdatetime_en
```
penrods: Go ahead and switch to using extract_datetime() (not the _en()) version once that change is made.

augustnmonteiro:
Another thing that would also be useful is split the function `extractdatetime_en` into some other functions so it can be reused for other languages and other functions

mycroft/util/parse.py
```diff
@@ -40,6 +44,27 @@ def extractnumber(text, lang="en-us"):
     return text
 
 
+def extract_datetime(text, anchorDate=None, lang="en-us"):
+
+    """Takes in a string and extracts a date and time.
+    Args:
+        text (str): the string to extract a date from
+        lang (str): the code for the language text is in
+    Return:
```
penrods: Add doc for the anchorDate arg.

The docstring you used for the extract_datetime_en() is actually really good, pull it up to this function and I think you should be good.  The _en() version doesn't really need the full docstring since it is more of a helper function, you can just document at the extract_datetime() level.

test/util/test_parse.py
```diff
@@ -40,6 +43,79 @@ def test_extractnumber(self):
...
+    def test_extractdatetime_en(self):
+
+        def extractWithFormat(text):
+            date = datetime(2017, 06, 27, 00, 00)
+            [extractedDate, leftover] = extract_datetime(text, date)
+            extractedDate = extractedDate.strftime("%Y-%m-%d %H:%M:%S")
+            return [extractedDate, leftover]
+
+        self.assertEqual(extractWithFormat("Set the ambush for 5 days from today")[0], "2017-07-02 00:00:00")  # noqa
+        self.assertEqual(extractWithFormat("Set the ambush for 5 days from today")[1], "set ambush")  # noqa
```
penrods: I don't want to be pedantic, but you could combine these into a single test function, e.g. something like:

```
def testExtract(text, expected_date, expected_leftover):
    res = extractWithFormat(text)
    self.assertEqual(res[0], expected_date)
    self.assertEqual(res[1], expected_leftover)

testExtract("Set the ambush for 5 days from today",
                "2017-07-02 00:00:00", "set ambush")
testExtract("What is the day after tomorrow's weather",
                "2017-06-29 00:00:00", "what is weather")
```

It'll be much more compact, less typing and more obvious.

mycroft/util/parse.py
```diff
@@ -164,6 +189,623 @@ def extractnumber_en(text):
...
+                    elif int(word) > 100 and \
+                            (wordPrev == "o" or wordPrev == "oh"):
+                        # 0800 hours (pronounced oh-eight-hundred)
+                        strHH = int(word)/100
+                        strMM = int(word) - strHH*100
+                        military = True
+                        if wordNext == "hours":
+                            used += 1
+                    elif wordNext == "hours" and word[0] != '0' and \
+                            (int(word) < 0100 and int(word) > 2400):
```
forslund: Do you mean `0100`(64 as numbers starting with leading `0` is interpreted as octal)

Anyway you shouldn't mix octal representation and normal decimal representation in the same if statement.

connor-penrod: I actually did not know it would interpret that as octal. I just figured it would be more clear that this is intended to parse military time. Will fix.

test/util/test_parse.py
```diff
@@ -40,6 +43,52 @@ def test_extractnumber(self):
...
+        testExtract("set up the appointment for 2 weeks and 6 days from Saturday", "2017-07-21 00:00:00", "set up appointment")  # noqa
+        testExtract("Begin the invasion at 03 45 on Thursday", "2017-06-29 03:45:00", "begin invasion")  # noqa
+        testExtract("Begin the invasion at o 800 hours on Thursday", "2017-06-29 08:00:00", "begin invasion")  # noqa
+        testExtract("Begin the invasion at 8 o'clock in the evening on Thursday", "2017-06-29 20:00:00", "begin invasion")  # noqa
+        testExtract("Begin the invasion at 8 in the evening on Thursday", "2017-06-29 20:00:00", "begin invasion")  # noqa
+        testExtract("Begin the invasion on Thursday at noon", "2017-06-29 12:00:00", "begin invasion")  # noqa
+        testExtract("Begin the invasion on Thursday at midnight", "2017-06-29 00:00:00", "begin invasion")  # noqa
+        testExtract("Begin the invasion on Thursday at 0500", "2017-06-29 05:00:00", "begin invasion")  # noqa
+        testExtract("remind me to wake up in 4 years", "2021-06-27 00:00:00", "remind me to wake up")  # noqa
+        testExtract("remind me to wake up in 4 years and 4 days", "2021-07-01 00:00:00", "remind me to wake up")  # noqa
```
forslund: Please break these up in two lines each so `# noqa` isn't required. Readability won't be hurt by

```python
        testExtract("Begin the invasion at 8 in the evening on Thursday",
                    "2017-06-29 20:00:00", "begin invasion")
```

mycroft/util/parse.py
```diff
@@ -164,6 +222,580 @@ def extractnumber_en(text):
...
+                        if wordNext == "o'clock":
+                            used += 1
+                        if wordNext == "in" or wordNextNext == "in":
+                            used += (1 if wordNext == "in" else 2)
+                            if (
+                                    wordNextNext and
+                                    wordNextNext in timeQualifier or
+                                    (words[words.index(wordNextNext)+1] and
+                                    words[words.index(wordNextNext)+1] in timeQualifier)  # noqa
+                                ):
```
forslund: In python there is no gain from having paranthesis around an entire if statement.

mycroft/util/parse.py
```diff
@@ -135,7 +193,7 @@ def extractnumber_en(text):
         aWords[count] = ""
 
         if (andPass):
```
forslund: parenthesis not required

mycroft/util/parse.py
```diff
@@ -164,6 +222,580 @@ def extractnumber_en(text):
     return val
 
 
+def extract_datetime_en(str, currentDate=None):
```
forslund: This function is huge! Is there any way to split it up in smaller functions?

mycroft/util/parse.py
```diff
@@ -164,6 +222,580 @@ def extractnumber_en(text):
...
+    if (
+            not found and
+            (
+                datestr == "" and timeStr == "" and
+                yearOffset == 0 and monthOffset == 0 and
+                dayOffset is False and hrOffset == 0 and
+                hrAbs == 0 and minOffset == 0 and
+                minAbs == 0 and secOffset == 0
+            )
+            ):
```
forslund: With if statements this big it might be better to have a separate function `date_found()` so the if statement becomes

```python
if not date_found():
```

The code gets more readable and the comment isn't really needed.

In this case you have many variables but if you make it a nested function you won't need to pass them.

mycroft/util/parse.py
```diff
@@ -164,6 +222,580 @@ def extractnumber_en(text):
...
+    # strip out 's, plurals, and ordinals (3rd becomes 3)
+    for idx, word in enumerate(words):
+        word = word.replace("'s", "")
+
+        ordinals = ["rd", "st", "nd", "th"]
+        if word[0].isdigit():
+            for ord in ordinals:
+                if ord in word:
+                    word = word.replace(ord, "")
+        words[idx] = word
```
forslund: This could for example easily be extracted to a helper function

ethanaward:
Did you ever look into the parsedatetime library that I mentioned? We already have it as a pip requirement, and it seems like it would simplify this a lot.

connor-penrod:
@ethanaward I'll just paste what I told Steve in regards to parsedatetime:

"
Alright, so on further testing I think we should stick with our own date/time parsing. 
The parsedatetime library failed on over half of the tests. Some of these were trivial, like parsedatetime interpreting "morning" as 6 AM instead of 8 AM, but others were more serious, like it misinterpreting "4th of July" as the 1st of July, or misinterpreting "2 weeks and 6 days from Saturday" (Saturday being 1st of July) as the 28th of July and not the 21st. Also, it seems inconsistent in what time it assigns sentences that don't specify a time. 

For example, "On Monday, buy groceries from Price Chopper" was interpreted with a correct date and a time of 00:00:00. 
However, "What's the weather next Thursday" was interpreted with a correct date but a time of 09:00:00.
This assigning of a 9 AM time happens a lot but not always, which could be very confusing for users.

And lastly, parsedatetime library doesn't work with military time expressions or other, more obscure ways of describing time. 
Phrases like "0800 hours", "03 45", "7 o'clock" and "o 800 hours" are badly misinterpreted.
"
Also apparently the library doesn't support natural language processing for non-English languages.

mycroft/util/parse.py
penrods: I take a large part of the blame for this.  The original PHP function was also a beast, but I left it that way since the monster was at least caged in this one function.

I'd rather get this in sooner than later, so I asked him to come back l later to do a refactoring.

mycroft/util/parse.py
connor-penrod: @forslund 
I agree with most of this. I literally directly translated this from Steve's prototype PHP code, so it's definitely a bit messy. I'll break it into subfunctions, and implement that `date_found()`

The parentheses around some of the `if` statements is intentional though. It allows me to use python's [implicit line continuation](https://docs.python.org/2.5/ref/implicit-joining.html), which is PEP8's [preferred method of line continuation](https://www.python.org/dev/peps/pep-0008/#maximum-line-length). 
And that `if (andPass)` is in a different utility function that I did not write.

penrods:
I'll add one other thought -- by implementing this here we provide a base for German, French, etc implementations.  The parsedatetime library is geared at written language, e.g. handling "5.7.2017" to convert it to July 5th, 2017.  But even in those locales that use this format, people don't say "five dot seven dot 2017", so it is of limited use for a verbal interface.

ethanaward:
Gotcha, yeah that makes a lot of sense. It seemed to be working pretty well from the couple of times I tested it, but that was likely just some sample bias. It also makes a lot of sense that we want to be pretty flexible with the language and format of potential times.

forslund:
I understand your reasoning and I'm fine with merging it in this state as long as we remember to refactor it in the future.

I think we could probably use parsedatetime to move some of the logic out of Mycroft but it needs some help to work well with natural speech. I did a date-extracter for the gcalendar-skill based on parsedatetime and some extra logic...

If this gets good and tidy we could probably try to contribute back to parsedatetime or possibly create a standalone module since it would probably be useful to others.

PR #873, opened 2017-06-30
#871 - Updating all skills

augustnmonteiro:
Removing the dirty repo check to update all skills

### DIFF
--- a/msm/msm
+++ b/msm/msm
@@ -166,7 +166,7 @@ cd "${mycroft_skill_folder}"
 for d in $(find "${mycroft_skill_folder}" -mindepth 1 -maxdepth 1 -type d |grep -v '.git'$ ); do
   if git -C "$d" rev-parse --git-dir > /dev/null 2>&1; then
     cd "${d}" 
-    if ! cat .git/info/exclude | grep "*.pyc" >> /dev/null; then
+    if ! grep -q '.pyc'$ .git/info/exclude; then
       echo "*.pyc" >> .git/info/exclude
     fi
     if [[ -z $(git status --porcelain) ]]; then
--- a/msm/msm
+++ b/msm/msm
@@ -166,8 +166,9 @@ cd "${mycroft_skill_folder}"
 for d in $(find "${mycroft_skill_folder}" -mindepth 1 -maxdepth 1 -type d |grep -v '.git'$ ); do
   if git -C "$d" rev-parse --git-dir > /dev/null 2>&1; then
     cd "${d}" 
-    rm .git/info/exclude
-    echo "*.pyc" >> .git/info/exclude
+    if ! cat .git/info/exclude | grep "*.pyc" >> /dev/null; then
+      echo "*.pyc" >> .git/info/exclude
+    fi
     if [[ -z $(git status --porcelain) ]]; then
       git fetch
       git reset --hard origin/master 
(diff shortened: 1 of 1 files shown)


kfezer:
Can you add a description to this @augustnmonteiro ?

augustnmonteiro:
Done

msm/msm
```diff
@@ -166,11 +166,14 @@ cd "${mycroft_skill_folder}"
 for d in $(find "${mycroft_skill_folder}" -mindepth 1 -maxdepth 1 -type d |grep -v '.git'$ ); do
   if git -C "$d" rev-parse --git-dir > /dev/null 2>&1; then
     cd "${d}" 
+    rm .git/info/exclude
```
penrods: Why would you remove an existing exclude file?  You just want to add *.pyc if it isn't in that file.  This is a little extra work, but if a dev has gone to the effort to add *.txt to an exclude then we don't want to wipe that out.

el-tocino:
Perhaps 
if ! grep -q '.pyc'$ .git/info/exclude ; then
  echo "*.pyc" >> .git/info/exclude
fi

msm/msm
```diff
@@ -166,11 +166,15 @@ cd "${mycroft_skill_folder}"
 for d in $(find "${mycroft_skill_folder}" -mindepth 1 -maxdepth 1 -type d |grep -v '.git'$ ); do
   if git -C "$d" rev-parse --git-dir > /dev/null 2>&1; then
     cd "${d}" 
+    if ! cat .git/info/exclude | grep "*.pyc" >> /dev/null; then
```
el-tocino: Uuoc, also should probably be grep -q?

msm/msm
```diff
@@ -166,11 +166,22 @@ cd "${mycroft_skill_folder}"
...
+    LOCAL=$(git rev-parse @)
+    REMOTE=$(git rev-parse "$UPSTREAM")
+    BASE=$(git merge-base @ "$UPSTREAM")
+
+    if ! grep -q '.pyc'$ .git/info/exclude; then
+      echo "*.pyc" >> .git/info/exclude
+    fi
+
+    # Checking if the repo isn't dirty or have commits to push
+    if [[ (-z $(git status --porcelain)) && !($LOCAL != $REMOTE && $REMOTE = $BASE) ]]; then
```
MatthewScholefield: Should we add `--untracked-files=no` to the `git status` command in case the skill creates new files?

augustnmonteiro: I think that the client has to now how to handle those situations, the problem is that we're using git for one thing that it isn't designed to do. A developer can create a skill and fix some bug and then rebase it if it happens the client should know how to update the skill If the user isn't a developer he doesn't know how to fix it.

el-tocino: Testing if we're not synced to something first maybe, then skip any further changes if not?

if ! git diff-index --quiet HEAD --; then
  continue (or return XYZ)
fi

augustnmonteiro: The skill storage is stored on ~/.mycroft/skill/<skill_name> so if the skill need to create they are created there

MatthewScholefield:
@augustnmonteiro Oh, ok that makes sense. But, for instance, what if a skill doesn't use our storage tools and just writes `os.basedir(sys.abspath(__file__))` which I'm pretty sure would write in the `/opt/mycroft/skills` directory? I don't think it would hurt to only check for tracked files.

augustnmonteiro:
@MatthewScholefield that make sense. if the skill creates a file there and msm check for tracked and untracked won't update the skill

msm/msm
penrods: LGTM

Issue #842, opened 2017-06-21
snowboy hotword-detection. build in as option or why not?


hersche:
hello all,

i would like to build in the snowboy hotword-detection. this has basicly one big reason: performance!
while pocketsphinx use a lot of cpu (near 100% on 1-2 cores), snowboy use not 10%, both on pi, both local/offline.. i quick-testet it personal too, definently faster reacting on both plattforms (pi, x86_64)

http://docs.kitt.ai/snowboy/

well, i would think this is a good idea, at least as option for pi's. if not whished, i would like to build in this option in my fork.

when i go to localreconizer, i could nearly replace the decoder, it's much simpler.. i have to add one file and a method for handling, that's it, basicly.. like here

https://github.com/Kitt-AI/snowboy/blob/master/examples/Python/demo.py#L31

well, the question for me is: what method i should call or how can i use the emitter in localreconizer (cause it doesn't exist there yet) for activating the recording-part, to make the pling, all this stuff..

how doing this from localreconizer as simple as possible?

thanks in advance!

hersche:
for eventually interessted, it's proof-of-concept-changed in my branch:
https://github.com/Skills-And-Translations/mycroft-core/tree/snowboy

MatthewScholefield:
@hersche [Psst](https://gist.github.com/anonymous/6915f546be6372f6257f7c8ab3882afe)

JarbasAI:
made a snowboy configurable for wake up and wakeword, have not tested it yet, should allow more than 1 wake word

snowboy recognizer - https://github.com/JarbasAI/JarbasAI/tree/patch-15/mycroft/client/speech/recognizer

listener changes - https://github.com/JarbasAI/JarbasAI/blob/patch-15/mycroft/client/speech/listener.py

snowboy in config - https://github.com/JarbasAI/JarbasAI/blob/patch-15/mycroft/configuration/mycroft.conf

hersche:
oh, nice, i should test this..

atm i'm a bit out of topic, so if someone else will finish it, it's ok. else i will take some time in about a month.

marksev1:
Jarbas nice, cant test am on vacation haha, and also this git is such a pain for me, I will never learn haha :)

arjenlentz:
I've opened an issue at the snowboy to get the license clarified.  The README.md is ambiguous, and source files don't contain a licensing header (just a copyright).

[...]

PR #876, opened 2017-07-03
Make Wake Word detection modular

MatthewScholefield:
This separates Pocketsphinx into its own class allowing other wake word recognizes to be switched in easier.


hersche:
could this step may closing #842 since the second module could use snowboy or so?
i dislike to reinvent the wheel more than necessary ;)

MatthewScholefield:
@forslund Fixed! :)

mycroft/client/speech/recognizer/pocketsphinx_recognizer.py
```diff
@@ -19,16 +19,18 @@
 import os
 import tempfile
 import time
-from os.path import join, dirname, abspath
 
+from os.path import join, dirname, abspath
 from pocketsphinx import Decoder
 
+from mycroft.client.speech.local_recognizer import LocalRecognizer
```
forslund: You missed a tiny thing =)

This should be
```python
from mycroft.client.speech.recognizer.local_recognizer import LocalRecognizer
```

MatthewScholefield: Good catch! Fixed.

forslund:
Looking good!

PR #440, opened 2016-12-30
#438 - Added KaldiSTT as a Speech to Text-Engine.

nilsseipel:
Kaldi gstreamer server has to run on localhost. Use this kaldi gstreamer server: https://github.com/alumae/kaldi-gstreamer-server 
Tested on a local machine with the kaldigstserver shipped with lucida (another AI). For using KaldiSTT you have to set the config for server to "update": false and the stt to "module": "kaldi" with "port" to your kaldigstserver port Number. Default is 8081.

PR #878, opened 2017-07-04
Kaldi STT

Teagan42:
Finish https://github.com/MycroftAI/mycroft-core/pull/440

Can someone with Kaldi please verify - I had to wipe my server the other day and haven't stood kaldi back up.

### DIFF
--- a/mycroft/stt/__init__.py
+++ b/mycroft/stt/__init__.py
@@ -116,8 +116,7 @@ class KaldiSTT(STT):
 
     def execute(self, audio, language=None):
         language = language or self.lang
-        config = ConfigurationManager.get().get("stt", {})
-        response = post(config.get("uri"), data=audio.get_wav_data())
+        response = post(self.config.get("uri"), data=audio.get_wav_data())
         return self.get_response(response)
 
     def get_response(self, response):
--- a/mycroft/stt/__init__.py
+++ b/mycroft/stt/__init__.py
@@ -22,6 +22,8 @@ from mycroft.api import STTApi
 from mycroft.configuration import ConfigurationManager
 from mycroft.util.log import getLogger
 
+import re
+
 __author__ = "jdorleans"
 
 LOG = getLogger("STT")
@@ -106,12 +108,31 @@ class MycroftSTT(STT):
         return self.api.stt(audio.get_flac_data(), self.lang, 1)[0]
 
 
+class KaldiSTT(STT):
+    def __init__(self):
+        super(KaldiSTT, self).__init__()
+
+    def execute(self, audio, language=None):
+        language = language or self.lang
+        config = ConfigurationManager.get().get("stt", {})
+        response = requests.post(self.config.get("uri"), data=audio.get_wav_data())
+        return get_response(response)
+
+    def get_response(response):
+        try:
+            hypotheses = response.json()["hypotheses"]
+            return re.sub(r'\s*\[noise\]\s*', '', hypotheses[0]["utterance"])
+        except:
+            return None
+
+
 class STTFactory(object):
     CLASSES = {
         "mycroft": MycroftSTT,
         "google": GoogleSTT,
         "wit": WITSTT,
-        "ibm": IBMSTT
+        "ibm": IBMSTT,
+        "kaldi": KaldiSTT
     }
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


mycroft/stt/__init__.py
```diff
@@ -106,12 +108,31 @@ def execute(self, audio, language=None):
...
 
 
+class KaldiSTT(STT):
+    def __init__(self):
+        super(KaldiSTT, self).__init__()
+
+    def execute(self, audio, language=None):
+        language = language or self.lang
+        config = ConfigurationManager.get().get("stt", {})
+        response = requests.post(self.config.get("uri"), data=audio.get_wav_data())
```
augustnmonteiro: Can you split this line, is more than 80 columns it brokes pep8, thank you

forslund:
I'm currently building Kaldi to give this a try. but it seems to take forever! If someone with an existing install would like to test this my struggling CPU would be happy =)

mycroft/stt/__init__.py
forslund: also: the module requests need to be imported at the top of the file. You can do

```python
from requests import post
```

And then shorten this line by just making it
`response = post(self.config.get("uri"), data=audio.get_wav_data())`

augustnmonteiro:
@forslund I mine just finished to compile, I'm downloading the models to run, but yeah, takes a lot of time

mycroft/stt/__init__.py
```diff
@@ -106,12 +108,31 @@ def execute(self, audio, language=None):
...
+        return get_response(response)
```
forslund: this should be
```python
        return self.get_response(response)
```

mycroft/stt/__init__.py
```diff
@@ -106,12 +108,31 @@ def execute(self, audio, language=None):
...
+
+    def get_response(response):
```
forslund: this should be 
```python
    def get_response(self, response):
```

Teagan42:
Sorry about that - wrote it real quick and didn't have a mycroft instance to test it against, in the middle of rebuilding some servers at home. Please check the latest commit

mycroft/stt/__init__.py
Teagan42: yup, yup and yup.

mycroft/stt/__init__.py
Teagan42: Context switching between python for some side projects and Java for work - my bad. addressed.

mycroft/stt/__init__.py
Teagan42: addressed.

mycroft/stt/__init__.py
```diff
@@ -106,12 +110,31 @@ def execute(self, audio, language=None):
         return self.api.stt(audio.get_flac_data(), self.lang, 1)[0]
 
 
+class KaldiSTT(STT):
+    def __init__(self):
+        super(KaldiSTT, self).__init__()
+
+    def execute(self, audio, language=None):
+        language = language or self.lang
+        config = ConfigurationManager.get().get("stt", {})
```
forslund: This config isn't actually used and can be removed

feitingen:
Great, it works for me with my kaldi setup!

May I suggest altering the comments in `mycroft/configuration/mycroft.conf`?
Specifically around 156:

`  // Speech to Text parameters`
`  // Override: REMOTE`
`  "stt": {`
`    // Engine.  Options: "mycroft", "google", "wit", "ibm"`
`    "module": "mycroft"`
`  },`


to something like
`  // Speech to Text parameters`
`  // Override: REMOTE`
`  "stt": {`
`    // Engine.  Options: "mycroft", "google", "wit", "ibm", "kaldi"`
`    "module": "mycroft"`
`    // URI required for kaldi`
`    // "uri": "http://localhost:8080/client/dynamic/recognize"`
`  },`

forslund:
Looking at the base class the config should really be something like
```json
"stt": {
  "module": "kaldi",
  "kaldi": {
    "uri": "http://localhost"
  }
}
```

using this setup, `self.config` (which gets mapped to stt['kaldi']) can be used instead of the local variable that's created.

This is the format used by the other STT backends that need extra parameters.

Teagan42:
@forslund addressed. You're right, I should have looked at the base class more carefully

forslund:
Nice! Thank you!

feitingen:
I checked out again, and still works nicely on my desktop.
Thank you!

Teagan42:
Can we get this merged up?

augustnmonteiro:
@Teagan42 thank you!!!

PR #882, opened 2017-07-05
Issue#850 Check for mimic.

el-tocino:
Since mimic is now available from packages on Fedora 26 (and probably elsewhere), or from previous build, query about building again if it's found.

Additionally, allow for mimic build to be skipped if correct parameter is passed to script.  

pip was called as both pip and pip2, I changed this to just pip for consistency. Both are called after the activate, so should be using the virtualenv's default pip?

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -40,6 +40,14 @@ else
     VIRTUALENV_ROOT="$WORKON_HOME/mycroft"
 fi
 
+# check for existing mimic (f26/previous build/etc)
+if hash mimic ; then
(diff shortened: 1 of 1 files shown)


aatchison:
Cool, I like it.
Let me test it here.

aatchison:
This works very nicely, thanks!

Blackbaud-RyanSnedegar:
NOTE:  I accidentally submitted portions of personal work from my own equipment while logged in to this work account instead of my personal account @el-tocino.  This was not done as part of my job or on work-time.

PR #888, opened 2017-07-05
Fully fix wake up functionality

MatthewScholefield:
- Increases threshold a little to make it possible to wake him up
 - Check if `None` before if block to prevent `NoneType` errors


forslund:
Looks good and works fine, Merging

PR #891, opened 2017-07-06
Wait until playback thread is done before sending audio_output_end

forslund:


forslund:
Closing, fixed in better way!

PR #892, opened 2017-07-06
Move audio output start and end messages to TTS class

MatthewScholefield:
This prevents them from being outputted too early if playback is run on another thread

In addition, this clears out redundant quiet checks


forslund:
Looks good to me! The only possible advantage my version has is that it limits the number of audio_output_start/end to once before and once when the queue is empty.

forslund:
I suggest we merge this one.

MatthewScholefield:
OK, I added a line to make it behave the same. (but still needs to be tested on a device)

PR #894, opened 2017-07-07
Check queue empty with self.queue.empty() instead of len()

### LINKED ISSUE #893: audio_end doesn't trigger in tts playback thread
I think a minor bug snuck into #892, Queue objects has no `len()` and the empty status needs to be checked with `.empty()`.

This means that the `...audio_output_end` message isn't sent.(and thus mic isn't unmuted, etc.)

forslund:
Resolves #893.


MatthewScholefield:
Tested and it now works properly.

PR #899, opened 2017-07-10
Add intent fallback system

MatthewScholefield:
This is a new system that allows other skills to register as fallbacks to handle general knowledge queries. For now, each skill self-assigns a priority where `0` is the `highest` and `100` is very low priority. When an intent failure occurs, each fallback gets run until one returns True meaning it found a response to speak to the user. Example usage:

**`skill-my-fallback/__init__.py`**:
```Python
class MyFallbackSkilll(MycroftSkill):
    def __init__(self):
        MycroftSkill.__init__(self, name="MyFallbackSkill")

    def initialize(self):
        self.register_fallback(self.handle_fallback, 80)

        if 'what is' in message.data['utterance']:
            self.speak_dialog('the answer is always 42')
            return True
        return False
```

This PR also removes the multi utterance intent fail. It only makes sense to emit an `intent_failure` regardless of the amount of intents, especially since we only ever receive one utterance anymore.

This PR relies on [this PR][skill-pr] in skill-wolfram-alpha. Instructions for testing:
```
git fetch
git checkout feature/fallbacks
mycroft_dir="$(pwd)"
cd /opt/mycroft/skills/skill-wolfram-alpha
git fetch
git checkout feature/fallbacks
cd "$mycroft_dir"
./mycroft.sh start
```
Then test functionality of Wolfram Alpha for questions like `what is a fox` (and maybe wolfram skill reloading when `/opt/mycroft/skill-wolfram-alpha/__init__.py` is edited).

[skill-pr]:https://github.com/MycroftAI/skill-wolfram-alpha/pull/3

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -20,6 +20,7 @@ import abc
 import imp
 import time
 
+import operator
 import os.path
 import re
 import time
@@ -187,6 +188,8 @@ class MycroftSkill(object):
     Skills implementation.
     """
 
+    fallback_handlers = {}
+
     def __init__(self, name, emitter=None):
         self.name = name
         self.bind(emitter)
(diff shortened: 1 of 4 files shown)


mycroft/skills/core.py
```diff
@@ -199,6 +202,38 @@ def __init__(self, name, emitter=None):
...
+        Fallback should receive message and return
+        a boolean (True if succeeded or False if failed)
+
+        Lower priority gets run first
+        0 for high priority 100 for low priority
+        """
+        while priority in MycroftSkill.fallback_handlers:
+            priority += 1
+
+        MycroftSkill.fallback_handlers[priority] = handler
```
augustnmonteiro: in this case, if have another skill with the same priority it's gonna override the first one.

MatthewScholefield: Well, it should keep incrementing the priority in the two lines of code above until there is no handler at that priority.

augustnmonteiro: Sorry, I didn't saw

forslund:
Would it be an idea to subclass a FallbackSkill from MycroftSkill with these new methods? MycroftSkill keeps growing (Partly my own fault) and this may be a good oppurtunity to split off some parts. Backwards compatibility might be tricky but worst case would be a fork of the original wolfram skill.

MatthewScholefield:
@forslund Yeah, I love the idea! Just implemented.

forslund:
@MatthewScholefield, looks great!

PR #900, opened 2017-07-10
Feature/issue 898 -- Create reboot handler for reboot on signal at start-up

aatchison:
This modified the enclosure client to look for a /tmp/mycroft/ipc/signal/reboot at start-up.
This is in case we need a reboot after update


aatchison:
no longer needed

Issue #887, opened 2017-07-05
msm may not handle mk1 correctly.


el-tocino:
https://community.mycroft.ai/t/mark-1-workon-command-not-found/1810
looks like need to account for picroft and mk1 similarly, though, shouldn't be too difficult to fix.

el-tocino:
Maybe....if [[ "$(hostname)" == 'picroft' ]] || [[ "$(hostname)" =~ "mark1" ]] &&  [[ -x /usr/local/bin/mycroft-wifi-setup-client ]] ;

nealmcb:
I'm not quite sure where you mean that to run, but note there's an underscore in the hostname "`mark_1`"
 (though it causes confusion with mdns....):

```
if [[ "$(hostname)" == 'picroft' ]] || [[ "$(hostname)" =~ "mark_1" ]] &&
 [[ -x /usr/local/bin/mycroft-wifi-setup-client ]]
```

el-tocino:
Nowhere yet, saving notes.  Agree that the underscore is probably not a good choice, I'd have gone with the hyphen if anything.

ChrisFernandez:
I just run into this issue trying to install some 3rd party skills on my new mark1

el-tocino:
Can you try this version of msm and see if it resolves the issue?
https://github.com/el-tocino/mycroft-core/blob/46d5fc7d3306b44b20fbc7f5f2ba3dc74b721e49/msm/msm

shaun040:
@el-tocino That did the trick! Able to install skills via msm now. Cheers!

nealmcb:
Worked for my Mark 1 also, which has an unchanged hostname. Thanks! 

But I left a comment on the code commit to try to make it more robust.

el-tocino:
Not sure there's a better way to handle it other than only checking for the wifi setup client, since both picroft and mk1 units can both have that changed.  I'd have to look up the way to edit docs or get one of the mycroft folks to do so to indicate that, perhaps.  Or have an /etc/mycroft/[my,pi]croft-release file or something that's part of those installs.

nealmcb:
Yes - I was thinking about something like a mycroft release file.

el-tocino:
PR merged, closing.

PR #907, opened 2017-07-12
Issue 887 -- msm mk1 fix

el-tocino:
This should help with the issue discussed here: https://github.com/MycroftAI/mycroft-core/issues/887
https://community.mycroft.ai/t/found-how-to-install-third-party-but-is-not-working/1830/3

Would like to see a more standard way of determining mk1/picroft; open to suggestions on that.

### DIFF
--- a/msm/msm
+++ b/msm/msm
@@ -29,8 +29,8 @@ if [[ ! -d "${mycroft_skill_folder}" ]] ; then
     exit 101
 fi 
 
-# picroft?
-if [[ "$(hostname)" == 'picroft' ]] || [[ -x /home/pi/bin/cli ]] ; then
+# picroft/mk1?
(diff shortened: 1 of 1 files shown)


ethanaward:
There also will need to be some kind of change for the upcoming debian/ubuntu package installs, as in those circumstances there is no virtualenv, all dependencies are taken care of via debian. That'll probably be a larger change than this though, especially because we need to think about how we want to deliver skills in those cases

el-tocino:
It would also be nice to have a more reliable way to determine what kind of install you have to allow for better handling of this, ie, have /etc/mycroft/release (package/picroft/mk1), everything else have ~/.mycroft/release or something. Keep install type, version, paths, username in it maybe.

ethanaward:
Yeah that's definitely true, I know we have pieces of that in places but it's definitely not standardized or to the point where we can really easily determine the platform.

Issue #914, opened 2017-07-14
Add script to update dev environment 


augustnmonteiro:
* When requirement.txt is changed the dev have to update the virtualenv

PR #915, opened 2017-07-14
#914 - Adding script to update the dev env

augustnmonteiro:
When a requirement is added the virtualenv have to be updated

* Updating dev virtual environment


aatchison:
Make sure you `chmod +x update-dev.sh`, it's not executable after my pull.

PR #916, opened 2017-07-14
Creating installation script

augustnmonteiro:
Creating an installation script to install all OS dependencies and prepare Mycroft virtualenv.

* Checking distros and using the right script to install dependencies
* Running dev_setup
* Updating Readme


penrods:
Check out the installer repo (https://github.com/MycroftAI/installers) we created recently.  The idea behind this is to be able to have a single place to pull down and setup all of the pieces in "one-click" as much as possible for all of the different ways Mycroft can be used.  E.g. KDE Plasmoid, Ubuntu Gnome widget, ubuntu non-graphical, Raspberry Pi, ...  I think this belongs in there.

penrods:
@augustnmonteiro please make this a PR in the https://github.com/MycroftAI/installers repo instead.  I'm closing this one.

PR #918, opened 2017-07-18
Make skill name parameter optional.

### LINKED ISSUE #906: Remove name parameter from MycroftSkill constructor
No reason to pass a name. Can use `self.__class__.__name__` instead. Do note, however, that for the reasonable future the constructor should still optionally accept the parameter as a means of backwards compatibility.

forslund:
Make name parameter optional for MycroftSkill.__init__().

If a name isn't specified, it defaults to the name of the class.

==== Fixed Issues ====
Resolves #906.

==== Documentation Notes ====
The skill creation documentation should mention that the name parameter
is only an option.

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -187,15 +187,16 @@ class MycroftSkill(object):
     Skills implementation.
     """
 
-    def __init__(self, name, emitter=None):
-        self.name = name
+    def __init__(self, name=None, emitter=None):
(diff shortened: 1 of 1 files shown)


penrods:
Let's beef up the PR commit message.  I don't know where it is not optional.  :)

forslund:
@penrods commit message is now updated according to the new template

PR #920, opened 2017-07-18
Make expect response trigger when mycroft stops speaking

### LINKED ISSUE #919: expect_response problems
if `expect_response` is used to trigger Mycroft listening this happens before Mycroft has finished speaking.

forslund:
Fixes bug that was introduced when moving to threaded approach for tts. Previously mycroft started listening as soon as tts had generated all voice data when `expect_response` was used to trigger listening. 

This PR makes listening trigger on `recognizer_loop:audio_output_end`. This should ensure that audio completes before the listening is triggered.
Resolves #919

### DIFF
--- a/mycroft/client/speech/main.py
+++ b/mycroft/client/speech/main.py
@@ -96,6 +96,13 @@ def handle_multi_utterance_intent_failure(event):
     mute_and_speak("Sorry, I didn't catch that. Please rephrase your request.")
 
 
+def _trigger_expect_response(message):
+    """
+        Makes mycroft start listening on 'recognizer_loop:audio_output_end'
+    """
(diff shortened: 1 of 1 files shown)


MatthewScholefield:
The code looks good. Is there a skill that uses this that I could quickly test with?

forslund:
This is my minimal test case: https://drive.google.com/open?id=0B6NfzYSsivqTUWV5MG5PTTh4SEU

MatthewScholefield:
Will merge once PR message includes the issue's description "if `expect_response` is used to trigger Mycroft listening this happens before Mycroft has finished speaking."

forslund:
Pr message updated!

PR #925, opened 2017-07-20
converse method

JarbasAI:
Add conversational support to skill system

The  most recently used skills now have an opportunity to preview all
utterances before they hit the intent system.

====  Tech Notes ==== 
Skills get a preview in the order of activation -- most recent first --
and if they can consume the utterance or ignore it.  If consumed,
processing stops.  If ignored, the next most recent skill gets a shot
at it.  Finally, if no skill consumes it the intent system takes over,
running as it always has.

Skills remain "active" for 5 minutes after last use.

A skill achieves this by implementing the converse() method, e.g.

    def def converse(self, utterances, lang="en-us"):
        if .... :
            return True  # handled, consume utterance
        else:
            return False  # not for this skill, pass it along


penrods:
Thank you for breaking this out!  This'll be way easier to digest.  I'll work to get this reviewed and ready for the next build.

JarbasAI:
i suppose we could avoid changing main, if we add listeners to core.py (handle_converse_request) and register the messages on bind(). then all skills would check if they were the target and if they were call converse method

any advantage in that approach over changing main?

- allow extra data to be passed in message.data or message.context

Issue #84, opened 2016-06-02
Giving Mycroft context


ethanaward:
Mycroft should be able to know when someone is talking about a previous query, e.g.

User: Mycroft, who is Lebron James?
...
User: Mycroft, how tall is he?
Mycroft: 6 feet 8 inches.

the7erm:
I think I have an idea for this but I'd like to get some input.

For starters let's take the model of what happens with humans.  Short term & long term memory.

Humans in general can hold about 5~6 items in their short term memory at a time.

We'll probably need to build a context dict, this will hold a small piece of 
info about what was said:
#### Context item
```python
{
    "skill": "wolfram_alpha",
    # insert joke here about how your wife remembers everything ...
    "scope": "public/private" # allow the developer of the skill
                              # to mark this information public or private
                              # if it's a public any other skill can read it
                              # if it's private no other skills can read it.
    "ttl": None, # time to live if it's None ... never forget
    "data": { # this is a dict that the skill sets so the skill author
              # controls what is remembered.
        "utterance": "Mycroft, who is Lebron James?",
        "response": {
            "said": "..."
        },
        "pronoun": {
            "he": "Lebron James"
        }
    },
    "timestamp": time.time()
}
```

I'm not sure about the best structure for this one. The way python handles objects you can have the same object exist in all the dicts.

#### Context containers 
```python
# skill_context would contain both the public & private entries
# later on we can build an interface that uses a database EG the long term
# memory.
skill_context = {
    "wolfram_alpha": [{...}, {...}],
    "random_skil_name": [{...}]
}

# This would be a list that you'd .append() items to.
public_context = [{...}, {...}]
```

#### What I have in mind
Here's what it'd look like when it's all put together ... I kinda think in code. 
```python
from mycroft_context import MycroftContext
mycroft_context = MycroftContext()

class MycroftSkill(object):
    def __init__(self, name, emitter=None):
        ...
        self.context = mycroft_context.get_context(self.name)
```

```python
class MySkill(MycroftSkill):
    def handle_intent_example(self, message):
        last_ten_items = self.context.recall()
        utterance = message.metadata['utterance']
        # TODO language_parser()
        # language_parser would separate the sentence into 
        # nouns, pronouns, verbs, adverbs etc
        # That's the hard part ... figuring out who "HE" is based
        # on the response.
        # That's going to require a ALOT of work because we'll have to have
        # a whole dictionary.
        # I recommend putting the language parser in an issue of it's own.
[...]

PR #934, opened 2017-07-29
Intent context (including conversations)

forslund:
This PR adds a simple context manager to the intent service allowing for some neat features. References #84.

- Context based conversation, intent handlers adds a requirement for a context when registering an intent handler. This context can then be provided by other intent_handlers using the `self.add_context()` method. (See examples below)
- Inject context such as location or time for other skills to use. This provides the possibility to have conversations like this

>  User: Hey Mycroft, what's the time in London
>  Mycroft: 10:42
>  User: Right, what's the weather over there?
>  Mycroft: Rain with a temperature of 14 degrees C

The Context manager will in default remember 3 entries back and with a time limit of 2 minutes. There is an option to configure the context management to be _greedy_, which will make it add every keyword as context. Mycroft will then guess correctly in a lot of cases but false positives will occur and the fallback skill(s) will not be triggered.

> User: Hey Mycroft, tell me a joke
> Mycroft: [Something about chuck norris]
> User: Hey Mycroft, tell me another
> Mycroft: [Chuck Norris does something new!]

# Context based conversation Examples

New Style using decorators

```python
from mycroft.util.log import getLogger
from mycroft.skills.context import *

LOGGER = getLogger(__name__)


class ContextTestSkill(MycroftSkill):
    def __init__(self):
        super(ContextTestSkill, self).__init__(name="ContextTest")

    @intent_handler(IntentBuilder('TeaIntent').require("TeaKeyword"))
    @adds_context('TeaContext')
    def handle_tea_intent(self, message):
        self.speak('Very well, would you like some sugar with that',
                   expect_response=True)

    @intent_handler(IntentBuilder('NoIntent').require("NoKeyword").
                                  require('TeaContext').build())
    @removes_context('TeaContext')
    def handle_yes_intent(self, message):
        self.speak('all right, here you go')

    @intent_handler(IntentBuilder('YesIntent').require("YesKeyword").
                                  require('TeaContext').build())
    @removes_context('TeaContext')
    def handle_no_intent(self, message):
        self.speak('I see, straight up it is')
```

Old style
```python
from mycroft.util.log import getLogger

LOGGER = getLogger(__name__)


class ContextTestSkill(MycroftSkill):
    def __init__(self):
        super(ContextTestSkill, self).__init__(name="ContextTest")

    def initialize(self):
        self.register_vocabulary(self.name, self.name + 'Keyword')
        tea_intent = IntentBuilder('TeaIntent'). \
            require("TeaKeyword").optionally(self.name + 'Keyword').build()
        no_intent = IntentBuilder('NoIntent'). \
            require("NoKeyword").require('TeaContext'). \
            optionally(self.name + 'Keyword').build()
        yes_intent = IntentBuilder('YesIntent'). \
            require("YesKeyword").require('TeaContext'). \
            optionally(self.name + 'Keyword').build()

        self.register_intent(tea_intent, self.handle_tea_intent)
        self.register_intent(no_intent, self.handle_no_intent)
        self.register_intent(yes_intent, self.handle_yes_intent)

    def handle_tea_intent(self, message):
        self.sugar_offer_given = True
        self.speak('Very well, would you like some sugar with that',
                   expect_response=True)
        # Add context TeaContext so intents requiring it can be triggered
        self.set_context('TeaContext')

    def handle_yes_intent(self, message):
        self.speak('all right, here you go')
        # Remove context preventing the intent handler to trigger
        self.remove_context('TeaContext')

    def handle_no_intent(self, message):
        self.speak('I see, straight up it is')
        # Remove context preventing the intent handler to trigger
        self.remove_context('TeaContext')
```

# Inner workings

The context manager is basically a stack of context keywords which are added to the current sentence when calculating the score of the intent matching. For each possible intent adapt checks for missing keywords and sends them into the context manager. The context manager searches for context frames with those keywords and returns the latest value for these along with a probability score. The score for the context frames decrease the further back in the frame stack a usable context was found.

The list is initially limited to the 3 last context frames and a timeout of 2 minutes to keep the number of matchings low.

# Settings

These are the defaults of the available settings.

```python
    "context": {
        "timeout": 2, # Timeout in minutes
        "max_frames": 3, # Maximum number of context keywords
        "keywords": ["Location"], # Keywords that will automatically added to context
        "greedy": false # Set to true to add all Keywords to context
    }
```

### DIFF
--- a/mycroft/skills/context.py
+++ b/mycroft/skills/context.py
@@ -1,0 +1,33 @@
+from functools import wraps
+
+"""
+    Helper decorators for handling context from skills.
+"""
+
+
+def adds_context(context, words=''):
+    """
+        Adds context to context manager.
+    """
+    def context_add_decorator(func):
+        @wraps(func)
+        def func_wrapper(*args, **kwargs):
+            ret = func(*args, **kwargs)
+            args[0].set_context(context)
+            return ret
+        return func_wrapper
+    return context_add_decorator
+
+
+def removes_context(context):
+    """
+        Removes context from the context manager.
+    """
+    def context_removes_decorator(func):
+        @wraps(func)
+        def func_wrapper(*args, **kwargs):
+            ret = func(*args, **kwargs)
+            args[0].remove_context(context)
+            return ret
+        return func_wrapper
+    return context_removes_decorator
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -316,6 +316,13 @@ class MycroftSkill(object):
                 logger.error('Could not enable ' + intent_name +
                              ', it hasn\'t been registered.')
 
+    def set_context(self, context, word=''):
+        self.emitter.emit(Message('add_context', {'context': context, 'word':
+                          word}))
+
+    def remove_context(self, context):
+        self.emitter.emit(Message('remove_context', {'context': context}))
+
     def register_vocabulary(self, entity, entity_type):
         self.emitter.emit(Message('register_vocab', {
             'start': entity, 'end': entity_type
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -317,10 +317,26 @@ class MycroftSkill(object):
                              ', it hasn\'t been registered.')
 
     def set_context(self, context, word=''):
+        """
+            Add context to intent service
+
+            Args:
+                context:    Keyword
+                word:       word connected to keyword
+        """
+        if not isinstance(context, basestring):
+            raise ValueError('context should be a string')
+        if not isinstance(word, basestring):
+            raise ValueError('word should be a string')
         self.emitter.emit(Message('add_context', {'context': context, 'word':
                           word}))
 
     def remove_context(self, context):
+        """
+            remove_context removes a keyword from from the context manager.
+        """
+        if not isinstance(context, basestring):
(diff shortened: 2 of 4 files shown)


forslund:
Sorry for not having a new style commit message did this a while ago, I will prepare one for a potential merge

clusterfudge:
I think I've got the gist of this, the PR is *super* pythonic, so I'm going to trust that it works without grokking every line :). I have a couple of questions:

1) Is it possible to bind multiple intents to a single handler?
2) Is it possible to conditionally clear the context? Perhaps an intent handler may need additional input based on the values of the input, like an insufficiently specific location.
3) It appears that the context annotations impact all the entities in the result. Things that are hyper-generic (time, location) seem appropriate to remain in context indefinitely. Is it possible to add/remove only portions of the context? How would we model that with this change?

forslund:
1. Yes multiple `intent_handler` (and `adds_context`) decorators can be applied to a single method
2. Not using the decorators. but the `self.add_context()` method can be conditionally called from within an intent_handler
3. We could either add each context frame to the manager with a separate timeout (or `None`)
Another method would be to have multiple context managers, with different characteristics (long term, short term and possibly more esoteric ones).

I'll keep thinking about number 3, there are probably better ways of doing it.

PR #938, opened 2017-08-01
Add intent fallback system

MatthewScholefield:
This is the same pull request from 3 weeks ago, but with `skill-wolfram-alpha` blacklisted and a new `fallback-wolfram-alpha` added to the default skills. Note: since Wolfram's API key is currently not working, we this cannot be tested.

**Original description:**
This is a new system that allows other skills to register as fallbacks to handle general knowledge queries. For now, each skill self-assigns a priority where 0 is the highest and 100 is very low priority. When an intent failure occurs, each fallback gets run until one returns True meaning it found a response to speak to the user. Example usage:

skill-my-fallback/\_\_init\_\_.py:

```Python
class MyFallbackSkilll(MycroftSkill):
    def __init__(self):
        MycroftSkill.__init__(self, name="MyFallbackSkill")

    def initialize(self):
        self.register_fallback(self.handle_fallback, 80)

        if 'what is' in message.data['utterance']:
            self.speak_dialog('the answer is always 42')
            return True
        return False
```
This PR also removes the multi utterance intent fail. It only makes sense to emit an intent_failure regardless of the amount of intents, especially since we only ever receive one utterance anymore.

When testing, check functionality of Wolfram Alpha for questions like what is a fox (and maybe wolfram skill reloading when /opt/mycroft/skill-wolfram-alpha/__init__.py is edited).

### DIFF
--- a/msm/msm
+++ b/msm/msm
@@ -46,7 +46,7 @@ else
   fi
 fi
 
-default_skills="skill-alarm skill-audio-record skill-configuration skill-date-time skill-desktop-launcher skill-ip skill-joke skill-hello-world skill-media skill-npr-news skill-naptime skill-pairing skill-personal skill-playback-control skill-reminder skill-installer skill-singing skill-speak skill-spelling skill-stop skill-stock skill-volume skill-weather skill-wiki skill-wolfram-alpha skill-mark1-demo"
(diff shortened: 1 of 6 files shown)


forslund:
Looks extremely good. Unfortunately wolfram alpha seems to be unavailable to me right now so I haven't been able to test it.

Also PR 145 on mycroft-skills (https://github.com/MycroftAI/mycroft-skills/pull/145) need to be merged before this one.

Would like to verify it with a working wolfram alpha before merging.

PR #939, opened 2017-08-01
Implement Padatious support (examples based intent parser)

MatthewScholefield:
This brings in Padatious support for Mycroft. Usage in a skill is as follows:

\_\_init\_\_.py:
```Python
class TomatoSkill(MycroftSkill):
    def __init__(self):
        MycroftSkill.__init__(self)

    def initialize(self):
        self.register_intent_file('what.is.intent', self.handle_what_is)
        self.register_intent_file('do.you.like.intent', self.handle_do_you_like)

    def handle_what_is(self, message):
        self.speak('A tomato is a big red thing')

    def handle_do_you_like(self, message):
        tomato_type = message.get('type')
        if tomato_type is not None:
            self.speak("Well, I'm not sure if I like " + tomato_type + " tomatoes.")
        else:
            self.speak('Of course I like tomatoes!')
```

vocab/en-us/what.is.intent:
```
What would you say a tomato is?
What's a tomato?
Describe a tomato.
What defines a tomato?
```

vocab/en-us/do.you.like.intent:
```
Are you fond of tomatoes?
Do you like tomatoes?
What are your thoughts on tomatoes?

Are you fond of {type} tomatoes?
Do you like {type} tomatoes?
What are your thoughts on {type} tomatoes?
```

For debugging, you can use `./mycroft.sh start && tail -f scripts/logs/mycroft-* | grep -i padatious`

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -198,6 +198,7 @@ class MycroftSkill(object):
         self.config_core = ConfigurationManager.get()
         self.config = self.config_core.get(self.name)
         self.dialog_renderer = None
+        self.vocab_dir = None
         self.file_system = FileSystemAccess(join('skills', self.name))
         self.registered_intents = []
         self.log = getLogger(self.name)
@@ -285,29 +286,8 @@ class MycroftSkill(object):
             self.register_intent(intent_parser, handler, need_self=True)
         _intent_list = []
 
-    def register_intent(self, intent_parser, handler, need_self=False):
-        """
-            Register an Intent with the intent service.
-
-            Args:
-                intent_parser: Intent or IntentBuilder object to parse
-                               utterance for the handler.
-                handler:       handler_function
-                need_self:     optional parameter, when called from a decorated
-                               intent handler the function will need the self
-                               variable passed as well.
-        """
-        if type(intent_parser) == IntentBuilder:
-            intent_parser = intent_parser.build()
-        elif type(intent_parser) != Intent:
-            raise ValueError('intent_parser is not an Intent')
-
-        name = intent_parser.name
-        intent_parser.name = str(self.skill_id) + ':' + intent_parser.name
-        self.emitter.emit(Message("register_intent", intent_parser.__dict__))
-        self.registered_intents.append((name, intent_parser))
-
-        def receive_handler(message):
+    def add_event(self, name, handler, need_self=False):
+        def wrapper(message):
             try:
                 if need_self:
                     # When registring from decorator self is required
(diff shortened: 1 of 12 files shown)


TREE-Edu:
Just curious, will adapt still work after this is merged or will Padatious be the new standard?  Thanks.

Mn0491:
Adapt will still work :)

forslund:
@MatthewScholefield while you were away devious forces (read me) changed some things in the fallback system to allow a standard shutdown method to clean up a skills all registered fallback handlers. I think this messed up your stuff (sorry),

to get functionallity back you could revert my changes or change [L54 in padatious_service.py](https://github.com/MycroftAI/mycroft-core/pull/939/files#diff-ab990fec3d51ab7df97acb4ab180a4daR54) to
```python
        FallbackSkill._register_fallback(self.handle_fallback, 5)
```

MatthewScholefield:
@forslund Fixed! Thanks for the heads up.

mycroft/skills/core.py
```diff
@@ -290,10 +286,24 @@ def receive_handler(message):
...
-            self.events.append((intent_parser.name, receive_handler))
+            self.emitter.on(name, wrapper)
+            self.events.append((name, wrapper))
+
+    def register_intent(self, intent_parser, handler, need_self=False):
+        name = intent_parser.name
+        intent_parser.name = self.name + ':' + intent_parser.name
+        self.emitter.emit(Message("register_intent", intent_parser.__dict__))
+        self.registered_intents.append((name, intent_parser))
+        self.add_event(intent_parser.name, handler)
```
forslund: This needs to pass on `need_self` to `add_event()` otherwise the decorated intent handlers will fail.

forslund:
I've been playing with the tomato skill and it's really cool and works very well.

I see the build_host_setup script for debian has been updated but not for other distros

MatthewScholefield:
Yeah, I started setting up a Fedora and Arch Linux vbox, but never finished. Does anyone have any other distro installed that can find the package for libfann-devel?

forslund:
I can check fedora for you

forslund:
`fann-devel` is the package in Fedora as you might have expected.

If it will start without libfann then it's ok. It looks like the dev_setup script also will handle this gracefully.

No need to fire up every OS to check the rest.

MatthewScholefield:
Thanks for checking Fedora. I've updated the build scripts and decorator. Now it should work to use:
```Python
@intent_file_handler('what.is.intent')
def handler_what_is(self, message):
    #...
```

penrods:
Packaging support was added for this, too. Let's get 'er merged.

PR #945, opened 2017-08-02
altered return codes for dupe/missing

el-tocino:
Altered return codes to make response clearer.


forslund:
That's terribly quick =)

forslund:
Looking good! Merging

PR #949, opened 2017-08-02
Restricting move version: mock==2.0.0

aatchison:


penrods:
Please enter reasons when making a change!

PR #957, opened 2017-08-03
Prevent unneeded mimic compilation in dev_setup.sh

penrods:
The code that detected existing mimic installs and asked for
confirmation before rebuilding had a few holes:
* A source install of mimic under the mycroft-core tree was missed
* Hitting "enter" or "spacebar" would be interpreted as "yes, rebuild",
  which didn't match the prompt.


penrods:
Ping

PR #965, opened 2017-08-04
Build mimic with cores determined by memory calc

el-tocino:
Use system memory to calculate number of cores to build with. Pass as a parameter to the install scripts. 

zeehio indicated that each core takes ~460mb of RAM to build. For padding, used 500mb to calculate number of cores. If sysmem divided by 500mb is less than nproc, then set CORES less than nproc. On a 12gb machine, for instance, it indicates max 24 cores.  System has 8 (4+hyperthreading), so it set CORES to 8.  On a pi3, with 128mb set for gpu mem, it calculates 1 (917504/500000 = 1), as we would prefer.  

The hash command can break the -Ee settings when it doesn't find a copy of mimic, so bypass the exit on error when checking for that and re-enable error bombing after.


dev_setup.sh
```diff
@@ -61,12 +61,12 @@ if [ ! -d "${VIRTUALENV_ROOT}" ]; then
 fi
 source "${VIRTUALENV_ROOT}/bin/activate"
 cd "${TOP}"
-easy_install pip==7.1.2 # force version of pip
-pip install --upgrade virtualenv
+#easy_install pip==7.1.2 # force version of pip
+#pip install --upgrade virtualenv
```
forslund: Did you mean to have these commented out?

dev_setup.sh
```diff
@@ -61,12 +61,12 @@ if [ ! -d "${VIRTUALENV_ROOT}" ]; then
...
 
 # install requirements (except pocketsphinx)
 # removing the pip2 explicit usage here for consistency with the above use.
-pip install -r requirements.txt 
+#pip install -r requirements.txt
```
forslund: and this?

el-tocino:
@forslund no, had them commented to test the math and script paramter bits and have now reverted that.

forslund:
Hi @el-tocino 
The calculation looks okey but the removed error check seems a bit weird. Will we notice if mimic is broken in this setup?

I started a rebase of this (excluding the msm fix) in https://github.com/MycroftAI/mycroft-core/tree/eltocino-mimic-cpu-calc-rebase where I removed the `set +eE` and Travis doesn't seem to have a problem with this: https://travis-ci.org/MycroftAI/mycroft-core/builds/262208211

Is the rebased branch OK for you?

el-tocino:
I had it turned off there for that since it was exiting when mimic wasn't found.  If it doesn't fail when running on a host that does not have mimic installed it should be ok.  Will verify a bit later.

el-tocino:
So that rebase worked.  I think that'll be fine then.  I also think we can skip the output redirect (2&>1) on the hash, it shows the error but doesn't fail, which seems ok.  I can move the msm changes to a new pr if you'd prefer, will close this one.

PR #973, opened 2017-08-09
General cleanup of Mycroft Skill Manager (msm)

penrods:
Clean-up of many things about MSM:
* Improve help message, documenting all options
* Make the DEFAULT_SKILLS list readable (wrapped lines)
* Prefix messages with ERROR or WARNING
* Fix return codes, they can't be > 255.  301 and 303 are now 201 and 203
* Cache list of skills (instead of hitting web repeatedly)
* Improve code spacing, comments
* Improve format of output messages
* Sort skills from 'list' and 'search' commands
* Return result codes (was always 0)

====  Tech Notes ====
The "install" skill will be updated to use result codes instead
of searching for specific message strings.  It was already
broken due to mismatch of the success message string.

==== Localization Notes ====
With error codes, output strings are less important

### DIFF
--- a/msm/msm
+++ b/msm/msm
@@ -21,13 +21,25 @@
 # @author Augusto Monteiro
 #
 # This script assists in the installation and management of
-# skills loaded from Github.  
+# skills loaded from Github.
+
+
+# These skills are automatically installed on all mycroft-core
+# installations.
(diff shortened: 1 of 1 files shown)


penrods:
I'll likely do more cleanup.  Still needs "uninstall", fixes for running under Picroft, and support for invoking requirements.sh.

augustnmonteiro:
LGTM!

PR #980, opened 2017-08-10
allow internal configuration changes

JarbasAI:
allow cached config to be updated from messagebus, from a skill or something else

===New Signals ===

message_type: configuration.update
message_data: {config:{} , save:False}
function: Triggers cached config update with new data, optionally save

=== Example Usage ===

voice control TTS skill

youtube: https://youtu.be/cKc_RxRbJ3U
github: https://github.com/JarbasAI/JarbasAI/tree/patch-15/jarbas_skills/skill_tts_control

### DIFF
--- a/mycroft/configuration/__init__.py
+++ b/mycroft/configuration/__init__.py
@@ -301,10 +301,14 @@ class ConfigurationManager(object):
         """
         ConfigurationManager.update(config)
         location = SYSTEM_CONFIG if is_system else USER_CONFIG
-        loc_config = load_commented_json(location)
-        with open(location, 'w') as f:
-            config = loc_config.update(config)
-            json.dump(config, f)
+        try:
+            LOG.info("Saving config")
+            loc_config = load_commented_json(location)
+            with open(location, 'w') as f:
+                config = loc_config.update(config)
+                json.dump(config, f)
+        except Exception as e:
+            LOG.error(e)
 
 
 class _ConfigurationListener(object):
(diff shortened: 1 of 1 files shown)


forslund:
Looks good! three comments:

- PEP8 issue (there's an blank line at the end of the file)
- the message name is very similar to the existing `configuration.updated` maybe this new message should be named `configuration.patch`
- Is there a risk that the patched configuration is overwritten when a change is made on the server and all files are reloaded?

JarbasAI:
regarding 3rd point, online config update may replace things, but with the save flag the changes would be reloaded

testing the save function seems to throw me an error saying that the file doesnt exist, probably something to do with permissions?

forslund:
I've never actually tried the save function, will play with it and see if I get the same issue...

forslund:
I think there might be an issue with the save feature. Firstly all running processes will try to write the config causing write issues and maybe other collisions. I'm not sure about using the save function at all. It seems like it nuked my local config

JarbasAI:
so should i just remove the save setting?
or add a method to skills core that sends the message and does the saving?

forslund:
Remove the saving.

If someone really wants to (even if it may be a bad idea), they can always do it using 
```python
        ConfigurationManger.save(self.base_config)
```

after sending the patch and waiting a bit.

forslund:
Tested and works as expected this is a nice improvement. Fix the pep8
`mycroft/configuration/__init__.py:321:20: W291 trailing whitespace`
and it's good to merge.

As a second step I see the addition of a separate config layer for changes made by skills. And then maybe add `patch()` and `reload()` (configuration.updated) as method in the `ConfigurationManager` (or somewhere under `mycroft.configuration`) object to make it a bit easier to access.

PR #987, opened 2017-08-13
fallback order override

JarbasAI:
sometimes we may not agree with dev choices on fallback priority, an option to override the fallback order in config would be good, so we can get updates without git pull complaining of stashing changes, also good for dev purposes

This pr is more for discussion than a serious proposal

- add fallback override flag in config, default false
- add fallback order list with skill folder names, default empty
- on register fallback track folder to handler
- on handler if override flag try ordered list
- if entry missing in ordered list try it next by skill load order
- if no override flag behave normally
- on shutdown also remove handler from tracked folder handlers

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -1301,6 +1301,9 @@ class FallbackSkill(MycroftSkill):
         by their priority.
     """
     fallback_handlers = {}
+    skills_config = Configuration.get().get("skills", {})
+    override_order = skills_config.get("override", False)
+    fallback_order = skills_config.get("fallback_order", [])
 
     def __init__(self, name=None, bus=None):
         MycroftSkill.__init__(self, name, bus)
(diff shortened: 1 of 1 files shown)


augustnmonteiro:
@JarbasAI thank you for the contribution, travis is getting those pep8 erros
```mycroft/skills/core.py:432:80: E501 line too long (81 > 79 characters)
mycroft/skills/core.py:433:42: E225 missing whitespace around operator
mycroft/skills/core.py:438:80: E501 line too long (92 > 79 characters)
mycroft/skills/core.py:442:80: E501 line too long (85 > 79 characters)
mycroft/skills/core.py:449:80: E501 line too long (84 > 79 characters)
mycroft/skills/core.py:462:80: E501 line too long (80 > 79 characters)
mycroft/skills/core.py:484:8: E114 indentation is not a multiple of four (comment)
mycroft/skills/core.py:544:1: W391 blank line at end of file```

File: mycroft/skills/core.py [hunk 1]
```diff
@@ -401,6 +401,9 @@ def shutdown(self):
 
 class FallbackSkill(MycroftSkill):
     fallback_handlers = {}
+    folders = {}
+    override = skills_config.get("fallback_override", False)
+    order = skills_config.get("fallback_priority", [])
```
JarbasAI: i like the override parameter, i may want to have the override list there but only toggle it sometimes, with the PR for changing config at runtime from skills this can even be changed vocally

forslund: Ok, Gotcha.

JarbasAI:
changes made

Issue #984, opened 2017-08-12
install_mimic.sh -- Mimic shouldn't have to clone all branches


aatchison:
It would be better if we used deplth=1.

forslund:
Fixed in #984. Closing

PR #988, opened 2017-08-14
#984 -- no need to clone all mimic branches for the dev_setup

aatchison:
Referenced issue:#984
--
==== Fixed Issues ====
984 - install_mimic.sh -- Mimic shouldn't have to clone all branches
====  Tech Notes ====
This modification clones the mimic repository with a depth of one in
order to reduce download times and local size. A side affect might be
that this would make it harder for developers to checkout different
mimic branches, as they would have to tech them first.


forslund:
Tested: Looks good!

Merging!

PR #990, opened 2017-08-15
Add class for handling downloads

forslund:
====  Tech Notes ====
Downloader() is a Thread subclass starting and handling a download. The
property done and status is set when the download is completed or
failed.
download manager ensuring continuation of started downloads (per process)
complete_action, a function to be called after download is complete can be
supplied to the downloader.

to download a file

```python
from time import sleep
from mycroft.utils.download import download

d = download('https://mycroft.ai', 'index.html')

while not d.done:
    sleep(0.5)
print "Done"
```

or

```python
from time import sleep
from mycroft.utils.download import download

def done_action(path):
    print '{} has been downloaded'.format(path)

d = download('https://mycroft.ai', 'index.html', done_action)

while not d.done:
    sleep(0.5)
```

### DIFF
--- a/mycroft/util/download.py
+++ b/mycroft/util/download.py
@@ -1,0 +1,104 @@
+from threading import Thread
+import requests
+import os
+from os.path import exists
+
+_running_downloads = {}
+
+
+def _get_download_tmp(dest):
+    tmp_base = dest + '.part'
+    if not exists(tmp_base):
+        return tmp_base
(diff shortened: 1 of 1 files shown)


augustnmonteiro:
@forslund why need a sleep if it has a callback?

forslund:
Sorry, bad example. the sleep is technically not needed.

PR #992, opened 2017-08-16
Add mycroft-core to virtualenv path

penrods:
Add the insertion of mycroft-core in the virtualenv created by
dev_setup.sh.  It becomes part of the python2.7 site-packages.
This is needed for running mycroft-core pieces (such as the
tests) directly from the command line, after performing
"workon mycroft".


forslund:
Seems to work fine now! Merging

forslund:
I take it back, didn't work but nose2 didn't have any problems for some reason

forslund:
Seems to need something like this:
```bash
PATH_FILE="${VIRTUALENV_ROOT}/lib/python2.7/site-packages/_virtualenv_path_extensions.pth"
if [ ! -f "$PATH_FILE" ] ; then
    echo "import sys; sys.__plen = len(sys.path)" > "$PATH_FILE" || return 1
    echo "import sys; new=sys.path[sys.__plen:]; del sys.path[sys.__plen:]; p=getattr(sys,'__egginsert',0); sys.path[p:p]=new; sys.__egginsert = p+len(new)" >> "$PATH_FILE" || return 1
fi

if ! grep -q "mycroft-core" $PATH_FILE; then
   echo "Adding mycroft-core to virtualenv path"
   sed -i.tmp '1 a\
'"$TOP"'
' "${PATH_FILE}"
fi
```

penrods:
After running dev_setup.sh does 'add2virtualenv' list your mycroft-core as an absolute path?

forslund:
Not sure what you mean. do you mean the warning that's generated when add2virtualenv converts a relative path to an absolute? 

The absolute path is however in the `_virtualenv_path_extensions.pth`

PR #996, opened 2017-08-17
Remove requrement to build Intents

forslund:
====  Tech Notes ====
IntentBuilder objects always need to be built into an Intent to be
usable in mycroft. Since this is always neccessary the code doing this
can be moved so users don't need to do this step.

This commit makes the `register_intent()` method check the handler
argument type, if it's an `IntentBuilder` object it builds the `Intent`, if
it's an `Intent` it'll continue as normal, if it's any other type it will
raise a `ValueError`

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -26,7 +26,7 @@ from os.path import join, dirname, splitext, isdir
 
 from functools import wraps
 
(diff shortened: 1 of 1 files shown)


Mn0491:
I like this change. Looks good to me.

Issue #1001, opened 2017-08-18
Multiple reloads at startup


forslund:
When skill service is loaded multiple reloads of all skills occur.

```
(mycroft) ake@roadrunner:~/projects/python/mycroft-core$ ./start.sh skills |& grep Reloading
2017-08-17 11:37:51,091 - Skills - DEBUG - Reloading Skill: skill-spelling
2017-08-17 11:37:53,222 - Skills - DEBUG - Reloading Skill: skill-spelling
2017-08-17 11:37:53,259 - Skills - DEBUG - Reloading Skill: skill-date-time
2017-08-17 11:37:53,351 - Skills - DEBUG - Reloading Skill: skill-ip
2017-08-17 11:37:53,368 - Skills - DEBUG - Reloading Skill: skill-singing
2017-08-17 11:37:55,423 - Skills - DEBUG - Reloading Skill: skill-date-time
2017-08-17 11:37:55,493 - Skills - DEBUG - Reloading Skill: skill-ip
2017-08-17 11:37:55,500 - Skills - DEBUG - Reloading Skill: skill-singing
2017-08-17 11:37:55,503 - Skills - DEBUG - Reloading Skill: skill-speak
2017-08-17 11:37:55,511 - Skills - DEBUG - Reloading Skill: skill-npr-news
2017-08-17 11:37:57,582 - Skills - DEBUG - Reloading Skill: skill-speak
2017-08-17 11:37:57,597 - Skills - DEBUG - Reloading Skill: skill-npr-news
2017-08-17 11:37:57,608 - Skills - DEBUG - Reloading Skill: skill-joke
2017-08-17 11:37:57,618 - Skills - DEBUG - Reloading Skill: skill-playback-control
2017-08-17 11:37:59,692 - Skills - DEBUG - Reloading Skill: skill-npr-news
2017-08-17 11:37:59,702 - Skills - DEBUG - Reloading Skill: skill-joke
2017-08-17 11:37:59,711 - Skills - DEBUG - Reloading Skill: skill-playback-control
2017-08-17 11:38:01,789 - Skills - DEBUG - Reloading Skill: skill-playback-control
2017-08-17 11:38:01,806 - Skills - DEBUG - Reloading Skill: rss_skill
2017-08-17 11:38:03,887 - Skills - DEBUG - Reloading Skill: skill-playback-control
2017-08-17 11:38:03,898 - Skills - DEBUG - Reloading Skill: rss_skill
[...]

PR #1002, opened 2017-08-18
Fix multiple reloads of skills at startup

forslund:
==== Fixed Issues ====
#1001

====  Tech Notes ====
Previously the skills were reloaded a couple of times during startup
since updates of the .pyc file and possible the settings.json file were
made.

This commit adds a bit finer control over which files to check.
Currently all files in skill root except ones ending in .pyc and the
settings.json are checked along with all visible subdirectories

### DIFF
--- a/mycroft/skills/main.py
+++ b/mycroft/skills/main.py
@@ -156,14 +156,30 @@ def check_connection():
 
 
 def _get_last_modified_date(path):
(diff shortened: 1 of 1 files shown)


forslund:
Please test carefully. I didn't find any cases where it failed to reload an updated skill but I may have missed something.

PR #1005, opened 2017-08-21
small fix

JarbasAI:
some bugs were introduced in converse pr when rebasing, fixed 2 missing instances of changing name to skill id and a bad call to time


forslund:
Looks good to me. Merging.

PR #1013, opened 2017-08-22
Messagebus interface

JarbasAI:
I propose two new classes, BusQuery and BusResponder

BusQuery sends a message and waits for the response
BusResponder listens for messages and sends responses

### DIFF
--- a/mycroft/messagebus/api.py
+++ b/mycroft/messagebus/api.py
@@ -31,7 +31,7 @@ class BusQuery():
         self.response = Message(None, None, None)
         if response_type is None:
             response_type = self.query_type + ".reply"
-        self.emitter.on(response_type, self._end_wait)
+        self.emitter.once(response_type, self._end_wait)
         self.emitter.emit(
             Message(self.query_type, self.query_data, self.query_context))
         self._wait_response(timeout)
--- a/mycroft/messagebus/api.py
+++ b/mycroft/messagebus/api.py
@@ -5,17 +5,14 @@ __author__ = "jarbas"
 
 
 class BusQuery():
-    def __init__(self, emitter, name, message_data=None, message_context=None):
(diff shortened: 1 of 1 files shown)


forslund:
I'm not 100% sure this is a good idea. Forcing an asynchronous system into a synchronous operation will probably not be good in the long run. We've done exactly this in a couple of places but I'm starting to feel that was a mistake.

The positive side of this approach is that it's easy to understand so in that way it's very good, I need to research what sort of performance hit we might see if this is picked up in many skills.

I would in any case look into using once instead of on when setting up the waiting task (the method will be triggered only once) and look into the possibility of using the `Message.reply()` method.

JarbasAI:
using once makes sense, will change that

the messagebus is a powerful mechanism, being able to force an asynchronous system into a synchronous operation is a very useful option to have, i have been using this in a lot of skills (which i tend to call "service_something") concurrently with success,

i agree this should not be a default, i don't like using this approach in converse method for instance, but it enables a lot more options and in the long run i think it is a useful tool to have

augustnmonteiro:
@JarbasAI @forslund I think that in this case we should use a tool that was made for that, websockets are really good if you don't need order or run sync, there are message brokers that solve this problem I would take a look at the Redis Pub/Sub https://redis.io/topics/pubsub

penrods:
Does https://github.com/MycroftAI/mycroft-core/pull/1423 make this obsolete?

JarbasAl:
not exactly, #1423 solves part of what this does, but this gives more control and options

i have included this in my utils package, usage example here https://github.com/JarbasAl/mycroft_jarbas_utils/tree/master/mycroft_jarbas_utils/messagebus

JarbasAI:
#1822 makes this obsolete

PR #1016, opened 2017-08-24
enclosure base class

JarbasAI:
A base class for enclosures, a new enclosure would override the desired methods

first step towards https://github.com/MycroftAI/mycroft-core/wiki/TASK-AbstractEnclosure


JarbasAI:
#1306 redoes this

PR #1024, opened 2017-08-29
Add sphinx documentation

MatthewScholefield:
It brings in support for readthedocs.io auto generated documentation. [Here's the link to the docs][rtd] (currently being generated off my fork.

[rtd]:http://mycroft-core.rtfd.io/


MatthewScholefield:
Yeah, the URLs will be:
master: http://mycroft-core.readthedocs.io/en/latest/
dev: http://mycroft-core-dev.readthedocs.io/en/latest/

After this goes in I can setup dev and master to actually build from the right place.

Issue #1022, opened 2017-08-28
Intent_service fails when a one_of rule is matched


deevrek:
Hello,
this issue occurs inside the update_context method:

>     def update_context(self, intent):
>             for tag in intent['__tags__']:
>                 context_entity = tag.get('entities')[0]
>                 if self.context_greedy:
>                     self.context_manager.inject_context(context_entity)
>                 elif context_entity['data'][0][1] in self.context_keywords:
>                     self.context_manager.inject_context(context_entity)
  
 _tags_ returned from adapt when a "one_of" rule is matched contains the entire Keyword match instead of the "entities" attribute.
The raised exception causes the failure of the skill.

I proposed a solution in the Adapt repository.
https://github.com/MycroftAI/adapt/pull/66

forslund:
Thanks for reporting this. In the meantime I shall add a check so that it doesn't break functionality.

forslund:
This was resolved in #1026, closing

PR #1026, opened 2017-08-29
Fix failure when using one_of in intents

forslund:
==== Fixed Issues ====
#1022 

====  Tech Notes ====
When a one_of intent is hit the intent returned by adapt doesn't look like normal require/optional intent parameters. This PR adds a check for entities before trying to accessing them when trying to update context.

This is a temporary workaround while it's determined if the adapt behaviour is correct or should be modified to conform to the normal format. (See issue 66 in the adapt repo), in any case it's a good sanity check

### DIFF
--- a/mycroft/skills/intent_service.py
+++ b/mycroft/skills/intent_service.py
@@ -181,7 +181,9 @@ class IntentService(object):
 
     def update_context(self, intent):
         for tag in intent['__tags__']:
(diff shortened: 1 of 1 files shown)


Mn0491:
@forslund should you add some comments mentioning that this is a temp workaround until we get the adapt situated?

forslund:
Good idea, I'll do that

PR #1028, opened 2017-08-29
Allow requests for STT from wave_file

JarbasAI:
Handler for STT requests for a wave audio file

Triggered by: "recognizer_loop:external_audio"
 Expected data fields:
         'wave_file' : ' path/to/file.wav'

Responds with:
          message.type: "recognizer_loop:external_audio.reply"
          message.data: "stt": transcription

### DIFF
--- a/mycroft/client/speech/listener.py
+++ b/mycroft/client/speech/listener.py
@@ -17,7 +17,6 @@
 
 
 import time
-import wave
 from Queue import Queue
 from threading import Thread
 
@@ -104,8 +103,11 @@ class AudioConsumer(Thread):
                         self.handle_external_audio_request)
 
     def read_wave_file(self, wave_file_path):
-        wr = wave.open(wave_file_path, 'r')
-        return wr.readframes(wr.getnframes()-1)
+        # use the audio file as the audio source
+        r = sr.Recognizer()
+        with sr.AudioFile(wave_file_path) as source:
+            audio = r.record(source)
+        return audio
 
     def handle_external_audio_request(self, event):
         wave_file = event.get("wave_file")
--- a/mycroft/client/speech/listener.py
+++ b/mycroft/client/speech/listener.py
@@ -17,6 +17,7 @@
 
 
 import time
+import wave
 from Queue import Queue
 from threading import Thread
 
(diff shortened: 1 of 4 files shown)


btotharye:
seems like there are some issues on this travis build that need to be check out @JarbasAI

penrods:
Pep8 strikes again!  

Also, can you include a little more documention on how this would be used.  I know this isn't all on you, but I want to start documenting the messagebus protocols.  So when you add a new message we have a definition of what data is expected.  E.g.  recognizer_loop:external_audio looks like it expects a data packet with the path to a WAV file.  For now, just include the doc in the PR since there isn't a definitive repository of these protocols (other than a spreadsheet I started a few weeks ago).

forslund:
Personally I'd prefer if this was implemented as a separate client. So that speech did speech, text did text and file (or what we might want to call it) did files.

The implementation as a stand alone client should be simple enough and frankly the speech client doesn't need any more complexity.

JarbasAI:
@forslund so a standalone .py that would create another STT factory just to handle this? as in a new running process?

im not sure i follow you on this one

forslund:
Exactly like that. A stand alone process recieving wav's sending it of to the configured STT, and then feeding the result forward to the intent service over the message bus.

The sst-factory is quite small and just returns a class sending of the file to a SST processor (local or online)

penrods:
Jarbas, are you looking to support input from two different audio sources at once (WAVs and mic stream both)?  Or would you only need to send WAVs in all the time?

JarbasAI:
fuu, checked out wrong branch working on the pt translation..sorry guys fixing this...

JarbasAI:
i like forslunds aproach for wave file reader with own stt better

the idea is that at any time a skill or anything else connected to the bus could request STT for any file, for example @btotharye mycroft-api project could use this

forslund:
@JarbasAI I've got an example of this pretty much ready. Gonna push to a repo later today

JarbasAI:
I have seen your wave file repo, i like it

My idea was to enable to get the transcription from anywhere, not necessarily emit it to the intent service after, for example a "transcribe this file" skill that would monitor a folder or something

forslund:
I haven't pushed the latest changes that would do something along those
lines. Matbe we can talk about it on mattermost in the next couple of days
and come up with something together. (At a wedding tonight)

Den 9 sep. 2017 9:06 em skrev "Jarbas" <notifications@github.com>:

JarbasAI:
closing, https://github.com/JarbasAI/mycroft-wave-client/blob/master/main.py

PR #1029, opened 2017-08-29
save utterances

JarbasAI:
new flag in listener config to allow for recording of utterances, can be used for collecting data for machine learning (voice prints / home baked STT)

new field in config

"listener" : {"record_utterances": false}


penrods:
If you can fix the pep8 errors I'll pull this in.  (autopep8 is pretty easy to use)

JarbasAI:
sorry on all the pep8 things, it usually happens when i make the whole PR in browser...

PR #1032, opened 2017-08-30
Add event scheduler

forslund:
====  Tech Notes ====
A single thread handling scheduled events. The skills interact with this
using the self.schedule_event() self.schedule_repeating_event
self.update_event() and self.remove_event().

This is an improvement over scheduledSkill since each skill creates
their own Thread and has to handle storing/restoring scheduled events
between starts.

All pending events are stored in a json file at shutdown for future
sessions.

Needs to be documented

==== Protocol Notes ====
new messagebus event handlers:
- mycroft.scheduler.schedule_event
- mycroft.scheduler.remove_event
- mycroft.scheduler.update_event

@penrods can you check so I touched all the parts you imagined.

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -452,7 +452,7 @@ class MycroftSkill(object):
                 'expect_response': expect_response}
         self.emitter.emit(Message("speak", data))
 
-    def speak_dialog(self, key, data={}, expect_response=False):
+    def speak_dialog(self, key, data=None, expect_response=False):
         """
             Speak sentance based of dialog file.
 
@@ -463,7 +463,7 @@ class MycroftSkill(object):
                                     response from the user and start listening
                                     for response.
         """
-
+        data = data or {}
         self.speak(self.dialog_renderer.render(key, data), expect_response)
 
     def init_dialog(self, root_directory):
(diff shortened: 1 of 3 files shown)


mycroft/skills/core.py
```diff
@@ -527,6 +527,73 @@ def shutdown(self):
...
+        event_data['time'] = time.mktime(datetime.timetuple())
+        event_data['event'] = name
+        event_data['repeat'] = repeat
+        event_data['data'] = data
+        self.emitter.emit(Message('mycroft.scheduler.schedule_event',
+                                  data=event_data))
+
+    def schedule_event(self, handler, datetime, data={}, name=None):
+        """
+            schedule a single event
```
penrods: Minor:  Uppercase the first word in these descriptions

mycroft/skills/core.py
```diff
@@ -527,6 +527,73 @@ def shutdown(self):
...
+
+            Args:
+                handler:    method to be called
+                datetime:   time for calling the handler
```
penrods: Lets write/doc it like this:
      when (datetime):  when the handler should be called
      data (dict, optional):  data to send when the handler is called
      name (str, optional):  friendly name for the event

So rename "datetime" to "when".  Make sense?

mycroft/skills/core.py
```diff
@@ -527,6 +527,73 @@ def shutdown(self):
...
+
+    def schedule_repeating_event(self, handler, datetime, frequency,
+                                 data={}, name=None):
+        """
+            schedule a repeating event
+
+            Args:
+                handler:    method to be called
+                datetime:   time for calling the handler
+                frequency:  time in sconds between calls
```
penrods: As above for frequency.  Can it be a float or just int?  Inquiring devs want to know.

mycroft/skills/core.py
forslund: Nemas problemas, I'll copy this to the wiki as an example for future reference

forslund:
Doc-strings updated

penrods:
Argh, Codacy _is_ right about the two default values data={}.  Bad practice.  Let's fix those real quick, just change the defaults to data=None, then add "if data=None: data={}".  Good to teach the children well.

forslund:
Yeah, deep down I know this but apparently I've gotten lazy. Good thing we got this bot now :)

PR #1033, opened 2017-08-30
Add portuguese translation to the dialogs

00tiagopolicarpo00:
Add portuguese translation to the dialogs


forslund:
Thanks for contributing these translations. I know absolutely no Portuguese but it looks good 👍 

Before we can merge it we'd like you to sign a CLA to avoid any possible legal issues: https://mycroft.ai/cla/

JarbasAI:
network and internet dialog could use some changes, there is a fine distinction that gets lost in translation

perhaps use "rede" instead of "internet" for network? Not sure on the best word, but they are not  the same thing :)

JarbasAI:
working on code translation to pt-pt to compliment this
https://github.com/JarbasAI/JarbasAI/tree/patch-14

PR #1039, opened 2017-08-31
Learning option for sharing wake word recordings

MatthewScholefield:
### Description ###

This adds an option in the config to upload wake words to a server. This is **disabled** by default. If the user chooses to, they can opt-in to help make Mycroft smarter. This adds the configuration section:
```Json
{
  "listener": {
    "wake_word_upload": {
      "enable": false,
      "server": "",
      "port": 0,
      "user": "",
      "folder": ""
    }
  }
}
```
With some prefilled options for the mycroft server. In addition,  there is a new `wakeword_rsa` resource for authentication with the server. This is copied to the user directory (`~/.mycroft/wakeword_rsa`) so that the permissions can be changed.

### Installation Notes ###
[After this PR][pr] gets merged, a new enclosure release must be made. Once this code makes its way onto devices, there will be a `TRAIN` option in the menu to enable this. After changing this option the user must restart the processes.

[pr]:https://github.com/MycroftAI/enclosure-mark1/pull/89


mycroft/client/speech/mic.py
```diff
@@ -164,20 +171,28 @@ def __init__(self, wake_word_recognizer):
...
         self.multiplier = listener_config.get('multiplier')
         self.energy_ratio = listener_config.get('energy_ratio')
         # check the config for the flag to save wake words.
-        self.save_wake_words = listener_config.get('record_wake_words', False)
+
         self.save_utterances = listener_config.get('record_utterances', False)
+        self.save_wake_words = listener_config.get('record_wake_words') \
+            or self.upload_config['enable']
+        self.upload_lock = Lock()
+        self.save_wake_words_dir = listener_config.get('record_dir')
```
penrods: Let's just use something like os.path.join(tempfile.gettempdir(), "mycroft_wake_word")

mycroft/configuration/mycroft.conf
```diff
@@ -111,6 +111,15 @@
     "sample_rate": 16000,
     "channels": 1,
     "record_wake_words": false,
+    "record_utterances": false,
+    "record_dir": "/tmp/mycroft_wake_word",
```
penrods: Don't need this after the above change to use tempfile

mycroft/client/speech/mic.py
```diff
@@ -308,6 +324,36 @@ def stop(self):
         """
         self._stop_signaled = True
 
+    def _upload_file(self, filename):
```
forslund: This is very loosely connected to the ResponsiveRecognizer and could very well be moved outside the class. Basically config and lock as far as I can see.

forslund: Not really a critical issue but something to think about for future refactoring

PR #1041, opened 2017-08-31
Fix errors during shutdown

forslund:
====  Tech Notes ====
Skills weren't cleanly shutdown. This ensures that skills are shutdown more cleanly and that shutdown procedure will not be interrupted by broken shutdown methods.

Shutdown of reload thread fixed

Unused globals removed


forslund:
There, pep8 fixed...

PR #1042, opened 2017-09-01
Tame the mycroft.sh script

penrods:
The mycroft.sh script uses the 'screen' command to run the various
processes that make up Mycroft Core.  When they are stopped with
"./mycroft.sh stop", it was using the screen 'quit' command.  This
ends the process abruptly, not allowing services to shutdown
cleanly.

This changes it to send a shutdown request first, equivalent of a
Ctrl-C.  If the service doesn't shutdown after 2 seconds it is
then killed.

This specifically prevents issues where a skill expected and did
not receive a call to the shutdown() method.

### DIFF
--- a/mycroft.sh
+++ b/mycroft.sh
@@ -111,7 +111,26 @@ function start-mycroft-debug {
 function stop-screen {
   for i in $(screen -ls "$1"); do
     if echo $i | grep -q $1; then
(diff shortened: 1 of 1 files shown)


forslund:
Rebased this on top of skills-shutdown PR and settings.json is now being written with no issue.

Looks good to me!

PR #1043, opened 2017-09-01
Write settings dict to disk more frequently

forslund:
====  Tech Notes ====
After an intent has been handled the settings will be stored if any changes are detected.

A force option was added to the .store method of the settings class which always writes the dict to disk even if no changes has been made. This is used during shutdown.

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -517,7 +517,7 @@ class MycroftSkill(object):
         shutdown all processes and operations in execution.
         """
         # Store settings
-        self.settings.store(force=True)  # Force saving of settings
+        self.settings.store()
 
         # removing events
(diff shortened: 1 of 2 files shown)


mycroft/skills/core.py
```diff
@@ -516,7 +517,7 @@ def shutdown(self):
         shutdown all processes and operations in execution.
         """
         # Store settings
-        self.settings.store()
+        self.settings.store(force=True)  # Force saving of settings
```
penrods: Let's drop the force.  I'm nervous about forcing a disk write by every single skill during shutdown scenarios, which is when the rug is most likely to be pulled out from under you -- impatient user unplugs device, script timeout passes and a kill gets executed, etc.

PR #1048, opened 2017-09-01
Fix msm from pulling back commits on another branch

MatthewScholefield:
Often, to do dev work on a skill, I go through the process:
```
git checkout -b feature/my-feature
git add __init__.py
git commit -m "Add my new feature!"
git push -u origin feature/my-feature
```
Unfortunately, as it is right now, msm will get rid of this new `Add my new feature` commit because it does a hard reset to `origin/master`. As goes without saying, this shouldn't happen. To solve this, we can check if we are currently on the master branch and not update unless that is so.

### DIFF
--- a/msm/msm
+++ b/msm/msm
@@ -223,8 +223,13 @@ function update() {
         echo "*.pyc" >> .git/info/exclude
       fi
 
-      # Check if the repo is dirty or has commits to push
-      if [[ (-z $(git status --porcelain --untracked-files=no)) && !($LOCAL != $REMOTE && $REMOTE = $BASE) ]]; then
+      BRANCH="$(git symbolic-ref HEAD 2>/dev/null)"
+      BRANCH="${BRANCH##refs/heads/}"
+
+      if [[ (-z $(git status --porcelain --untracked-files=no)) &&  # No Modified files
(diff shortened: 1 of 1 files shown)


augustnmonteiro:
In this case if a developer do a forced push to his skill and it generates any conflict with the location version the --ff-only will never update the skill

MatthewScholefield:
@augustnmonteiro I think we went through this before. It's never acceptable for a developer to force push to a public repo. If you ever do a search for any git command on StackOverflow, there are lots of people who warn never to force push to a public git repository. For example, look at the yellow box on [this GitHub.com help page](https://help.github.com/articles/changing-a-commit-message/#amending-older-or-multiple-commit-messages). So, do you agree that there's a greater chance of annoying developers whose commits get removed than of a skill developer who unacceptably rebases a public skill repository?

augustnmonteiro:
@MatthewScholefield I agree that people should not do that, but are we going to do the code based in what people should do or prevent failures?

el-tocino:
Check the commit it's on, and if it doesn't match the version listed on the skills repo, and it's being run as an interactive run, query the user if you want to checkout the designated skills version?

MatthewScholefield:
@augustnmonteiro OK, I've updated it so it keeps the same functionality but checks if on master. I'll still never agree on this topic but oh well xD.

Issue #1056, opened 2017-09-05
Mathematical parser


JarbasAI:
in the format / parse files (not sure where this one would go) it would make sense to have helper functions to process natural mathematical speech, these would be also dependent on language

proof of concept:
https://pastebin.com/6Z1GX0NG

output:
https://pastebin.com/TiyvLiKC

```
normalized string: 1 times 7 plus 2 multiply by 2
raw string: one times seven plus two multiply by two
processing: [[0, '+', 'next'], [0, '+', '7.0 + next'], [0, '+', 4.0]]
result: 7.0+0+4.0
processing: [[0, '+', 'next'], [0, '+', 11.0]]
result: 11.0
vars: {}
```

forslund:
Closing Issue since we're archiving the repo

PR #1057, opened 2017-09-05
Mathematics from natural language support #1056 

JarbasAI:
solve math from natural language utils #1056 

adding more functionality bit by bit, suggestions welcome

unittests included

### new methods in util/format.py ###

```
extract_expression()
solve_expression()
```

### usage ###

```
extract_expression("one dog plus one dog plus two frogs")  =   [['1dog', '+', '1dog'], ['prev', '+', '2frogs']]

solve_expression("one times seven plus two multiply by two")  =      '11'
```

### DIFF
--- a/mycroft/util/format.py
+++ b/mycroft/util/format.py
@@ -294,6 +294,7 @@ class StringOperation():
         res = res.replace("++", "+")
         res = res.replace("--", "-")
         res = res.replace("/1", "")
+
         while len(res) > 2 and res[0] in ["+", "0"] and res[1] != ".":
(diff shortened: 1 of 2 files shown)


Mn0491:
@JarbasAI would this make more sense as a skill?

JarbasAI:
I am working on a fallback skill that would use this, but this is more of a tool to be used anywhere, much like the extract date_time, it is also dependent on language and this seems to be the approach for language parsing support

JarbasAI:
not ready for PR, many changes to come...

PR #1058, opened 2017-09-05
Notification on dependency changes + Padatious upgrade

MatthewScholefield:
This update hashes `requirements.txt` and `dev_setup.sh` in the `.installed` file to notify the user when there is a change to the dependencies. This also upgrades Padatious to both test the feature, and so that users won't have to run `dev_setup.sh` a second time for the upgrade.

Regardless of the current state, the first time it is required to run `dev_setup.sh` to generate the hash file.

To test:
```Bash
git fetch
git checkout feature/notify-update
git reset --hard HEAD~1
./dev_setup.sh -sm  # Simulate having an already setup git clone from another version
git pull  # Pull down the commit to increment Padatious version number
./mycroft.sh start  # Should get notification to update dependencies
./dev_setup.sh -sm
./mycroft.sh start  # Should no longer get notification
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ logs/*
 mycroft/audio-accuracy-test/data/*
 scripts/*.screen
 doc/_build/
+.installed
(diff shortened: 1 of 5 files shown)


Mn0491:
I like this idea! LGTM!

PR #1060, opened 2017-09-07
Ignore enclosure.mouth.display message by default

forslund:
====  Tech Notes ====
Add 'enclosure.mouth.display' to ignore_logs default configuration since
it will get called alot and spam the skills log.


Mn0491:
I was worried people would view this as spammy! LGTM

PR #1061, opened 2017-09-07
Centralize APIS section

JarbasAI:
This adds two variables to MycroftSkill class

self.APIS_config = self.config_core.get("APIS", {}) # exposes all API keys

self.API = self.APIS_config.get(self.name) # api field with SkillName

API keys can often used in more than one skill, to use same key in more than one skill requires developer to know in which part of config file it is stored and manually accessing it, or to duplicate the key entry

in long term this makes it messy for user-setup and dev-usage

issue #540

my use-case:

´´´
"APIS": {
"WolframAlphaAPI": "key",
"WeatherAPI": "key",
"GraphAPI": "key",
"MashapeAPI": "key",
"WikimapiaAPI": "key",
"CloudsightAPI": "key",
"RedditAPI": "key",
"CloudsightSecret": "secret",
"RedditSecret": "secret"
},
´´´

### DIFF
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -205,6 +205,12 @@
     "intent_cache": "~/.mycroft/intent_cache",
     "train_delay": 4
   },
+
+  // Centralized API section
+  "APIS": {
+      "WolframAlphaSkill": "",
+      "WeatherSkill": ""
+  },
   // =================================================================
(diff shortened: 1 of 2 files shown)


aatchison:
I like this idea @penrods , thoughts?

JarbasAI:
now that skills use settings i think it makes even more sense, no api keys all over the place, easy to keep private

PR #1062, opened 2017-09-07
Hotword factory

JarbasAI:
Implements hot word factory class to allow any kind of hot word engines

included snowboy support if user installs it, i know snowboy isn't free but neither are most of the STT options


mycroft/client/speech/hotword_factory.py
```diff
@@ -0,0 +1,142 @@
...
+    def create_hotword(hotword, lang="en-us"):
+        LOG.info("creating " + hotword)
+        config = ConfigurationManager.get().get("hot_words", {})
+        module = config.get(hotword).get("module")
+        config = config.get(hotword, {"module": module})
+        clazz = HotWordFactory.CLASSES.get(module)
+        return clazz(hotword, config, lang=lang)
+
+    @staticmethod
+    def create_wake_word(lang="en-us"):
```
MatthewScholefield: Can `create_wake_word`, `create_hotword`, and `create_standup_word` be put into one function? For instance, `create_wake_word` could just be:
```
conf = ConfigurationManager.get()
return HotWordFactory.create_hotword(conf['wake_word'], conf['threshold'], conf['module'])
```

mycroft/configuration/mycroft.conf
```diff
@@ -111,14 +111,28 @@
     "sample_rate": 16000,
     "channels": 1,
     "record_wake_words": false,
-    "wake_word": "hey mycroft",
-    "phonemes": "HH EY . M AY K R AO F T",
+    "record_utterances": false,
```
MatthewScholefield: I think this is an unrelated addition

mycroft/configuration/mycroft.conf
```diff
@@ -111,14 +111,28 @@
     "sample_rate": 16000,
     "channels": 1,
     "record_wake_words": false,
-    "wake_word": "hey mycroft",
-    "phonemes": "HH EY . M AY K R AO F T",
```
MatthewScholefield: Not using this key will ignore server changes. Can you add support for using this value. Otherwise, anyone who's customized it will lose the setting.

mycroft/configuration/mycroft.conf
```diff
@@ -111,14 +111,28 @@
...
     "multiplier": 1.0,
-    "energy_ratio": 1.5
+    "energy_ratio": 1.5,
+     // wake word module
+    "wake_word_config": {
+        "module": "pocketsphinx",
+        "wake_word": "hey mycroft",
+        "phonemes": "HH EY . M AY K R AO F T",
+        "threshold": 1e-90,
+        "sample_rate": 16000
```
MatthewScholefield: Sample rate should probably be part of the listener config since there's no way to change that per hot word.

mycroft/client/speech/hotword_factory.py
```diff
@@ -0,0 +1,142 @@
...
+class HotWordFactory(object):
+    CLASSES = {
+        "pocketsphinx": PocketsphinxHotWord,
+        "snowboy": SnowboyHotWord
+    }
+
+    @staticmethod
+    def create_hotword(hotword, lang="en-us"):
+        LOG.info("creating " + hotword)
+        config = ConfigurationManager.get().get("hot_words", {})
```
MatthewScholefield: Not that big of a deal, but you can always make the assumption config values (that you add to `mycroft.conf`) exist. No need to do `, {})`. Can even do `ConfigurationManager.get()['hot_words']`, but that's up to you.

Same thing below.

JarbasAI: hot words section may not exist, i added it expecting multiple hot word support, but wake/wake up word are mandatory and have their own field in listener config

mycroft/configuration/mycroft.conf
JarbasAI: yup, this one escaped me, should i still leave it? it is a parameter i added in another already merged PR but did not add to config file

mycroft/configuration/mycroft.conf
JarbasAI: changing

mycroft/client/speech/hotword_factory.py
JarbasAI: changing

mycroft/configuration/mycroft.conf
JarbasAI: added support and TODO flag so it can be removed later

mycroft/client/speech/hotword_factory.py
```diff
@@ -0,0 +1,143 @@
...
+from os.path import dirname, exists, join, abspath
+import os
+import time
+import tempfile
+
+__author__ = 'seanfitz, jdorleans, jarbas'
+
+LOG = getLogger("HotwordFactory")
+
+BASEDIR = dirname(abspath(__file__)).join("recognizer")
```
MatthewScholefield: If you are going to append "recognizer" here, call it something like `RECOGNIZER_DIR` for clarity.

mycroft/client/speech/hotword_factory.py
```diff
@@ -0,0 +1,143 @@
...
+        (fd, file_name) = tempfile.mkstemp()
+        words = key_phrase.split()
+        phoneme_groups = phonemes.split('.')
+        with os.fdopen(fd, 'w') as f:
+            for word, phoneme in zip(words, phoneme_groups):
+                f.write(word + ' ' + phoneme + '\n')
+        return file_name
+
+    def create_config(self, dict_name, config):
+        model_file = join(BASEDIR, 'recognizer', 'model', self.lang, 'hmm')
```
MatthewScholefield: This is why the test is failing. Right now this gets turned into `join('.. speech/recognizer', 'recognizer', ...`.

mycroft/client/speech/hotword_factory.py
```diff
@@ -0,0 +1,143 @@
...
+        if not exists(model_file):
+            LOG.error('PocketSphinx model not found for ' + str(self.lang))
+            model_file = join(BASEDIR, 'model', 'en-us', 'hmm')
```
MatthewScholefield: Make sure this line matches the one above. (Right now it doesn't include the extra `, 'recognizer'`.

mycroft/client/speech/listener.py
```diff
@@ -27,15 +27,13 @@
...
 import mycroft.dialog
 from mycroft.client.speech.mic import MutableMicrophone, ResponsiveRecognizer
-from mycroft.client.speech.recognizer.pocketsphinx_recognizer \
-    import PocketsphinxRecognizer
 from mycroft.configuration import ConfigurationManager
 from mycroft.messagebus.message import Message
 from mycroft.metrics import MetricsAggregator
 from mycroft.session import SessionManager
 from mycroft.stt import STTFactory
-from mycroft.util import connected
```
JarbasAI: codacy is complaining on this unused import, even tough i have removed it!

mycroft/client/speech/listener.py
```diff
@@ -200,35 +199,39 @@ def _load_config(self):
...
-    def create_wakeup_recognizer(self, rate, lang):
-        wake_word = self.config.get('standup_word', "wake up").lower()
-        phonemes = self.config.get('standup_phonemes', "W EY K . AH P")
-        threshold = self.config.get('standup_threshold', 1e-18)
-        return PocketsphinxRecognizer(wake_word, phonemes,
-                                      threshold, rate, lang)
+        LOG.info("creating wake word engine")
+        word = "hey mycroft"  # self.config.get("wake_word", "hey mycroft")
+        # TODO remove this, only for server settings compatibility
+        phonemes = self.config.get("phonemes")
```
MatthewScholefield: One more thing, could we add threshold backwards compatibility here just like `phonemes`? Otherwise, for some simpler wake words, without the changed threshold it will start triggering like crazy.

JarbasAI: done

JarbasAI:
added phoneme length calculation in base HotWord class, it would be a problem eventually for non pocket-sphinx modules

mycroft/client/speech/listener.py
```diff
@@ -211,26 +211,31 @@ def _load_config(self):
...
-    def create_wakeup_recognizer(self, rate, lang):
-        wake_word = self.config.get('standup_word', "wake up").lower()
-        phonemes = self.config.get('standup_phonemes', "W EY K . AH P")
-        threshold = self.config.get('standup_threshold', 1e-18)
-        return PocketsphinxRecognizer(wake_word, phonemes,
-                                      threshold, rate, lang)
+        LOG.info("creating wake word engine")
+        word = self.config.get("wake_word", "hey mycroft")
+        # TODO remove this, only for server settings compatibility
+        phonemes = self.config.get("phonemes")
```
MatthewScholefield: There's no config value for `"phonemes"` so it sets `phonemes` to None.

JarbasAI: that is intended, phonemes/thresh should be None except for remote config update compatibility, travis is failing on line 61 of hot_word factory, but i do not understand why since

self.phonemes = self.config.get("phonemes", "HH EY . M AY K R AO F T")
self.num_phonemes = len(self.phonemes.split())

should not allow phonemes to be None

penrods:
Something weird is happening in Travis CI, as this passes local tests.  Let's pull this in and see if it breaks things.

penrods:
I think I see the issue now...

See below.  There is no default for "phonemes" in the mycroft.conf anymore, so it gets set to None.  Then that config gets passed on to create_hotword(), which reads the None back out of it and attempts to call split().

I'll restore the default in the mycroft.conf.


 def create_wake_word_recognizer(self):
        # Create a local recognizer to hear the wakeup word, e.g. 'Hey Mycroft'
        LOG.info("creating wake word engine")
        word = self.config.get("wake_word", "hey mycroft")
        # TODO remove this, only for server settings compatibility
        phonemes = self.config.get("phonemes")            # <--- this returns None
        thresh = self.config.get("threshold")
        config = self.config_core.get("hotwords", {word: {}})
        config[word]["phonemes"] = phonemes             # <---- this sets "phonemes" to none in the config
        config[word]["threshold"] = thresh
        if phonemes is None or thresh is None:
            config = None
        return HotWordFactory.create_hotword(word, config, self.lang)

JarbasAI:
so should it be

def create_wake_word_recognizer(self):
# Create a local recognizer to hear the wakeup word, e.g. 'Hey Mycroft'
LOG.info("creating wake word engine")
word = self.config.get("wake_word", "hey mycroft")
# TODO remove this, only for server settings compatibility
if self.config.get("phonemes") and self.config.get("threshold"):
     config = self.config_core.get("hotwords", {word: {}})
     config[word]["phonemes"] = phonemes
     config[word]["threshold"] = thresh
else:
     config = None
return HotWordFactory.create_hotword(word, config, self.lang)

do i need to do anything? (since it is marked as merged)

penrods:
Nah, I think this should be good now.  I'm testing 'dev' now.

PR #1067, opened 2017-09-11
Run requirements.sh when installing skills if exists

MatthewScholefield:
This brings in support for running a `requirements.sh` file to setup things like native dependencies for skills.


Mn0491:
@MatthewScholefield is there any permissions issues

PR #1069, opened 2017-09-12
speak method improvements, mute and metadata

JarbasAI:
changes to speak method message.data:
- add "metadata" dict to allow skills to send extra info #281 

changes to speak method message.context:
- add "mute" to allow skills to not trigger TTS execution (browser client/server architecture)
- add "more_speech" to allow listeners to know more speak is coming

changes audio service to take into account the mute flag
changed speak_dialog data={} argument to data = None and  if data is None data = {}

I believe mute/more_speech should be in context and not in data, same for expect_response

### DIFF
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -78,6 +78,7 @@ def mute_and_speak(utterance, mute):
 
         Args:
             utterance: The sentence to be spoken
+            mute:      Do not execute tts
     """
     global tts_hash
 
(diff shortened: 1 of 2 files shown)


forslund:
I don't think these should be in the context of the message. Context should be used to direct messages and maybe return client specific state in a potential multi user setting. So I'd like the parameters you added moved to data.

I can't see `more_speech` being used anywhere, can you give an example on how it's supposed to be used? Is it for non-audio clients and similar that should send the entire message in one go?

JarbasAI:
more_speech is meant for anything listening to speak messages to know there is more data coming to be processed, suppose you have some skill processing all utterances for some reason like sentiment analisys / diagnostics / caching, this would signal that the next speak message should be grouped/processed together with the previous one like if it was a single answer

i can change the other fields to data, i still think it feels right at context instead of data (and am using it that way) but i guess it comes to personal preference and can't really give you a sound argument

PR #1073, opened 2017-09-13
mycroft-use script for Mark 1 and picroft

Mn0491:
mycroft-use script is used to switch from stable and unstbale builds for the mycroft core. It also points the unstable build to use the home-test.mycroft.ai cluster instead of production cluster.

mycroft-packaging repo needs to be merged to master for the packaging to take place.


forslund:
Rebase the branch and the failing tests should be fixed.

Mn0491:
@forslund the build failed again. is it because of these unit tests?

PR #1076, opened 2017-09-14
Make test_api slightly more robust

forslund:
====  Tech Notes  ====
- Restore ConfigurationManager after mocking it
- enforce value of cached pairing status


forslund:
==== Tech Notes ====

- Restore ConfigurationManager after mocking it
- Enforce value of cached pairing status

PR #1079, opened 2017-09-14
Add messages for skill handler's starting/stopping

forslund:
====  Tech Notes ====
Add message bus messages when a skill handler is starting to execute and
a separate message for when skill handler completes. Usable if grouping
speech in certain clients for example. Can also be of use for audio
ducking and similar processes.

==== Protocol Notes ====
Adds the following messages:
mycroft.skill.handler.start
mycroft.skill.handler.complete

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -156,6 +156,24 @@ def create_skill_descriptor(skill_folder):
     return {"name": basename(skill_folder), "info": info}
 
 
+def get_handler_name(handler):
+    """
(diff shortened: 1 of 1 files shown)


JarbasAI:
damn, you beat me to it xD i had this one ready to PR

the one thing i did differently was that i also added a .fail message mycroft.skill.handler.failure with the error in data

forslund:
Great minds think alike :)

I'll add something along those lines.

PR #1084, opened 2017-09-15
Feature/jarbas priority skills

forslund:
Rebase of @JarbasAI's PR #858 

- updated skill id method
- added pairing to default priority list
- changed waiting for all skills to wait for priority skills before starting the network detector to increase startup time when pairing is needed.


forslund:
travis got interrupted at some point it seems

JarbasAI:
much cleaner than my approach :)

PR #1086, opened 2017-09-15
Bugfix/hotword threshold

forslund:
Hotword threshold was forcibly set to `None` when missing in config. (as was the case for the default config) crashing the mycroft-speech-client

### DIFF
--- a/mycroft/client/speech/listener.py
+++ b/mycroft/client/speech/listener.py
@@ -226,8 +226,10 @@ class RecognizerLoop(EventEmitter):
         phonemes = self.config.get("phonemes")
         thresh = self.config.get("threshold")
         config = self.config_core.get("hotwords", {word: {}})
-        config[word]["phonemes"] = phonemes
-        config[word]["threshold"] = thresh
+        if phonemes:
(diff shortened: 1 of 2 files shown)


forslund:
Needs to be pulled before release (or equivalent fix)

JarbasAI:
may i suggest instead 

```
def create_wake_word_recognizer(self):
    LOG.info("creating wake word engine")
    word = self.config.get("wake_word", "hey mycroft")
    if self.config.get("phonemes") and self.config.get("threshold"):
         config = self.config_core.get("hotwords", {word: {}})
         config[word]["phonemes"] =self.config.get("phonemes")
         config[word]["threshold"] = self.config.get("threshold")
    else:
         config = None
    return HotWordFactory.create_hotword(word, config, self.lang)
```

the idea is to eventually remove those

i would actually remove that whole part like the TODO note says, it is just because mathew asked for compatibility because of server updates, so consider the value if it is there, but expect it to not be there

phonemes and threshold should be inside "hotwords" from now on, not in "listener", "listener" should only tell which hotword to use as wakeword

PR #1093, opened 2017-09-17
Add 'msm remove', improve error handling and parsing

penrods:
Semi-major rework of the Mycroft Skill Manager (MSM), including:
* Add parse skill name strings support for quoted strings, e.g.:
    msm install "daily meditation"
  Previously it would search for "daily" and also search for
  "meditation", but both independently.
* All commands return distinct error codes.  Commands with
  that perform multiple things, such as 'msm default' will
  continue on a failure of one skill and return the highest
  error code ultimately.
* Add 'msm remove', essentially the inverse of 'msm install'
* Changed the example install URL to a penrods repo since the
  ethanward repo no longer exists.
* Reworked skill list caching to behave properly with $( )
  subprocess operations.
* Output for "install" and "remove" is now between hyphen
  lines, making them easier to parse.
* Revamped many informational and error messages
* Published under Apache 2.0 license

A corresponding update to the skill-installer will take
advantage of these changes.

### DIFF
--- a/msm/msm
+++ b/msm/msm
@@ -1,27 +1,49 @@
 #!/bin/bash
 
-# Copyright 2016 Mycroft AI, Inc.
+# Copyright 2017 Mycroft AI Inc.
 #
-# This file is part of Mycroft Core.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
 #
-# Mycroft Core is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
+#    http://www.apache.org/licenses/LICENSE-2.0
 #
-# Mycroft Core is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with Mycroft Core.  If not, see <http://www.gnu.org/licenses/>.
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
+script=${0}
+script=${script##*/}
 
-# @author Augusto Monteiro
-#
-# This script assists in the installation and management of
-# skills loaded from Github.
+function help() {
+  echo "${script}: Mycroft Skill Manager"
+  echo "usage: ${script} [option] [repo | name]"
+  echo
+  echo "Options:"
+  echo "  default                  installs the default skills, updates all others"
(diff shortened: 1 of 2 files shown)


msm/msm
```diff
@@ -176,56 +260,74 @@ function install() {
       fi
     fi
     if [[ -f "requirements.txt" ]]; then
-      echo "Installing requirements..."
-      if [[ "${picroft}" == 'false' ]]; then
+      echo "  Installing requirements..."
+      if [[ "${picroft}" == "false" ]]; then
```
forslund: Something is weird here. On my desktop system the script entered the else branch of this if-statement and installed requirements using `sudo`

Debug output:
```
+ echo '  Installing requirements...'
  Installing requirements...
+ [[ '' == \f\a\l\s\e ]]
+ sudo pip install -r requirements.txt
```

msm/msm
```diff
@@ -35,121 +57,179 @@ DEFAULT_SKILLS="skill-alarm skill-audio-record skill-configuration "\
...
-  # loop through arguments, treat each as a independent request
-  while [[ $# -gt 0 ]] ; do
-    cd "${mycroft_skill_folder}"
+  git_line=$(echo "$skill" | sed 's/\:.*//')
+  if [[ "${skill}" ==  *$'\n'* ]]; then
+    # The str matches multiple skill repos, don't install
+    # due to ambiguity.
+    #
+    echo "Multiple matches for '${str}', be more specific."
+    echo "----------------------------------------------------------------------"
```
forslund: More than 80 characters

msm/msm
forslund: I guess it's that the `picroft` variable has been renamed `picroft_mk1`

msm/msm
```diff
@@ -176,56 +260,74 @@ function install() {
...
           if workon mycroft ; then
             if ! pip install -r requirements.txt ; then
-              echo "ERROR: Unable to install requirements for skill ${iskill}!"
+              echo "ERROR: Unable to install requirements for skill '${name}'"
               deactivate mycroft
               return 121
             fi
           else
-            echo "ERROR: Unable to activate mycroft virtualenv!"
+            echo "ERROR: Unable to activate Mycroft virtualenv, requirements not installed."
```
forslund: Should this be small letters since the Mycroft virtualenv name is actually `mycroft`

msm/msm
```diff
@@ -242,41 +344,145 @@ function update() {
...
+      exit_code=$?
+      ;;
+  "default")
+      echo "=== Checking for default skills"
+      INSTALLING_DEFAULTS="true"
+      get_skill_list
+      exit_code=$?
+      if [[ ${exit_code} -gt 0 ]] ; then
+         echo "${script}: error ${exit_code}"
+         exit ${exit_code}
```
forslund: These error printout/exit could be moved to a separate function since they're duplicated for each msm command

msm/msm
penrods: Fixed

msm/msm
penrods: Fixed

msm/msm
penrods: Fixed

msm/msm
penrods: I'm not clear on what you mean here

forslund: Nothing really important but one could create a function wrapping the above lines since they exist a couple of times in the script.

```bash
function exit_error() {
       echo "$1: error $2"
       exit $2
}

PR #1095, opened 2017-09-18
Add new LOG class

MatthewScholefield:
This brings in a new way to log. Rather than putting `getLogger(__name__)` at the top of every file, you simply log with `LOG.debug('message')`. Here's an example:

```Python

def run():
    LOG.debug("Here's a debug message")
    LOG('CUSTOM_NAME').info('Another message')
    custom_logger = LOG.create_logger('MY_NAME')
    custom_logger.warn('warning?!')

run()
```

Output:
```Bash
14:03:44.061 - __main__:run:6 - DEBUG - Here's a debug message
14:03:44.061 - CUSTOM_NAME - INFO - Another message
14:03:44.061 - MY_NAME - WARNING - warning?!
```

As you can see, the default format is `module:function:line_number`. Since this isn't necessarily a high priority, I'm not assigning anyone yet.

As a side effect of touching all the import lines, Codacy complained of all the unused imports that already existed in the code. Because of this, this PR also removes all unused imports from the code base.


mycroft/util/log.py
```diff
@@ -14,33 +14,116 @@
...
-FORMAT = '%(asctime)s - %(name)s - %(levelname)s - %(message)s'
-logging.basicConfig(format=FORMAT, level=logging.getLevelName(log_level))
-logger = logging.getLogger("MYCROFT")
+    _custom_name = None
 
+    @classmethod
+    def init(cls):
+        sys_config = '/etc/mycroft/mycroft.conf'
+        level_name = load_commented_json(sys_config)['log_level'] \
+            if isfile(sys_config) else 'DEBUG'
```
forslund: This is a personal opinion, I find the ternary operation a bit confusing when split on different lines like this.

mycroft/util/log.py
```diff
@@ -14,33 +14,116 @@
...
+        cls.ch = logging.StreamHandler(sys.stdout)
+        cls.ch.setFormatter(formatter)
+        cls.create_logger('')  # Enables logging in external modules
 
-    :param name: Module name for the logger
+    @classmethod
+    def create_logger(cls, name):
+        l = logging.getLogger(name)
+        l.propagate = False
+        if not hasattr(cls, 'level'):
```
forslund: Also a personal preference, using hasattr makes reading the code a bit harder than

instanciating level to `None` and using

```python
     if not cls.level:
         ...
```

My personal opinion is that hasattr should mainly be used if there is uncertainty about what object is being operated on.

But personal opinion so feel free to ignore :)

forslund:
I'm not sure about how atomic python calls are, could there be some threading issues here since the class variables are basically global?

mycroft/util/log.py
```diff
@@ -14,33 +14,116 @@
...
 
-FORMAT = '%(asctime)s - %(name)s - %(levelname)s - %(message)s'
-logging.basicConfig(format=FORMAT, level=logging.getLevelName(log_level))
-logger = logging.getLogger("MYCROFT")
+    _custom_name = None
 
+    @classmethod
+    def init(cls):
+        sys_config = '/etc/mycroft/mycroft.conf'
+        level_name = load_commented_json(sys_config)['log_level'] \
```
forslund: I believe this will fail if a sys config exitst but lacks a `log_level` entry  should probably have `.get('log_level', 'info')` or similar.

forslund:
Regarding docstrings, you should be able to forward the docstrings from logging.Logger.* to your implementations either with a decorator or just something like  `method.__doc__ = logger.Logging.method.__doc__` after the method has been declared. Maybe prepend a sentence saying this call wraps a call to Logger.

mycroft/util/log.py
MatthewScholefield: Here's my take on the subject:

Do we want to support there being a malformed system config? I'm pretty sure if someone deleted that there's a ton of other places in the code it would fail. Since the config gets updated at the same time as the code, I would think we should be able to assume it to always be there. Also, in the case that in the future a developer moves the log_level config value but forgets to update it here, rather than subtly introducing a big it will error out at the exact spot in the code that needs to be fixed.

mycroft/util/log.py
MatthewScholefield: Good idea! Yeah, hasattr is pretty ugly

mycroft/util/log.py
forslund: I'm not sure if a missing entry is considered malformed.

The `/etc/mycroft/mycroft.conf` is not used in any setups based on the github repo and the current template for the file does not include the log_level parameter so it would break logging on most installations.

MatthewScholefield: Yeah, you're right; I didn't realize it's only loading the system config. Will change.

MatthewScholefield:
I've added a unit test to verify there's no threading issues. I've also added the docstrings, to methods, but done dynamically to reduce code duplication.

forslund:
Cool beans, as no one ever said. Looks good, I'll test run it properly tomorrow and will merge if nothing unexpected pops up!

Good thing testing for threading issues. I haven't got a good grip on python threading, with C EVERYTHING is a risk for threading issues :)

mycroft/messagebus/service/main.py
```diff
@@ -14,16 +14,13 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with Mycroft Core.  If not, see <http://www.gnu.org/licenses/>.
-
 import tornado.ioloop as ioloop
 import tornado.web as web
-import tornado.autoreload as autoreload
```
forslund: This is missing and hinders the messagebus service from starting

MatthewScholefield: Good catch! It's strange that despite the alias, it still imports the original name. Either way, I've changed it to use the alias now (and changed `import mod.mod as mod` to `from mod import mod`).

forslund:
This may be a good idea, my main concern would be that it gives a false sense of that we actually have good tests. Current ~40% status only lie a little bit. That said just loading the code is also a valid test.

Shouldn't a linter catch these type of things?

PR #1096, opened 2017-09-18
Fix install from within Mycroft on Mark1/Picroft

penrods:
MSM was invoking 'sudo chown' every time it installs new skills.
However the skill folder is already under the correct user when it
has been installed this way, so the chown is unnecessary.  Now we
check for the user and group before performing the action.

Also removed Wolfram as a default skill.  It is still available for
install, just is not a default.


forslund:
Looks good, tested on Debian

PR #1099, opened 2017-09-20
alter msm virtualenv handling to allow for docker/other installs.

el-tocino:
Instead of failing for no virtualenv found, continue and leave the vwrap flag as false. Stop initiating virtualenv every install call, start it before we handle arguments instead, and deactivate after.  

Also changed some variables from upper to lower case and adjusted a couple of names as @aatchison wished for a while ago.  

Of note, line 322 was not liked at all by shellcheck. If this is to parse the output of the enclosed command, should be changed to `! $()`? See also https://stackoverflow.com/questions/17191622/why-would-i-not-leave-extglob-enabled-in-bash

### DIFF
--- a/msm/msm
+++ b/msm/msm
@@ -48,7 +48,7 @@ function help() {
 
 # These skills are automatically installed on all mycroft-core
 # installations.
-DEFAULT_SKILLS="skill-alarm skill-audio-record skill-configuration "\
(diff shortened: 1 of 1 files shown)


el-tocino:
Investigating better virtualenv handling.

el-tocino:
Removing logs for virtualwrapper things, and updated the check for #1100 as well.  Hopefully this works better?

PR #1103, opened 2017-09-21
add amazon Polly TTS support

JarbasAI:
add support for Polly TTS engine

logs error when missing python requirements (boto3)


JarbasAI:
included in #1112

PR #1104, opened 2017-09-22
add SSML support for TTS

JarbasAI:
some TTS engines support SSML to modulate speech

- flag in config for TTS engines, support_SSML, default=False
- allow TTS engines to define supported/ extra ssml tags in config
- in speak check for SSML, if current TTS does not support SSML remove all ssml tags (regex)
- if current TTS supports SSML, check if it supports all tags, remove invalid ones

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -470,6 +470,44 @@ class MycroftSkill(object):
         re.compile(regex_str)  # validate regex
         self.emitter.emit(Message('register_vocab', {'regex': regex_str}))
 
+    def check_for_ssml(self, text):
+        """ checks if current TTS engine supports SSML , if it doesn't
+        removes all SSML tags, if it does removes unsupported SSML tags,
+        returns processed text """ 
+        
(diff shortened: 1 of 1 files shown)


forslund:
Cool!

I would rather see this in the tts module instead of in the skill module. Right in the execute method perhaps first in the execute method of the tts base class.

In mimic at least the non-supported tags will just be ignored so no real need to filter out unsupported tags. Just a filter to remove all tags using re.sub would be needed for modules without SSML support.

JarbasAI:
will move into tts class

i think unsupported tags could be useful because we never know which TTS will be used, some may just read the unsupported tags

JarbasAI:
as a side effect it gives some protection against bad formatting

this will also remove html tags if you speak(html_str), is this undesired?

MatthewScholefield:
Okay, the content of the PR has been brought into #1545. I was planning on just pushing to this branch, but it looks like that's disabled because the fork this is based off of is archived.

Issue #1105, opened 2017-09-22
Customized wake word broken in 0.8.22


smpickett:
Customized wake words no longer work; the 'config' doesn't get configured with the wake word. See a debugging log below:
`2017-09-22 03:11:07,132 - mycroft.client.speech.listener - INFO - creating wake word engine`
`2017-09-22 03:11:07,133 - mycroft.client.speech.listener - INFO - ### phonemes HH EY . V IH K T AO R IY AH`
`2017-09-22 03:11:07,134 - mycroft.client.speech.listener - INFO - ### word hey victoria`
`2017-09-22 03:11:07,134 - mycroft.client.speech.listener - INFO - ### config {u'wake up': {u'threshold': 1e-20, u'phonemes': u'W EY K . AH P', u'module': u'pocketsphinx'}, u'hey mycroft': {u'threshold': 1e-90, u'phonemes': u'HH EY . M AY K R AO F T', u'module': u'pocketsphinx'}}`

`Traceback (most recent call last):
  File "/usr/local/bin/mycroft-speech-client", line 14, in <module>
    load_entry_point('mycroft-core==0.8.22', 'console_scripts', 'mycroft-speech-client')()`
`  File "/usr/local/lib/python2.7/site-packages/mycroft_core-0.8.22-py2.7.egg/mycroft/client/speech/main.py", line 140, in main
    loop = RecognizerLoop()`
`  File "/usr/local/lib/python2.7/site-packages/mycroft_core-0.8.22-py2.7.egg/mycroft/client/speech/listener.py", line 193, in __init__
    self._load_config()`
`  File "/usr/local/lib/python2.7/site-packages/mycroft_core-0.8.22-py2.7.egg/mycroft/client/speech/listener.py", line 211, in _load_config
    self.wakeword_recognizer = self.create_wake_word_recognizer()`
`  File "/usr/local/lib/python2.7/site-packages/mycroft_core-0.8.22-py2.7.egg/mycroft/client/speech/listener.py", line 230, in create_wake_word_recognizer
    LOG.info("### config {}".format(config[word]))`
`KeyError: u'hey victoria'`

forslund:
Thanks for reporting!

Does it revert to standard Hey Mycroft or does it silently crash?

smpickett:
It causes the mycroft-speech-service to fail to boot, so it is completely unresponsive. If the default is configured at home.mycroft.ai (with 'hey mycroft') then it works fine.

[...]

PR #1106, opened 2017-09-22
Make sure structure for wake word config entry exist before use

forslund:
==== Fixed Issues ====
#1105

====  Tech Notes ====
Verify that the sub-config for the set wake word exist before writing to
it.


penrods:
Sounds like this doesn't fix #1105 completely, we'll keep investigating

PR #1108, opened 2017-09-22
fallback start, end, failure messages

JarbasAI:
add messages for intent_failure start end and failure

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -37,7 +37,6 @@ from mycroft.util.log import LOG
 
 __author__ = 'seanfitz'
 
-
 MainModule = '__init__'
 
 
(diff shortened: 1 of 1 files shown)


mycroft/skills/core.py
```diff
@@ -661,15 +665,27 @@ def make_intent_failure_handler(cls, ws):
...
+            ws.emit(Message("mycroft.skill.handler.start",
+                            data={'handler': "fallback"}))
             for _, handler in sorted(cls.fallback_handlers.items(),
                                      key=operator.itemgetter(0)):
                 try:
                     if handler(message):
+                        #  indicate completion
+                        ws.emit(Message(
+                            'mycroft.skill.handler.complete',
+                            data={'handler': get_handler_name(handler)}))
```
forslund: The handler name of the complete should match the name given in `mycroft.skill.handler.start` (fallback) Having the name of the actual handler is probably a good idea so I suggest a separate field 'fallback_handler' for the name of the handler.

forslund:
Looking Good @JarbasAI, thanks!

PR #1111, opened 2017-09-23
check for language support on skill loading

JarbasAI:
on load skill do not load skills that don't support current language

if no "regex" and no "vocab" folders exist -> all languages supported
if "vocab" or "regex" folder for self.lang exist -> language is supported

#621 same as this closed and unmerged PR

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -122,6 +122,8 @@ def load_skill(skill_descriptor, bus, skill_id, BLACKLISTED_SKILLS=None):
                 callable(skill_module.create_skill)):
             # v2 skills framework
             skill = skill_module.create_skill()
+            if not skill.is_current_language_supported(path):
+                LOG.info("SKILL DOES NOT SUPPORT CURRENT LANGUAGE")
             skill.settings.allow_overwrite = True
             skill.settings.load_skill_settings_from_file()
             skill.bind(bus)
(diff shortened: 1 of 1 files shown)


penrods:
I'm not 100% certain about this feature.  I'm a native english speaker, so I can't really judge it fairly.  Would nobody ever want a skill that isn't in their native tongue?  I.e. if there was a skill that only supported German responses but it did something I really want (e.g. control some unique piece of IoT from Germany), I think I'd want access to it.

Of course, this might be a practical matter.  Likely the STT will never generate the correct text for match in Adapt, so the skill might be functionally useless anyway.

I'd like to hear the experience of others.  @forslund @augustnmonteiro @JarbasAI ?

JarbasAI:
the skill maker can always override the method and return True inconditionally if the skill doesn't care of language

overall i think this is desired, the skill will be hard to trigger and just be loaded and not used, using up memory and increasing boot time

i imagine it being frustrating for users aswell trying to use it / triggering it by accident

penrods:
I think the real frustration point is the install, not the load.  If I "successfully" install a skill but then it silently doesn't load, *that* would be confusing.

I expect that we will be able to support multiple languages simultaneously.  So I'm questioning that this change makes things actually better.

forslund:
Personally I'd prefer falling back to english if my language isn't supported. Would allow some usage instead of no usage.

The interference might be a valid point though but that will have to be seen and could be handled during intent parsing.

JarbasAl:
@forslund  you could never fallback to english because the STT would not output english text

unless we add a layer to detect language from sound and change STT

penrods:
At one point there was a demo someone created of a system that auto-switched based on the spoken language.  This is interesting, but going to be some time before it is really ready for prime time.

@JarbasAI, you have been working with autotranslation, right?  How is that experience?  I'm thinking about a layer in the mycroft.dialog.get() and such that can use a translation service and cache the results when no current-language version of the resource requested is found.  That seems like it would be a better experience (assuming the translation wasn't horrible) and would be a better path going forward where we will be able to dynamically switch lang based on what the STT detects.  This is why we pass around a lang value instead of just using a single global setting all the time.

JarbasAl:
i used auto translation at several levels, auto translating regex and .voc files does not work very well

i made a AutotranslatableSkill class that translates at runtime if it detects a different language in text , and also do the reverse for fallbacks that require a certain language at input https://github.com/JarbasAl/auto_translatable_skills at the end of this blog post you have the repos

my changes were at the speak  method, not in the dialog, this accounts for runtime generated speech. i do not have any sort of caching of results done

I am currently training a model to recognize language from audio in order to change STT engine

PR #1115, opened 2017-09-26
Add flag to allow running dev_setup.sh as root

forslund:
====  Tech Notes ====
Added flag --allow-root to circumvent the requirement to run ./dev_setup
as a non-root user.

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -27,8 +27,42 @@
 # exit on any error
 set -Ee
 
-if [ $(id -u) -eq 0 ]; then
+show_help() {
+        echo "dev_setup.sh: Mycroft development environment setup"
+        echo "Usage: dev_setup.sh [options]"
+        echo
+        echo "Options:"
+        echo "    -r, --allow-root  Allow to be run as root (e.g. sudo)"
+        echo "    -sm               Skip building mimic"
+        echo "    -h, --help        Show this message"
+        echo
+        echo "This will prepare your environment for running the mycroft-core"
+	echo "services. Normally this should be run as a normal user,"
+	echo "not as root/sudo."
+}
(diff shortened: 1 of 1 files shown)


el-tocino:
As you could potentially have both -sm and --allow-root, might be better to have distinct argument handling done earlier in the script.

penrods:
I do like having a help option and placing that help text at the very top of the code.  See 
https://github.com/MycroftAI/mycroft-core/blob/dc4a28c8fa9fcc4a7d0374192de4a81a1941e0d5/start-mycroft.sh

So let's add something like this near the very top:

```bash
show_help() {
        echo "dev_setup.sh: Mycroft development environment setup"
        echo "usage: dev_setup.sh [options]"
        echo
        echo "Options:"
        echo "    -r or -allow-root  Allow setup to be run as root (e.g. sudo)"
        echo "    -sm                Skip compiling mimic"
        echo
        echo "This will prepare your environment for running the mycroft-core services"
        echo "Normally this should be run as a normal user, not sudo/root."
}

opt_skipmimic=false
opt_allowroot=false

for var in "$@"
do
    if [[ ${var} == "-h" ]] || [[ ${var} == "--help" ]] ; then
        show_help
        exit 0
    fi

    if [[ ${var} == "-r"]] || [[ ${var} == "--allow-root" ]] ; then
        opt_allowroot=true
    fi

    if [[ ${var} == "-sm" ]] ; then
        opt_skipmimic=true
    fi
done


...

if [[ "${opt_skipmimic}" == true ]] ; then
   ...
```

forslund:
Totally forgot that the script already took an argument. will fix

forslund:
Ok, Updated!

PR #1117, opened 2017-09-26
Add workaround for inspect-decorator problem

forslund:
====  Tech Notes ====
when using multiple decorators on a method inspect will return incorrect
values despite `@wraps`. This causes the intent handler to fail to
execute.

`@decorator` can't really be of help here since it doesn't handle
decorators with arguments (as far as I can understand)

This is a workaround using the fact that the argument count will be zero
on methods with multiple decorators, and basically tries the usual
signatures.

@MatthewScholefield you have more experience with inspect than I do, is there a better solution apart from switching to python 3.5+ where `@wraps` is more complete?

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -344,6 +344,13 @@ class MycroftSkill(object):
                         handler(self, message)
                     elif len(getargspec(handler).args) == 1:
                         handler(self)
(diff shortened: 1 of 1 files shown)


btotharye:
Verified this fix works since I hit it today on a skill I was setting up, testing the PR code and it works as expected :)

PR #1119, opened 2017-09-29
Fix configuration manager saving

MatthewScholefield:
`dict.update()` returns None which previously deleted any saved config.


forslund:
Good catch!

we might want to go one step further and use the merge_conf method (L100) otherwise the learning update will overwrite any other changes to the listener sub-dict (for example).

Won't be a problem in most cases but if the method starts to be used more...

MatthewScholefield:
@forslund Nice catch as well! Fixed.

forslund:
Tested, works like a charm! Excellent! Merging!

PR #1120, opened 2017-09-29
Fix log autocomplete

MatthewScholefield:
Due to the use of `setattr()`, IDEs like PyCharm couldn't use autocomplete on the `LOG` class. This unrolls the loop so that the logging attributes appear in the autocomplete menu.


forslund:
Merging since this is approved!

PR #1122, opened 2017-09-30
minor grep tweak

el-tocino:
quietly grep instead.

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -117,13 +117,13 @@ else
   # first, look for a build of mimic in the folder
   has_mimic=""
   if [[ -f ${TOP}/mimic/bin/mimic ]] ; then
-      has_mimic=$( ${TOP}/mimic/bin/mimic -lv | grep Voice )
+      has_mimic=$( ${TOP}/mimic/bin/mimic -lv | grep -q Voice )
   fi
 
   # in not, check the system path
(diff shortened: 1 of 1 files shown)


MatthewScholefield:
I think since it's executing grep in `$()`, the output goes to the variable rather than the terminal (and the return code is discarded). Because of this, making it quiet seems to make it never detect mimic because `has_mimic` is always empty and evaluates to false. Let me know if this sounds right.

penrods:
I believe @MatthewScholefield is correct.   I verified with a quick test script under KDE Neon (basically Ubuntu).   @el-tocino, does this give you different results in your environment?  If so, what are you running under?

My test_quiet.sh code:
```bash
has_voice=$( mimic -lv | grep Voice )
has_voix=$( mimic -lv | grep Voix )
has_voice_quiet=$( mimic -lv | grep -q Voice )
has_voix_quiet=$( mimic -lv | grep -q Voix )

echo "has_voice= '${has_voice}'"
echo "has_voix = '${has_voix}'"
echo "has_voice_quiet= '${has_voice_quiet}'"
echo "has_voix_quiet = '${has_voix_quiet}'"
```
Output of `bash test_quiet.sh`:
```
has_voice= 'Voices available: ap slt kal awb kal16 rms awb_time '
has_voix = ''
has_voice_quiet= ''
has_voix_quiet = ''
```

PR #1123, opened 2017-10-02
Chromecast quit current app before starting new playback

forslund:
====  Tech Notes ====
Before the current chromecast application was quit when mycroft was started which caused some interference. Now the current app is quit right before starting playback on the device instead.


aatchison:
Good good. This was an issue with a community member.

PR #1125, opened 2017-10-03
Upgrade Padatious to 0.3.2 which brings in entity support

MatthewScholefield:
This upgrades Padatious which adds support for defining entity examples when declaring example-based intents. For example:

## Defining Entities
**`weather.intent`**
```
Is  {weekday}?
Is today {weekday}?
```
**`weekday.entity`**
```
Monday
Tuesday
...
```

## Intent Expansion
In addition, to reduce repetition, you can write out values with the or symbol in parentheses like this:

### Long Version
**`food.order.intent`**
```
Order some {food}.
Order some {food} from {place}.
Grab some {food}.
Grab some {food} from {place}.
```

### Short Version
**`food.order.intent`**
```
(Order | Grab) some {food} (from {place} | )
```

Another feature brought in is the ability to match numbers by using `#`. For example:

**`call.intent`**
```
Call {number}.
Call the phone number {number}.
```

**`number.intent`**
```
+### (###) ###-####
+## (###) ###-####
+# (###) ###-####
(###) ###-####
###-####
###-###-####
###.###.####
### ### ####
##########
```

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -403,19 +403,62 @@ class MycroftSkill(object):
     def register_intent_file(self, intent_file, handler, need_self=False):
         """
             Register an Intent file with the intent service.
+            For example:
+
+            === food.order.intent ===
+            Order some {food}.
+            Order some {food} from {place}.
+            I'm hungry.
+            Grab some {food} from {place}.
+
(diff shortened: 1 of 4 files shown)


Mn0491:
LGTM!

PR #1132, opened 2017-10-05
Padatious fixes

MatthewScholefield:
This fixes using the wrong method to register intents with Padatious and increments the version from `0.3.2` to `0.3.3` which adds a bugfix for calculating intents when the container is empty.


Mn0491:
LGTM! merging

PR #1143, opened 2017-10-08
Setup skill before installing in skill dir

### LINKED ISSUE #1137: msm install / update issue
The basic issue is loading/reloading the skill before msm has installed requirements (if new requirements are needed) causing the skill to not load at all. (Issue Identified by @penrods)

Install skill:
When msm is installing a new skill it clones directly into the skill directory resulting that the reload skill thread will load the skill as soon as it's there without looking if msm is done.

Update skill:
Currently this is not a problem because on update setup scripts don't seem to be run, which is a problem in itself. but will need to be handled as well at some point

Solutions:
Filesystem: msm cloned into /tmp and all setup scripts (pip install -r requirements.txt, requirements.sh) are run and the skill is then moved to the skill directory. For update I think adding support for a LOCK file in the skill directory. msm places the lock file in the directory before starting the update procedure and removes it afterwards. the reload thread looks for this file and if it's present no reload will be performed.

Another solution is to download and run the requirement.txt requirements.sh scripts before cloning / updating a skill.

@el-tocino, @btotharye since you're part of the skills team maybe you have some more insights into this?

forslund:
==== Fixed Issues ====
Partially resolves #1137

====  Tech Notes ====
Clones the skill into a tmp directory, runs pip install requirements.txt and requirements.sh and then moves the skill to the mycroft skill dir.

### DIFF
--- a/msm/msm
+++ b/msm/msm
@@ -250,8 +250,9 @@ function install() {
     fi
 
     echo "Installing from: ${repo}"
-    git clone "${repo}" >> /dev/null
-    if ! cd "${name}" ; then
+    temp_folder=`mktemp -d`
+    git clone "${repo}" ${temp_folder} >> /dev/null
+    if ! cd "${temp_folder}" ; then
       echo "ERROR: Unable to access directory ${name}!"
       return 102
     fi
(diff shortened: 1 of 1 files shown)


forslund:
Differnt path selected. Closing

PR #1147, opened 2017-10-10
Updating version when device is update

augustnmonteiro:
Updating core and enclosure versions when device starts, so we have the current information on the backend


mycroft/configuration/mycroft.conf
```diff
@@ -107,7 +107,7 @@
   "websocket": {
     "host": "0.0.0.0",
     "port": 8181,
-    "route": "/core",
+    "route": "/core"    ,
```
MatthewScholefield: Random extra space

PR #1148, opened 2017-10-10
wss support for websocket

JarbasAI:
support for secure websocket

for discussion!

how should the certificates be handled? 

i included a small script  to create a self signed cert

Changed some things in order to avoid this error with self signed certs

`SSLError: [SSL: TLSV1_ALERT_UNKNOWN_CA] tlsv1 alert unknown ca (_ssl.c:581)`

not sure [this](https://github.com/MycroftAI/mycroft-core/pull/1148/files#diff-a262263624537add2e6640c6acf1e993R118) is safe or desired,  it's worse having insecure ssl than no ssl, i do not know enough of the subject to judge the safety of this

currently checks config for flag to generate self-signed and for paths to key and cert

### DIFF
--- a/mycroft/messagebus/client/ws.py
+++ b/mycroft/messagebus/client/ws.py
@@ -14,6 +14,7 @@
 #
 import json
 import time
+import ssl
 from multiprocessing.pool import ThreadPool
 from threading import Event
 import traceback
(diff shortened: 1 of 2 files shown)


augustnmonteiro:
@JarbasAI hey man, thanks for the PR :+1: 
Can you rebase it, the header files was changed and we removed the authors from the files. thanks

JarbasAI:
can someone confirm the ssl is safe? including the self-signed certificate script?

JarbasAl:
closing this one, its a moot point in the larger picture, https://github.com/Nhoya/MycroftAI-RCE was basically ignored, but a deeper change is needed, i solved this differently in chatterbox and might revisit porting our changes later

PR #1151, opened 2017-10-12
improved skill settings

Mn0491:
improved skill settings
====  Tech Notes ====
added capability to auto upload changes from settingsmeta.json to home.mycroft.ai

If a developer make changes to the settingsmeta.json, then this will be auto uploaded to home.mycroft.ai

NOTE: currently the auto delete of previous settingsmeta.json from device function is commented out until the api end point is implemented in the backend. For now once you make changes to the settingsmeta.json, you will have an entirely new entry in home.mycroft.ai. You will need to delete the old one manually on home.mycroft.ai.

==== Protocol Notes ====
hash and uuid are now stored as variables in files located in ~/.mycroft/skills/{skill-name}

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -603,7 +603,7 @@ class MycroftSkill(object):
         """
         # Store settings
         self.settings.store()
-
+        self.settings.is_alive = False
         # removing events
         for e, f in self.events:
             self.emitter.remove(e, f)
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -254,7 +254,7 @@ class MycroftSkill(object):
         try:
             return self._settings
         except:
-            self._settings = SkillSettings(self._dir)
+            self._settings = SkillSettings(self._dir, self.name)
             return self._settings
 
     def bind(self, emitter):
--- a/mycroft/skills/settings.py
+++ b/mycroft/skills/settings.py
@@ -53,7 +53,7 @@ class SkillSettings(dict):
         self._settings_path = join(directory, 'settings.json')
         self._meta_path = join(directory, 'settingsmeta.json')
         self._api_path = "/" + self._device_identity + "/skill"
-
+        self.is_alive = True
         self.loaded_hash = hash(str(self))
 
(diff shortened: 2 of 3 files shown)


forslund:
I've just skimmed the code so I might be totally off but won't the Timers hang around after skill shutdown? endlessly repeat? each time a skill is restarted a new one will occur?

Also the tail reactivated timer is quite unreliable when trying to shutdown since cancel won't work while the timed function is executed.

I'll look in more detail tomorrow morning.

forslund:
Did some basic testing and the timers are hanging around after settings goes out of scope. This means each time a skill get's reloaded another timer will be started.

The timers also hinders the process to exit cleanly, This can be quickly fix by setting the `daemon` property of the timers to `True`

Mn0491:
@forslund @MatthewScholefield fixed the timer issue. can you review

forslund:
This solves part of the problem. This will allow the skill process to terminate despite timers being active.

However the Timer still seems to keep going after skill shutdown as far as I can see (and I might very well be missing something).

Probably not a big deal in most cases but not a good situation in the long run. I have an idea to deal with this using weakrefs but won't have time to deal with this tonight. I can get up early tomorrow morning and give it a go but if you think this is acceptable for this release go ahead and merge.

MatthewScholefield:
@forslund I'm pretty sure this won't continue after restarting the skills service because this example python code will immediately exit without printing `hello`:

```Python2
#!/usr/bin/env python2
from __future__ import print_function
from threading import Timer


t = Timer(1, print, args=('hello',))
t.daemon = True
t.start()
```

forslund:
Not restarting the process, that is all good. Reloading the skill. This happens when a skill is changed for example at an update.

Mn0491:
@forslund matthew came up with an idea to set a flag in skill settings called is_alive, and we only poll if is_alive = True, on shutdown in the core.py we can call self.settings.is_alive = False, and that should stop the polling.

forslund:
That will work just fine. (I really just wanted a reason to use weakrefs :) )

Add that and merge!

Issue #967, opened 2017-08-05
Eyes never stop spinning on startup (Mark 1)


kjcole:
Mycroft appears have become a bit slower to start, and now displays rolling eyes continuously and the mouth never appears. The menus ILLUM, et al, still appear, and TEST runs through its set of routines okay... I think. The OFF and REBOOT menu choices do not appear to work.

I can still SSH in, and still Mycroft still listens and speaks.

(Earlier today he was displaying the time continuously, but no longer.)

```
$ uname -a
Linux mark_1 4.9.35-v7+ #1014 SMP Fri Jun 30 14:47:43 BST 2017 armv7l GNU/Linux
```

```
$ lsb_release --all
No LSB modules are available.
Distributor ID:	Raspbian
Description:	Raspbian GNU/Linux 8.0 (jessie)
Release:	8.0
Codename:	jessie
```

```
$ dpkg -l mycroft-core
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name           Version      Architecture Description
+++-==============-============-============-=================================
ii  mycroft-core   0.8.20       armhf        mycroft-core
```

forslund:
Hi,

can you still interact with Mycroft? He can tell you the time but not show it?

If that's the case it sounds like the encolsure service isn't started properly. You can check if the process is running with `ps aux | grep enclosure`, if it's running you should see something similar to

```
mycroft    939  4.1  2.1 286992 20460 ?        Sl   06:37   0:04 python2.7 /usr/local/bin/mycroft-enclosure-client
```

If not try running `sudo /etc/init.d/mycroft-enclosure-client start` and you'll see the error information on the command line.

I had a similar issue and I had messed up the config-file causing the launch of the enclosure to halt.

kjcole:
I haven't altered the config file.  Whatever the problem was, it happened consistently yesterday across power cycling the machine (unplugging it completely).  However, this morning the eyes are stable, the mouth is working, etc.  It seems he was having a bad day?

forslund:
Bad days happens to us all :)

[...]

PR #1160, opened 2017-10-16
Prevent the "rolling eye" syndrome - Issue #967

penrods:
When (re)booting a Mark 1 unit will show rolling eyes until it reaches
a "ready" state.  This happens by sending a command to the Arduino.
There is also code that prevents sending commands our the serial port
if not running on a Mark 1.  In certain situations, the message
indicating that the Mark 1 Arduino was found was posted to the
messagebus before it was fully open.  When this was missed, the system
didn't think it was on a Mark 1 and the command to stop the eyes from
rolling (and for further interactions with the Mark 1 hardware) were
not sent.

The Mark 1 Arduino detection is now triggered when the messagebus
'open' notification is generated rather than when the object is
constructed.

==== Fixed Issues ====
#967 - Eyes never stop spinning on startup (Mark 1)

### DIFF
--- a/mycroft/client/enclosure/__init__.py
+++ b/mycroft/client/enclosure/__init__.py
@@ -232,7 +232,7 @@ class Enclosure(object):
 
     def __init__(self):
         self.ws = WebsocketClient()
-        self.ws.open("open", self.on_ws_open)
+        self.ws.on("open", self.on_ws_open)
 
         ConfigurationManager.init(self.ws)
         self.config = ConfigurationManager.instance().get("enclosure")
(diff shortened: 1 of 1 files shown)


mycroft/client/enclosure/__init__.py
```diff
@@ -232,12 +232,19 @@ class Enclosure(object):
 
     def __init__(self):
         self.ws = WebsocketClient()
+        self.ws.open("open", self.on_ws_open)
```
MatthewScholefield: Probably should be `self.ws.on(`

PR #1161, opened 2017-10-16
Fix for install of MSM defaults

penrods:
Indentation added in a recent PR (#1152) introduced a shell script
syntax error on a long line with continuations.  Spaces are not
allowed before the subsequent line's double-quotes.

==== Fixed Issues ====
Indentation added in a recent PR (#1152) introduced a shell script
syntax error on a long line with continuations.  Spaces are not
allowed before the subsequent line's double-quotes.


forslund:
Tested and works. Merging

PR #1165, opened 2017-10-17
Return if audio is None in listener

MatthewScholefield:
This is necessary because if the listener receives a stop signal, it returns `None`.

### DIFF
--- a/mycroft/client/speech/listener.py
+++ b/mycroft/client/speech/listener.py
@@ -101,6 +101,9 @@ class AudioConsumer(Thread):
         except Empty:
             return
 
+        if audio is None:
+            return
+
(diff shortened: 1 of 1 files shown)


Mn0491:
LGTM

forslund:
Looks sane, both Queue empty and None will result in the same outcome.

I see one issue I didn't see before. you need to specify timeout=0.5 in self.queue.get() (timeout is the second argument)

so it should be `self.queue.get(timeout=0.5)`

I had constant lockups when testing and this fixed it, it's a wonder that it have worked as well as it has.

forslund:
Ok I'll send a PR with that change then

Mn0491:
oh shoot

PR #1166, opened 2017-10-17
Fix timeout of self.queue.get() in consumer

forslund:


Mn0491:
ok we tried on interpreter and it does what we want it to do. LGTM

PR #1169, opened 2017-10-18
Add https support to audio service

roadriverrail:
The audio service could not handle https URLs.  VLC can support them,
so https was added to its supported_uris.  Additionally, the play( )
function in the audio service could not actually correctly search the
backends for a supported uri, so the code there has been fixed


forslund:
Works like a charm. Many thanks! Merging

PR #1170, opened 2017-10-19
Fall back to default wake word on invalid config

MatthewScholefield:
Also format to lowercase to prevent decoder error if wake word has capitals

### Testing procedure
```bash
./start-mycroft.sh all && tail -f scripts/logs/mycroft-voice.log
#  - Create invalid config in home.mycroft.ai (set phonemes to something like "ZZZ")
#  - "hey mycroft, update config"
# You should see an exception in the logs and it should no longer respond
#  - Ctrl+C

git fetch
git checkout feature/robust-listener
./start-mycroft.sh all && tail -f scripts/logs/mycroft-voice.log
# You should still see exceptions in the logs, but it should still respond
#  - Reset hom.mycroft.ai to valid config
#  - "hey mycroft, update config"
# You should no longer see an exception in the logs and it should still respond
```

### DIFF
--- a/mycroft/client/speech/hotword_factory.py
+++ b/mycroft/client/speech/hotword_factory.py
@@ -57,7 +57,7 @@ class PocketsphinxHotWord(HotWordEngine):
         self.num_phonemes = len(self.phonemes.split())
         self.threshold = self.config.get("threshold", 1e-90)
         self.sample_rate = self.listener_config.get("sample_rate", 1600)
-        dict_name = self.create_dict(key_phrase, self.phonemes)
+        dict_name = self.create_dict(self.key_phrase, self.phonemes)
         config = self.create_config(dict_name, Decoder.default_config())
         self.decoder = Decoder(config)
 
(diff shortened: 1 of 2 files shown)


MatthewScholefield:
In response to @pep8speaks, in this case I think a broad except is necessary because the instantiation can be of any class type and thus throw any exception that should be handled.

forslund:
Agreed, a catch-all is good here.

I was just looking at this kind of thing today in another part of the code and googled upon this: http://effbot.org/zone/stupid-exceptions-keyboardinterrupt.htm, they argue that KeyboardInterrupt should often be re-raised.

forslund:
Tested this and it looks good.

However when it sends the wakeword to the server for learning it's tagged with the `wake word` set at home.mycroft.ai and not _hey mycroft_ which is the wake word that's actually used.

Is this something that should be changed or is this acceptable behaviour? Feels to me it might be confusing when tagging.

MatthewScholefield:
@forslund Nice catches. Although it's unlikely a `KeyboardInterrupt` will be thrown in the split second it initializes the Hotword, I've added the except clause just in case. Also, I now have it pull the wake word name from the wake word recognizer to prevent mismatch.

MatthewScholefield:
Mwahahaha, no one will ever be able to tell if I actually deleted Codacy's comment or if it was another automated self-deletion.

forslund:
Indeed in this case it's highly unlikely that the exception should matter, I mainly brought it up as something we might think of in the future.

Anyways, this looks very good now so.... Merging!

PR #1173, opened 2017-10-22
Feature/subscription voice

forslund:
Download specific subscriber voice as needed. Currently supporting x86_64 and armv7l.

Download will launch in background when mimic tts is loaded, when download is finished. During download the Alan Pope voice will be used, when finished the voice will switch to the newly downloaded voice.


JarbasAI:
are these additional paid voices for mimic?

penrods:
@JarbasAI, yep!  First is an American female voice.

PR #1174, opened 2017-10-23
phonemes from string util

JarbasAI:
Add util to get phonemes from any string, useful for example for a skill that changes wake word (now it is possible to change config at runtime)

checks cmu dict from nltk for phonemes, if not available guesses them

print "jarbas", get_phonemes("jarbas")
print "alexa", get_phonemes("alexa")
print "hey mycroft", get_phonemes("hey mycroft")
print "raspi", get_phonemes("raspi")

jarbas JH AE R B AE S
alexa AH0 L EH1 K S AH0
hey mycroft H H   E Y 1 . M   Y   K   R   O W   F   T
raspi R AE S P IH


forslund:
Hi JarbasAI!

A couple of issues here, nltk isn't in the requirements.txt file, nltk requires the dataset for the cmudict to be downloaded separately so for this to work in the general case the download should be performed in the dev_setup script.

JarbasAI:
@forslund changed to a lighter requirement

PR #1176, opened 2017-10-27
Fix enclosure client broken by #1110

penrods:
Syntax error, needed Configuration.get().get("enclosure")
instead of just Configuration.get("enclosure")


forslund:
Tested and is working on my machine.

PR #1177, opened 2017-10-27
Fix PEP8 and unittest errors from PR 1049

penrods:
PR 1049 introduced several cosmetic PEP8 errors that were easily fixed.
Additionally there are unittests that include non-ASCII characters which are
failing.  As Pt-PT support is a work-in-progress, I just commented them out
with TODOs next to them.

### DIFF
--- a/mycroft/util/parse.py
+++ b/mycroft/util/parse.py
@@ -238,13 +238,13 @@ def extract_datetime_en(str, currentDate=None):
 
     def date_found():
         return found or \
-               (
-                   datestr != "" or timeStr != "" or
-                   yearOffset != 0 or monthOffset != 0 or
-                   dayOffset is True or hrOffset != 0 or
-                   hrAbs != 0 or minOffset != 0 or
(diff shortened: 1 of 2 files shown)


penrods:
@JarbasAI check out the unittest failures, please.

forslund:
Looks good!

JarbasAI:
will investigate tests failing, i'm pretty sure they weren't before so perhaps i messed up the PR itself with old code...

forslund:
@jarbasai from my brief checks it's a file encoding issue. All non-ascii characters shows as <?> when the file is in utf-8. Do you know if you're using one of the Latin-* encodings or something?

JarbasAI:
if i am it is not on purpose, doesn't # -*- coding: iso-8859-15 -*- at the top of the file solve this?

JarbasAI:
i am not sure how to fix these, it seems to be encoding related, i just replaced the ? in a couple tests to the correct char and those no longer fail

how can i help solving this?

PR #1178, opened 2017-10-28
Fixes for startup when not paired

penrods:
Several attempts to connect to the backend fail before pairing has
been completed, which was producing errors that prevented the
messagebus and audio services from starting up.

Now the DeviceApi().is_subscriber property returns False if the
check fails, assuming unpaired == not subscribed.

Also stopped logging the call stack on the HTTPError exception, it
made an expected situation look like a major crash.


penrods:
@forslund You can check this out later.  Taking it to continue the validation.

forslund:
Looks good!

PR #1183, opened 2017-10-28
Fix no-internet prompt

penrods:
The prompt during skill downloads was occurring even when the "speak" flag was
set to False.  Now it is honored.

Also removed the "no network connection.dialog" which essentially was a copy of
the "not connected to the internet.dialog" file.


forslund:
Tested and is working as expected.

PR #1184, opened 2017-10-28
PocketSphinx STT

JarbasAI:
add pocketsphinx as an offline solution for STT

### DIFF
--- a/mycroft/stt/__init__.py
+++ b/mycroft/stt/__init__.py
@@ -220,6 +220,16 @@ class GoVivaceSTT(TokenSTT):
         return response.json()["result"]["hypotheses"][0]["transcript"]
 
 
+class PocketSphinxSTT(BasicSTT):
+    def __init__(self):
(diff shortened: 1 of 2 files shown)


JarbasAI:
should i add .lm a .dict files for english by default?

penrods:
Those dicts are pretty big, right?  I don't really want to bloat the repo if we can avoid it and instead pull down models from elsewhere on demand.

JarbasAI:
32.4MB for the english .lm.bin and .dict

updating PR for .bin support

zenogantner:
If you are worried about bloating the repo, you could look at git-lfs as a solution.
Mozilla's DeepSpeech uses it, too.

JarbasAl:
closing this one, will make a plugin once #2594 is in

JarbasAl:
@renekliment and anyone else interested, a plugin for this is now available https://github.com/OpenJarbas/mycroft-stt-plugin-pocketsphinx

PR #1185, opened 2017-10-29
Inform user about skill installation at first start

forslund:
I haven't had time to test this thoroughly, seems to be ok from the 5 tests I've run but should be tested further before merge.

If no skills are present don't immediately try to start pairing, instead inform user that the skills will be installed and launch pairing after that.

### DIFF
--- a/mycroft/res/text/en-us/installing default skills.dialog
+++ b/mycroft/res/text/en-us/installing default skills.dialog
@@ -1,0 +1,1 @@
+Please wait while default skills are installed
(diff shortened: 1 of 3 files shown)


forslund:
I realized just now that it should be limited to speak only once and not on subsequent retries. Will update...

penrods:
I want to build on this to provide clean feedback points (booting, looking for wifi, connected, paired, initial download / updating skills, all skills loaded, etc), along with appropriate feedback in the user interface.  E.g. changing the eye spin and/or colors.

penrods:
I think the feedback in #1308 takes care of this issue.  Can we close this one out unmerged?

forslund:
Yes!

PR #1192, opened 2017-10-31
Removing default skill pip dependencies

aatchison:
I have removed what I believe to be the all the pip dependencies for the default skills


forslund:
I've tested it as well and seems to work just fine. Since @MatthewScholefield has approved it I'll go ahead and merge this

PR #1193, opened 2017-11-01
Feature/status event scheduler

Mn0491:
This PR gives an api for skill developers to query event status from the event scheduler

```
(_datetime, time_left) = self.get_event_status(name_of_scheduled_event)

#_datetime is the datetime object that the event will go off
# time_left is the seconds left in epoch time till the event go off
```

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -388,10 +388,10 @@ class MycroftSkill(object):
             Args:
                 name: Name Intent
         """
-        for e, f in self.events:
-            if name in e:
-                self.events.remove((e, f))
-                self.emitter.remove(e, f)
+        for _name, _handler in self.events:
+            if name == _name:
+                self.events.remove((_name, _handler))
+                self.emitter.remove(_name, _handler)
 
     def register_intent(self, intent_parser, handler, need_self=False):
         """
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -381,6 +381,18 @@ class MycroftSkill(object):
             self.emitter.on(name, wrapper)
             self.events.append((name, wrapper))
 
+    def remove_event(self, name):
+        """
+            Removes an event from emitter and events list
+
+            Args:
+                name: Name Intent
+        """
+        for e, f in self.events:
+            if name in e:
+                self.events.remove((e, f))
+                self.emitter.remove(e, f)
+
     def register_intent(self, intent_parser, handler, need_self=False):
         """
             Register an Intent with the intent service.
@@ -700,6 +712,7 @@ class MycroftSkill(object):
                 name (str):   Name of event
         """
         data = {'event': self._unique_name(name)}
+        self.remove_event(name)
         self.emitter.emit(Message('mycroft.scheduler.remove_event', data=data))
 
 
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -703,7 +703,7 @@ class MycroftSkill(object):
         }
         self.emitter.emit(Message('mycroft.schedule.update_event', data=data))
 
-    def cancel_scheduled__event(self, name):
+    def cancel_scheduled_event(self, name):
         """
             Cancel a pending event. The event will no longer be scheduled
             to be executed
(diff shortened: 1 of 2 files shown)


mycroft/skills/core.py
```diff
@@ -699,9 +712,44 @@ def cancel_event(self, name):
...
+
+            Return:
+                (datetime, int) (tuple): datetime and time left in epoch time
+        """
+        event_name = self._unique_name(name)
+        data = {'name': event_name}
+
+        # this is needed because message.data could return None
+        # and that would make the while loop infinite
+        self.event_status = "None"
```
forslund: I would personally have used the `None` type as a default value

mycroft/skills/core.py
```diff
@@ -699,9 +712,44 @@ def cancel_event(self, name):
...
+            time_left_in_seconds = event_time - current_time
+            date = datetime.datetime.now() + time_left
+            self.event_status = (date, time_left_in_seconds)
+
+        emitter_name = 'mycroft.event_status.callback.{}'.format(event_name)
+
+        self.emitter.once(emitter_name, callback)
+        self.emitter.emit(Message('mycroft.scheduler.get_event', data=data))
+
+        while True:
```
forslund: This should probably have a timeout of say two seconds, and do sleeps between checks to avoid hogging the cpu:

```python
start_wait = time.time()
while event_status is None and time.time() - start_wait < 2.0:
    time.sleep(0.1)
return event_status
```
(Having it return `None` on failure is a common approach in python)

forslund: Even better would be if it raised an exception on timeout. Not sure which one would be approprate. IOError or some from the websocket client?

Mn0491:
@forslund did some changes. Also i realized this PR also contains the remove_event code from the other PR. Should we just close the other PR and have this one for both?

forslund:
Yes, might be better to close that.

mycroft/skills/core.py
```diff
@@ -699,9 +711,45 @@ def cancel_event(self, name):
...
+                time_left (int): the time left in seconds
+        """
+        event_name = self._unique_name(name)
+        data = {'name': event_name}
+
+        # making event_status an object so it's refrence can be changed
+        event_status = [None]
+
+        def callback(message, event_status):
+            event_time = int(message.data[0][0])
```
forslund: Needs check if `message.data is not None` since that is a valid response if event doesn't exist. If it is `None` set event_status[0] to False or 0 to indicate that it doesn't exist.

(Timeout exception is a bit misleading if the actual issue is that the handler doesn't exist)

mycroft/skills/core.py
```diff
@@ -699,9 +711,45 @@ def cancel_event(self, name):
...
+            Return:
+                time_left (int): the time left in seconds
+        """
+        event_name = self._unique_name(name)
+        data = {'name': event_name}
+
+        # making event_status an object so it's refrence can be changed
+        event_status = [None]
+
+        def callback(message, event_status):
```
forslund: event_status don't seem to be needed as an argument

Mn0491: this is actually needed to be able to change the value of the event_status. If this is not passed this way to the callback then when we say event_status = whatever, it won't reference the same variable.

Mn0491:
@forslund made some changes and also changed the function name for scheduled events to be more expressive of what it's doing. let me know what you thiink

mycroft/skills/core.py
forslund: Hmm, when I tested it wasn't needed. As long the outer_var is a mutable (in this case a list) it should work.

```python
def outer():
    outer_var = [None]
    def inner():
        outer_var[0] = 'Modified'

    inner()
    print outer_var[0]

outer()
```

I'll retest but it's not really important, this way works as well.

forslund: Ok retried and it doesn't seem necessary but never mind both ways work.

mycroft/skills/core.py
```diff
@@ -691,17 +703,59 @@ def update_event(self, name, data=None):
...
+        emitter_name = 'mycroft.event_status.callback.{}'.format(event_name)
+        self.emitter.once(emitter_name,
+                          lambda message: callback(message, event_status))
+        self.emitter.emit(Message('mycroft.scheduler.get_event', data=data))
+
+        start_wait = time.time()
+        while event_status[0] is None and time.time() - start_wait < 3.0:
+            time.sleep(0.1)
+        if event_status[0] is False:
+            raise Exception("Scheduled Event not found for {}".format(name))
```
forslund: Sorry if I was unclear (or confused, or both) earlier. I meant that the function should return `None` if there is no scheduled event.

A use case I can imagine is to check if an event is scheduled at all and checking if The `self.get_scheduled_event_status(event) is None` is a pretty simple way to do it.

Timeout is an actual error while "event isn't scheduled" is a valid status (at least in my mind).

What's your opinion?

mycroft/skills/core.py
Mn0491: ahh ok thanks!

mycroft/skills/core.py
Mn0491: makes a lot of sense to me!

Mn0491:
@forslund let's try this again!

forslund:
Awesome, looks and works great! Merging

PR #1199, opened 2017-11-03
Precise Wake Word Listener

MatthewScholefield:
This adds precise as an optional wake word listener instead of Pocketsphinx. For testing, it has been made default by modifying the config with `{"hotwords": {"hey mycroft": {"module": "precise"}}}`. Once this has been tested, I can remove the `Make Precise default` commit before a merge.

Notes on testing: It downloads a 70 MB file the first time it runs. On the unit, this will be a 30MB file and it will notify the user with text on the screen. During this initial download period, Mycroft will not respond to the wake word.

### DIFF
--- a/mycroft/client/speech/hotword_factory.py
+++ b/mycroft/client/speech/hotword_factory.py
@@ -195,10 +195,7 @@ class PreciseHotword(HotWordEngine):
                 self.cooldown -= 1
                 self.has_found = False
                 continue
-            if float(line) > 0.5:
-                self.has_found = True
-            else:
-                self.has_found = False
+            self.has_found = float(line) > 0.5
 
     def update(self, chunk):
         self.proc.stdin.write(chunk)
--- a/mycroft/client/speech/hotword_factory.py
+++ b/mycroft/client/speech/hotword_factory.py
@@ -144,6 +144,10 @@ class PreciseHotword(HotWordEngine):
         except OSError:
             pass
 
+        exe_file = expanduser('~/.mycroft/precise/' + self.exe_name)
+        if isfile(exe_file):
+            return exe_file
+
         import platform
         import stat
 
@@ -152,7 +156,7 @@ class PreciseHotword(HotWordEngine):
             Popen('echo "' + cmd + '" > /dev/ttyAMA0', shell=True)
 
         arch = platform.machine()
-        exe_file = expanduser('~/.mycroft/precise/' + self.exe_name)
+
         url = self.url_base + 'dist/' + arch + '/' + self.exe_name
 
         snd_msg('mouth.text=Updating Listener...')
--- a/mycroft/client/speech/hotword_factory.py
+++ b/mycroft/client/speech/hotword_factory.py
@@ -111,8 +111,10 @@ class PreciseHotword(HotWordEngine):
     def __init__(self, key_phrase="hey mycroft", config=None, lang="en-us"):
         super(PreciseHotword, self).__init__(key_phrase, config, lang)
         self.update_freq = 24  # in hours
-        self.url_base = ('https://raw.githubusercontent.com/'
-                         'MycroftAI/precise-data/')
+
+        precise_config = Configuration.get()['precise']
+        self.dist_url = precise_config['dist_url']
+        self.models_url = precise_config['models_url']
         self.exe_name = 'precise-stream'
 
         ww = Configuration.get()['listener']['wake_word']
@@ -157,7 +159,7 @@ class PreciseHotword(HotWordEngine):
 
         arch = platform.machine()
 
-        url = self.url_base + 'dist/' + arch + '/' + self.exe_name
+        url = self.dist_url + arch + '/' + self.exe_name
 
         snd_msg('mouth.text=Updating Listener...')
         self.download(url, exe_file)
@@ -182,7 +184,7 @@ class PreciseHotword(HotWordEngine):
             if get_time() - stat.st_mtime < self.update_freq * 60 * 60:
                 return
         name = name.replace(' ', '%20')
-        url = self.url_base + 'models/' + name
+        url = self.models_url + name
         self.download(url, file_name)
         self.download(url + '.params', file_name + '.params')
 
--- a/mycroft/client/speech/hotword_factory.py
+++ b/mycroft/client/speech/hotword_factory.py
@@ -16,9 +16,16 @@ import tempfile
 import time
 
 import os
-from os.path import dirname, exists, join, abspath
+from os.path import dirname, exists, join, abspath, expanduser, isdir, isfile
+
+from os import mkdir
+from time import time as get_time
 
 from mycroft.configuration import Configuration
+from subprocess import Popen, PIPE, call
+from threading import Thread
+
+from mycroft.util import resolve_resource_file
 from mycroft.util.log import LOG
 
 
@@ -39,6 +46,9 @@ class HotWordEngine(object):
 
     def found_wake_word(self, frame_data):
         return False
+
+    def update(self, chunk):
+        pass
 
 
 class PocketsphinxHotWord(HotWordEngine):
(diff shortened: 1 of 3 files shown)


mycroft/client/speech/hotword_factory.py
```diff
@@ -97,6 +107,90 @@ def found_wake_word(self, frame_data):
...
+        self.download(url + '.params', file_name + '.params')
+
+    def check_stdout(self):
+        while True:
+            line = self.proc.stdout.readline()
+            if self.cooldown > 0:
+                self.cooldown -= 1
+                self.has_found = False
+                continue
+            if float(line) > 0.5:
```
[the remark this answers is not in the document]
MatthewScholefield: Anyone know what Codacy is talking about here? `bool(0.4)` evaluates to `True`.

forslund:
I'm having a bit of trouble getting precise running on my machine (Debian jessie:

From speech client (after download is complete)

```
[25960] Error loading Python lib '/tmp/_MEI0vQ7v1/libpython3.5m.so.1.0': dlopen: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.25' not found (required by /tmp/_MEI0vQ7v1/libpython3.5m.so.1.0)
Exception in thread Thread-14:
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/ake/projects/python/mycroft-core/mycroft/client/speech/hotword_factory.py", line 190, in check_stdout
    if float(line) > 0.5:
ValueError: could not convert string to float:
```

I get the same issue if I run the precise-stream from the commandline as well.

I'm on 64-Bit debian Stretch right now, will try on my Ubuntu desktop later tonight.

Another thing, could this use the downloader in utils to consolidate the code a bit? Or is it too clunky?

MatthewScholefield:
@forslund About the error, it's because I built the executable on too new of an OS. I'm working on running automation scripts in an Ubuntu 14.04 docker container which should fix this issue.

Also, I didn't even realize we had a download util. However, things would get a little more ugly if I split the code up since it is non blocking. But, I'm fine modifying Downloader to add a blocking parameter even though that sort of defeats the purpose of the class since it inherits from the Thread class.

forslund:
Point well made.

Will patiently wait for the updated executable :)

MatthewScholefield:
@forslund OK, it should be fixed. I rebased so you will have to `git fetch && git checkout feature/precise && git reset --hard origin/precise`

mycroft/client/speech/hotword_factory.py
```diff
@@ -97,6 +107,104 @@ def found_wake_word(self, frame_data):
...
+
+        import platform
+        import stat
+
+        def snd_msg(cmd):
+            """Send message to faceplate"""
+            Popen('echo "' + cmd + '" > /dev/ttyAMA0', shell=True)
+
+        arch = platform.machine()
+        exe_file = expanduser('~/.mycroft/precise/' + self.exe_name)
```
forslund: This path is not one of the paths checked in resolve_resource_file() so the exe is downloaded each startup.

mycroft/client/speech/hotword_factory.py
```diff
@@ -97,6 +107,104 @@ def found_wake_word(self, frame_data):
...
+        url = self.url_base + 'dist/' + arch + '/' + self.exe_name
+
+        snd_msg('mouth.text=Updating Listener...')
```
forslund: Could this be done using the enclosure api? Right now I get an error message saying it can't open the /dev/ttyAMA0 device. (also I think the /dev/ttyAMA0 should not be hardcoded, instead the enclosure->port value from the config might be better)

MatthewScholefield: I guess, but that gets messy because then the websocket has to get passed from `RecognizerLoop` to the `HotwordFactory` to the `HotwordEngine` base class, with a handler being written to communicate enclosure text commands from the internal recognizerloop `EventEmitter` to the actual websocket. Do you think it'd better to just remove it from communicating to the arduino?

mycroft/client/speech/hotword_factory.py
MatthewScholefield: Should be resolved t=now. Good catch!

mycroft/client/speech/hotword_factory.py
```diff
@@ -97,6 +107,104 @@ def found_wake_word(self, frame_data):
...
+        os.chmod(exe_file, os.stat(exe_file).st_mode | stat.S_IEXEC)
+        Popen('echo "mouth.reset" > /dev/ttyAMA0', shell=True)
+        return exe_file
+
+    @staticmethod
+    def download(url, filename):
+        import shutil
+        from urllib2 import urlopen
+        LOG.info('Downloading: ' + url)
+        req = urlopen(url)
```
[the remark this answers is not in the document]
MatthewScholefield: Since the prefix is fixed to `https://`, I don't think this is an issue.

mycroft/client/speech/hotword_factory.py
forslund: We can leave it as is for the time being(maybe use the config value instead of hardcoded path), this was not a big issue, just something I thought of when looking at the output. Feedback to the user is good providing an explanation when the device won't respond.

I'm a bit surprised that it works though, I'd expect the enclosure process to block the serial device and hinder other processes to write to it.

forslund:
Works! I'll test more tomorrow morning but if Michael tests it and wants to merge I trust his judgment. I've not yet tested broken downloads for example.

JarbasAI:
is precise only trained for "hey mycroft" or can we change it?

MatthewScholefield:
@JarbasAI Right now, hey mycroft is the only trained wake word, but it loads from a model file according to the wake word configuration value, meaning we can easily add more in the future. I'm working on building a static executable to easily train new wake words.

JarbasAI:
great :) was wondering if you would go the snowboy route and have them trainable in your website or something

is it easy to train new words?

how much wake word samples do you need to achieve good results?

is the precise repo up to date-ish? could i use that repo's code to train a new model and play around?

MatthewScholefield:
@JarbasAI The repo isn't up to date yet. There's still some work left to ease the process of training. At the moment, it requires a fair amount of data to prevent false positives which is what I'm working on automating. But, for the time being, testing it as an experimental feature will serve to collect the data to prevent false positives.

forslund:
I've tested some more and it seems pretty good. I'm still a bit worried about the direct write to the display but I've not tried it on an actual Mark-1 so there might be some magic I'm not aware of

Mn0491:
So holding I think we should hold off on pulling this in until we figure out the issues with the models

MatthewScholefield:
Even if we don't make it default? Since it downloads new models every day,
we should be able to improve it after the merge where we can test on the
device, and it won't affect regular users. However, at the very least, I
have to make one more commit to this branch changing the source to s3.

forslund:
(and make it not-default)

forslund:
Is it the fact that precise is default that makes Travis fail or is that something else?

forslund:
Is it the fact that precise is default that makes Travis fail or is that something else?

Mn0491:
I didn't catch the non-default part! i agree with merging it on non-default.

Mn0491:
I didn't catch the non-default part! i agree with merging it on non-default.

MatthewScholefield:
@forslund Well, it's failing because it's default and the tests don't use the new .update mechanism.

forslund:
*Sigh*, sorry github (or the universe) was acting up

forslund:
@MatthewScholefield Ok I'll make a not to revisit those tests

MatthewScholefield:
@JarbasAI Sorry, I thought I replied, but maybe it didn't go through. Well, since we still support pocketsphinx, those that choose to contribute their data are already building up a data for new wake words. However, right now it's not the easiest to train new wake words because it has a bunch of false positives at first. However, I'm fixing up a solution to this by feeding it a bunch of unrelated data to inhibit activation, but for the moment, the plan is to perfect `hey mycroft` by deploying it as an optional module so that people can begin testing, feeding data back into the model, to reduce false positives.

ppisljar:
Is precise open sourced as well ? (cant find it)

PR #1204, opened 2017-11-07
Move session config from global scope

forslund:
====  Tech Notes ====
As soon as mycroft, mycroft.api or mycroft.skills.core were imported the
configuration was loaded (including reaching out to the web config).
This will reduce unneccessary configuration loadings and make it easier
to handle configuration in the tests.


penrods:
I like where you are going with this ... I hate those globals.  I'll give it a shot as soon as I can.

PR #1205, opened 2017-11-08
Feature/keep transcriptions

GuyDaniels:
Changes core to optionally (based on signal files) write out either or both audio files and text transcription files which are both written to /var/log/mycroft.  Default is set to 'off'.
Note: Associated with skill-keep-transcriptions which provides intents to start/stop the audio/text transcriptions.

### DIFF
--- a/mycroft/client/speech/transcribe.py
+++ b/mycroft/client/speech/transcribe.py
@@ -10,7 +10,8 @@ import os
 
 __author__ = "reginaneon"
 
-text_permission = check_for_signal('transcribe_text_permission', 0) # starting default is off
+text_permission = check_for_signal(
+    'transcribe_text_permission', 0) # starting default is off
 audio_permission = check_for_signal('keep_audio_permission', 0)
 # text_permission = create_signal('transcribe_text_permission')
 # audio_permission = create_signal('keep_audio_permission')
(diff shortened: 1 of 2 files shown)


JarbasAI:
how exactly would this be turned on and off? i like the idea :)

GuyDaniels:
JarbasAI : The skill-keep-transcriptions turns the file signals off/on.  It's on a skills pull request - if I did that correctly..

JarbasAI:
couldnt Transcribe use @classmethod to avoid having to instantiate it? it doesnt really need self if i got it correctly

mycroft/client/speech/transcribe.py
```diff
@@ -0,0 +1,83 @@
...
+
+class Transcribe:
+    """
+    Name: Transcribe
+    Purpose: Writes the transcription file.
+    Imports data to the new, more precise text file, containing
+             the lines needed.
+    """
+    @classmethod
+    def write_transcribed_files(self, audio, text):
```
forslund: Does this have to be in a class? The self reference isn't used.

GuyDaniels: No, it doesn't need to be in a class.  Do you think I should just move write_transcribed_files() over into listener.py?  That works for me...

PR #1207, opened 2017-11-08
Fix wake word record length

MatthewScholefield:
Currently, wake word recordings that are used when the user opts in are too short (0.1 seconds or less). This fixes that. This also changes the value in the config to be in seconds. @forslund Let me know if there's a reason we should keep this as 120. I wasn't really sure what unit that was in, so I could be messing something up.

## Steps for Testing

**Before Patch:**
 - Comment [os.remove on line 338 of mic.py](https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/client/speech/mic.py#L338)
 - After `hey mycroft, <anything>`, the wav file stored in `/tmp/mycroft_wake_words/` should be really short

`git checkout bugfix/record-length`

**After Patch:**
 - Comment the same line of mic.py
 - After `hey mycroft, <anything>`, the wav file stored in `/tmp/mycroft_wake_words/` should be of long length and hold all of the wake word plus a bit before in case of multiple attempted activations.

### DIFF
--- a/mycroft/client/speech/mic.py
+++ b/mycroft/client/speech/mic.py
@@ -32,6 +32,7 @@ from speech_recognition import (
 )
 
 from mycroft.configuration import Configuration
+from mycroft.identity import IdentityManager
 from mycroft.session import SessionManager
 from mycroft.util import (
     check_for_signal,
@@ -430,19 +431,22 @@ class ResponsiveRecognizer(speech_recognition.Recognizer):
                 # file.
                 if self.save_wake_words and said_wake_word:
                     audio = self._create_audio_data(byte_data, source)
-                    stamp = str(int(1000 * get_time()))
-                    uid = SessionManager.get().session_id
+
                     if not isdir(self.save_wake_words_dir):
                         mkdir(self.save_wake_words_dir)
-
                     dr = self.save_wake_words_dir
+
                     ww = self.wake_word_name.replace(' ', '-')
-                    filename = join(dr, ww + '.' + stamp + '.' + uid + '.wav')
-                    with open(filename, 'wb') as f:
+                    stamp = str(int(1000 * get_time()))
+                    sid = SessionManager.get().session_id
+                    uid = IdentityManager.get().uuid
+
+                    fn = join(dr, '.'.join([ww, stamp, sid, uid]) + '.wav')
+                    with open(fn, 'wb') as f:
                         f.write(audio.get_wav_data())
 
                     if self.upload_config['enable'] or self.config['opt_in']:
-                        t = Thread(target=self._upload_file, args=(filename,))
+                        t = Thread(target=self._upload_file, args=(fn,))
                         t.daemon = True
                         t.start()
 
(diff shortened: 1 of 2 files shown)


forslund:
I believe the unit was ms (120/1000.0 => seconds). If I recall correctly 120ms was a suggested mean phoneme time for a typical speaker + 20 percent.

mycroft/configuration/mycroft.conf
```diff
@@ -125,7 +125,7 @@
       "user": "precise",
       "folder": "/home/precise/wakewords"
     },
-    "phoneme_duration": 120,
+    "phoneme_duration": 0.2,
```
penrods: Comment this default.  0.2 what?  

What was the reason for changing units?  What happens if someone customized this setting?  Why not stick with milliseconds?

forslund:
I did some testing on dev and feature/precise and can't seem to reproduce this.

Could it be that the num_phonemes is actually 0 (or possibly 1) in your test case?

The cast to int when converting to bytes is a really good catch. Is it that or upping the time per phoneme that's the core fix?

mycroft/client/speech/mic.py
```diff
@@ -335,7 +336,7 @@ def _upload_file(self, filename):
                       keyfile + ' ' + fn + ' ' + address
                 if os.system(cmd) == 0:
                     del self.filenames_to_upload[i]
-                    os.remove(fn)
+                    #os.remove(fn)
```
forslund: I think this sneaked in unintentionally

forslund:
Tested this with the "Hey Victoria" wake word (HH EY . V IH K T AO R IY AH) and for some reason it won't trigger. "Hey Mycroft" works fine and on dev Hey Victoria works. Will continue to investigate.

mycroft/client/speech/mic.py
```diff
@@ -283,7 +284,7 @@ def decrease_noise(level):
 
     @staticmethod
     def sec_to_bytes(sec, source):
-        return sec * source.SAMPLE_RATE * source.SAMPLE_WIDTH
+        return int(sec * source.SAMPLE_RATE * source.SAMPLE_WIDTH)
```
forslund: This needs to be rounded up for it to work consistently

either
`return int(math.ceil(sec * source.SAMPLE_RATE * source.SAMPLE_WIDTH))`
or the c hackers choice
`return int(sec * source.SAMPLE_RATE * source.SAMPLE_WIDTH + 0.5)`

MatthewScholefield: Hmm, this is strange because it indicates that changing a value by one sample (1/16000 of second) is making a difference. Could you by chance find which call to sec_to_bytes caused that?

forslund: The relevant call is at L366 in mic.py 

I think it's if the resulting byte size is odd, and since we pull the data from the rear of the byte array every 16 bit sample will be mangled.

forslund: which means my suggested fix is probably wrong instead we need to check that `bytes % sample_width == 0` and then append `sample_width - bytes % sample_width` padding if it's not?

MatthewScholefield:
@penrods Changed back to milliseconds and added a comment in the config to show the unit.

forslund:
Since at @penrods concerns have been addressed and the PR looks good. I'm merging this.

PR #1209, opened 2017-11-09
Bugfix/fix speech listener reload 1208

GuyDaniels:
This is how I was able to get reload to work.  This may well not be the best way - perhaps a bit of a kludge.
This reload is complicated by the differences between picroft/mycroft and the dev environments.  So I'm open to any suggestions.  It would be great if the original code worked!


GuyDaniels:
How about this approach?  I leave listener.reload() alone and add listener.restart() instead.  

I probably (maybe) could make skip wake words work without the restart, but I also have an up-coming pr for switching to local STT (pocketsphinx) which, when controlled by a skill, would need also to restart the speech client to re-init the recognizers and audio objects.

PR #1210, opened 2017-11-09
Feature/skip wake words

GuyDaniels:
Skipping wake words does just that.  It keeps recording, after 3 seconds of silence (or whatever the setting), it stops and then starts recording again.  There are a few comments in listener.process() where one can bypass the STT of the silence, which it might not be if the utterance happens to be exactly 3 secs.  Any other length of audio gets transcribed and sent on to skills.

Skipping wake words relies on the reload function working in order to start/stop it via a skill. That commit is included in this.  I'm not sure if that's how to handle commit dependencies, newby that I am to git.

### DIFF
--- a/mycroft/client/speech/listener.py
+++ b/mycroft/client/speech/listener.py
@@ -136,9 +136,11 @@ class AudioConsumer(Thread):
         if self._audio_length(audio) < self.MIN_AUDIO_SIZE:
             LOG.warning("Audio too short to be processed")
         else:
-            if len(audio.frame_data) != 65538 and \
-                    len(audio.frame_data) != 96258:  # 3 seconds silence
-                self.transcribe(audio)
+            # skip transcribing silence
+            # 3 seconds silence,
+            # matches RECORDING_TIMEOUT_WITH_SILENCE in mic.py
+            # if len(audio.frame_data) != 96258:
+            self.transcribe(audio)
 
     def transcribe(self, audio):
         text = None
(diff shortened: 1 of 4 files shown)


forslund:
Hi @GuyDaniels I'm not sure this is the best way to handle this. For one the laptop/desktop doesn't necessary need to be running a git install and a mark-1 may run a git checkout. (I do that from time to time to check some new feature)

Secondly it would be interesting to know where the reload hangs. Currently the recognizer loop should reload if the listener configuration is changed. Did you try to have the reload-handler reset the `loop._config_hash` and have it reloaded in the expected sequence?

Alternatively you could use the configuration "patch" feature to enable skip_wake_word. Have it as a config under listener and it should reload as well.

forslund:
Also thanks for raising this issue and helping to fix it, it is much appreciated

GuyDaniels:
So I separated reload and restart and did a little testing.  

When I do an emit of 'recognizer_loop:reload' for my skill I can debug and step through it in reload.   But It kills the speech client at some point (which I can't find), meaning it's gone - must be restarted.  I'm not sure where it actually dies and I don't see any 'crash' log anywhere.  I could debug further with try...catch...

When reload is run 'as expected' then it does reload the config, but I don't see it re-init the recognizer loop.  Seems to work ok for the new configurations, but it doesn't re-initialize the objects I need and skip_wake_words fails to work.

The skip-wake-words function only works with a restart.

I get it that the restart needs to different things depending on how mycroft was installed, but I'm not sure what to do about that.

PR #1211, opened 2017-11-09
Add optional skill announcement message

penrods:
A Skill can now implement the get_intro_message() method, which can return a
string to be spoken the very first time a skill is run.  This is intended to
be an announcement, such as further steps necessary to setup the skill.

Also stopped generation of the Error message when the expected StopIteration
occurs on a intent failure.  This confused new developers and poluted the logs.

Finally, corrected some documentation typos.


forslund:
Looks good, I've got an idea but I'll submit that as a separate PR.

Merging this!

PR #1217, opened 2017-11-09
Remove all mycroft packages individually

forslund:
Autoremove don't seem to be reliable enough (maybe due to the older version of the script installing mycroft-core separately). This will remove the required packages individually.

Has not been tested on picroft.


MatthewScholefield:
Would it work to instead do:

```bash
invoke_apt remove
sudo apt-get autoremove -y
```

That should remove all mycroft packages properly.

PR #1218, opened 2017-11-10
Handle missing messagebus in a correct way

forslund:
==== Tech notes ====
The send script will throw an IOerror exception if the bus service isn't
started. This correctly catches it and emits a single warning.


forslund:
Merging since it's been laying approved for some time.

PR #1230, opened 2017-11-16
Add send_email method to skills

MatthewScholefield:
Update: It now works for me.
### Usage
```Python
self.send_email('Title', '<b>content</b><br>here')
```

### Example Skill
**\_\_init__.py**
```Python
from mycroft import MycroftSkill, intent_file_handler


class EmailSkill(MycroftSkill):
    def __init__(self):
        MycroftSkill.__init__(self)

    @intent_file_handler('send.email.intent')
    def handle_send_email(self, message):
        self.send_email('Your email', message.data.get('content', '...'))
        self.speak_dialog('sent.email')


def create_skill():
    return EmailSkill()
```

**send.email.intent**
```
send an email saying {content}.
send the email {content}.
```

sent.email.dialog**
```
I've sent the email.
The email has been sent.
```

### DIFF
--- a/mycroft/api/__init__.py
+++ b/mycroft/api/__init__.py
@@ -162,6 +162,13 @@ class DeviceApi(Api):
                      "enclosureVersion": version.get("enclosureVersion")}
         })
 
+    def send_email(self, title, body, sender):
+        return self.request({
+            "method": "PUT",
+            "path": "/" + self.identity.uuid + "/message",
+            "json": {"title": title, "body": body, "sender": sender}
+        })
+
     def get(self):
         """ Retrieve all device information from the web backend """
         return self.request({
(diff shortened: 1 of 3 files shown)


forslund:
I've tested it and the call to the backend returns a 200 status but so far I've not received any e-mail.

Two things:
- Test case (at least for the api)
- Docstrings

MatthewScholefield:
@forslund I've added docstrings, and it should be fixed now. Also, how should we do a test case without actually sending an email?

forslund:
In this case the tests would be pretty easy. Mock out the `Api.request()` method and verify that it's called and the expected parameter is passed in the correct format. This would be mainly for protection against sloppy coders like me in the future :)

Looking at the possibilities here, you could do that for a mocked "valid" `IdentityManager.uuid` and an uninstantiated (which would be a `DeviceIdentity().uuid`)

MatthewScholefield:
@forslund I've added a unit test for the `send_email` part of `DeviceApi`, similar to `STT`.

PR #1234, opened 2017-11-17
Enhance CLI and screen corruption fix

penrods:
Several refinements:
* Remove the "What time is it" preloaded example
* Page up/down now moves by 1/2 the # visible log lines instead of always 10 lines
* Reduce the frequency of full-screen redraws to 10 secs instead of 5 (because less needed with the corruption fix)
* Add the version of mycroft-core in the upper-right corner
* FIX: screen redraw now uses a lock, preventing corruption from drawing simultaneously from multiple threads


MatthewScholefield:
The `oldest` marker seems to no longer be latched to the top. Is this intentional?

PR #1235, opened 2017-11-17
Remove pystache

MatthewScholefield:
On certain platforms it escapes quotes which causes problems. Since we only use it for something as simple as key value replacements, it doesn't make sense to include it as a dependency.

@forslund Any idea why this randomly started happening for me? It sounded very much like a Python 3 issue which is why I'm mentioning you. For me, the following is messed up:

`workon mycroft && python2`
```Python
import pystache
pystache.render('{{key}}', {'key': "'"})
```

### DIFF
--- a/mycroft/dialog/__init__.py
+++ b/mycroft/dialog/__init__.py
@@ -16,7 +16,6 @@ import random
 from io import open
 
 import os
-import pystache
 
 from mycroft.util import resolve_resource_file
 from mycroft.util.log import LOG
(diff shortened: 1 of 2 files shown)


forslund:
I have no good answer for you. I've seen this when I've been running under python3, according to this https://stackoverflow.com/questions/34829472/pystache-without-escaping-unescaped pystache should escape the characters unless specified not to with triple curly-braces `{{{key}}}`

```
>>> pystache.render('{{{key}}}', {'key': "'"})
u"'"
```

I wonder if it could be a file encoding issue? Or maybe something different I know I triggered it when playing with futures and switched to utf8-strings.

forslund:
Anyway, reducing number of dependencies are always a good idea.

forslund:
Tested and working as expected. Merging. (with the additional bonus that unparsed entries will remain in speech, might be an improvement for troubleshooting skills)

PR #1240, opened 2017-11-20
Translate German

gras64:
German recognizer files


forslund:
Thanks for this pull request.

I wonder could the dev_setup script be modified to download these instead of having them included in the git repository?

penrods:
I would prefer to support it that way, also.  I don't want the size of the repo to explode with a bunch of binary models.  Where did you obtain the PocketSphinx model files @gras64 ?  Is there a stable location where we can download the latest on demand?

gras64:
I have my files here:.
https://sourceforge.net/projects/cmusphinx/files/Acoustic%20and%20Language%20Models/German/cmusphinx-de-voxforge-5.2.tar.gz/download

and I have a bad script in the dev-setup.sh built. I hope I can help with that.

PR #1242, opened 2017-11-21
Improve mouth synchronization with speech

penrods:
When the TTS engine provides visemes to the faceplate, the information
passed along consists of the mouth shape and the duration to display it.
When the system get backed-up for some reason (e.g. the CPU is briefly
overloaded), the code would attempt to catch up the animation but would
still send the 'expired' viseme across the serial port to the faceplate
with no wait-time.  This results in a fast-moving mouth to catch up,
which isn't very pleasing.

Now the viseme is passed along with an expiration date, so if the time
to display it has already passed then the viseme code gets thrown away
instead of being sent across the (relatively slow) serial port.  This
allows better catch-up.


forslund:
Looks good, merging.

PR #1243, opened 2017-11-21
Fix dialog renderer

penrods:
The recent change to remove Pystache introduced a bug in cases where
the passed-in key/value data included non-strings for the value.  This
showed up in the weather skill which passed in "lat" and "lon" as floats.
Now the value is explicitly converted using str().


MatthewScholefield:
Do we need to worry about the key not being a string?

penrods:
Interesting -- in JSON the key has to be a string (https://www.json.org/) but I played with Python and it allows numeric keys.  So yup, I'll str(() that too.

Learn something ever day.

MatthewScholefield:
Lgtm

PR #1244, opened 2017-11-22
Add 'mycroft.mic.listen' messagebus command

penrods:
Add support for 'mycroft.mic.listen' on the messagebus to trigger the system
to listen for STT processing.  This can be posted on the messagebus by outside
systems, such as a physical or GUI Listen button.


penrods:
This was requested by @AIIX for the KDE plasmoid, but is generally useful for implementing hardware trigger buttons and other GUI triggers.

PR #1247, opened 2017-11-22
Add general metrics reporting

MatthewScholefield:
This can be used from skills to report arbitrary JSON metrics with `self.report_metric('usage-type', ['ask.time'])`. It can also be used internally via `from mycroft.metrics import report_metric; report_metric('metric_type', {'metric': 'json'})`.

**Note**: This relies on #1230 (causing the unrelated email commits).

### DIFF
--- a/mycroft/metrics/__init__.py
+++ b/mycroft/metrics/__init__.py
@@ -18,12 +18,23 @@ import time
 
 import requests
 
+from mycroft.api import DeviceApi
 from mycroft.configuration import Configuration
 from mycroft.session import SessionManager
 from mycroft.util.log import LOG
 from mycroft.util.setup_base import get_version
 
-config = Configuration.get().get('server')
+
+def report_metric(name, data):
+    """
+    Report a general metric to the Mycroft servers
+
+    Args:
+        name (str): Name of metric
+        data (dict): JSON dictionary to report. Must be valid JSON
+    """
+    if Configuration().get()['opt_in']:
+        DeviceApi().report_metric(name, data)
 
 
 class Stopwatch(object):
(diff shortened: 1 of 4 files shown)


mycroft/metrics/__init__.py
```diff
@@ -104,9 +109,10 @@ def publish():
 
 
 class MetricsPublisher(object):
-    def __init__(self, url=config.get("url"), enabled=config.get("metrics")):
-        self.url = url
-        self.enabled = enabled
+    def __init__(self, url=None, enabled=None):
```
forslund: Nitpicking:
I think `False` should be used instead of `None` the effect is the same but slightly more readable. (should be immutable just as None so no issue to have as default for the argument.

mycroft/skills/core.py
```diff
@@ -324,6 +327,21 @@ def converse(self, utterances, lang="en-us"):
         """
         return False
 
+    def report_metric(self, name, data):
+        """Report a skill metric to the Mycroft servers"""
```
forslund: Describe the arguments as well, that data should be a dict for example

mycroft/metrics/__init__.py
```diff
@@ -18,12 +18,17 @@
...
+from mycroft.api import DeviceApi
 from mycroft.configuration import Configuration
 from mycroft.session import SessionManager
 from mycroft.util.log import LOG
 from mycroft.util.setup_base import get_version
 
-config = Configuration.get().get('server')
+
+def report_metric(name, data):
+    """Report a general metric to the Mycroft servers"""
```
forslund: Nitpicking: describe arguments

MatthewScholefield:
Should be good now.

forslund:
Awesome! Merging!

PR #1250, opened 2017-11-22
Fix viseme bug from PR #1242

penrods:
I merged code and left out the 'until' value from the api call.  This
corrects that and deals with a missing 'until' cleaner.


forslund:
Looks good. Merging

PR #1255, opened 2017-11-23
Better support for creating distribution packages and having skills folders not at /opt/mycroft/skills

antlarr-suse:
These commits try to allow users to define their own skills folder so /opt/mycroft/skills is not hardcoded (althought will still be used by default). This will allow me to create openSUSE packages for mycroft so regular users can execute all mycroft services (and plasmoid).

### DIFF
--- a/msm/msm
+++ b/msm/msm
@@ -87,18 +87,23 @@ fi
 # Determine if on picroft/mk1?
 picroft_mk1="false"
 vwrap="true"
+use_virtualenvwrapper="$(get_config_value '.enclosure.use_virtualenvwrapper' 'true')"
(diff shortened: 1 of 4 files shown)


forslund:
I've only glanced over this yet but I think the event_scheduler storage file isn't covered here. See [mycroft/skills/event_scheduler.py](https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/skills/event_scheduler.py#L27)

JarbasAl:
jq seems to fail with commented json used in the config files, 

parse error: Invalid numeric literal at line 2, column 2

PR #1257, opened 2017-11-24
Added more utterances for misunderstood requests 

InconsolableCellist:
Not sure why there are so many commits shown in this PR, but there's only one file listed in the diff, and it's my changes: 

```
Previously a misunderstood command would result in Mycroft replying with
just one utterance; this rapidly became annoying and repetitive.

==== Localization Notes ====
Just like before, the new utterances need to be localized (existing TODO
in the code)
```


forslund:
Thanks for the PR I think it might be better to use the dialog system to handle this:

Seems like we prepared for that but it never was used:
edit `/mycroft/res/text/en-us/i didn't catch that.dialog` and (any other language) and add your strings

`import mycroft.dialog` at the top of the file, then replace

```python
data = {'utterance':
            "Sorry, I didn't catch that. Please rephrase your request."}
```

with

```python
data = {'utterance': mycroft.dialog.get('i didn't catch that.dialog')}
```

You can see it's used similarly used in [skills/main](https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/skills/main.py#L229)


To remove the extra commits you should rebase against the current upstream dev branch. It'll be a bit tricky since you seem to have made the changes in the dev branch (and not a feature branch). So if you like I can clean it up before merging.

penrods:
We actually have a pending Skill for this:  https://github.com/MycroftAI/mycroft-skills/pull/186

We will likely use this in the future to help identify and recommend skills.  You mind if I just close this PR?

PR #1259, opened 2017-11-26
Python 2/3 compatibility

forslund:
Adding python 2/3 compatibility to, speech client, audio, skills and messagebus services.

Most of the changes are very trivial, for example using the more modern exception syntax (`except Exception as e:` instead of `except Exception, e:`. I've been using http://python-future.org/compatible_idioms.html as a guide and tried to use the "best" solutions for compatibility.

The cli still misbehaves under python 3 and there is a name-space collision with signal when running the audiotest but I believe this is a good start and will be good to have in preparation for the move to python3.

### DIFF
--- a/mycroft/client/enclosure/__init__.py
+++ b/mycroft/client/enclosure/__init__.py
@@ -14,7 +14,7 @@
 #
 import subprocess
 import time
-from Queue import Queue
+import sys
 from alsaaudio import Mixer
 from threading import Thread, Timer
 
@@ -35,6 +35,10 @@ from mycroft.util import play_wav, create_signal, connected, \
     wait_while_speaking
 from mycroft.util.audio_test import record
 from mycroft.util.log import LOG
+if sys.version_info[0] < 3:
+    from Queue import Queue
+else:
+    from queue import Queue
 
 
 class EnclosureReader(Thread):
--- a/mycroft/client/speech/mic.py
+++ b/mycroft/client/speech/mic.py
@@ -455,7 +455,7 @@ class ResponsiveRecognizer(speech_recognition.Recognizer):
                     ww_module = self.wake_word_recognizer.__class__.__name__
 
                     ww = self.wake_word_name.replace(' ', '-')
-                    md = md5(ww_module).hexdigest()
+                    md = md5(ww_module.encode('utf-8')).hexdigest()
                     stamp = str(int(1000 * get_time()))
                     sid = SessionManager.get().session_id
                     aid = self.account_id
--- a/mycroft/client/speech/mic.py
+++ b/mycroft/client/speech/mic.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import sys
 import audioop
 import collections
 import datetime
@@ -134,6 +135,13 @@ class MutableMicrophone(Microphone):
         return self.muted
 
 
+def get_silence(num_bytes):
+    if sys.version_info[0] < 3:
+        return '\0' * num_bytes
+    else:
+        return bytes(num_bytes)
+
+
 class ResponsiveRecognizer(speech_recognition.Recognizer):
     # Padding of silence when feeding to pocketsphinx
     SILENCE_SEC = 0.01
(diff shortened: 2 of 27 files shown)


forslund:
I think the reported codacy issues can be ignored,

one is a reopening of an older issue (using md5) and the other is a redefinition of range (which is intended)

mycroft/client/enclosure/__init__.py
```diff
@@ -35,6 +35,10 @@
     wait_while_speaking
 from mycroft.util.audio_test import record
 from mycroft.util.log import LOG
+if sys.version_info[0] < 3:
```
MatthewScholefield: I think we should decide to either always check `sys.version_info` or always try and except.

mycroft/client/speech/mic.py
```diff
@@ -134,6 +135,13 @@ def is_muted(self):
         return self.muted
 
 
+def get_silence(num_bytes):
+    if sys.version_info[0] < 3:
+        return '\0' * num_bytes
+    else:
+        return bytes(num_bytes)
```
MatthewScholefield: We can support both in one line with `return b'\0' * num_bytes`

mycroft/messagebus/send.py
```diff
@@ -38,17 +38,17 @@ def main():
...
-        print "Usage: python -m mycroft.messagebus.send message"
-        print "       python -m mycroft.messagebus.send message JSON-string\n"
-        print "Examples: python -m mycroft.messagebus.send mycroft.wifi.start"
-        print "Ex: python -m mycroft.messagebus.send speak " \
-            "'{\"utterance\" : \"hello\"}'"
+        print("Command line interface to the mycroft-core messagebus.")
+        print("Usage: python -m mycroft.messagebus.send message")
+        print("       python -m mycroft.messagebus.send message JSON-string\n")
+        print("Examples: python -m mycroft.messagebus.send mycroft.wifi.start")
+        print("Ex: python -m mycroft.messagebus.send speak "
```
MatthewScholefield: For Python 2 compatibility here do we need to `from __future__ import print_function`?

mycroft/skills/intent_service.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 #
 import time
-
+from builtins import range
```
MatthewScholefield: Can we move this by `from past.builtins` and maybe make it explicitly `from future.builtins`?

mycroft/messagebus/send.py
forslund: Not when making python2 legacy code compatible with python3.

If we were backporting from python3 where python3 features had been used we'd need to use `print_function`.

mycroft/skills/intent_service.py
forslund: Sure, `future.builtins.range is builtins.range` evaluates to True so shouldn't be an issue

mycroft/client/enclosure/__init__.py
forslund: Good call. Will update to use sys.version... and not try/except

mycroft/messagebus/send.py
forslund: But I'll add it to be on the safe side

MatthewScholefield: Oh yeah, only really a problem with something like `print('var', var)`.

forslund:
I think that's all of the issues.

MatthewScholefield:
[Something's wrong with the skill reloading](https://hastebin.com/nuziqiniqi.rb) when I try it in Python 3.

forslund:
Oh good catch, it's probably the `__pycache__` directory. This probably happened the first time I ran with python3 as well.

I'll delete the directories and see if I can reproduce.

MatthewScholefield:
I couldn't find a start-mycroft-py3.sh script so I modified the scripts to do that. It could somehow be related. I pushed my changes to `python3-full`. Most notable, there was this python path code that I couldn't get to work well so I changed it to essentially `pip install -e .` so it makes symlinks of mycroft to the virtualenv rather than adding the folder to the virtualenv's path.

forslund:
I haven't included that in the PR since I don't think this PR since it's still a bit of a hack (manual installation of the virtualenv, etc), and I didn't want to do that since I don't think python 3 isn't a good user experience yet (and making bash scripts is hard!)

My intention here was to fix the little "easy" things so python 3 can be run (with some manual intervention) so we can start to debug the harder issues.

I'll try your setup and see if I can reproduce the issue with that.

forslund:
Hmm, still can't reproduce with the dev-setup-script you provided.

I used a clean `mycroft3` virtualenv (`mkvirtualenv -p /usr/bin/python3 mycroft3`) as base and had to manually install the requirements.txt to make the the script complete.

PR #1260, opened 2017-11-26
bing tts

JarbasAl:
add bing tts engine

### DIFF
--- a/mycroft/tts/bing_tts.py
+++ b/mycroft/tts/bing_tts.py
@@ -25,7 +25,7 @@ class BingTTS(TTS):
         self.config = Configuration.get().get("tts", {}).get("bing", {})
         api = self.config.get("api_key")
         self.bing = Translator(api)
-        self.gender = self.config.get("gender", "male")
+        self.gender = self.config.get("gender", "Male")
         self.format = self.config.get("format", "riff-16khz-16bit-mono-pcm")
 
     def get_tts(self, sentence, wav_file):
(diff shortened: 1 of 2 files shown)


penrods:
Main question about this one is with regards to the bingtts requirements.  I started a similar discussion in #1262, so let's just about it there.

forslund:
Tested this and ran into some trouble. I couldn't get the bingtts module to work at all with my access key.

Using the sample script from Microsoft, the access key works fine: (https://github.com/Azure-Samples/Cognitive-Speech-TTS/blob/master/Samples-Http/Python/TTSSample.py)

forslund:
Figured it out, the gender value is case sensitive.

mycroft/tts/bing_tts.py
```diff
@@ -0,0 +1,60 @@
...
+
+class BingTTS(TTS):
+    def __init__(self, lang, voice):
+        super(BingTTS, self).__init__(lang, voice, BingTTSValidator(self))
+        self.type = 'wav'
+        from bingtts import Translator
+        self.config = Configuration.get().get("tts", {}).get("bing", {})
+        api = self.config.get("api_key")
+        self.bing = Translator(api)
+        self.gender = self.config.get("gender", "male")
```
forslund: gender is case sensitive and need to be "Male"

forslund:
Looks good, thank you! Merging!

PR #1261, opened 2017-11-26
ibm watson tts

JarbasAl:
add ibm watson tts engine support

### DIFF
--- a/mycroft/tts/__init__.py
+++ b/mycroft/tts/__init__.py
@@ -333,14 +333,15 @@ class TTSFactory(object):
     from mycroft.tts.mary_tts import MaryTTS
     from mycroft.tts.mimic_tts import Mimic
     from mycroft.tts.spdsay_tts import SpdSay
-
+    from mycroft.tts.ibm_tts import WatsonTTS
     CLASSES = {
         "mimic": Mimic,
         "google": GoogleTTS,
         "marytts": MaryTTS,
         "fatts": FATTS,
         "espeak": ESpeak,
-        "spdsay": SpdSay
+        "spdsay": SpdSay,
+        "watson": WatsonTTS
     }
 
     @staticmethod
--- a/mycroft/tts/ibm_tts.py
+++ b/mycroft/tts/ibm_tts.py
@@ -1,0 +1,55 @@
+# Copyright 2017 Mycroft AI Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from mycroft.tts import TTSValidator
+from mycroft.tts.remote_tts import RemoteTTS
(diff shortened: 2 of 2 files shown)


mycroft/tts/__init__.py
```diff
@@ -333,14 +333,15 @@ class TTSFactory(object):
     from mycroft.tts.mary_tts import MaryTTS
     from mycroft.tts.mimic_tts import Mimic
     from mycroft.tts.spdsay_tts import SpdSay
-
+    from mycroft.tts.ibm_tts import WatsonTTS
```
penrods: The means WatsonTTS should be added to the requirements.txt, correct?

mycroft/tts/ibm_tts.py
```diff
@@ -0,0 +1,55 @@
...
+class WatsonTTSValidator(TTSValidator):
+    def __init__(self, tts):
+        super(WatsonTTSValidator, self).__init__(tts)
+
+    def validate_lang(self):
+        # TODO
+        pass
+
+    def validate_connection(self):
+        # TODO
```
penrods: Is there any API you can hit on Watson as a validator?  Most of the other TTS just do a quick call to a "getversion" of some sort, throwing an error if there is no response to that.

mycroft/tts/__init__.py
JarbasAI: hum no, WatsonTTS is the engine class i made, only requirements is requests, i used the standard remoteTTS class available

mycroft/tts/ibm_tts.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def __init__(self, lang, voice="en-US_AllisonVoice",
+                 url="https://stream.watsonplatform.net/text-to-speech/api"):
+        super(WatsonTTS, self).__init__(lang, voice, url, '/v1/synthesize',
+                                        WatsonTTSValidator(self))
+        self.type = "wav"
+        self.config = Configuration.get().get("tts", {}).get("watson", {})
+        user = self.config.get("user")
+        password = self.config.get("password")
+        self.auth = (user, password)
```
forslund: I'd like this to be `self.auth = HTTPBasicAuth(user, password)` and change the `self.session.get()` call in `__request` in `remote_tts.py` to send `auth=self.auth`

mycroft/tts/ibm_tts.py
```diff
@@ -0,0 +1,55 @@
...
+class WatsonTTS(RemoteTTS):
+    PARAMS = {'accept': 'audio/wav'}
+
+    def __init__(self, lang, voice="en-US_AllisonVoice",
+                 url="https://stream.watsonplatform.net/text-to-speech/api"):
+        super(WatsonTTS, self).__init__(lang, voice, url, '/v1/synthesize',
+                                        WatsonTTSValidator(self))
+        self.type = "wav"
+        self.config = Configuration.get().get("tts", {}).get("watson", {})
+        user = self.config.get("user")
```
forslund: I wonder if we should look for "username" as well since then the credentials from Watson can just be copy-pasted into the config.

mycroft/tts/ibm_tts.py
```diff
@@ -0,0 +1,55 @@
...
+        params['text'] = sentence.encode('utf-8')
+        return params
+
+
+class WatsonTTSValidator(TTSValidator):
+    def __init__(self, tts):
+        super(WatsonTTSValidator, self).__init__(tts)
+
+    def validate_lang(self):
+        # TODO
```
forslund: Should at least verify that the config exists (url, username, password) since those are all needed for this to work.

forslund:
Thanks @JarbasAI, merging.

PR #1263, opened 2017-11-27
Only emit start/stop messages for skill handlers by default (not all event handlers)

forslund:
When creating event handlers with add_event now by default there are no messages about start and completion of the handler.

The handler info base name is now passed as an argument to the method allowing for custom messages. skill intent handlers still report skill.handler.start and skill.handler.complete but handlers created by skill creators using `add_event` and scheduled events wont send these start/stop by default (but could instead trigger for example skill.scheduled_event.start/complete)

This makes it easier to separate skill handlers status reports on the message bus and would make metrics easier to collect.

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -558,26 +558,31 @@ class MycroftSkill(object):
             text = f.read().replace('{{', '{').replace('}}', '}')
             return text.format(**data or {}).split('\n')
 
-    def add_event(self, name, handler, need_self=False, once=False):
+    def add_event(self, name, handler, need_self=False,
+                  handler_info=None, once=False):
         """
             Create event handler for executing intent
 
             Args:
-                name:       IntentParser name
-                handler:    method to call
-                need_self:  optional parameter, when called from a decorated
-                            intent handler the function will need the self
-                            variable passed as well.
-                once:       optional parameter, Event handler will be removed
-                            after it has been run once.
+                name:           IntentParser name
+                handler:        method to call
+                need_self:      optional parameter, when called from a
+                                decorated intent handler the function will
+                                need the self variable passed as well.
+                once:           optional parameter, Event handler will be
+                                removed after it has been run once.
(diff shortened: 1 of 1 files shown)


mycroft/skills/core.py
```diff
@@ -275,11 +277,23 @@ def settings(self):
...
+    def emitter(self):
+        """ Backwards compatibility. """
+        LOG.info('self.emitter is depreciated, use self.messagebus instead')
+        return self.messagebus
+
+    @emitter.setter
+    def emitter(self, value):
+        """ Backwards compatibility. """
+        LOG.info('self.emitter is depreciated, use self.messagebus instead')
+        self.messagebus = value
```
augustnmonteiro: in this case wouldn't be self.mesagebus.emit(value) ?

forslund: This is to fake the set operation

```python

skill.emitter = test_emitter
```

Would throw an exception if the setter method didn't exist. This makes it so that anything that tries to set `.emitter` actually sets `.messagebus`

MatthewScholefield:
Why rename to `messagebus`? We already use `ws` and `emitter` in other places so if we want to change to that we should make sure to do it everywhere to stay consistent.

forslund:
It was something @penrods mentioned in chat a while back. That messagebus would make a more intuitive name for the emitter in the skill. As it was sort of related to this I felt it was a good starting point.

Changing it globally is out of scope for this PR so I'll revert the name change.

forslund:
It feels a bit late to include this one in the release so I'm untagging it.

penrods:
@forslund and @MatthewScholefield -- is this good to merge now?

MatthewScholefield:
Lgtm

forslund:
Will rebase.

forslund:
Rebased. The coverage test fails since the PR was rebased to include the corrected coverage.

forslund:
Ping @MatthewScholefield

PR #1264, opened 2017-11-27
allow phonemes in any lang

JarbasAl:
info:

wakewords now work even if the main language is changed, as long as the phonemes are accurate everything should just work

changes made:

This allows hotword engines to explicitly request a language in config file

solves:

English phonemes may not exist in other language, which would break wakeword when changing main language


forslund:
Looks good and useful. Could you remove line 37 and change line 46 to 

```python
        self.lang = str(self.config.get("lang", str(lang))).lower()
```

forslund:
Looks good, merging.

PR #1266, opened 2017-11-28
Allow spaces around key in dialog template file

forslund:
pystache allowed an array of different variations on the dialog _key_ syntax

{{key}}
{{    key   }}
{{{ key }}}

This was lost in the latest change which causes some skills to fail.

This adds a regular expression to handle more of these cases correctly (all of above works fine).

The issue was originally reported by *tjoen* in the mattermost chat.

### DIFF
--- a/mycroft/dialog/__init__.py
+++ b/mycroft/dialog/__init__.py
@@ -16,6 +16,7 @@ import random
 from io import open
 
 import os
+import re
 
 from mycroft.util import resolve_resource_file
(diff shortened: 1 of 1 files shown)


JarbasAl:
does this also apply to padatious {entity} ?

penrods:
@JarbasAI no, the Pystache library was only used by the dialog rendering, nothing around Padatious.

I'm merging this since it was approved.

PR #1271, opened 2017-11-30
Make reference leak detection more reliable

forslund:
The reference count sometimes reported that references were remaining
even if they actually weren't. By enforcing a garbage collector run
after shutting down a skill the false positives are reduced if not
removed all together.

As this is mainly for developers and can reduce performance it is now only active if a "debug" config value is set.

### DIFF
--- a/mycroft/skills/main.py
+++ b/mycroft/skills/main.py
@@ -42,6 +42,7 @@ ws = None
 event_scheduler = None
 skill_manager = None
 
+DEBUG = Configuration.get().get("debug", False)
 skills_config = Configuration.get().get("skills")
 BLACKLISTED_SKILLS = skills_config.get("blacklisted_skills", [])
 PRIORITY_SKILLS = skills_config.get("priority_skills", [])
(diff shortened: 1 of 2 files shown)


mycroft/skills/main.py
```diff
@@ -41,6 +42,7 @@
 event_scheduler = None
 skill_manager = None
 
+DEBUG = Configuration.get().get("debug", False)
```
MatthewScholefield: Shouldn't we add this to `mycroft.conf`?

mycroft/skills/main.py
```diff
@@ -278,14 +280,16 @@ def _load_or_reload_skill(self, skill_folder):
...
 
-            # Remove two local references that are known
-            refs = sys.getrefcount(skill["instance"]) - 2
-            if refs > 0:
-                LOG.warning(
-                    "After shutdown of {} there are still "
-                    "{} references remaining. The skill "
-                    "won't be cleaned from memory."
-                    .format(skill['instance'].name, refs))
+            if DEBUG:
```
MatthewScholefield: Just to clarify, `gc.collect()` is normally automatically run, right? So all this does is make sure it's run right before checking references?

forslund: Yes that's precisely it.

mycroft/skills/main.py
forslund: Ok, mycroft.conf is now updated. You're probably right that it's good to have in the config as a reference.

JarbasAl:
i tried that before and there were still references left, didnt see a difference using gc.collect(), but i think it was because of unremoved listeners, need to check it to be sure and will give feedback

PR #1272, opened 2017-11-30
Correct test coverage

forslund:
Previously the coverage counted the entire mycroft core including things
such as the docs and the tests themself. The inclusion of the actual
tests in the coverage metric scewed the metric quite a bit and the
inclusion of irrelevant files made the information presented by
coveralls cluttered.


forslund:
The drop in coverage is OK since it's more true.

penrods:
LGTM

PR #1273, opened 2017-11-30
Attempt to allow CLI to display utf-8 encoded text.

forslund:
If default locale supports utf-8 re-encode strings to utf8.

I'm not sure how much of a hack this is but it seems to work.

### DIFF
--- a/mycroft/client/text/main.py
+++ b/mycroft/client/text/main.py
@@ -37,6 +37,11 @@ from mycroft.messagebus.message import Message              # nopep8
 from mycroft.util import get_ipc_directory                  # nopep8
 from mycroft.util.log import LOG                            # nopep8
 
+import locale                                               # nopep8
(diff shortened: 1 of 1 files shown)


penrods:
I attempted to test on a Pi3 but using two terminals.  One was running mycroft-cli-client, the other a nano session on the /var/log/mycroft-speech-client.log.  I pasted the word "cédille" into the nano terminal as a new line and saved the file.  It didn't load/translate it correctly.

This system was set to locale UTF-8, 'en-GB'.  So not 100% certain how this _should_ behave.

yleny-dev:
For the patch, https://github.com/MycroftAI/mycroft-core/pull/1273/commits/8254c1778cdf9fd82ecd09f34bc5707a9ca573a2 (Attempt to allow CLI to display utf-8 encoded text. )
When I patch the latest Mycroft-core source code with it and I type command in the cli, the cli crash.
Python 2.7.13

PR #1274, opened 2017-11-30
Added a new feature to... nevermind

penrods:
The utterance processing now looks at the tail end of the transcribed phrase, ignoring the utterance if it ends with "nevermind" or "cancel" or "ignore that".  This is for use case like:
   Hey Mycroft, can you tell me the...ummm...oh, nevermind

I have some concerns about a few corner cases that a skill might want to use.
For example:
 * Hey Mycroft, play Nirvana's Nevermind
 * Hey Mycroft, I want that last request to cancel
But I think this is something that can normally be designed around easily.

====  Tech Notes ====
Added mycroft.dialog.get_all(phrase, lang, context).  This lets you load all of the rendered versions, not just a single randomly selected one.

This class has the experiment of returning a value when a matching .dialog file isn't found.  Instead it returns ['phrase'], replacing the . with spaces.  So if I add:
    mycroft.dialog.get_all("thank.you") and there is no resource file found named 'thank.you.dialog', it will still return successfully:  ['thank you']

==== Localization Notes ====
This adds the new "nevermind" core resource string.

### DIFF
--- a/mycroft/dialog/__init__.py
+++ b/mycroft/dialog/__init__.py
@@ -12,23 +12,30 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import os
 import random
 from io import open
-
-import os
 
 from mycroft.util import resolve_resource_file
 from mycroft.util.log import LOG
 
 
-__doc__ = """
-
-"""
-
-
 class MustacheDialogRenderer(object):
     """
-    A dialog template renderer based on the mustache templating language.
+    A dialog template renderer based on the {{mustache}} templating language.
+    
+    The {{mustache}} language is a simple scheme that supports simple,
+    position-independent templating.  We use this same basic concept, but only
+    implement the variable replacement mechanism -- no HTML quoting with triple
+    brackets, sections, etc.
+
+    A difference between this and typical mustache mechanism is output
+    randomization. Only a single line is ever returned.  If the template file
+    contains multiple lines, one is randomly selected (unless a specific one
+    is requested).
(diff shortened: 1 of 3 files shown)


JarbasAl:
maybe having a convention for always ending with "that", it sounds natural and should reduce collisions. the examples you gave would work

i think cancel and nevermind are too generic, "cancel that", "nevermind that", "ignore that" and maybe also "X that order" or "X this order"

these sound natural, until we get a more robust approach it would be better to "train the user" to interact in a certain way than mysteriously failing here and there.

penrods:
Can you provide some good collision examples?  Remember that this only matches the very end of the transcription, so you can say "Hey Mycroft, cancel my appointment with Jimbo" without any issue.  I am stuck coming up with real-world phrases that people might attempt to use that this would prevent.  It feels like "cancel" might cause conflicts, but I can't come up with realistic phrases that might end in 'nevermind'.  If you can give me some potential phrases, I could be swayed.

I agree that this is a "train the user" solution for now.

JarbasAI:
@penrods 

What is another word for nevermind?
What is the opposite of nevermind?
What is the meaning of the word nevermind?
How do you pronounce the word nevermind?

i would train the user to always end with X "that" "this" or "that order"

JarbasAI:
what if you checked for neverminds after determining intent but before message broadcast and allow individual intents to disable this setting?

penrods:
I think this one is of limited utility, so closing for now.

PR #1276, opened 2017-12-01
Fix skill metrics

MatthewScholefield:
Otherwise, skill metric reporting always errors out because of an invalid endpoint.


forslund:
Seems to work as I would expect. Merging.

PR #1278, opened 2017-12-04
Add get_response method as a synchronous alternative to converse

MatthewScholefield:
Documentation:
```python
def get_response(self, dialog, data=None):
    """
    Ask and return a response from the user for the given dialog

    Args:
        dialog (str): Dialog file to read to the user
        data (str): data used to render the dialog
    Returns:
        any: None or transcription user's reply
    """
    pass
```

Example usage:
```python
color = self.get_response('what.is.your.favorite.color')
if color:
    self.speak('My favorite color is also ' + color)
```

### DIFF
--- a/mycroft/client/speech/listener.py
+++ b/mycroft/client/speech/listener.py
@@ -152,6 +152,7 @@ class AudioConsumer(Thread):
                 text = "pair my device"  # phrase to start the pairing process
                 LOG.warning("Access Denied at mycroft.ai")
         except Exception as e:
+            self.emitter.emit('recognizer_loop:speech.recognition.unknown')
             LOG.error(e)
             LOG.error("Speech Recognition could not understand audio")
         if text:
--- a/mycroft/client/speech/main.py
+++ b/mycroft/client/speech/main.py
@@ -56,6 +56,10 @@ def handle_utterance(event):
     ws.emit(Message('recognizer_loop:utterance', event))
 
 
+def handle_unknown():
+    ws.emit(Message('mycroft.speech.recognition.unknown'))
+
+
 def handle_speak(event):
     """
         Forward speak message to message bus.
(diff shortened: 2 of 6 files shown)


mycroft/client/speech/main.py
```diff
@@ -56,6 +56,10 @@ def handle_utterance(event):
     ws.emit(Message('recognizer_loop:utterance', event))
 
 
+def handle_unknown():
+    ws.emit(Message('speech.recognition.unknown'))
```
forslund: @penrods suggested using recognizer_loop:x internally in the speech client and using mycroft.x.y on the interprocess message bus.

MatthewScholefield: OK, this has been changed to `recognizer_loop:speech.recognition.unknown` and `mycroft.speech.recognition.unknown`. Let me know if any other name sounds better.

mycroft/client/speech/listener.py
```diff
@@ -152,6 +152,7 @@ def transcribe(self, audio):
                 text = "pair my device"  # phrase to start the pairing process
                 LOG.warning("Access Denied at mycroft.ai")
         except Exception as e:
+            self.emitter.emit('recognizer_loop:peech.recognition.unknown')
```
penrods: Is "peech" intentional?

mycroft/client/speech/main.py
```diff
@@ -56,6 +56,10 @@ def handle_utterance(event):
     ws.emit(Message('recognizer_loop:utterance', event))
 
 
+def handle_unknown():
+    ws.emit(Message('mycroft.mycspeech.recognition.unknown'))
```
penrods: Is "mycspeech" intentional?  Looks like a search/replace error.

MatthewScholefield: Lol, no idea how I failed at typing that much. Fixed now.

mycroft/skills/core.py
```diff
@@ -320,13 +322,121 @@ def converse(self, utterances, lang="en-us"):
...
+
+        # TODO: Load with something like mycroft.dialog.get_all()
+        cancel_voc = 'text/' + self.lang + '/cancel.voc'
+        with open(resolve_resource_file(cancel_voc)) as f:
+            cancel_words = list(filter(bool, f.read().split('\n')))
+
+        def is_cancel(utterance):
+            return utterance in cancel_words
+
+        def on_valid_default(utterance):
```
MatthewScholefield: Sounds a bit funny since the original parameter isn't called `on_valid`.

PR #1279, opened 2017-12-05
Add mycroft.awoken message when listener is awoken

forslund:
Instead of speaking directly from the listener send a message that the
naptime_skill can use to trigger speech and/or other indications that
the listener is now awake.

### DIFF
--- a/mycroft/client/speech/listener.py
+++ b/mycroft/client/speech/listener.py
@@ -114,7 +114,7 @@ class AudioConsumer(Thread):
         if self.wakeup_recognizer.found_wake_word(audio.frame_data):
             SessionManager.touch()
             self.state.sleeping = False
-            self.__speak(mycroft.dialog.get("i am awake", self.stt.lang))
+            self.emitter.emit('recognizer_loop:awoken')
             self.metrics.increment("mycroft.wakeup")
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


mycroft/client/speech/listener.py
```diff
@@ -114,7 +114,7 @@ def wake_up(self, audio):
         if self.wakeup_recognizer.found_wake_word(audio.frame_data):
             SessionManager.touch()
             self.state.sleeping = False
-            self.__speak(mycroft.dialog.get("i am awake", self.stt.lang))
+            self.emitter.emit('mycroft.awoken')
```
penrods: Within the listener bus, use "recognizer_loop:awoken"

mycroft/client/speech/main.py
```diff
@@ -135,6 +141,7 @@ def main():
     loop.on('recognizer_loop:utterance', handle_utterance)
     loop.on('speak', handle_speak)
     loop.on('recognizer_loop:record_begin', handle_record_begin)
+    loop.on('mycroft.awoken', handle_awoken)
```
penrods: Switch to:  "recognizer_loop:awoken"

forslund:
PR updated according to comments

PR #1280, opened 2017-12-05
Add support for unnamed intents

penrods:
This allows skill writers to ignore naming intents.  Combined with a
forthcoming change to Adapt that creates a default of None for IntentBuilder()

This allows the current:
    @intent_handler(IntentBuilder("CurrentWeatherIntent").require(
        "Weather").optionally("Location").build())
    def handle_current_weather(self, message):
        ...

To become:
    @intent_handler(IntentBuilder().require("Weather").optionally("Location"))
    def handle_current_weather(self, message):
        ...

Which will automatically name the Intent "handle_current_weather".

Also dropped the log message in the default initialize() method since it is
common to not override it now.

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -295,11 +295,11 @@ class MycroftSkill(object):
 
     def initialize(self):
         """
-        Initialization function to be implemented by all Skills.
+        Initialization function, run after fully constructed
 
         Usually used to create intents rules and register them.
         """
-        LOG.debug("No initialize function implemented")
+        pass
 
     def get_intro_message(self):
         """
@@ -458,6 +458,10 @@ class MycroftSkill(object):
             intent_parser = intent_parser.build()
         elif type(intent_parser) != Intent:
             raise ValueError('intent_parser is not an Intent')
+
+        if not intent_parser.name:
+            # Default to the handler's function name if None or ""
+            intent_parser.name = handler.__name__
 
         name = intent_parser.name
         intent_parser.name = str(self.skill_id) + ':' + intent_parser.name
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -400,6 +400,8 @@ class MycroftSkill(object):
                         except TypeError:
                             handler(self)
                     else:
+                        LOG.error("Unexpected argument count:" +
+                                  str(len(getargspec(handler).args)))
                         raise TypeError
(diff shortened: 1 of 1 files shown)


mycroft/skills/core.py
```diff
@@ -295,11 +295,11 @@ def detach(self):
 
     def initialize(self):
         """
-        Initialization function to be implemented by all Skills.
+        Initialization function, run after fully constructed
```
forslund: We might want to clarify this even more. Since the current use is to register handlers for the message bus and run methods requiring the messagebus (register entities with padatious, etc)

_Initialization function, will be run after skill is fully constructed and registered with the mycroft messagebus._

mycroft/skills/core.py
```diff
@@ -459,6 +459,10 @@ def register_intent(self, intent_parser, handler, need_self=False):
         elif type(intent_parser) != Intent:
             raise ValueError('intent_parser is not an Intent')
 
+        if not intent_parser.name:
+            # Default to the handler's function name if None or ""
+            intent_parser.name = handler.__name__
+
         name = intent_parser.name
         intent_parser.name = str(self.skill_id) + ':' + intent_parser.name
```
forslund: This could be a bit more compact and just as easy to understand:

```python
        # Default to the handler's function name if None or ""
        name = intent_parser.name or handler.__name__
        intent_parser.name = str(self.skill_id) + ':' + name
```

mycroft/skills/core.py
penrods: Good call, I reworked the comments altogether.

mycroft/skills/core.py
penrods: Done.  You are so Pythonic.  ;-)

mycroft/skills/core.py
```diff
@@ -459,7 +458,9 @@ def register_intent(self, intent_parser, handler, need_self=False):
         elif type(intent_parser) != Intent:
             raise ValueError('intent_parser is not an Intent')
 
-        name = intent_parser.name
+        # Default to the handler's function name if none given
+        name = intent_parser.name or handler.__name__
+
         intent_parser.name = str(self.skill_id) + ':' + intent_parser.name
```
forslund: this should be

```python
intent_parser.name = str(self.skill_id) + ':' + name
```

after the previous change

PR #1284, opened 2017-12-06
Tweaks for Command Line Interface (CLI)

penrods:
Several minor changes to the CLI:
* Placed lock on screen refresh from a new log message
* Slowed frequency of mic updates to limit opportunity for screen corruption.
  Still happening occasionally, which is vexing.
* Up/Down arrow now scroll logs by a single line
* Ctrl+P/Ctrl+N (Previous/Next) now scroll through the command history
* Ensured that the "Oldest" line is always the first in the log


forslund:
Looks good.

However I find it a bit confusing that the scroll direction of the single line (arrow keys) and the page scroll (page up/down) are different.

On my laptop (and my previous) the UP shares key with PG-UP and DOWN shares with PG DOWN which might make this feel even weirder.

penrods:
Whoa, the PgUp/PgDn behavior is supposed to be flipped.  I have tried this on different machines and apparently didn't get that correct in this code I'm committing.  I'll update!

penrods:
I corrected the behavior, give 'er another shot when you can @forslund

forslund:
Behaves as expected now! Merging

PR #1287, opened 2017-12-07
Add phonetic spelling option to TTS engines

MatthewScholefield:
This can be used to resolve common pronunciation issues with TTS engines. Any engine can pass a parameter to the base constructor to disable this.


forslund:
27 days later: Looks good, merging!

PR #1290, opened 2017-12-07
Global Sync for skill settings

Mn0491:
====  Tech Notes ====
If there are multiple devices with the same skill, the devices will now automatically share the settings set in home.mycroft.ai. If the settingsmeta.json changes for one device, then there will be an additional settings box for that skill in home.mycroft.ai until they other skills are updated to the same settingsmeta.json

You can also make changes to the settings and it will push it up to the web UI if the device is the creator of the skill. Eventually, we will want any device to be able to change settings if they are linked to the specific skill settings but the backend currently does not allow that.

### DIFF
--- a/mycroft/skills/settings.py
+++ b/mycroft/skills/settings.py
@@ -18,6 +18,18 @@
     system to provide a GUI interface, described by meta-data described in
     an optional 'settingsmeta.json' file.
 
+    As it stands today, if you have multiple devices with the same skill
+    installed, they will share the same settings if the settings is can be
+    configurable on the web (anything sent by settingsmeta.json). If a skill
+    modify a setting on the client that can be configured in the web, the web
+    will update with the settings modified from the skill. Currently ONLY
+    the creator of the skill can modify the the web settings from the skill
+    it self. The creator is defined as the device that is the first one to
+    upload the settingsmeta.json to the web ui. This is temporary until
+    some mechanism are put into place in the backend to allow any device to
+    modify the settings configurable from the web.
+
+
     Usage Example:
         from mycroft.skill.settings import SkillSettings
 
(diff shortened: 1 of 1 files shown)


mycroft/skills/settings.py
```diff
@@ -302,24 +412,56 @@ def _put_metadata(self, settings_meta):
...
             with open(self._settings_path, 'w') as f:
                 json.dump(self, f)
             self.loaded_hash = hash(str(self))
+
+        if self._should_upload_from_change:
+            settings_meta = self._load_settings_meta()
+            hashed_meta = self._get_meta_hash(settings_meta)
+            uuid = self._load_uuid()
+            if uuid is not None:
+                LOG.info("deleting meata data for {}".format(self.name))
```
forslund: meata data: very protein rich data :)

mycroft/skills/settings.py
```diff
@@ -228,6 +290,18 @@ def _is_new_hash(self, hashed_meta):
...
 
+    def update_remote(self):
+        """ update settings state from server """
+        skills_settings = None
+        if self.get('is_synced'):
+            skills_settings = self._request_other_settings(hashed_meta)
+        if not skills_settings:
+            skills_settings = self._request_my_settings(hashed_meta)
+
+        if skills_settings is not None:
```
forslund: could be simplified to
`else:`

mycroft/skills/settings.py
```diff
@@ -67,41 +70,36 @@ def __init__(self, directory, name):
...
-                    LOG.info("seems like it got deleted from home... " +
-                             "sending settingsmeta.json for " +
-                             "{}".format(self.name))
-                    new_uuid = self._send_settings_meta(
-                        settings_meta, hashed_meta)
-                    self._save_uuid(new_uuid)
-                    self._save_hash(hashed_meta)
+                # first look at all other devices on user account to see
+                # if the settings exist. if it does then sync with this device
+                if skill_settings is not None:
```
forslund: could maybe be
`if skill_settings:`

mycroft/skills/settings.py
Mn0491: The control flow here is actually needed. the first if statement checks for skill settings that belong to other devices it can sync to. It can return None. the second if statements check to see if the skill settings exist for itself. It can also return None. if both returns None then don't save the settings

forslund: Right right, so it could be `if skill_settings`

PR #1291, opened 2017-12-07
Fix lock around speech.

forslund:
the lock could be taken by a waiting thread between sentences in a multi-sentenced utterance. This locking method will allow the entire utterance to be synthezised before handling next.

### DIFF
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -46,9 +46,6 @@ def handle_speak(event):
     global _last_stop_signal
 
     with lock:
-        # Mild abuse of the signal system to allow other processes to detect
-        # when TTS is happening.  See mycroft.util.is_speaking()
-
         utterance = event.data['utterance']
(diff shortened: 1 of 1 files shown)


mycroft/audio/speech.py
```diff
@@ -45,38 +45,40 @@ def handle_speak(event):
...
-            except KeyboardInterrupt:
-                raise
-            except:
-                LOG.error('Error in mute_and_speak', exc_info=True)
-            if _last_stop_signal > start or check_for_signal('buttonPress'):
-                break
-    else:
-        mute_and_speak(utterance)
+    with lock:
+        # Mild abuse of the signal system to allow other processes to detect
```
MatthewScholefield: These comments should probably be kept above next to `_last_stop_signal`

forslund: Maybe they should be removed completely? they're not really related to anything here. The isSpeaking signal is now set in the tts.execute()

PR #1293, opened 2017-12-08
Handle failure in _save_uuid

forslund:
_save_uuid() may return None on error. This needs to be caught to prevent skill from failing.


Mn0491:
LGTM

PR #1294, opened 2017-12-08
added missing variable fixes

Mn0491:
update_remote had a couple missing variables.


forslund:
Looks good to me.

A question:

does the settingsmeta file always exist when the _load_settings_meta() method is called. if not the method could trigger an exception

Mn0491:
@forslund  good call, i can throw something in for that

forslund:
Nice! Merging

PR #1296, opened 2017-12-08
Bugfix/polling fixes

Mn0491:
have a try catch block to polling if the internet is lost. also moved some code into a function to keep code DRY. added a check to remove now_owner flag is an upload is a success because if an upload is successful then the skill becomes the new owner.


Mn0491:
it will only do the call if there is a settingsmeta.json and it does on load

forslund:
Why doesn't a settingsmeta exist on load?

PR #1300, opened 2017-12-10
Protect against bad settingsmeta.json

penrods:
When there is an error in settingsmeta.json, the load of the skill
would fail.  Now it generates an error message but continues on.
Additionally the exception is caught and now displays information
about where the error in the JSON is (line and column).


penrods:
Will this conflict with #1300 ?  Shall we merge?

PR #1301, opened 2017-12-12
fix pt-pt unicode troubles

JarbasAI:
Fix the unicode problems in the pt-pt parsing and unittests


JarbasAI:
pep8 bot, you missed a few!

forslund:
Thanks @JarbasAI! Merging!

PR #1302, opened 2017-12-12
fix _hack_check_for_duplicates

JarbasAI:
minor fix, the self. was missing in this bit of code, however this is a hack for backwards compatibility that maybe could be removed instead of syntax fixed?


forslund:
Nice find. This may be good to keep in for now since people have been told to hold off on using their mark-1's until 28 of Februari 2018, so there will be a bunch of units that has the issue this hack fixes coming online still.

PR #1304, opened 2017-12-12
Feature/overwritten settings

Mn0491:
values in settings.json were being overwritten if skills initialize settings values in `__init__`. This will prevent that from happening if the value already exists in settings.json.


forslund:
Can you do a rebase against dev to clean up the commit history here?

Am I correct in my understanding that a setting set in `__init__` of a skill will be overwritten after __init__ is over?

Could we instead have an `allow_overwrite` flag that's initialized to `False` and the `__setitem__` only performs the action if the key doesn't exist or `allow_overwrite` is True. Something like
```
    def __setitem__(self, key, value):
        if `self.allow_overwrite or key not in self`
            return super(SkillSettings, self).__setitem__(key, value)
```

This would allow people who wants to reset variables in `__init__` to do so by setting the flag manually.

In any case this behaviour is pretty nonstandard and needs to be well documented.

penrods:
Tossing in my $0.02 on this PR:

We started this work after this fairly standard use case occurred (pseudo code):

class MySkill:
      self.setting['somesetting'] = 'default default'

"somesetting" gets initialized the very first time the skill is run.  If it gets modified during the course of the lifetime of the skill, that modification is serialized and the next time you reboot the value of 'somesetting' is NOT the 'default default' -- it is loaded from the disk instead after the init.

So far, that seems alright.  But if later on I decide to expose that value via a webUI and create a settingsmeta.json that includes 'somesetting', this behavior changed.  My initialization in the constructor began wiping out the saved setting.

forslund:
I understand the issue here and I see the value in this. I'm just proposing a different implementation.

penrods:
After this change, the behavior becomes:

```
class MySkill(MycroftSkill):
    def __init__(self):
        # This means you are creating a persisted value and
        # the initialization is for the beginning of the life of
        # this value -- which extends across reboots.
        self.setting['somesetting'] = 'default default'

    def initialize(self):
        # During initialize() timeframe (or after) you can make changes
        # to the setting value.
        if is_monday():
            self.setting['somesetting'] = 'reset'
```

I think this allows you a skill writer to achieve anything the want, although we might need to document this better.

forslund:
Yes, I agree. Not much of a discussion here. I just suggested a simpler implementation to get the exact same result.

penrods:
If we added your code and initialize allow_overwrite to False, would it remain that way?  If so, then you couldn't just write ```self.settings['somesetting'] = 3``` in a handler, right?  You'd have to flip the allow_overwrite flag.  Or were you thinking that flag would be flipped after init time or something?

forslund:
I was thinking doing like the current implementation is doing. flipping it over after the `__init__` is done (where this version copies back the old info from the backup)

Mn0491:
@forslund that makes sense i'll do it that way

Mn0491:
@forslund @penrods did the change. Also when you merge make sure to squash and merge

forslund:
LGTMAWIMT (looks good to me and works in my test)

PR #1305, opened 2017-12-13
Add platform type/build to info sent to register

penrods:
Add platform information (type and build #) to the messages
send to the server to pair and to update version info.

### DIFF
--- a/mycroft/api/__init__.py
+++ b/mycroft/api/__init__.py
@@ -144,21 +144,35 @@ class DeviceApi(Api):
 
     def activate(self, state, token):
         version = VersionManager.get()
+        platform = "unknown"
+        platform_build = ""
+        if "enclosure" in config:
+            platform = config.get("enclosure").get("platform", "unknown")
+            platform_build = config.get("enclosure").get("platform_build", "")
         return self.request({
             "method": "POST",
             "path": "/activate",
             "json": {"state": state,
                      "token": token,
                      "coreVersion": version.get("coreVersion"),
+                     "platform": platform,
+                     "platform_build": platform_build,
                      "enclosureVersion": version.get("enclosureVersion")}
         })
 
     def update_version(self):
         version = VersionManager.get()
+        platform = "unknown"
+        platform_build = ""
+        if "enclosure" in config:
+            platform = config.get("enclosure").get("platform", "unknown")
+            platform_build = config.get("enclosure").get("platform_build", "")
         return self.request({
             "method": "PATCH",
             "path": "/" + self.identity.uuid,
             "json": {"coreVersion": version.get("coreVersion"),
+                     "platform": platform,
+                     "platform_build": platform_build,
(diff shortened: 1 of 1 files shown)


penrods:
@augustnmonteiro and @matheuslima let me know if this looks correct.  I had it sitting on my local machine and wanted to get the code off of it for review or trash.

penrods:
I should expand on this -- there should be no reason to set the device type manually in home.mycroft.ai when you register a device.  This change is intended to get the information to the server, and obviously the server will have to listen for it.  Does this change make sense?

mycroft/api/__init__.py
```diff
@@ -144,21 +144,35 @@ def get_code(self, state):
...
+        platform = "unknown"
+        platform_build = ""
+        if "enclosure" in config:
+            platform = config.get("enclosure").get("platform", "unknown")
+            platform_build = config.get("enclosure").get("platform_build", "")
         return self.request({
             "method": "PATCH",
             "path": "/" + self.identity.uuid,
             "json": {"coreVersion": version.get("coreVersion"),
+                     "platform": platform,
```
augustnmonteiro: the platform is the called type on the backend so if should be
`"type": platform,`

mycroft/api/__init__.py
```diff
@@ -144,21 +144,35 @@ def get_code(self, state):
...
+                     "platform_build": platform_build,
                      "enclosureVersion": version.get("enclosureVersion")}
         })
 
     def update_version(self):
         version = VersionManager.get()
+        platform = "unknown"
+        platform_build = ""
+        if "enclosure" in config:
+            platform = config.get("enclosure").get("platform", "unknown")
```
augustnmonteiro: the platforms is a enum with the types `Mycroft, Desktop, Picroft, Other`

mycroft/api/__init__.py
```diff
@@ -144,21 +144,35 @@ def get_code(self, state):
...
+            platform_build = config.get("enclosure").get("platform_build", "")
         return self.request({
             "method": "PATCH",
             "path": "/" + self.identity.uuid,
             "json": {"coreVersion": version.get("coreVersion"),
+                     "platform": platform,
+                     "platform_build": platform_build,
```
augustnmonteiro: we don't have this field on the backend

penrods: Let's add it!

penrods:
If we are changing the backend to eliminate "type" as an enum, then we can go ahead and leave this as a new string named "platform" and "platform_build".  This will be ignored today, but once the backend has made this migration it will start picking up the new values.

forslund:
@penrods I think the consensus was that this should be merged and the backend behavior should be updated accordingly.

There is however an error  (config is undeclared when used to get the platform type), if you'll fix that I'll merge this

penrods:
Fixed!

forslund:
Looks good. Merging

PR #1306, opened 2017-12-13
abstract enclosure

JarbasAl:
this PR adds a base class for the enclosure and changes the Mark1Enclosure to use this

on enclosure start it checks the config file for platform and loads a specific enclosure, currently mark1 or dummy

I am creating my own jarbas enclosure and using this as a starting point

i do not own a Mark1, someone please test this


forslund:
We've been talking about moving this out of core entirely. just keeping the display manger and api (moving them to `/mycroft/enclosure`). Have a local branch with this change but I'm waiting for the #1259 to be approved.

So Maybe my work could be replaced by this. but then the base class should be moved to `/mycroft/enclosure`

It should not check for enclosure type to load. The entire enclosure process would be replaced for different enclosures.

forslund:
Question: Doesn't this mean that each subclass implements all the handlers (even the ones they aren't using)?

JarbasAl:
the handlers are there for every class, but all they do is "pass" unless overrided (like the Mark1Enclosure now does)

The mark1 enclosure actually implements 4 Enclosure... i would be moving those into a single one next

My idea was to expand the main enclosure to a "factory" like TTS, hotwords and STT, that would always run (by default with a dummy)

forslund:
I understand your idea about a factory but I don't think it's a good idea here. The implementation can vary quite a lot and be quite platform dependent.

Having all handlers registered and executing nothing seems like a waste and feels a bit clumsy and keeping code separation using different classes is in my opinion a good idea.

forslund:
`mycroft/enclosure/`

should only have the baseclass, api and displaymanager. The actual implementation of the mark1 should remain where it was. Since we need to keep backward compatibility the api and displaymanger need to have dummy copies (in the old location)

JarbasAl:
moved directory structure from mycroft/client/enclosure to mycroft/enclosure

note: the naptime skill would need changes and the sleep animation should go into the enclosure, it would still work as is, but enclosures should be able to override this and the needed signals for that have been added

should i remove the edits to main.py ?

what should i do about the handlers then? i was thinking of replacing the "pass" with a log so we can debug if things are being hit or not

forslund:
Ok, I would have the Weather class, Eye class etc not inherit from Enclosure. Instead I'd have an Mark1Enclosure that instanciates the Eyest, Weather etc classes and stores them internally and has the messagebus handlers that basically calls the appropriate object method.

forslund:
Ok, there's a bunch of replaced `connect` to `_connect`

you'll need to keep backwards compatibility with the old layout by having an api.py in mycroft/client/enclosure

containing
```python
from mycroft.api import *
```

and the same for display_manager.py

The Mark1Enclosure could technically be renamed back to Enclosure and do an
```python
from mycroft.enclosure import Enclosure as BaseEnclosure

class Enclosure(BaseEnclosure):
```

thing. But I'm not sure that is necessary. if there are calls into the class.

JarbasAl:
should i include some log with a deprecation warning in the old files so it appears anytime the old stuff is imported?

maybe do the same with EnclosureClass and send a warning log with the new usage

turboproc:
Hi, would really appreciate some restructuring here. In the process of including the ReSpeaker (4 mics  + 12 leds). Not too difficult from a coding point of view (libraries exist already) but getting this in the MyCroft structure as it stands is rather challenging. So having some abstraction layer that is independent of the physical enclosure at its specific drivers might be an interesting way forward.

JarbasAl:
https://github.com/MycroftAI/mycroft-core/pull/1760 handles most of this

PR #1308, opened 2017-12-13
Add boot sequence visual indicators

penrods:
Creating several visual feedback points to let users know what is
occurring in the boot sequence.  This also expects a new version
of the enclosure.

* On shutdown and reset, change eye color to a dark gray
* On initial startup the enclosure will have left the eyes at a
  yellow color, indicating that it is on but not fully running yet.
  When the enclosure comes up (and doesn't kick off the out-of-box
  or the notice that they need to run wifi setup) it will begin to
  scroll an "UPDATING" message.

This assumes that someone else will remove the "UPDATING" message.
That gets handled by the Mycroft-Mark-1 skill, which resets the
mouth and sets the eye color to default.


forslund:
Looks good, merging.

PR #1310, opened 2017-12-15
CLI: interpretation of VT100 keys was flipped

penrods:
The mapping from a VT100 key to NCURSES PgUp/PgDn was inverted,
resulting in different behavior for a keyboard and monitor plugged
in to the Pi than when SSH'ed in to the unit.


forslund:
Quick test, seems OK. Merging

PR #1311, opened 2017-12-15
Fix misspelling in a text resource

penrods:
Fixes a misspelling and removes the "i am awake.dialog" which is
no longer used.  (The dialog is handled by the (Sleep skill)[https://github.com/MycroftAI/skill-naptime] instead.


forslund:
Ok!

PR #1312, opened 2017-12-15
Add new parse function fuzzy_match

penrods:
A fuzzy_match() produces a basic ratio match of two strings.
The result is between 0.0 and 1.0, where 1.0 is a perfect
match.


penrods:
The encoding on the file got hosed, throwing this PR away...

PR #1313, opened 2017-12-15
Add MycroftSkill.translate_namedvalues()

penrods:
This method will load a translateable (and expandable) list of names
and values from the dialog/xx-xx/ folder of a skill.  For example:

dialog/en-en/Colors.value
```
# List colors and their hex RGB values
alice blue, #F0F8FF
antique white, #FAEBD7
aqua, #00FFFF
```


forslund:
Looks good, merging!

PR #1314, opened 2017-12-15
Add message.utterance_remainder() method

penrods:
This helper will return the portion of an utterance not
consumed by the Adapt parser already.  For example,
"turn on the kitchen light" would have a remainder of
"the kitchen" if there was an Intent with entities that
matched "turn on" and "light".

### DIFF
--- a/mycroft/messagebus/message.py
+++ b/mycroft/messagebus/message.py
@@ -16,13 +16,14 @@ import json
 
 
 class Message(object):
-    """This class is used to minipulate data to be sent over the websocket
+    """Holds and manipulates data sent over the websocket
 
-        Message objects will be used to send information back and fourth
-        between processes of mycroft service, voice, skill and cli
+        Message objects will be used to send information back and forth
+        between processes of Mycroft.
+
     Attributes:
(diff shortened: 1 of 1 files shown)


mycroft/messagebus/message.py
```diff
@@ -127,3 +127,20 @@ def publish(self, type, data, context=None):
...
+        For intents get the portion not consumed by Adapt.
+
+        For example: if they say 'Turn on the family room light'
+        and there are entity matches for "turn on" and "light",
+        then it will leave behind "the family room".
+
+        Returns:
+            str: Leftover words or None if not an utterance.
+        """
+        utt = message.data["utterance", None]
```
forslund: This is probably not correct syntax.

penrods: Cleaned up and tested with the Wink skill (replacing its own get_remainder() implementation

forslund:
I'm not sure about the placement here. The Message is intended as a neutral structure to describe any message on the messagebus, this is rather specific to the messages from the intent_service to the skills.

I'll merge this for now but we should add some "message class" system

PR #1316, opened 2017-12-19
fixed skill settings migration for web ui

Mn0491:
==== Fixed Issues ====
The settings migration stopped working due to this code in settings.py  

```python
skill.allow_overwrite = False
```

This is because when a settingsmeta get's modified, the settings in settings.json is prevented from loading onto the skillsettings class because the flag for allow_overwrite was set to false. This change will now load settings from settings.json on startup, as well as load settings from settings.json before any migration takes place.


penrods:
I'm OK with this one if you are, @forslund

PR #1325, opened 2017-12-22
Create ISSUE_TEMPLATE.md

KathyReid:


forslund:
This might be a bit long and as it'll be in the textbox the user will basically need to clear it out. Can this be added as a section in the CONTRIBUTING.md and a simpler template that can be used directly be put in place of this?

KathyReid:
Hi @forslund I see your point. I'm in two minds on this one. 

Points for having a detailed ISSUE.md: 
* Tries to set a strong standard for Issues that are raised, making them easier to replicate and resolve
* Tries to make sure we have all the information that's needed, reducing the contact time with the submitter, which can often cause delays. 

Points against having a detailed ISSUE.md: 
* The submitter has to select all and delete before submitting an Issue, which is poor UX

I'm leaning towards having a more detailed ISSUE.md. 

Are there other points that we may not have considered?

penrods:
Looking at similar and popular projects:

Tensorflow (https://github.com/tensorflow/tensorflow/issues/new)
Similar, lots of descriptive text.

Puppeteer (https://github.com/GoogleChrome/puppeteer/issues/new)
Also fairly wordy, fewer examples by sections to fill out

Flask (https://github.com/pallets/flask/issues/new)
Similar to Puppeteer


This might be on the more wordy side, but I don't think it is that unusual or will be a significant hassle for someone reporting a bug.

forslund:
My point is that it should be more like the flask example (and our own PR template) where there are headings to help.

PR #1327, opened 2017-12-22
Bugfix/msm list

forslund:
Msm failed to list many skills as installed if the submodule name != repository name, this includes most of the default skills (which has been renamed from skill-X to mycroft-X while the repo name is still skill-X)

This PR makes msm use the submodule _path_ when checking if a skill is installed.

This PR also removes the write to `~/count.txt` (I've been wondering why this file turns up all the time)

### DIFF
--- a/msm/msm
+++ b/msm/msm
@@ -638,15 +638,25 @@ case ${OPT} in
           exit ${exit_code}
       fi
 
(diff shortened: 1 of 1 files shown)


forslund:
I'm sure there's better ways of doing this (I'm terrible with bash)

PR #1332, opened 2017-12-29
feature add it-it support

aleale99:
insert it-it support to nice_number function


aleale99:

PR #1334, opened 2017-12-29
feature : initial insert italian tests for nice_number

aleale99:
Inserted and tested all the it cases, but the Italians do not like to use  fractions different from 1/2 ..1/3 and quarter
in italian if you say "six and seven eleven" you're definitely an alien


aleale99:

PR #1335, opened 2017-12-29
Flexible lang dialect support

JarbasAI:
support language code variations, adds a utility function and implements it in skills core where relevant

these check for language code variations, lets say we want to access "path/en-uk" , but it does not exist, this PR adds methods to check for "path/en" , "path/en-us" etc, and returns the best language alternative

skills in "en-us" will still work if i configure mycroft for "en-uk", or "es-es" skills will work if i configure mycroft lang for "es"

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -575,7 +575,8 @@ class MycroftSkill(object):
         """
         if old_dirname:
             # Try the old directory (dialog/vocab/regex)
-            root_path = get_language_dir(join(self.root_dir, old_dirname), self.lang)
+            root_path = get_language_dir(join(self.root_dir, old_dirname),
+                                         self.lang)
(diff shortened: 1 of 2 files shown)


JarbasAl:
https://www.reddit.com/r/Mycroftai/comments/7v0nom/noob_where_are_all_the_configuration_options/dtrom1n/

```
For example I am in the UK I see in the example there is the line

"lang": "en-us",
Naturally I assume that the UK equivalent is:

"lang": "en-uk",
But this causes my device to no understand any command.
```

neoscreenager:
Try en-gb

JarbasAl:
en-gb, en-uk, en-us, en-au, en ..... they all should work by default @neoscreenager , the difference should be minimal, no reasons for things to fail to load

or should we copy and have the exact same files in slightly differently named folders?

PR #1336, opened 2017-12-29
Update parse.py for lang it_it

aleale99:
start normalize functions for italian language

there is a warning : "We've detected the file encoding as ISO-8859-1. When you commit changes we will transcode it to UTF-8."
i can't verify if it has altered the text  in Portuguese


JarbasAI:
the encoding change seems to have broken the portuguese unittests

forslund:
I've suggested a change in the structure to limit these kinds of problems. See #1338

aleale99:

PR #1342, opened 2018-01-01
Auto update optional

JarbasAI:
this allows users to disable the msm auto update feature, msm can still be used, but will not be invoked by skills main

for users looking for more security this is a desired option, as illustrated by the evil skill :)

### DIFF
--- a/mycroft/configuration/config.py
+++ b/mycroft/configuration/config.py
@@ -138,34 +138,8 @@ class RemoteConf(LocalConf):
     """
     def __init__(self, cache=None):
         super(RemoteConf, self).__init__(None)
-
         cache = cache or '/opt/mycroft/web_config_cache.json'
-
-        try:
-            # Here to avoid cyclic import
-            from mycroft.api import DeviceApi
-            api = DeviceApi()
-            setting = api.get_settings()
-            location = api.get_location()
-            if location:
-                setting["location"] = location
-            # Remove server specific entries
-            config = {}
-            translate_remote(config, setting)
-            for key in config:
-                self.__setitem__(key, config[key])
-            self.store(cache)
-
-        except HTTPError as e:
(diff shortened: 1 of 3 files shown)


forslund:
Good idea!

I'd like a couple of small modifications:
- instead of returning directly in the download_skills method I think it would be clearer if it wasn't called at all so update l330 to

```python
            if time.time() >= self.next_download and AUTO_UPDATE:
                self.download_skills()
```

- Instead of loading the autoupdate config value only at startup, load it before the check so we perhaps can add a voice command to start/stop updating during runtime.

JarbasAI:
should i do             

       # Update skills once an hour
        if time.time() >= self.next_download:
            # check if auto update is enabled 
            AUTO_UPDATE = Configuration.get().get("skills", {}).get(
                "auto_update", True)
            if AUTO_UPDATE:
                self.download_skills()
or

       # check if auto update is enabled
        AUTO_UPDATE = Configuration.get().get("skills", {}).get(
            "auto_update", True)
        
        # Update skills once an hour
        if time.time() >= self.next_download and AUTO_UPDATE:
            self.download_skills()

i like it best only checking the flag before download but after the time out, otherwise the logs get a little polluted with config loading messages


so it checks the flag once an hour instead of once every 2 seconds

forslund:
Rebased and merged in #1344. Thanks @JarbasAI !

tlc:
I just ran into a problem that might be related to this.  Then again I'm new to Mycroft.

On a new install of mycroft (on Ubuntu 16.04), I ran _msm/msm default_, but mycroft wouldn't load any skills except _pairing_.  I looked at the skills loading and saw it wasn't getting past this:

>  if (exists(SKILLS_DIR) and
>         (self.next_download or not update)):

After I changed my config to _auto_update: false_, they loaded.

PR #1350, opened 2018-01-07
loaded skills bus messages

JarbasAl:
## Description
Adds bus messages for skill.loaded, skill.shutdown and skill.load.fail

## How to test
see new messages appearing in the logs

CLA [ x] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -926,10 +926,6 @@ class MycroftSkill(object):
         except:
             LOG.error("Failed to stop skill: {}".format(self.name),
                       exc_info=True)
-        self.emitter.emit(Message("skill.shutdown",
-                                  {'folder': self._dir.split("/")[-1],
-                                   "id": self.skill_id,
-                                   "name": self.name}))
 
     def _unique_name(self, name):
         """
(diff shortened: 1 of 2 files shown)


forslund:
Part of this overlaps with work done in #1318 so I might merge this into that PR and fix the conflicts.

JarbasAl:
i missed #1318, just needed this for the intent data PR #1351

forslund:
My PR got merged before I had time to update it with this pr. Can you rebase this and resolve the conflict?

I also would like to have the shutdown message in the SkillManager and not within the shutdown method. (To make it more consistent)

The message names should start with `mycroft.skills`.

JarbasAl:
i had the message there at first, but maybe skills can call the shutdown method somehow (self shutdown or other shenanigans ) and the message is still good to have

forslund:
Even if the skills call their own shutdown method that won't unload them. They will still be registered as loaded in the skills service and hang around in memory. I'd like the message more as an indicator of what the skill load/reload system does

forslund:
Looks good, can you fix the remaining Pep8 issue (too many blank lines after shutdown message)

forslund:
Merged via #1396

Issue #1330, opened 2017-12-24
dependencies for openSUSE Tumbleweed


JenaPlinsky:
Hi guys 
... I can not find these dependencies for openSUSE Tumbleweed can someone help me ?

- virtualenvwrapper

- libglib2.0-devel

- libssl-devel

- portaudio19-devel

ChrisFoster4:
After some Googling/DDGing I think I have found the packages. I'm not running openSuse so I can't test building with  these packages but I hope this helps. Post any further issues you have.

[virtualenvwrapper](https://opensuse.pkgs.org/42.3/opensuse-oss/python-virtualenvwrapper-4.3.2-5.1.noarch.rpm.html)

[libglib2.0-devel](https://software.opensuse.org/package/glib2-devel)

[libssl-devel](https://opensuse.pkgs.org/42.2/opensuse-update-oss/libressl-devel-2.5.3-5.6.1.x86_64.rpm.html)

portaudio19-devel - Should be installable as libportaudio2. I found it in [this ](https://en.opensuse.org/Archive:Package_list_12.3)package list and [this site](https://opencpn.org/wiki/dokuwiki/doku.php?id=opencpn:developer_manual:developer_guide:compiling_linux) told me that it may be called "libportaudio2" in non Debian distros.

ChrisFoster4:
Ok I have got it to run in an openSuse Tumbleweed virtual machine by adding the approriate zypper commands to dev_setup.sh. I also have to run the `update-dev.sh` script before running the `start-mycroft.sh debug` script.Could you try using the modified dev_setup.sh I have attached to get Mycroft working. Good luck.
[dev_setup.zip](https://github.com/MycroftAI/mycroft-core/files/1591639/dev_setup.zip)

JenaPlinsky:
@ChrisFoster4 
unfortunately I still have problems anyway thank you very much for your help

```
[sudo] password di root: 
Caricamento dati del repository in corso...
Lettura dei pacchetti installati in corso...
'glib2-devel' è già installato.
Nessun aggiornamento candidato per 'glib2-devel-2.54.2-361.1.x86_64'. La versione più recente disponibile è già installata.
'python-virtualenvwrapper' è già installato.
Nessun aggiornamento candidato per 'python-virtualenvwrapper-4.7.1-2.3.noarch'. La versione più recente disponibile è già installata.
'libressl-devel' è già installato.
[...]

PR #1355, opened 2018-01-09
Add support for installing on openSUSE

forslund:
## Description

Running the dev_setup.sh script and getting mycroft up and running on openSUSE was not working (See #1330 and #1340), these modifications are tested to work on a fresh install of OpenSUSE Tumbleweed.

- Use Zypper to install packages
- use primary group of current user when changing owner of /opt/mycroft (openSuse has no user specific group by default)

## How to test
The big thing to test if `prepare_msm.sh` uses the correct group after this change. In my tests (debian and openSuse) this method seem to work.

CLA [Yes]

### DIFF
--- a/scripts/prepare-msm.sh
+++ b/scripts/prepare-msm.sh
@@ -31,11 +31,12 @@ chmod +x ${DIR}/../msm/msm
 
 # Determine which user is running this script
 setup_user=$USER
+setup_group=$( id -gn $USER )
 
 # change ownership of ${mycroft_root_dir} to ${setup_user } recursively 
 function change_ownership {
-    echo "Changing ownership of" ${mycroft_root_dir} "to user:" ${setup_user} "with group:" ${setup_user}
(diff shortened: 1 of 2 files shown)


scripts/prepare-msm.sh
```diff
@@ -31,11 +31,12 @@ chmod +x ${DIR}/../msm/msm
 
 # Determine which user is running this script
 setup_user=$USER
+setup_group=`id -gn $USER`
```
MatthewScholefield: I know at the moment mycroft scripts use some of both, but I think we should transition to `$()` command substitution to make the code more consistent. Let me know what you think.

forslund: Reading up on this I can't disagree. Seems like the back-ticks are depreciated even. Will fix

forslund: The PR has been updated

MatthewScholefield:
Looks good, merging!

PR #1359, opened 2018-01-11
Disable overflow exceptions when reading mic

### LINKED ISSUE #1348: Raspbian audio input freeze
As described in forum posts https://community.mycroft.ai/t/speech-rec-for-mycroftpi-using-kinect/1267/7 and https://community.mycroft.ai/t/mic-just-stopped-working-weird-right-and-will-not-come-back-local-usb-or-mic-jack/2191/17 the audio input on a raspberry pi installation goes dead after a few minutes. Mycroft cannot record any more audio input and the mic level display in the cli just freezes.

It occurred to me both with a regular picroft image on a rpi3 as well as an install on raspbian stretch from the master branch of the repo with two different usb microphones. In my setup the bug is reproducable every time when starting a mycroft instance, however it does not seem to be widespread, so i don´t know what is causing it specifically.

In the code the problem originates in the MutableStream class where at some point calls to the pyaduio `stream.read()` method only result in IOErrors. IOErrors are thrown on input overflows as described in http://stackoverflow.com/questions/10733903/pyaudio-input-overflowed. The AudioProducer Thread therefore only handles those errors, no audio is processed.

This seems to be a pyaudio internal issue on the raspberry pi. 
It can however be fixed easily by passing a parameter `exception_on_overflow=False` to the pyaudio `stream.read()` method.

forslund:
## Description
This PR sets the input stream to ignore buffer overflows by default. (The exception can be turned on again using a config parameter)
Disabling this exception should free up the audio producer thread to continue deliver the available sound data fixing the instances where the the mic stops working after a while.

Should fix #1348.

Many thanks to @mikonse for troubleshooting and suggesting this fix.

## How to test
Load the code on a Picroft and assert that the mic continues to operate after

CLA [Yes]

### DIFF
--- a/mycroft/client/speech/mic.py
+++ b/mycroft/client/speech/mic.py
@@ -61,7 +61,18 @@ class MutableStream(object):
     def unmute(self):
         self.muted = False
 
-    def read(self, size):
+    def read(self, size, of_exc=False):
+        """
+            Read data from stream.
+
+            Arguments:
+                size (int): Number of bytes to read
+                of_exc (bool): flag determining if the audio producer thread
+                               should throw IOError at overflows.
+
+            Returns:
+                Data read from device
+        """
         frames = collections.deque()
         remaining = size
         while remaining > 0:
@@ -69,7 +80,8 @@ class MutableStream(object):
             if to_read == 0:
                 sleep(.01)
                 continue
-            result = self.wrapped_stream.read(to_read)
+            result = self.wrapped_stream.read(to_read,
+                                              exception_on_overflow=of_exc)
             frames.append(result)
             remaining -= to_read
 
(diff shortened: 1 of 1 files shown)


mycroft/client/speech/mic.py
```diff
@@ -199,7 +213,7 @@ def __init__(self, wake_word_recognizer):
 
     @staticmethod
     def record_sound_chunk(source):
-        return source.stream.read(source.CHUNK)
+        return source.stream.read(source.CHUNK, self.overflow_exc)
```
penrods: self. flops in the staticMethod.  Need to add this as a class member that defaults to None, then switch here and the constructor to use  ```ResponsiveRecognizer.overflow_exc``` instead of ```self.overflow_exc```

forslund: Thwarted again by my nemesis `staticmethod`, will fix.

forslund: In the end I found the cleanest way to fix this to be just removing the `@staticmethod` decorator. There was no particular reason for this to not be a proper class method as far as I could see.

forslund:
I ended up removing the `@staticmethod` since there was no real reason to keep it as one if it needed access to an object member.

PR #1364, opened 2018-01-16
Bugfix/scheduled event

forslund:
## Description
Previously the scheduled event handlers were never cleaned up after use so if the same event was scheduled multiple times the handler would be called multiple times.

Basically now there's no need to manually cancel an expired timer.

## How to test
Make a short test skill similar to

```python
# -*- coding: utf-8 -*-
from datetime import datetime, timedelta

class TestSkill(MycroftSkill):
    @intent_handler(IntentBuilder('').require('Test'))
    def test_handler(self, message):
        LOG.info('Event in 10 secs!')
        self.schedule_event(self.handler,
                            datetime.now() + timedelta(seconds=10),
                            name='test-event')

    def handler(self, message):
        LOG.info('!!!!!!!!!!!! SCHEDULED HANDLER !!!!!!!!!!!!')

def create_skill():
    return TestSkill()
```

and assure that the `handler` is only called once after each run of `test_handler` and/or check that other skills like mycroft-timer still works as intended

CLA [Yes]

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -596,24 +596,26 @@ class MycroftSkill(object):
             text = f.read().replace('{{', '{').replace('}}', '}')
             return text.format(**data or {}).split('\n')
 
-    def add_event(self, name, handler, need_self=False):
+    def add_event(self, name, handler, need_self=False, once=False):
         """
             Create event handler for executing intent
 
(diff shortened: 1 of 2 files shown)


forslund:
Ping @Mn0491

PR #1365, opened 2018-01-17
is_remote_list check

JarbasAI:
## Description

when getting remote configuration all lists are considered a list of dicts, this does not allow lists to be sent to the device

this PR adds a method to check if a list is remote formatted or a normal list

If using [another backend](https://github.com/JarbasAl/personal-mycroft-backend/) this is good functionality to have, and also allows mycroft-home to expand better without accidentally screwing the configs

in my case i was sending a list of blacklisted skills from my backend to the device when i noticed this

## How to test

on remote config update verify nothing different happens

CLA [**yes** ] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

### DIFF
--- a/mycroft/configuration/config.py
+++ b/mycroft/configuration/config.py
@@ -51,7 +51,7 @@ def is_remote_list(values):
     for v in values:
         if not isinstance(v, dict):
             return False
-        if not "@type" in v.keys():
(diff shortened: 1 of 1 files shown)


forslund:
Seems to be working fine. Thanks

PR #1366, opened 2018-01-17
local Deep speech

JarbasAI:
## Description

i doubt this will be useful anytime soon, remote is the way to go for deep speech, but for those wanting offline stt that can handle it

adds local deepspeech STT option now that it can be pip installed

auto downloads and extracts model

## How to test

find an enclosure that can handle deep speech

change config to "deepspeech"

CLA [**yes** ] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

### DIFF
--- a/mycroft/stt/deepspeech_stt.py
+++ b/mycroft/stt/deepspeech_stt.py
@@ -35,9 +35,14 @@ class DeepSpeechSTT(STT):
         self.N_FEATURES = self.config.get("n_features", 26)
         self.N_CONTEXT = self.config.get("n_context", 9)
 
-        self.model = self.config.get("model", join(dirname(__file__),
-                                                   "deepspeech",
-                                                   "output_graph.pb"))
+        self.model = self.config.get("model", "deepspeech_0.1.0")
+        if self.model == "deepspeech_0.1.0":
+            self.model = join(dirname(__file__), "deepspeech",
+                              "output_graph.pb")
+            auto_dl = True
+        else:
+            auto_dl = False
+
         self.lm = self.config.get("lm", join(dirname(__file__),
                                              "deepspeech",
                                              "lm.binary"))
@@ -49,8 +54,8 @@ class DeepSpeechSTT(STT):
                                                  "trie"))
         self.downloaded = False
         self.dl = None
-        auto_dl = self.config.get("auto_download", False)
-        if self.is_ready(not auto_dl):
+
+        if self.model_downloaded():
             self.load_model()
         elif auto_dl:
             LOG.info("Downloading model")
@@ -58,8 +63,12 @@ class DeepSpeechSTT(STT):
             while not self.downloaded:
                 sleep(1)
 
-            if self.is_ready(True):
-                self.load_model()
+            if not self.model_downloaded():
+                raise AssertionError(self.model + " does not exist"
+                                     ", download a pre-trained model with \n "
+                                     "wget -O -" + DeepSpeechSTT.URL + "| "
+                                     "tar xvfz -")
+            self.load_model()
 
     def load_model(self):
         LOG.info('Loading model from file %s' % (self.model))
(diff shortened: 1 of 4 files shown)


mycroft/stt/deepspeech_stt.py
```diff
@@ -0,0 +1,131 @@
...
+                                             "lm.binary"))
+        self.alphabet = self.config.get("alphabet", join(dirname(__file__),
+                                                         "deepspeech",
+                                                         "alphabet.txt"))
+        self.trie = self.config.get("trie", join(dirname(__file__),
+                                                 "deepspeech",
+                                                 "trie"))
+        self.downloaded = False
+        self.dl = None
+        auto_dl = self.config.get("auto_download", False)
```
clusterfudge: What is the case that we don't auto-download? I'd say unit tests maybe, but there are none for this class. Since this has to be explicitly configured, can we just assume that if the model doesn't exist we should fetch it?

mycroft/stt/deepspeech_stt.py
```diff
@@ -0,0 +1,131 @@
...
+from mycroft.util.log import LOG
+from timeit import default_timer as timer
+from os.path import exists, join, dirname
+from mycroft.util.download import download
+import tarfile
+from time import sleep
+
+
+class DeepSpeechSTT(STT):
+    URL = "https://github.com/mozilla/DeepSpeech/releases/download/v0.1.0/" \
```
clusterfudge: how big is this model? are we sure that existing mark1s and picrofts can download this without accidentally filling disk?

mycroft/stt/deepspeech_stt.py
```diff
@@ -0,0 +1,131 @@
...
+            LOG.error(e)
+            if is_critical:
+                raise
+            return False
+        return True
+
+    def download(self):
+        LOG.info("starting model download")
+        target_folder = join(dirname(__file__), "deepspeech")
+        self.dl = download(DeepSpeechSTT.URL, target_folder, self._extract)
```
clusterfudge: why is this async? Can we just use a sync version instead?

mycroft/stt/deepspeech_stt.py
```diff
@@ -0,0 +1,131 @@
...
+        lm_load_start = timer()
+        self.ds.enableDecoderWithLM(self.alphabet, self.lm, self.trie,
+                                    self.LM_WEIGHT,
+                                    self.WORD_COUNT_WEIGHT,
+                                    self.VALID_WORD_COUNT_WEIGHT)
+        lm_load_end = timer() - lm_load_start
+        LOG.debug('Loaded language model in %0.3fs.' % (lm_load_end))
+
+    def is_ready(self, is_critical=False):
+        try:
```
clusterfudge: I'm not a fan of this pattern of throwing, catching, and conditionally rethrowing. Why don't we just return success or failure and then throw at the caller if necessary?

mycroft/stt/deepspeech_stt.py
```diff
@@ -0,0 +1,131 @@
...
+        self.alphabet = self.config.get("alphabet", join(dirname(__file__),
+                                                         "deepspeech",
+                                                         "alphabet.txt"))
+        self.trie = self.config.get("trie", join(dirname(__file__),
+                                                 "deepspeech",
+                                                 "trie"))
+        self.downloaded = False
+        self.dl = None
+        auto_dl = self.config.get("auto_download", False)
+        if self.is_ready(not auto_dl):
```
clusterfudge: can we change this so it's a little more declarative?

```
if not model.downloaded():
    model.download()

self.load_model()
```

mycroft/stt/deepspeech_stt.py
JarbasAI: 2.1 GB

mycroft/stt/deepspeech_stt.py
JarbasAI: i just used the official download util, this way we would get any updates to the download code instead of me re implementing it

mycroft/stt/deepspeech_stt.py
JarbasAI: this way we know what failed in more detail, i will throw a generic model not found exception instead

forslund:
I think it's better to use an external server (PR #1370) which can be run on the same machine in a docker container or some remote host. Is there any specific benefit to running it within the Mycroft virtualenv / process?

MB2412:
i changed the config file to a personal server STT deepspeech and i launched the server using this command in an other terminal "deepspeech-server --config config.json" but when i run mycroft , it shows me this error "temporary failure in name resolution " . thank you for your help
29920781_10211594618437359_682507320_n
i ve already tested that with a wave file and it ran perfectly ! but when i try to mix it with mycroft it wont work even after changing the required configurations

PR #1369, opened 2018-01-18
Implement max-line to limit memory usage

penrods:
The major point of the PR is to limit the memory usage of the CLI by
implementing a maximum log limit.  It defaults to 5000.

Other changes:
* Add "--debug" option to support troubleshooting/debugging the CLI itself
* Add support for jumping to the top (Ctrl+T/Ctrl+PgUp) or bottom (Ctrl+B/Ctrl+PgDn) of the logs.
* Remove the "OLDEST" message from the log.  It was really no longer necessary since the log navigation issues got straightened out, and it complicated the max log line logic.


forslund:
Looks good!

PR #1373, opened 2018-01-19
Restore captured text on cli exceptions

MatthewScholefield:
## Description
If the CLI throws an exception, it currently gets hidden, unless specifically called with `--debug`. This PR reveals all saved text in an exception hook so that nothing gets hidden on an error.

## How to test
 - Insert a random `print('xyz'); raise Exception('Ahhhh!')` somewhere in the `mycroft/client/text/main.py`
 - Run the cli (`./start.sh debug`) and you should see both `xyz` and the stack trace for the exception

 - [x] CLA

### DIFF
--- a/mycroft/client/text/main.py
+++ b/mycroft/client/text/main.py
@@ -12,35 +12,39 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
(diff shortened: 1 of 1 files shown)


forslund:
Works great, merging.

PR #1377, opened 2018-01-23
fix msm install order

JarbasAl:
## Description

requirements.sh is installed after requirements.txt

pip packages that require system packages will fail to install


forslund:
Looks good, many thanks. Merging

PR #1388, opened 2018-01-28
Run pip after any update to skills managed by msm.

### LINKED ISSUE #1387: msm does not re-run pip when skills are updated.
I just pushed some changes to my skill, and ran `msm update` on my Mark 1 to test the new features. After restarting the skills service, my skill would not load, as it was still trying to use an old version of a library I use in my skill. Removing and re-installing the skilled resolved the issue, however, I believe there is a bug in the `update` feature.

Taking a peak at the `msm` script, it looks like this is because `pip install -r requirements.txt` is not necessarily run after each update. I believe this should run any time the repo for the skill is updated.

ChristopherRogers1991:
Run pip after any updates to skills, to ensure that libraries are
updated per the lateset version of the skill. This may cause pip to run
twice (if mycroft has re-PIPed), however, given that the cost of this
is very low (basically a no-op on the second run), and the re-PIP logic
is flagged for removal, it seems safest to not touch that logic for
now.

==== Fixed Issues ====

## Description
Fixes #1387 

## How to test
1) Install a skill using `msm`.
1) Updates it's `requirements.txt` to include a new library.
1) Run `msm update`
1) Run `pip freeze`

Without this fix, the new library should not be installed. With this fix, the new library will be installed.

CLA [X]

### DIFF
--- a/msm/msm
+++ b/msm/msm
@@ -471,6 +471,7 @@ function update() {
         git fetch
         git reset --hard origin/master
         rm -f *.pyc
+        run_pip "${d}"
       else
         echo "Ignoring ${d}, skill has been modified."
       fi


JarbasAI:
should the same be done for requirements.sh ?

ChristopherRogers1991:
Hmm... Hadn't thought of that, but yeah, potentially. There may be more of
a cost to running those scripts though. Pip behaves in a predictable way,
but those scripts could be doing anything. Running them repeatedly could
result in unintended consequences. Best solution might be to try to
determine if the script has changed, and only run it if it has.

penrods:
For now I'd rather keep this as PIP only.  Rerunning the requirements.sh is definitely riskier.  I'm also looking at a big rewrite of MSM, probably moving to Python instead of a shell script to give us more flexibility.

PR #1394, opened 2018-01-31
Feature/format time number

penrods:
Add formatting utilities:
* mycroft.util.format.nice_time()
* mycroft.util.format.pronunce_number()
* unittests for above

Also renamed the private helper method convert_number() to
_convert_to_mixed_fraction()


forslund:
Did a couple of tweaks but it looks good now. Merging.

PR #1404, opened 2018-02-02
Check that value exists before checking for change

forslund:
## Description
An error occured when testing skill settings when checking if the settings should be sent when trying to upload a field without a value. To guard against this there is a check if the field has a value before checking.

CLA [Yes]


Mn0491:
LGTM! let's merge and release

PR #1408, opened 2018-02-05
Allow clients to target the replies from Mycroft

forslund:
## Description
The speech handling now checks the message context if it's the intended target for the message and will only speak in the following conditions:

- Explicitly targeted i.e. the target include `"audio"`
- target is set to `None`
- target is missing completely

The idea is that for example when the android app is used to access Mycroft the device at home shouldn't start to speak.

### Targeting Theory
- The context target parameter in the original message can be set to list with any number of intended targets:
```python
    ws.emit(Message('recognizer_loop:utterance', data, context={'destination': ['audio', 'kde']))
```
- A missing target or if the target is set to `None` is interpreted as a multicast and should trigger all output capable processes (be it the *mycroft-audio* process, a web-interface, the KDE plasmoid or maybe the android app)

CLA [Yes]

### DIFF
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -49,6 +49,13 @@ def handle_speak(event):
     Configuration.init(bus)
     global _last_stop_signal
 
+    # if the message is targeted and audio is not the target don't
+    # don't synthezise speech
+    if (event.context and 'destination' in event.context and
+            event.context['destination'] and
+            'audio' not in event.context['destination']):
+        return
+
     # Get conversation ID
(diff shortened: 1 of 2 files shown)


JarbasAI:
now the devs just need to get used to use message.reply instead of creating a new Message objects :D

AIIX:
This target feature seems similar to my request for adding an alternative speak method, https://github.com/MycroftAI/mycroft-core/issues/1414 hence adding the request here to have a  condition added where utterance is not emitted over the messagebus and instead will only have an audio response. Use case is in the issue above with a video where this is useful

PR #1415, opened 2018-02-08
Setup python3

forslund:
## Description
- Fixes a few incompatibilities with python 3 to make the stack running fairly well
- Sets up the virtualenv with python3
This also adds the possibility to set the environment variable MYCROFT_VENV to separate the virtualenv for python3 from the stable version.

Example:
```bash
export MYCROFT_VENV=mycroft-18.02
./dev_setup.sh
./start-mycroft.sh all
```

- Upgrade tests and travis

CLA [Yes]

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -91,7 +91,7 @@ install_deps() {
     fi
 
     if found_exe apt-get; then
-        $SUDO apt-get install -y git python python-dev python-setuptools python-virtualenv python-gobject-dev virtualenvwrapper libtool libffi-dev libssl-dev autoconf automake bison swig libglib2.0-dev s3cmd portaudio19-dev mpg123 screen flac curl libicu-dev pkg-config automake libjpeg-dev libfann-dev build-essential jq
+        $SUDO apt-get install -y git python python-dev python-setuptools python-venv python-gobject-dev libtool libffi-dev libssl-dev autoconf automake bison swig libglib2.0-dev s3cmd portaudio19-dev mpg123 screen flac curl libicu-dev pkg-config automake libjpeg-dev libfann-dev build-essential jq
     elif found_exe pacman; then
         $SUDO pacman -S --needed git python2 python2-pip python2-setuptools python2-virtualenv python2-gobject python-virtualenvwrapper libtool libffi openssl autoconf bison swig glib2 s3cmd portaudio mpg123 screen flac curl pkg-config icu automake libjpeg-turbo base-devel jq
     elif found_exe dnf; then
@@ -109,20 +109,18 @@ install_deps() {
     fi
 }
 
-install_deps
+TOP=$(cd $(dirname $0) && pwd -L)
+VIRTUALENV_ROOT=${VIRTUALENV_ROOT:-"${TOP}/.venv"}
+
+install_venv() {
+    python3 -m venv .venv/
+    curl https://bootstrap.pypa.io/get-pip.py | .venv/bin/python
+}
+#install_deps
 
 # Configure to use the standard commit template for
 # this repo only.
 git config commit.template .gitmessage
-
-TOP=$(cd $(dirname $0) && pwd -L)
-
-MYCROFT_VENV=${MYCROFT_VENV:-"mycroft"}
-if [ -z "$WORKON_HOME" ]; then
-    VIRTUALENV_ROOT=${VIRTUALENV_ROOT:-"${HOME}/.virtualenvs/$MYCROFT_VENV"}
-else
-    VIRTUALENV_ROOT="$WORKON_HOME/$MYCROFT_VENV"
-fi
 
 # Check whether to build mimic (it takes a really long time!)
 build_mimic='y'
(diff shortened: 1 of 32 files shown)


MatthewScholefield:
I have a few general comments:
 - Since this is in the 18.02 branch, I think we should ditch any python 2.7 compatibility code since it would only add bloat and potential bugs down the road
 - Proposal:
   - Do we want to use python 3's `venv`? I heard it was the best way to do virtualenvironments in Python 3.3+
   - What if we switch the virtualenv location to within the Mycroft directory (somewhere like `.venv/` or `venv/`? This would solve the issue of conflicting with the Python 2 virtualenv. I find this makes sense because the virtual environment is only ever used by mycroft-core. And, you can quickly run the local environment without activating with commands like `.venv/bin/python`.

forslund:
I've got no experience with venv but from a quick google it seems to be a good. Will try it out in more detail tomorrow.

I'll remove the compatibility where I remember/find it, it's couldn't be more than 50 lines of code in total.

MatthewScholefield:
Also, as a reference, you could look at how it's done in the [training-server](https://github.com/MycroftAI/training-server/blob/master/setup.sh#L10) setup.sh code.

forslund:
Thanks

mycroft/skills/settings.py
```diff
@@ -94,7 +94,7 @@ def __init__(self, directory, name):
         self._settings_path = join(directory, 'settings.json')
         self._meta_path = join(directory, 'settingsmeta.json')
         self.is_alive = True
-        self.loaded_hash = hash(str(self))
+        self.loaded_hash = hash(frozenset(self.items()))
```
MatthewScholefield: Since this will fail with recursive dicts, what if we use `hash(json.dumps(self, sort_keys=True))`?

Also, make sure to keep this consistent with the method used below in `self._is_stored()`.

dev_setup.sh
```diff
@@ -109,20 +109,19 @@ install_deps() {
     fi
 }
 
-install_deps
+TOP=$(cd $(dirname $0) && pwd -L)
+VIRTUALENV_ROOT=${VIRTUALENV_ROOT:-"${TOP}/.venv"}
+
+install_venv() {
+    python3 -m venv .venv/
```
MatthewScholefield: `.venv/` here and below should be `"$VIRTUALENV_ROOT"` if we are still making it configurable.

mycroft/skills/settings.py
forslund: I noticed the issue with recursive dicts but obviously failed to fix all of them. Sorted json.dumps should be the cleanest solution. Will switch over to that.

dev_setup.sh
forslund: Good catch.

After looking at msm we might want to remove the override for simplicity.

msm/msm
```diff
@@ -364,14 +356,14 @@ function run_pip() {
...
-        if [[ "${VIRTUAL_ENV}" =~ .mycroft$ ]] ; then
+        if [[ "${VIRTUAL_ENV}" =~ .mycroft-core/.venv$ ]] ; then
           if ! pip install -r requirements.txt ; then
             echo "ERROR: Unable to install requirements for skill '${name}'"
             send_install_fail "${name}" 121
             return 121
           fi
         else
-          if workon mycroft ; then
+          if source ${ACTIVATE} ; then
```
MatthewScholefield: `ACTIVATE` could have spaces so this should be wrapped in quotes.

test/unittests/skills/core.py
```diff
@@ -27,6 +27,9 @@
     load_vocab_from_file, load_vocabulary, MycroftSkill, \
     load_skill, create_skill_descriptor, open_intent_envelope
 
+if sys.version_info[0] >= 3:
```
MatthewScholefield: Can be removed

test/unittests/client/audio_consumer_test.py
```diff
@@ -13,14 +13,18 @@
...
-from Queue import Queue
 
 import speech_recognition
 from os.path import dirname, join
 from speech_recognition import WavFile, AudioData
 
 from mycroft.client.speech.listener import AudioConsumer, RecognizerLoop
 from mycroft.stt import MycroftSTT
+import sys
+if sys.version_info[0] < 3:
```
MatthewScholefield: Can be removed

dev_setup.sh
```diff
@@ -109,20 +109,19 @@ install_deps() {
     fi
 }
 
+TOP=$(cd $(dirname $0) && pwd -L)
+VIRTUALENV_ROOT=${VIRTUALENV_ROOT:-"${TOP}/.venv"}
+
+install_venv() {
+    python3 -m venv ${VIRTUALENV_ROOT}/
```
MatthewScholefield: Quotes needed around variable.

dev_setup.sh
```diff
@@ -109,20 +109,19 @@ install_deps() {
...
+    curl https://bootstrap.pypa.io/get-pip.py | .venv/bin/python
```
MatthewScholefield: To be configurable this should end with:
```bash
"$VIRTUALENV_ROOT/bin/python"
```

Issue #1411, opened 2018-02-07
Leave all nice_number formatting to language 


zecakeh:
I'm working on the french translation and while creating tests I realized that nice_number is returning decimal numbers with an english formatting. By that I mean that in french for decimals we use a comma instead of a dot, and for thousands separator we use unbreakable spaces rather than commas (that one is just a comment for now). And this is handled directly in the main nice_number function.

Since it looks like the output of this function is directly fed to a display or TTS, the numbers should be formatted according to the language.

If I'm right, I could send a pull request with the changes. Should the english formatting also be included in the other already included languages, waiting for a maintainer to change it if necessary but not breaking anything in the meantime?

zecakeh:
What about using the python string format function? It uses the formatting of the language setting to display numbers nicely. That way it would automagically work for digits whatever the language is, even if it's not implemented.

zecakeh:
As said in PR #1416 , it's not recommended to use the python <code>format</code> function because it involves changing the default locale of the program and it might break things.

Another option would be to use [Babel](http://babel.pocoo.org/en/latest/). It provides functions to format numbers, dates and other things by providing the locale as an argument.

So is it worth including this library ?
It could be used to format data that's supposed to be displayed without needing for developers to code a locale-specific function, they would only need to help TTS. Besides, even if there is no TTS helper, it should help it since I guess, for example, that a french TTS would expect to find decimal numbers with a comma and not a dot.

forslund:
Hi zecakeh, the language modules have been extracted into [lingua-franca](https://github.com/MycroftAI/lingua-franca/). Much development has happened on this front, but I think this discussion is probably better to have over there. I'm closing this issue and you can raise the remaining parts you think needs to be handled over there.

PR #1416, opened 2018-02-08
Issue-1375 - Fix and complete french translation

zecakeh:
## Description
fixes Issue # 1375
Update the current dialog texts and create the ones missing
Update format and parse by including format_fr and parse_fr

## How to test
test_parse_fr.py and test_format_fr.py included in unittests/util

CLA [Yes]


mycroft/util/lang/format_en.py
```diff
@@ -71,18 +72,34 @@
 }
 
 
-def nice_number_en(mixed):
+def nice_number_en(number, speech, denominators):
```
forslund: The comment below isn't quite up to date with these changes, can you update that as well?

zecakeh: It's just the description of the function, right ? The args and return value are up to date.
Also there's talk in the Issue #1411 that instead of moving the number formatting code to each language's code, we should just put a failsafe for every language in the main function in case it's not yet implemented.

forslund: Yes, the args and return value lookes good I just meant the (1, 1, 3) part.

And, ah humm I see what you mean now. It might be more clear to have it structured like you have here and then we can have a simpler nice_number in format.py. unless the text formatting becomes trivial with the use of `format` you suggest.

zecakeh: This use of <code>format</code> would actually mean the <code>nice_number</code> function would work for displays out of the box for every language and the language specific function would only be needed for speech to help the TTS.

forslund: Sounds good!

zecakeh: I just read on [python's locale doc](https://docs.python.org/2/library/locale.html#background-details-hints-tips-and-caveats) that it's not recommended to change the locale setting on the fly since it's expensive on resources and as a global value, it will affect the entire program, even temporarily.

Should it be set at startup when reading the conf file to allow us to use the format function for each language without worrying? Or should we just drop this for now and keep what's been done?

forslund:
A good plan. There's some conflicts at the moment could you try to resolve them?

zecakeh:
@forslund Conflicts solved and I updated the swedish translation to be also compatible with this PR

forslund:
Noticed that nice_number() now fails for sv-se it's missing the import of `convert_to_mixed_fraction`

zecakeh:
Oops you're right... Didn't notice it since there's no test_format_sv. Just fixed it!

forslund:
Yeah, that's my bad for not writing robust enough tests. Tack så mycket!

Merging!

PR #1420, opened 2018-02-13
Reduce CLI log clutter from settings,  etc

penrods:
* Drop settings INFO messages to DEBUG
* Add DEBUG filter in the CLI by default


forslund:
Looks great, the CLI is now eerily quiet. Merging

PR #1424, opened 2018-02-15
Add interface to Mark 1 faceplate capabilities

penrods:
This adds API interfaces for two Mark 1 faceplate capabilities to the
mycroft.client.enclosure.EnclosureAPI()

EnclosureAPI.setpixel(index, r,g,b)
- Set individual eye pixels to any color.  The indices go from 0-23 with
  the 0-12 corresponding to the right eye and 11-23 to the left.

EnclosureAPI.fill(percentage)
- Fill the eyes to a percentage, for use in meters or countdowns.  The
  right eye is 0-50%, the left eye also fills if going up to 100%.

### DIFF
--- a/mycroft/client/enclosure/api.py
+++ b/mycroft/client/enclosure/api.py
@@ -171,7 +171,7 @@ class EnclosureAPI:
             volume (int): 0 to 11
         """
         if volume < 0 or volume > 11:
-            raise ValueError('volume ({}) must be between 0-100'.
+            raise ValueError('volume ({}) must be between 0-11'.
                              format(str(volume)))
         self.ws.emit(Message("enclosure.eyes.volume", {'volume': volume}))
 
--- a/mycroft/client/enclosure/api.py
+++ b/mycroft/client/enclosure/api.py
@@ -117,6 +117,29 @@ class EnclosureAPI:
         self.ws.emit(Message("enclosure.eyes.color",
                              {'r': r, 'g': g, 'b': b}))
 
+    def eyes_setpixel(self, idx, r=255, g=255, b=255):
+        """Set individual pixels of the Mark 1 neopixel eyes
+        Args:
+            neopixel_idx (int): 0-11 for the right eye, 12-23 for the left
+            r (int): The red value to apply
+            g (int): The green value to apply
+            b (int): The blue value to apply
+        """
+        if idx < 0 or idx > 23:
+            return
+        self.ws.emit(Message("enclosure.eyes.setpixel",
+                             {'idx': idx, 'r': r, 'g': g, 'b': b}))
+
+    def eyes_fill(self, percentage):
+        """Use the eyes as a type of progress meter
+        Args:
+            amount (int): 0-49 fills the right eye, 50-100 also covers left
+        """
+        if percentage < 0 or percentage > 100:
+            return
+        self.ws.emit(Message("enclosure.eyes.fill",
+                             {'percentage': percentage}))
+
     def eyes_brightness(self, level=30):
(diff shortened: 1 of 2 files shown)


mycroft/client/enclosure/api.py
```diff
@@ -117,6 +117,29 @@ def eyes_color(self, r=255, g=255, b=255):
         self.ws.emit(Message("enclosure.eyes.color",
                              {'r': r, 'g': g, 'b': b}))
 
+    def eyes_setpixel(self, idx, r=255, g=255, b=255):
+        """Set individual pixels of the Mark 1 neopixel eyes
+        Args:
+            neopixel_idx (int): 0-11 for the right eye, 12-23 for the left
```
MatthewScholefield: Docstring doesn't match.

mycroft/client/enclosure/api.py
```diff
@@ -117,6 +117,29 @@ def eyes_color(self, r=255, g=255, b=255):
...
+            r (int): The red value to apply
+            g (int): The green value to apply
+            b (int): The blue value to apply
+        """
+        if idx < 0 or idx > 23:
+            return
```
MatthewScholefield: Would it make sense to `raise ValueError` here to let the programmer know they've made a mistake?

mycroft/client/enclosure/api.py
```diff
@@ -117,6 +117,29 @@ def eyes_color(self, r=255, g=255, b=255):
...
+        self.ws.emit(Message("enclosure.eyes.setpixel",
+                             {'idx': idx, 'r': r, 'g': g, 'b': b}))
+
+    def eyes_fill(self, percentage):
+        """Use the eyes as a type of progress meter
+        Args:
+            amount (int): 0-49 fills the right eye, 50-100 also covers left
+        """
+        if percentage < 0 or percentage > 100:
+            return
```
MatthewScholefield: Same `ValueError` idea as above

mycroft/client/enclosure/api.py
```diff
@@ -146,6 +169,8 @@ def eyes_volume(self, volume):
         Args:
             volume (int): 0 to 11
         """
+        if volume < 0 or volume > 11:
+            return
```
MatthewScholefield: Same

mycroft/client/enclosure/eyes.py
```diff
@@ -67,6 +69,24 @@ def color(self, event=None):
...
+            g = int(event.data.get("g", g))
+            b = int(event.data.get("b", b))
+        color = (r * 65536) + (g * 256) + b
+        self.writer.write("eyes.set=" + str(idx) + "," + str(color))
+
+    def fill(self, event=None):
+        amount = 0
+        if event and event.data:
+            percent = int(event.data.get("percentage", 0))
+            amount = round(23.0 * percent / 100.0)
```
MatthewScholefield: This doesn't seem to cause troubles, but I think we should cast to int since in python 2, `round` still returns a float.

penrods:
Addressed issues

mycroft/client/enclosure/api.py
```diff
@@ -146,6 +170,9 @@ def eyes_volume(self, volume):
         Args:
             volume (int): 0 to 11
         """
+        if volume < 0 or volume > 11:
+            raise ValueError('volume ({}) must be between 0-100'.
```
forslund: between 0 and 11?

MatthewScholefield: @penrods After that change, it looks good to go.

penrods:
Thanks for keeping me honest, guys.  Merging.

PR #1425, opened 2018-02-15
Fix cancel_scheduled_event for non-repeating events by updateing pyee to v5.0.0

forslund:
## Description
The old version of pyee (1.0.1) could not remove events registered as "once" (instead of "on")

This fixes cancelling scheduled events 

## How to test
Run the update-dev.sh script and create a test skill like below. Check that the *!!!!!!!!!!!! SCHEDULED HANDLER !!!!!!!!!!!!* message isn't shown.

```python
# -*- coding: utf-8 -*-
from datetime import datetime, timedelta
import time

class EventTestSkill(MycroftSkill):
    @intent_handler(IntentBuilder('').require('Testing'))
    def test_handler(self, message):
        LOG.info('Event in 10 secs!')
        self.schedule_event(self.handler,
                            datetime.now() + timedelta(seconds=10),
                            name='test-event')
        time.sleep(2)
        self.cancel_scheduled_event('test-event')

    def handler(self, message):
        LOG.info('!!!!!!!!!!!! SCHEDULED HANDLER !!!!!!!!!!!!')

def create_skill():
    return EventTestSkill()
```
CLA [Yes]

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -661,7 +661,7 @@ class MycroftSkill(object):
                     pass
                 try:
                     self.emitter.remove(_name, _handler)
-                except ValueError:
+                except (ValueError, KeyError):
                     LOG.debug('{} is not registered in the emitter'.format(
                               _name))
 
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -664,6 +664,8 @@ class MycroftSkill(object):
                 except (ValueError, KeyError):
                     LOG.debug('{} is not registered in the emitter'.format(
                               _name))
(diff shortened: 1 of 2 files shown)


penrods:
Hmm, this is odd.  On my Mark 1 I can cancel the event even without this code.  It doesn't seem dangerous to update, however.

forslund:
Hmm, let me recheck

forslund:
Found the issue and it wasn't what I thought, likely a rebasing issue. The ability to remove event handlers set up by `once` is however good since otherwise they may linger and cause multiple triggers.

Will update this PR.

PR #1426, opened 2018-02-15
Add default when for schedule_repeating_event()

penrods:
MycroftSkill.schedule_repeating_event(handler, when, frequency) now
will compute the 'when' as now+frequency seconds if None is passed.


forslund:
Works very well. Merging.

PR #1428, opened 2018-02-15
Fix self parameter in decorator functions with single argument

MatthewScholefield:
## Description
This fixes the self parameter in decorator functions with single argument.

## How to test
Create a skill that has a handler with a decorator but no message parameter:
```Python
    @intent_file_handler('my.intent')
    def handle_me(self):
        self.speak('hello')
```
You can use the `repeat recent` skill, which behaves this way. (`install repeat`, `what did I just say?`)
When activating this handler before the change, it errors saying `Message` has no attribute `speak`. After this change, it should work properly and not error.

 - [x] CLA


forslund:
Looks good, merging.

Issue #1083, opened 2017-09-14
CLI crashes if chat history height set too high


MatthewScholefield:
With the feature introduced by #1081, typing something like `:history 1000` will crash the CLI.

penrods:
I debated about what to do here and took the lazy/easy "asking it to do stupid things might provide stupid results", especially since this is a programmer tool.  But I probably should cap the use of this value based on the minimum number of lines needed to render other screen elements.

zachwelch:
With PR #1436 merged, this issue can be closed.

forslund:
Ok, closing!

PR #1436, opened 2018-02-21
issue #1083: limit history size to fix crash

zachwelch:
The calulated limit ensures that at least one message will be displayed.
Using a hard-coded constant is ugly, but such are scattered throughout
the script.  Untangling those constants has been left for future patches.

## Description
Fixes issue #1083

## How to test
Run the CLI and type `: history 1000`.

Without patch, CLI crashes inside a curses library call.

With patch, CLI displays maximal chat history lines and the most recent log message.

Requested


forslund:
Tested and works nicely, will merge as soon as the CLA is processed.

Thank you very much for helping out here!

PR #1439, opened 2018-02-23
Improve docs within mycroft-conf

penrods:
Clarified the merging process.  Also remove unused skill-specific settings.


forslund:
Small fix needed, but good now. Merging

PR #1351, opened 2018-01-07
Intent service data

JarbasAl:
## Description

adds new bus messages to allow getting data from intent service, for example in a manifest skill or in converse method to know what intent would trigger

works together with #1350 

"skill.manifest" - get loaded skills 
"skill.manifest.response" - {"skill id": "skill name"}
"vocab.manifest" - get loaded vocab
"vocab.manifest.response" - {"word": "MatchingKeyword"}
"intent.manifest" - get loaded intents
"intent.manifest.response" - {"skill id": ["intent1", "intent2]}
"intent.get" - get intent that will trigger - {"utterance": "an utterance", "lang":"en-us"}
"intent.response" - {"utterance": "utterance", "intent_data":{message of intent}}

## How to test

emit one of the messages and catch the response

## Contributor license agreement signed?
CLA [x ] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

PR #1440, opened 2018-02-24
Intent service data v2

zachwelch:
## Description
This PR replaces PR #1351 with freshly rebased and cleaned patches. See that PR for more information.

CLA [YES ]

### DIFF
--- a/mycroft/skills/intent_service.py
+++ b/mycroft/skills/intent_service.py
@@ -177,6 +177,25 @@ class IntentService(object):
         """
         return self.skill_names.get(int(skill_id), skill_id)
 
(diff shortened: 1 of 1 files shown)


zachwelch:
FWIW, I found a minor bug in this implementation, while using it to implement an introspection skill. When skills are reloaded, the relevant list in the `intent_map` did not get cleared. Each time, the intents for the skill were re-registered and added to the end of that list.

As such, I think this PR needs to be closed until such a time as I post a revised branch that fixes that bug (and any others that I might discover while working on the introspection skill). I will re-open this PR (or file a new one?) when the patches are ready for further consideration. Sorry for the noise.

PR #1441, opened 2018-02-24
Change precise executable to a folder

MatthewScholefield:
This prevents self-extraction of the precise executable each bootup which reduces SD card load and lowers boot time. On a technical side, this changes the self contained executable to a .tar.gz which is extracted once during first boot.

### DIFF
--- a/mycroft/client/speech/hotword_factory.py
+++ b/mycroft/client/speech/hotword_factory.py
@@ -16,17 +16,18 @@ import tempfile
 import time
 
 import sys
+from tempfile import NamedTemporaryFile
+
 import os
 from os.path import dirname, exists, join, abspath, expanduser, isdir, isfile
 
-from os import mkdir
+from os import mkdir, getcwd, chdir
 from time import time as get_time
 
 from mycroft.configuration import Configuration
 from subprocess import Popen, PIPE, call
 from threading import Thread
 
-from mycroft.util import resolve_resource_file
 from mycroft.util.log import LOG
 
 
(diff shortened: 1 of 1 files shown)


mycroft/client/speech/hotword_factory.py
```diff
@@ -142,16 +143,17 @@ def get_model_info(self):
         return model_name, model_path
 
     def find_download_exe(self):
-        exe_file = resolve_resource_file(self.exe_name)
-        if exe_file:
-            return exe_file
         try:
-            if call(self.exe_name + ' < /dev/null', shell=True) == 0:
+            if call(['command', '-v', self.exe_name], stdout=PIPE) == 0:
```
forslund: Will this ever work without `shell=True`? `command` is not an application as far as I can read but implemented in the shell?

```python
>>> from subprocess import call, PIPE
>>> call(['command', '-v', 'bash'], stdout=PIPE)
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/subprocess.py", line 523, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/usr/lib/python2.7/subprocess.py", line 711, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1343, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
>>> call(['command', '-v', 'bash'], shell=True, stdout=PIPE)
0
>>>
```

MatthewScholefield: Nice catch. Fixed.

forslund:
Looks good, merging!

PR #1443, opened 2018-02-25
Fix hash method to work with python3

antlarr:
## Description
In python3, unicode objects have to be encoded before being
hashed with hashlib so this commit fixes the hash method so
it works with both python2 and python3.

## How to test
I run mycroft-core with python3 and saw an exception being raised:
TypeError: Unicode-objects must be encoded before hashing
After applying this commit, it works successfully.

### DIFF
--- a/mycroft/skills/settings.py
+++ b/mycroft/skills/settings.py
@@ -307,7 +307,10 @@ class SkillSettings(dict):
 
     def hash(self, str):
         """ md5 hasher for consistency across cpu architectures """
-        return hashlib.md5(str).hexdigest()
+        if type(str) == bytes:
+            return hashlib.md5(str).hexdigest()
+        else:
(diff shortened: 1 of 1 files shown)


forslund:
Hi @Antlarr, thanks for the PR. We've prepared a bunch of fixes for python 3 in the 18.02 branch. Can you check that the issue remains in there? and in that case resubmit for that branch?

antlarr:
Ah, I noticed the python3 branch (which is very old) but not that 18.02 branch. When is it planned to be merged to dev ?

antlarr:
Btw, the issue is fixed in the 18.02 branch. Great. Althought for python3 to work, the skillls need also to support python3 with fixes such as https://github.com/MycroftAI/fallback-wolfram-alpha/pull/11

penrods:
Yes, before 18.02 we have officially been a

PR #1444, opened 2018-02-26
only NTP sync on certain enclosures

zachwelch:
## Description
This patch eliminates a 15 second delay when starting the skill server on platforms that do not implement a handler for the `system.ntp.sync` message.

## How to test
Run it on Linux and watch how much faster it starts up

CLA [YES]

### DIFF
--- a/mycroft/skills/main.py
+++ b/mycroft/skills/main.py
@@ -133,9 +133,13 @@ def check_connection():
             # Skip the sync message when unpaired because the prompt to go to
             # home.mycrof.ai will be displayed by the pairing skill
             enclosure.mouth_text(mycroft.dialog.get("message_synching.clock"))
+
         # Force a sync of the local clock with the internet
-        ws.emit(Message("system.ntp.sync"))
-        time.sleep(15)   # TODO: Generate/listen for a message response...
+        config = Configuration.get()
+        platform = config['enclosure'].get("platform", "unknown")
+        if platform in ['mycroft_mark_1', 'picroft']:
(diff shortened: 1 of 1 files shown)


mycroft/skills/main.py
```diff
@@ -133,9 +133,13 @@ def check_connection():
             # Skip the sync message when unpaired because the prompt to go to
             # home.mycrof.ai will be displayed by the pairing skill
             enclosure.mouth_text(mycroft.dialog.get("message_synching.clock"))
+
         # Force a sync of the local clock with the internet
-        ws.emit(Message("system.ntp.sync"))
-        time.sleep(15)   # TODO: Generate/listen for a message response...
+        config = Configuration.get()
+        platform = config.get("enclosure", {}).get("platform", "unknown")
```
MatthewScholefield: We can assume `"enclosure"` exists in the config since it is written in `mycroft/configuration/mycroft.conf`, so this can be written as `config["enclosure"].get(...`.

mycroft/skills/main.py
```diff
@@ -133,9 +133,13 @@ def check_connection():
...
+        if platform in ['mark-1', 'picroft']:
```
MatthewScholefield: The official platform names are actually `mycroft_mark_1` and `picroft`. This was confusing because they are actually documented wrong in the config file (but I've created [a PR](https://github.com/MycroftAI/mycroft-core/pull/1448/files) to fix this).

MatthewScholefield:
Thanks for the changes. Merging

PR #1445, opened 2018-02-26
improve output when no words are transcribed

zachwelch:
## Description
This patch improves the logging output of the speech client, displaying a meaningful message when no words are transcribed.

## How to test
Say `Hey Mycroft` and say nothing.  Without this patch, the log file records an `IndexError` message that can confuse users. With the patch, it logs "no words were transcribed".

CLA [Y]


MatthewScholefield:
Nice improvement!

PR #1446, opened 2018-02-27
Remove redundant except (SystemExit, KeyboardInterrupt) clauses

MatthewScholefield:
## Description
I realized that it's not actually necessary to add the SystemExit and KeyboardInterrupt except clauses if the broad clause uses `except Exception`. This works because SystemExit and KeyboardInterrupt do not inherit from the base `Exception` class

## How to test
Make sure the wake word still falls back to `hey mycroft` when using incorrect phonemes in Pocketsphinx.

### DIFF
--- a/mycroft/audio/main.py
+++ b/mycroft/audio/main.py
@@ -78,8 +78,6 @@ def get_services(services_folder):
                     continue
                 try:
                     services.append(create_service_descriptor(name))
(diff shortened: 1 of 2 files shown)


forslund:
Teaches me for trusting what I read on the net :)

Looks good, merging.

PR #1448, opened 2018-02-27
Fix platform comment in config

MatthewScholefield:
## Description
This adds the real, valid options to the config to prevent confusion

## How to test
Make sure the config loads properly and something like `what time is it` still works


forslund:
Looks good and loads as expected!

PR #1452, opened 2018-02-28
Call callback when remote settings are inited

forslund:
## Description
When a fresh image first updates the settings after pairing it is handled somewhat differently and this was missed in the original implementation of the callback handling.

This minor change includes this case as well.

## How to test

Make sure callbacks still function as they normally do.

CLA [Yes]


penrods:
Merging!

PR #1453, opened 2018-02-28
Msm update improvements

MatthewScholefield:
## Description
This improves msm update with the following changes:
 - Run git updates in parallel (~6x faster)
 - Ignore updating git repos that use ssh (this prompts for a password in most cases)
 - Clarify update logic
 - Nicer output (shows what updates and what doesn't)

## How to test
Run `msm/msm update` and compare speeds. Also, make skill modifications to verify msm skips those skills.

### DIFF
--- a/msm/msm
+++ b/msm/msm
@@ -439,54 +439,113 @@ function search() {
   done < <(grep -i "${search_string}" <<< "${search_list}")
 }
 
+function should_update_git() {
+  local branch=$(git symbolic-ref --short HEAD)
+  if [ "$branch" != "master" ]; then
+    return 1
+  fi
+
+  local changed_files=$(git status --porcelain --untracked-files=no)
+  if [ -n "$changed_files" ]; then
+    return 1
+  fi
+
+  local upstream=${1:-'@{u}'}
+  local local=$(git rev-parse @)
+  local remote=$(git rev-parse "$upstream")
+  local base=$(git merge-base @ "$upstream")
+  if [ "$local" != "$remote" ] &&  # New commits
+     [ "$remote" = "$base" ]  # Has not diverged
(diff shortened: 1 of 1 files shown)


penrods:
Let's merge this AFTER the initial 18.02 release

msm/msm
```diff
@@ -439,61 +439,120 @@ function search() {
...
-         echo "Re-running PIP on requirements.txt"
-      fi
+    git_dir=$(git -C "$d" rev-parse --git-dir 2>&1)
+    if [ "$git_dir" = ".git" ]; then
+      update_skill "$d" &
+      pids="$pids $!"
     fi
   done
 
+  wait $pids
```
forslund: The wait statement doesn't seem to work quite right:
```
./msm/msm: line 548: wait: `22635 22642 22647 22657 22667 22676 22683 22694 22707 22714 22721 22732 22746 22754 22762 22770 22778 22793 22804 22815 22828 22837 22856 22869 22879 22910 22932 22945 22957 22974 22986 23001 23011 23028 23038 23054 23066 23077 23096 23124 23133': not a pid or valid job spec
```

And no wait occurs.

MatthewScholefield: Hmm, seems to be a difference between out bash versions. See if the change I made fixes this for you.

forslund:
Doesn't quite work for me still. Seems like all pids become a single string instead of an array. With some experimenting I got something that seems to be working for me (TM): https://github.com/forslund/mycroft-core/commit/76e0eba41c62f812082ef5d3a1bf7aef5763123f

MatthewScholefield:
@forslund Looks like that also works for me so I've pulled in that commit.

PR #1454, opened 2018-03-01
Use function attributes for intent decorators

MatthewScholefield:
## Description
This prevents needing to use a shared list which misbehaves when multiple skills initialize at once

There's a chance this resolves issues with intents activating the wrong skill

## How to test
Test activating both Adapt and Padatious intents. Make sure there are no new errors in the logs.


forslund:
That was one of my initial thoughts, that much of the special logic for need_self could be removed in the add_event method. but probably a good call to leave that out for now.

PR #1456, opened 2018-03-01
Add proper not loaded message

MatthewScholefield:
## Description
Once skills load, skill-unknown will handle all failed fallbacks. The only time this is spoken is when skills still aren't loaded yet

## How to test
Ask Mycroft a question before the bootup sequence is finished

### DIFF
--- a/mycroft/client/speech/main.py
+++ b/mycroft/client/speech/main.py
@@ -15,6 +15,7 @@
 import sys
 from threading import Thread, Lock
 
+from mycroft import dialog
 from mycroft.client.enclosure.api import EnclosureAPI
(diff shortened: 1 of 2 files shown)


mycroft/client/speech/main.py
```diff
@@ -80,8 +81,7 @@ def handle_speak(event):
 def handle_complete_intent_failure(event):
     LOG.info("Failed to find intent.")
     # TODO: Localize
```
penrods: You can remove this TODO comment with this change

MatthewScholefield: Fixed!

PR #1458, opened 2018-03-01
Refactor handler argument calling

MatthewScholefield:
## Description
This simplifies the logic in `add_event`, specifically for calling the handler. It also removes the `has_self` parameter of `add_event` which could break any skills that use it with that parameter.

## How to test
Test if skills work with this change

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -580,6 +580,7 @@ class MycroftSkill(object):
 
         def wrapper(message):
             skill_data = {'name': get_handler_name(handler)}
+            stopwatch = Stopwatch()
             try:
                 message = unmunge_message(message, self.skill_id)
                 # Indicate that the skill handler is starting
@@ -588,7 +589,6 @@ class MycroftSkill(object):
                     msg_type = handler_info + '.start'
                     self.emitter.emit(Message(msg_type, skill_data))
 
-                stopwatch = Stopwatch()
                 with stopwatch:
                     is_bound = bool(getattr(handler, 'im_self', None))
                     num_args = len(getargspec(handler).args) - is_bound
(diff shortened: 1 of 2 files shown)


mycroft/skills/core.py
```diff
@@ -591,45 +590,21 @@ def wrapper(message):
...
                 # Convert "MyFancySkill" to "My Fancy Skill" for speaking
-                handler_name = re.sub("([a-z])([A-Z])", "\g<1> \g<2>",
-                                      self.name)
-                # TODO: Localize
-                self.speak("An error occurred while processing a request in " +
-                           handler_name)
-                LOG.error("An error occurred while processing a request in " +
-                          self.name, exc_info=True)
+                handler_name = re.sub(r"([a-z])([A-Z])", r"\1 \2", self.name)
+                data = {'skill': handler_name}
```
forslund: This overwrites the data prepared previously for the skill.handler.started / .stopped messages (the `name` key will be slightly different)

mycroft/skills/core.py
```diff
@@ -591,45 +590,21 @@ def wrapper(message):
...
-                            try:
-                                handler(self, unmunge_message(message,
-                                                              self.skill_id))
-                            except TypeError:
-                                handler(self)
-                        else:
-                            LOG.error("Unexpected argument count:" +
-                                      str(len(getargspec(handler).args)))
-                            raise TypeError
+                    is_bound = bool(getattr(handler, 'im_self', None))
```
forslund: May possibly be simplified slightly by using 
`is_bound = hasattr('im_self')`

MatthewScholefield: From what I understand, if a class method is given accessed statically via `ClassName.function`, `im_self` will exists but be `None`.

forslund: Ah, right you are. Would fail for classmethods

mycroft/skills/core.py
MatthewScholefield: Nice catch! Fixed.

forslund:
Excellent! Merging!

PR #1459, opened 2018-03-02
Fix local service crash when servers are down

MatthewScholefield:
## Description
When the servers are not reacting properly, server requests can throw a `ReadTimeout` exception which was previously not handled in a few places. This PR changes the except to cover all request exceptions, in particular, preventing a crash if the speech service restarts during a server failure.

## How to test
 - Mess up the servers
 - Restart your services
 - Fix the servers
 - Make sure the speech client still works

### DIFF
--- a/mycroft/client/speech/listener.py
+++ b/mycroft/client/speech/listener.py
@@ -17,7 +17,7 @@ from threading import Thread
 import sys
 import speech_recognition as sr
 from pyee import EventEmitter
-from requests import HTTPError
+from requests import RequestException
 from requests.exceptions import ConnectionError
 
 import mycroft.dialog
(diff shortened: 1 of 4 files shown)


mycroft/api/__init__.py
```diff
@@ -86,7 +86,7 @@ def get_response(self, response):
                 and not response.url.endswith("auth/token"):
             self.refresh_token()
             return self.send(self.old_params)
-        raise HTTPError(data, response=response)
+        raise RequestException(data, response=response)
```
forslund: What is the reasoning for raising RequestException instead of HTTPError here? since HTTPError is a subclass of RequestException is shouldn't be strictly necessary right?

In any case I think the pairing and configuration skills need to be updated before merging this since they check for HTTPError.

MatthewScholefield: Nice catch. Yeah, it doesn't make much sense to broaden the raise clause; only the except clauses. I've changed it to how it was before.

forslund:
Looks good! Merging

PR #1468, opened 2018-03-12
Converse deactivate  calback

JarbasAl:
## Description
adds a callback for when a skill is removed from active skills list

allows for example a skill to keep itself always active, or reset states if it was expecting to use converse method but didn't 

Used in my [utils_package](https://github.com/JarbasAl/mycroft_jarbas_utils/tree/master/mycroft_jarbas_utils/skills) to allow you to do things like passive skills that are always active

## How to test

make a skill, override on_deactivate method to Log something, trigger an intent from that skill, wait 5 minutes, check that log is there

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -261,7 +261,7 @@ class MycroftSkill(object):
         if emitter:
             self.emitter = emitter
             self.enclosure = EnclosureAPI(emitter, self.name)
-            self.add_event("converse.deactivate", self.on_deactivate)
+            self.add_event("converse.deactivate", self._deactivate_skill)
             self.__register_stop()
 
     def __register_stop(self):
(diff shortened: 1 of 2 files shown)


JarbasAl:
i have an idea for a skill that needs this to ensure it stays on top of active skill list, it would intercept all utterances and sometimes mutate them, there are a few applications for this:

    secret speech, you can have mycroft do something totally unrelated to what you just asked for specific utterances "what time is it?" -> starts cracking wifi and says the time
    auto correct common failures, if the stt often thinks that you said something you never say you can change it back to the intended utterance
    expanded vocabulary, you can trigger skills in a specific way, "play morning jams" -> "play X from youtube in Y place"

i could serve some local web ui for you to easily add new utterances, in your phone for example

JarbasAl:
this would be really useful for some skills im working on. Im creating a daemon to make my skill active every minute, id rather have a callback than a thread...

i can rebase this, if someone reviews it...

forslund:
I think a callback is a good addition. Now I know more about the converse system and could do a preliminary review (letting a Mycroft member have the final word)

PR #1477, opened 2018-03-14
Slowing the skill setting poll to once every 5 min

penrods:
Experiment to see if this is overloading the backend.


forslund:
Merging

PR #1487, opened 2018-03-17
fix audio service previous

JarbasAl:
## Description
Previous track is not working because audio service uses .prev() and the base class was updated to .previous()

## How to test
play some playlist with audio service, test by saying "next track" and "previous track"


forslund:
Thanks, looks more correct! Merging

PR #1488, opened 2018-03-17
clean audio service shutdown

JarbasAl:
## Description
Adds a shutdown method to AudioBackend, this may be required by some backends that interact with other programs

Also cleanly removes listeners

## How to test
Currently no audio backend needs this, in the mpv backend PR #1489  this is necessary and correctly performs shutdown, without this audio continues playing

### DIFF
--- a/mycroft/audio/main.py
+++ b/mycroft/audio/main.py
@@ -449,6 +449,31 @@ class AudioService(object):
                 self.pulse_quiet = self.pulse_lower_volume
                 self.pulse_restore = self.pulse_restore_volume
 
+    def shutdown(self):
+        for s in self.service:
+            try:
+                LOG.info('shutting down ' + s.name)
+                s.shutdown()
+            except Exception as e:
+                LOG.error('shutdown of ' + s.name + ' failed: ' + str(e))
+
+        # remove listeners
+        self.ws.remove('mycroft.audio.service.play', self._play)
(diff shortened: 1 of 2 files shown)


mycroft/audio/main.py
```diff
@@ -449,6 +449,28 @@ def setup_pulseaudio_handlers(self, pulse_choice=None):
                 self.pulse_quiet = self.pulse_lower_volume
                 self.pulse_restore = self.pulse_restore_volume
 
+    def shutdown(self):
+        for s in self.service:
+            LOG.info('shutting down ' + s.name)
+            s.shutdown()
```
forslund: Add a try-except block here so a failure in one backend won't cause others to not shutdown.

JarbasAl:
changes made, should i also add a call to the stop method before shutdown? or do i assume the shutdown method always implements it like it should?

forslund:
I think assuming the backend handles stop is the best way. Stop shouldn't be forced by the system here

JarbasAl:
maybe add a call to stop in the backend template class in case some backend doesn't implement this method?

forslund:
That could be a good idea. I guess the times you don't want to do stop is when working against a "remote" backend like the chromecast or mopidy and in these cases they can implement a shutdown that don't call stop.

So go ahead and add that to the service class!

forslund:
Rebased and is now merging!

PR #1490, opened 2018-03-17
feature: audio backend mplayer

JarbasAl:
## Description

Adds mplayer audio backend

## How to test

install mplayer and python api

`sudo apt-get install mplayer`

`pip install git+https://github.com/JarbasAl/py_mplayer`

Change your config file 

```
"Audio": {
    "backends": {
      "local": {
        "type": "mpg123",
        "active": true
      },
      "vlc": {
        "type": "vlc",
        "active": true
      },
      "mplayer": {
        "type": "mplayer",
        "active": true
      }
    },
    "default-backend": "mplayer"
  }
```

## Contributor license agreement signed?
CLA [ yes] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

PR #1492, opened 2018-03-19
bugfix: honor "active" flag for audio backends

JarbasAl:
## Description
all audio backends are loaded,  the "active" flag is currently ignored, this PR fixes this

chromecast uses an auto discover mechanism and ignores this

#1489 and #1490 need changes to reflect this

## How to test
in config file set a audio backend to not active, verify it does not load in the audio service logs

### DIFF
--- a/mycroft/audio/services/mopidy/__init__.py
+++ b/mycroft/audio/services/mopidy/__init__.py
@@ -115,6 +115,7 @@ class MopidyService(AudioBackend):
 def load_service(base_config, emitter):
     backends = base_config.get('backends', [])
     services = [(b, backends[b]) for b in backends
-                if backends[b]['type'] == 'mopidy']
+                if backends[b]['type'] == 'mopidy' and
+                backends[b].get('active', True)]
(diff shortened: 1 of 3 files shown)


forslund:
The active flag was kind of orphaned (and should have been removed from the config as well) when the backends went from explicit declarations to discovery based. Do you have a need to deactivate backends completely?

JarbasAl:
i would use it during dev, but since config can be changed at runtime i can easily imagine a skill that manages backends by toggling this flag, and some may conflict with each other

also allows to provide some default configs without having the backend load

forslund:
Ok...

Then I suggest we use `backends[b].get('active', True)` so if config exists but active is unset it's treated as active.

JarbasAl:
done, should i update the other PRs then?

forslund:
Shouldn't be needed since the requirements aren't there they won't be loaded...

JarbasAl:
and the logs will get dirty with exceptions and you flooded with even more questions : D

forslund:
Looks good, merging

PR #1493, opened 2018-03-20
allow google stt by default, using test key

JarbasAl:
## Description

if no key is provided for google stt, the speech recognition python package uses a default test key that works

because of the type casting to string an invalid key is passed, this PR changes it in order for google STT to work out of the box

## How to test

change stt module to google, verify it works without key

### DIFF
--- a/mycroft/stt/__init__.py
+++ b/mycroft/stt/__init__.py
@@ -84,6 +84,8 @@ class KeySTT(STT):
 class GoogleSTT(TokenSTT):
     def __init__(self):
         super(GoogleSTT, self).__init__()
+        if self.token == "None":
+            self.token = None
 
     def execute(self, audio, language=None):
         self.lang = language or self.lang


forslund:
speechrecognition writes:
*The Google Speech Recognition API key is specified by ``key``. If not specified, it uses a generic key that works out of the box. This should generally be used for personal or testing purposes only, as it **may be revoked by Google at any time**.*

As I've understood it this is for persons testing out the speechrecognition module and need a quick start before they get their own key, so I'm not sure we should allow it.

JarbasAl:
going trough the trouble of changing stt in a config file qualifies as personal use doesn't it?

better have it and get it revoked than not have it, also i assume the package would replace it if revoked?

google moved to google voice cloud services, which uses a json config, im not sure you can even get an api key

this module is pretty much getting deprecated i guess, so there is no harm in squeezing the juice out of it

forslund:
That's my opinion, the purpose of the default key is to do initial tests with the speechrecognition module.

MatthewScholefield:
One thing to add on this topic, however, is that it does seem hacky that we cast None to a string in the STT module. Perhaps we should look at removing that, and then adding an explicit exception to prevent using the default key.
As a side note, I also agree we should explicitly disallow this, not for Google's sake, but so that a few users don't end up switching to the default key, getting it revoked so that no other users of the `speech_recognition` library can use it.

Issue #1494, opened 2018-03-21
Bug: opt_in not being honored in intent_service for send_metrics


InconsolableCellist:
The configuration parameter "opt_in" (added in 0569ba40f7213e) isn't being honored in the send_metrics function in intent_service.py:238, contrary to the comment (added in c68ad44b1c64).

Steps to reproduce:

1. Verify "opt_in" is set to false in mycroft.conf:39 (top level)
2. Start mycroft in a manner that lets you see the logs, and issue an utterance

Observed behavior:

While handling an utterance the log prints out "Sending metric," indicating that the code in handle_metrics is executing, including the call to report_timing, and thus eventually the call to the API to make a request to the external Mycroft endpoint.

Expected behavior:

No such metrics are sent when opt_in isn't set to true, and no log message is displayed

forslund:
I tested it, and no metrics is actually sent. The check is done in the send function:
https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/metrics/__init__.py#L38

The debug message is a bit misleading though...

penrods:
I verified and commented with details on the PR.  The metrics are not sent, but I recommend we remove that debugging log.  @InconsolableCellist can modify his PR if he'd like to do that.

PR #1495, opened 2018-03-21
Bugfix for 1494: opt_in not honored in intent

InconsolableCellist:
The configuration parameter "opt_in" (added in 0569ba4) wasn't being
honored in the send_metrics function in intent_service.py:238, contrary
to the comment (added in c68ad44).

==== Fixed Issues ====
1494

## How to test
See #1494 message for steps to reproduce and expected behavior. Change opt_in parameter to true and false, verify "sending metrics" appears in the former and not the latter.

NO: Request submitting, waiting on a response from team.

### DIFF
--- a/mycroft/skills/intent_service.py
+++ b/mycroft/skills/intent_service.py
@@ -126,6 +126,7 @@ class ContextManager(object):
 class IntentService(object):
     def __init__(self, emitter):
         self.config = Configuration.get().get('context', {})
+        self.config_core = Configuration.get()
         self.engine = IntentDeterminationEngine()
 
         # Dictionary for translating a skill id to a name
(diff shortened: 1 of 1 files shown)


penrods:
Hey @InconsolableCellist, I commented on this in the forum last night but will repeat here:

That log message is misleading.  If you look at the implementation of report_metric you'll see that the very first thing it does is check the "opt_in" value (much like your code).  https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/metrics/__init__.py#L38 

When Opt In is on I see messages like this in my logs:
```
 https://api.mycroft.ai:443 "POST /v1/device/12345678-1234-1234-1234-123456789012/metric/timing HTTP/1.1"
```

But after I toggle the Opt In on the web, save, and execute "Hey Mycroft, update your configuration" I no longer get these in my logs.  So I am certain that we are honoring the Opt In flag.   We really just need to dump that poorly worded debugging message.  The additional check of the Opt In value is unnecessary.

InconsolableCellist:
Thanks, not sure how I missed that. I'm proposing a much simpler change: modifying the text to just avoid this confusion in the future. To avoid muddying up the git commit history with a check-in and then a commit that undoes it, I'm making a new PR and closing this one.

PR #1497, opened 2018-03-21
Simple log output change to prevent confusion

InconsolableCellist:
Modified the debugging message output after an utterance is complete, to
make it clear that metric data is only sent if opt_in is enabled.

This addresses #1494, which I filed in error, thinking that opt_in was
never checked (due to reading the log output and not knowing it was
checked elsewhere)

==== Fixed Issues ====

==== Localization Notes ====
Slight change to existing, unlocalized text

CLA [YES]

### DIFF
--- a/mycroft/skills/intent_service.py
+++ b/mycroft/skills/intent_service.py
@@ -241,7 +241,7 @@ class IntentService(object):
 
         NOTE: This only applies to those with Opt In.
         """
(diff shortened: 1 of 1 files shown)


InconsolableCellist:
@penrods FYI here's the new PR, super simple change. Also signed the CLA last night.

penrods:
LGTM!

PR #1498, opened 2018-03-22
Fix msm lock hanging forever

MatthewScholefield:
## Description
Sometimes the `send_end_update` would happen before the `send_start_update` and due to the `install_started` being `false`, `msm.updated` would never get sent

## How to test
Start the Mycroft services a few times. You will notice, without this change, once every few bootups skils never load. With this change, they consistently load.


forslund:
Looks good, merging

PR #1500, opened 2018-03-23
Reduce audio service logging verbosity

forslund:
## Description
- Duplicates and non-informative messages removed
- Do not output stack traces when failing to load a backend
- Many info messages reduced to debug level

## How to test
Start the audio service and make sure it's working, see that it's not quite as verbose as it used to be.

CLA [Yes]

### DIFF
--- a/mycroft/audio/main.py
+++ b/mycroft/audio/main.py
@@ -112,8 +112,9 @@ def load_services(config, ws, path=None):
         try:
             service_module = imp.load_module(descriptor["name"] + MAINMODULE,
                                              *descriptor["info"])
-        except Exception:
-            LOG.error('Failed to import module ' + descriptor['name'])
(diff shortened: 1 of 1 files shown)


mycroft/audio/main.py
```diff
@@ -124,15 +122,13 @@ def load_services(config, ws, path=None):
                 s = service_module.autodetect(config, ws)
                 service += s
             except Exception:
-                LOG.error('Failed to autodetect...',
-                          exc_info=True)
+                LOG.error('Failed to autodetect...')
```
MatthewScholefield: Should we at least print `e.__name__ + ': ' + str(e)` or something, here and below?

forslund: Probably a good idea, yes

PR #1502, opened 2018-03-23
AudioService - no old style class

JarbasAl:
## Description

Audio service does not inherit from "object" , this was causing me some random troubles when sub-classing it and using super

use-case (fixed to not use super anymore) -
 https://github.com/JarbasAl/mycroft_jarbas_utils/blob/master/mycroft_jarbas_utils/skills/audio.py#L10

## How to test

check everything works the same


forslund:
Good catch! Merging!

PR #1504, opened 2018-03-23
honor skill update flag

JarbasAl:
## Description

#1342 is ignored in direct update needed, this keeps messing up my dev environment by downloading skills on boot

## How to test
disable auto skills update, check that on boot skills are not auto downloaded (in fresh install, else delete .msm file and some default skill)

CLA [ x] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

### DIFF
--- a/mycroft/skills/main.py
+++ b/mycroft/skills/main.py
@@ -67,19 +67,23 @@ def direct_update_needed():
     dot_msm = join(SKILLS_DIR, '.msm')
     hours = skills_config.get('startup_update_required_time', 12)
     LOG.info('TIME LIMIT {}'.format(hours))
-    # if .msm file is missing or older than 1 hour update skills
-    if (not exists(dot_msm) or
-            os.path.getmtime(dot_msm) < time.time() - 60 * MINUTES * hours):
-        return True
-    else:  # verify that all default skills are installed
-        with open(dot_msm) as f:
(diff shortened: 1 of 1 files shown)


forslund:
I can't quite repeat the issue...I've tried a couple of times now, both with timelimit and removing `.msm`.

My procedure:
Add 
```json
  "skills": {
    "auto_update": false
  }
```
to my mycroft.conf and then
```
rm /opt/mycroft/skills/.msm
rm -rf /opt/mycroft/skill-version-checker
./start-mycroft.sh skills
```
Check that version checker isn't re-added
(and check logging for a debug message I've added)

`download_skills()` is only called from one place as far as I can see and will only be called if the update flag is True, is there something wrong with this logic or is there some other place where the method is called?

I think the check can be inlined on [L259](https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/skills/main.py#L259) instead of adding the logic to the update_needed() method.

`if update and direct_update_needed():`

Keeping the logic for if updates are disabled in the SkillManager makes more more sense than adding it to the logic for checking the flag-file in the direct_update_needed() function

JarbasAl:
i think this might have been because of #1513, because i did not have the default skills in my dev dir it never finished booting up, closing for now, will reopen if i manage to replicate

PR #1505, opened 2018-03-24
Pulseaudio update

forslund:
## Description
Replaces the current approach of modifying streams on the go with an approach using pulse audio's built in support for audio ducking. This will mute or lower the volume of streams with certain media roles (music and video by default).

To trigger the ducking/corking when mycroft is listening the audioservice launches `paplay` with a dummy input.

## How to test
Load the module-role-ducking:
`pactl load-module module-role-ducking`
Enter the following into your mycroft.conf:
```json
  "Audio": {
    "pulseaudio": true
  }
```

Start up Mycroft and start the npr-news. Why playing say "Hey Mycroft", the news volume should be reduced while mycroft listens. The volume should also be lowered when Mycroft starts speaking.

CLA [Yes]

### DIFF
--- a/mycroft/audio/audioservice.py
+++ b/mycroft/audio/audioservice.py
@@ -416,6 +416,7 @@ class AudioService(object):
                        self._restore_volume)
         self.ws.remove('recognizer_loop:record_end', self._restore_volume)
         self.ws.remove('mycroft.stop', self._stop)
+        self._restore_volume()
 
 
 def main():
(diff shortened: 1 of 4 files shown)


j1nx:
Ok, then I leave the rebase to you so I can just easily download the PR as patch and apply for some testing on my own.

Thx.

PS. I think this PR also makes it easier to start playing with other PulseAudio plugins right?

forslund:
I think so. It basically names the mycroft streams consistently so pulse plugins can handle them separately.

forslund:
@j1nx sorry for the delay. It's now rebased and tested.

j1nx:
@forslund OK, great! Atm not able to test this out, bu should be anytime now. Will report back as soon as I have some feedback.

PR #1508, opened 2018-03-28
Add proper messages when the backend replies incorrectly

MatthewScholefield:
## Description
This PR notifies the user whether either the backend is down or there is no internet connection on both bootup and when speech to text fails.

## How to test
 - Check if it replies something sensible when either the network is down or the server url is wrong.


forslund:
Looks good to me! Merging

PR #1509, opened 2018-03-28
Wait for messagebus before emitting a message

MatthewScholefield:
## Description
This PR waits for the messagebus connection to open before emitting messages. Previously the behavior was to silently return without sending the message.

## How to test
 - Make sure all functionality is normal
 - Try to send a message right after creating a websocket client and make sure it doesn't eat it up


forslund:
Looks excellent. Merging.

PR #1519, opened 2018-04-03
Shutdown improvements

MatthewScholefield:
## Description
This cleans up the shutdown process so that skill shutdown methods are called properly. It is now no longer necessary to call `MycroftSkill.shutdown` from within skill shutdown methods. Other notable changes include:
 - `create_daemon` method used to clean up create daemon threads
 - `create_echo_function` used to reduce code duplication with messagebus
 echo functions
 - `wait_for_exit_signal` used to wait for ctrl+c (SIGINT)
 - `reset_sigint_handler` used to ensure SIGINT will raise `KeyboardInterrupt`

## How to test
Start and stop all services. Make sure each process both shuts down properly and doesn't consistently have to kill the services (there is a new message to show when this happens)

### DIFF
--- a/mycroft/audio/main.py
+++ b/mycroft/audio/main.py
@@ -18,24 +18,24 @@
     This handles playback of audio and speech
 """
 import imp
-import json
 import sys
 import time
-
 from os import listdir
+
 from os.path import abspath, dirname, basename, isdir, join
 
 import mycroft.audio.speech as speech
 from mycroft.configuration import Configuration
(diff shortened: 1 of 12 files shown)


MatthewScholefield:
@penrods Okay, this has been updated to both perform proper behavior when shutting down and use a renamed `_shutdown` method of `MycroftSkill`.

forslund:
Looks pretty good and works as intended. The codacy things should be ok. I added some extra safety when calling the skill specific shutdown method to allow the generic one to always execute.

Merging

PR #1520, opened 2018-04-03
Feature/api etags

MatthewScholefield:
## Description
This adds support for html ETags which decrease the bandwidth of requests by replying with an empty response when the request contains the same content as it did before.

## How to test
 - Make sure the config and skill settings still update properly on the production servers
 - Switch the `server.url` config value to `https://api-test.mycroft.ai` and test if settings and skill settings still works
 - Verify etags update the first time and the say the settings haven't changed subsequent times in the logs
Tip: Decrease polling seconds in `mycroft/skills/settings.py:399`

### DIFF
--- a/mycroft/api/__init__.py
+++ b/mycroft/api/__init__.py
@@ -14,6 +14,7 @@
 #
 from copy import copy
 
+import json
 import requests
 from requests import HTTPError, RequestException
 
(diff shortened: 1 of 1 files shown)


forslund:
Looks good as far as I can see. I was a bit worried about the shared dicts could be a problem but I guess the worst that can happen is that data is re-fetched from the server...

forslund:
Merging!

PR #1431, opened 2018-02-16
First cut at skill tester

CarstenAgerskov:
Minimum viable solution for running existing testcases (json files).
As you can see, the current to do list below has a number of bullets on it. Because of that, you may want to delay the merge of this PR, or you can test and merge now, at your convenience.

Currently considered additions:
TODO: Make template for testing one skill only, for the skill developer to use
TODO: Pass something to intent, that tells that this is a test run. The skill intent can then avoid side effects
TODO: add optional timeout parameter to test_case
TODO: Check that all intents are checked (what about context)
TODO: Add command line utility to test an event against a test_case, allow for debugging tests
TODO: Add support for expected response, and others
TODO: Detect skill execution end, to decide if a test has failed, without waiting for timeout 

## Description
This PR runs all the existing tests in each skills test/intent folder.

## How to test
Execute unit tests using mycroft-core/test/integrationtests/skills/discover_tests.py

## Contributor license agreement signed?
CLA [ ] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

PR #1527, opened 2018-04-09
Feature/improved skill tester

forslund:
## Description
This is an updated version of #1431. Many thanks to @CarstenAgerskov for doing the main work here.

Changes are:
- Adding support for Padatious and fallbacks.
- Using pytest to do test discovery via parametrized tests
- Possibility to set up test environment
- `expected_dialog` and `expected_data` structures in the intent.json files

### Steps to test
 - `git checkout feature/improved-skill-tester`
 - `./dev_setup.sh -sm` (`pytest` is a new requirement)
 - `./start-mycroft.sh unittest`
 - Make sure the tests all succeed
 - `workon mycroft`
 - `pytest test/integrationtests/skills/discover_tests.py`
 - Now, many of these should fail, but make sure at least a few succeed (`.` means success and `F` means fail). Example correct output:
```
================================================================================================ test session starts ================================================================================================
platform linux2 -- Python 2.7.14, pytest-3.4.2, py-1.5.2, pluggy-0.6.0
rootdir: /path/to/mycroft-core, inifile:
collected 99 items                                                                                                                                                                                                  

test/integrationtests/skills/discover_tests.py F....F......F.........FFFFFFFF..FF.......F.F...FFFFFFFF.FF..........FFFFF......FFFF.F.FFFFFFFFFFF..

# Then a bunch of junk
```

CLA [Yes]

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -194,6 +194,10 @@ if ! pip install -r requirements.txt; then
     fi
 fi
 
+if ! pip install -r test-requirements.txt; then
+  echo "Warning test requirements wasn't installed, Note: normal operation should still work fine..."
+fi
+
 SYSMEM=$(free|awk '/^Mem:/{print $2}')
 MAXCORES=$(($SYSMEM / 512000))
 CORES=$(nproc)


MatthewScholefield:
What's the status of this? Do we think it's ready to be reviewed?

forslund:
I think so, it works well enough so...

KathyReid:
Not sure if I'm helping or hindering here, I went to test this to try to help using the "Steps to test". Pulled the feature branch in, ran `dev_setup.sh` OK, but it did _not_ install `pytest`. 

```
kathyreid@kathyreid-N76VZ:~/Dropbox/PHPworkspace/mycroft-core$ ./start-mycroft.sh unittest
./start-mycroft.sh: line 166: pytest: command not found
```

forslund:
The testing requirements are separated from the normal requirements so they are not installed by default. (since they shouldn't be installed when doing packaging and when just running it as a user)

You need to activate the virtual env and then install test-requirements.txt:

```
source .venv/bin/activate
pip install -r test-requirements.txt
```

KathyReid:
OK that was super helpful @forslund - I've att'd my `unittest` results in the file below. 

[kathy-unittest.txt](https://github.com/MycroftAI/mycroft-core/files/2013944/kathy-unittest.txt)

MatthewScholefield:
Should we install the `test-requirements.txt` by default in `dev_setup.sh`? The reason would be that `dev_setup.sh` isn't run when packaging, and `pytest` is light enough that it shouldn't matter that it's installed on a few people using a `git` clone who aren't developers.

forslund:
Can't argue with that reasoning. Especially since now we want developers to test their skills so...

PR #1528, opened 2018-04-09
Skill settings improvements

MatthewScholefield:
## Description
 - Support triggering skill settings update via a message
 - Decrease polling to 60 seconds now the etag support reduced the load on the servers

## How to test
 - Check if the initial skill settings download
 - Manually trigger skill settings update via `workon mycroft; python mycroft/messagebus/send.py "mycroft.skills.settings.update"`
 - Ensure after 1 minute each skill still only updates once

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -207,6 +207,7 @@ class MycroftSkill(object):
         self.name = name or self.__class__.__name__
         # Get directory of skill
         self._dir = dirname(abspath(sys.modules[self.__module__].__file__))
+        self.settings = SkillSettings(self._dir, self.name)
 
         self.bind(emitter)
         self.config_core = Configuration.get()
@@ -249,15 +250,6 @@ class MycroftSkill(object):
     def lang(self):
         return self.config_core.get('lang')
 
-    @property
-    def settings(self):
-        """ Load settings if not already loaded. """
-        try:
-            return self._settings
-        except:
-            self._settings = SkillSettings(self._dir, self.name)
-            return self._settings
-
     def bind(self, emitter):
         """ Register emitter with skill. """
         if emitter:
@@ -268,6 +260,11 @@ class MycroftSkill(object):
                            self.handle_enable_intent)
             self.add_event('mycroft.skill.disable_intent',
                            self.handle_disable_intent)
+
+            name = 'mycroft.skills.settings.update'
+            func = self.settings.run_poll
+            emitter.on(name, func)
+            self.events.append((name, func))
 
     def __register_stop(self):
         self.stop_time = time.time()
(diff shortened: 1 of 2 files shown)


mycroft/skills/core.py
```diff
@@ -268,6 +260,9 @@ def bind(self, emitter):
                            self.handle_enable_intent)
             self.add_event('mycroft.skill.disable_intent',
                            self.handle_disable_intent)
+            n, f = 'mycroft.skills.settings.update', self.settings.run_poll
```
penrods: Nice!  Are you intending to emit the "mycroft.skills.settings.update" from somewhere specific?  Seems like this should be generated as part of the "Update configuration" handler in skill-configure.

NOTE:  I'm personally not a huge fan of the
```python
n,f = something, something_else
```
notation instead of
```python
n = something
f = something_else
```

penrods: Also, be sure to add the message to https://docs.google.com/spreadsheets/d/1RsBpMOGx4o8A_zF_UELYQSLK8DlIScHGnfzJ3iSN0AE/edit#gid=0

MatthewScholefield: Added to the doc, and you are correct about emitting it in skill-configuration. I have a local change that does that. Also, I was in between the two ways, but used the side by side method to make it analogous to `.emit(name, func)`

penrods: I see the .emit(name, func) parallel, but I think that assignment construct is unique enough to Python that it could be a little confusing and it doesn't seem to add a ton of clarity.  Just my $0.02, not a huge deal.

mycroft/skills/settings.py
```diff
@@ -93,18 +92,28 @@ def __init__(self, directory, name):
...
 
+    def run_poll(self, _=None):
+        """Immediately poll the web for new skill settings"""
+        if self._poll_timer:
+            self._poll_timer.cancel()
+            self._poll_skill_settings()
+
+    def stop_polling(self):
+        if self._poll_timer:
+            self._poll_timer.cancel()
```
forslund: The `cancel` will only work if the method is waiting to start executing _poll_skill_settings. Since the `_poll_skill_settings` does a couple of remote calls it might take some time to execute and there is a risk that the setting poll isn't stopped.

forslund:
My concerns are thoroughly squashed :)

MatthewScholefield:
This one will have to wait until there's some restructuring of the unit tests. This is because registering something with the emitter for skill settings requires skill settings to be instantiated during unit tests which requires the full config to be there. A few unit tests overwrite the config which causes the tests to fail.

forslund:
Good good, Merging this now.

PR #1536, opened 2018-04-12
run time skills config changes

JarbasAl:
## Description

skills configuration was read on boot only

this PR makes skills directory and blacklisted skills list be read at runtime

the reasoning behind this is that one may want to blacklist a skill at runtime

skills directory was also made configurable, but changing it at runtime will require more work (shutting down all skills)

## How to test

check everything works the same

you can play with [py_msm](https://github.com/JarbasAl/py_msm/) and check that blacklisting triggers a skill shutdown through triggering a skill reload, then on loading the skill should stay down because of the blacklist update

       from py_msm import MycroftSkillsManager
       msm = MycroftSkillsManager()
       msm.add_to_blacklist("skill-speak")

### DIFF
--- a/mycroft/skills/main.py
+++ b/mycroft/skills/main.py
@@ -58,7 +58,13 @@ MINUTES = 60  # number of seconds in a minute (syntactic sugar)
 
 
 def get_skills_dir():
-    return Configuration.get().get("skills", {}).get("directory", '/opt/mycroft/skills')
+    skills_dir = Configuration.get().get("skills", {})\
+        .get("directory", '~/.mycroft/jarbas_skills')
+    if "~" in skills_dir:
+        skills_dir = os.path.expanduser(skills_dir)
+    if not exists(skills_dir):
+        os.makedirs(skills_dir)
+    return skills_dir
 
 
 def get_blacklisted_skills():
--- a/mycroft/skills/main.py
+++ b/mycroft/skills/main.py
@@ -58,8 +58,15 @@ MINUTES = 60  # number of seconds in a minute (syntactic sugar)
 
 
 def get_skills_dir():
-    skills_dir = Configuration.get().get("skills", {})\
-        .get("directory", '/opt/mycroft/skills')
+    # TODO support configurable directory,
+    # the "directory" configuration variable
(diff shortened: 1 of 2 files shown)


mycroft/skills/main.py
```diff
@@ -49,23 +49,35 @@
...
+installer_config = Configuration.get().get("SkillInstallerSkill", {})
 MSM_BIN = installer_config.get("path", join(MYCROFT_ROOT_PATH, 'msm', 'msm'))
 
-MINUTES = 60  # number of seconds in a minute (syntatic sugar)
+MINUTES = 60  # number of seconds in a minute (syntactic sugar)
+
+
+def get_skills_dir():
+    skills_dir = Configuration.get().get("skills", {})\
+        .get("directory", '~/.mycroft/jarbas_skills')
```
forslund: The default skills dir should probably be something neutral...or maybe not have a default at all letting it error if the directory isn't set in the config.

JarbasAl: the directory would probably be set on the enclosure config, perhaps the default could just be ~/skills ? not sure if erroring out if the best thing to do, it should work out of the box on first run before users even look at configs

forslund:
Seems like the "directory" configuration variable is included in the blob sent by the server. This messed up my setup something horrible before I figured it out :)

mycroft/skills/main.py
```diff
@@ -49,23 +49,42 @@
...
-
-installer_config = Configuration.get().get("SkillInstallerSkill")
+installer_config = Configuration.get().get("SkillInstallerSkill", {})
 MSM_BIN = installer_config.get("path", join(MYCROFT_ROOT_PATH, 'msm', 'msm'))
 
-MINUTES = 60  # number of seconds in a minute (syntatic sugar)
+MINUTES = 60  # number of seconds in a minute (syntactic sugar)
+
+
+def get_skills_dir():
```
forslund: I guess this is mainly placeholder for future updates which is fine.

forslund:
@JarbasAI much of this has been implemented as part of the move to the new msm. Should we close this or should this be rebased?

PR #1539, opened 2018-04-14
Feature: Allow comments in voc and regex files

### LINKED ISSUE #1531: Allow comments in dialog and voc files
Having a way to comment dialog and voc files would ease the translation process as the original developer can provide context useful for the translator, and the translator can explain things that motivate his/her choice of words.

I am thinking for instance of voc files where similar words to the target word are used to trigger the skill in order to workaround speech to text errors. Documenting that the meaning of those words should not be translated in a comment would be practical.

Example: the "high" word triggers the hello world skill because "high" is similar to "hi". In the Spanish translation we have "hola" (for "hello") and we should have "ola" (literally "wave") because it sounds like "hola"

Other file formats usually used for translation purposes allow developer/translator comments (see the .po file format for instance)

zeehio:
When loading voc and regex files, lines starting with "#" are now
ignored, so developers and translators can use them to document their
decisions.

Fixes #1531 

## How to test
Use a comment (line starting with "#") in a voc file.

CLA [x]

### DIFF
--- a/mycroft/skills/skill_data.py
+++ b/mycroft/skills/skill_data.py
@@ -37,6 +37,8 @@ def load_vocab_from_file(path, vocab_type, emitter):
     if path.endswith('.voc'):
         with open(path, 'r') as voc_file:
             for line in voc_file.readlines():
+                if line.startswith("#"):
+                    continue
                 parts = line.strip().split("|")
                 entity = parts[0]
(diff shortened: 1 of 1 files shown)


forslund:
Tested and working.
Thanks! Merging.

(Notice the clever rhyme)

PR #1545, opened 2018-04-18
Add SSML support for TTS

MatthewScholefield:
## Description
This brings in @JarbasAI's SSML support PR. Here's the original description:
> some TTS engines support SSML to modulate speech
> - make TTS Factory pass config to Engine
> - add method validate_ssml to TTS, checks for SSML, if current TTS does not support SSML remove all ssml tags (regex)
> - if current TTS supports SSML, check if it supports all tags, remove invalid ones
> - validate ssml in audio service before calling execute()

Additional changes are:
 - Parameter to TTS class used to specify supported SSML tags

## How to test
Add a `self.speak` to something like the hello world skill and make sure the following works:
```xml
<speak>Each morning when I wake up, <prosody rate='x-slow'>I speak quite slowly and deliberately until I have my coffee.</prosody></speak>
```
Also, make sure other TTS engines still work due to the constructor changes.

### DIFF
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -114,7 +114,7 @@ def mute_and_speak(utterance, ident):
 
     LOG.info("Speak: " + utterance)
     try:
-        tts.validate_and_execute(utterance, ident)
+        tts.execute(utterance, ident)
     finally:
         lock.release()
 
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -113,7 +113,10 @@ def mute_and_speak(utterance, ident):
         tts_hash = hash(str(config.get('tts', '')))
 
     LOG.info("Speak: " + utterance)
-    tts.execute(utterance, ident)
+    try:
+        tts.validate_and_execute(utterance, ident)
+    finally:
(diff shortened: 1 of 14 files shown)


mycroft/audio/speech.py
```diff
@@ -113,7 +114,10 @@ def mute_and_speak(utterance, ident):
         tts_hash = hash(str(config.get('tts', '')))
 
     LOG.info("Speak: " + utterance)
-    tts.execute(utterance, ident)
+    try:
+        tts.execute(utterance, ident)
```
forslund: This eats any indication that something has failed. An except statement with an error log of the exception might be a good idea.

MatthewScholefield: There's no `except:` clause on this so any exceptions will still get raised. However, do you think we should log exceptions and suppress them here?

forslund: I had some problems with Bing failing silently (due to expired credentials) and assumed it was here, buuut I think that was my own fault since I can see the stack trace now...

I think I was looking at the cli and and the audio log isn't picked up there. So all is good, sorry for the delay in merging :)

PR #1262, opened 2017-11-26
polly tts

JarbasAl:
add polly tts engine option

PR #1550, opened 2018-04-19
Fixes #1549 Addition of Ivona as tts engine

### LINKED ISSUE #1549: Addition of Ivona as tts engine
# Feature Request

Addition of Ivona as tts engine

Wanna work on this issue.

bhaveshAn:
## Description
Fixes #1549 Addition of Ivona as tts engine

## How to test

 @forslund @penrods @aatchison Please review. thanks.

### DIFF
--- a/mycroft/tts/ivona_tts.py
+++ b/mycroft/tts/ivona_tts.py
@@ -1,5 +1,4 @@
 from mycroft.configuration import Configuration
-import subprocess
 from mycroft.tts import TTS, TTSValidator
 import pyvona
 
--- a/mycroft/tts/ivona_tts.py
+++ b/mycroft/tts/ivona_tts.py
@@ -1,0 +1,42 @@
+from mycroft.configuration import Configuration
+import subprocess
+from mycroft.tts import TTS, TTSValidator
+import pyvona
+
+
+class Ivona(TTS):
+    def __init__(self, lang, config):
+        super(Ivona, self).__init__(lang, config, IvonaValidator(self))
+        self.config = Configuration.get().get("tts", {}).get("ivona", {})
+        self.access_key = self.config.get("access_key")
+        self.secret_key = self.config.get("secret_key")
+        self.gender = self.config.get("gender")
+        self.engine = pyvona.create_voice(self.access_key, self.secret_key)
+        if self.gender == 'female':
+            self.engine.voice_name = 'Salli'
+        else:
+            self.engine.voice_name = 'Joey'
+
+    def execute(self, sentence, ident=None):
+        self.begin_audio()
+        self.engine.speak(sentence)
+        self.end_audio()
+
+
+class IvonaValidator(TTSValidator):
+    def __init__(self, tts):
+        super(IvonaValidator, self).__init__(tts)
+
+    def validate_lang(self):
+        # TODO
+        pass
+
+    def validate_connection(self):
+        try:
+            self.engine.speak(sentence)
+        except:
+            raise Exception(
+                'ESpeak is not installed. Run: sudo apt-get install espeak')
+
+    def get_tts_class(self):
+        return Ivona
--- a/mycroft/tts/ivona_tts.py
+++ b/mycroft/tts/ivona_tts.py
@@ -35,7 +35,7 @@ class IvonaValidator(TTSValidator):
             self.engine.speak(sentence)
         except:
             raise Exception(
-                'ESpeak is not installed. Run: sudo apt-get install espeak')
+                'pyvona is not installed')
 
     def get_tts_class(self):
         return Ivona
--- a/mycroft/tts/ivona_tts.py
+++ b/mycroft/tts/ivona_tts.py
@@ -16,10 +16,8 @@ class Ivona(TTS):
         else:
             self.engine.voice_name = 'Joey'
 
-    def execute(self, sentence, ident=None):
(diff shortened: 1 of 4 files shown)


mycroft/tts/ivona_tts.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self, tts):
+        super(IvonaValidator, self).__init__(tts)
+
+    def validate_lang(self):
+        # TODO
+        pass
+
+    def validate_connection(self):
+        try:
+            self.engine.speak(sentence)
```
MatthewScholefield: `self.engine` doesn't exist as far as I can tell so it will always raise the exception that `pyvona is not installed`. Perhaps this should be something like:
```python3
with tempfile.TemporaryFile(suffix='.ogg') as f:
    self.tts.engine.fetch_voice_fp('test', f)
```

mycroft/tts/ivona_tts.py
```diff
@@ -0,0 +1,41 @@
...
+        self.access_key = self.config.get("access_key")
+        self.secret_key = self.config.get("secret_key")
+        self.gender = self.config.get("gender")
+        self.engine = pyvona.create_voice(self.access_key, self.secret_key)
+        if self.gender == 'female':
+            self.engine.voice_name = 'Salli'
+        else:
+            self.engine.voice_name = 'Joey'
+
+    def execute(self, sentence, ident=None):
```
MatthewScholefield: Looking at pyvona, it seems that it requires an extra dependency, `pygame`, if you use `engine.speak()`. Instead, why don't you override Mycroft TTS's [`get_tts()`](https://github.com/bhaveshAn/mycroft-core/blob/dab45c026850a9d3dfe8a25dcecea71fbf9bdeb4/mycroft/tts/mimic_tts.py#L141) method and use `engine.fetch_voice('text', '/file/name')`? Using `fetch_voice` doesn't require the extra dependency.

mycroft/tts/ivona_tts.py
```diff
@@ -0,0 +1,41 @@
...
+        super(IvonaValidator, self).__init__(tts)
+
+    def validate_lang(self):
+        # TODO
+        pass
+
+    def validate_connection(self):
+        try:
+            self.engine.speak(sentence)
+        except:
```
MatthewScholefield: Even though we don't do it in some of the other classes, this should probably be at least `except Exception:` so that it doesn't catch things like `SyntaxError`.

bhaveshAn:
Hi @MatthewScholefield please review the code.

mycroft/tts/ivona_tts.py
```diff
@@ -0,0 +1,40 @@
+from mycroft.configuration import Configuration
+from mycroft.tts import TTS, TTSValidator
+import pyvona
+
+
+class Ivona(TTS):
+    def __init__(self, lang, config):
+        super(Ivona, self).__init__(lang, config, IvonaValidator(self))
+        self.config = Configuration.get().get("tts", {}).get("ivona", {})
```
MatthewScholefield: One last change you'll need is to supply the file extension type by declaring `self.type = '<audio_type>'` [like this](https://github.com/MycroftAI/mycroft-core/blob/b63928888bb81537f011f522dfefc7de1bab5b2c/mycroft/tts/google_tts.py#L24). One thing I noticed is that we only support playing `wav` and `mp3` files so you'll have to make the type `mp3`, also change `ogg` to `mp3` down below, and also change the engines codec with `self.engine.codec = 'mp3'`.

Let me know if this makes sense.

bhaveshAn: Can you see now.

JarbasAl:
wasnt this replaced by polly?

https://github.com/MycroftAI/mycroft-core/pull/1262

MatthewScholefield:
@bhaveshAn Hmm, yeah it does look like Ivona is the older version. Are you alright if we close this in favor of Jarbas's #1262 ?

bhaveshAn:
Polly is updated version but pyvona is without any bug too.

Also Polly has requirements like https://github.com/aws-samples/amazon-polly-sample/blob/master/requirements.txt
> You will need an AWS account and an RSS feed. Some technical experience is required to setup your own instance of the app, but you don't have to write any code. Once setup, it can be used by anyone using a standard Podcast player.

What about keeping both tts engines to provide users with more options.

JarbasAl:
but you can't get new ivona accounts?

if the old ones still work i think we should have both

penrods:
With no CLA I cannot accept these changes.   I'm closing unmerged, but am happy to revisit if you would like to request a CLA via https://mycroft.ai/cla

penrods:
CLA has been provided,

krisgesling:
Hi there, not sure why this didn't get merged previously, but we're planning to add a [plugin system for audio backends, STT and TTS services](https://github.com/MycroftAI/mycroft-core/pull/2594). 

This system will allow users (or developers configuring Mycroft for use in other projects) to pull in these services as they're needed rather than having them all included by default on every Mycroft installation. The aim is to keep mycroft-core lighter but still provide the same level of features and customization through the provision of these plugins.

So I'm going to close this PR but only because it won't need to be included in mycroft-core. We'll be transitioning all of these out to be independent plugins as per #2701 and can include this service in that work.

Thanks

PR #1558, opened 2018-04-26
Fix google tts playing audio

forslund:
## Description
The file extention of the audio was not set properly in the google_tts causing it to fail to playback the synthezised audio.

Big thanks to @livey0u for helping to identify the problem!

## How to test
Add the following block to the mycroft.conf
```json

  "tts": {
    "module": "google"
  }
```
and restart the audio process. Verify that the TTS is not mimic and is working as expected

CLA [Yes]


forslund:
Excellent, thanks for testing! Merging

PR #1560, opened 2018-04-26
Add versioning to msm

MatthewScholefield:
## Description
This makes msm stick to the version listed in the skills repo under the 18.02 branch. If an installed skill isn't on the skill repo, it will continue to update on its own like normal.

## How to test
 - Run `msm/msm update`
 - Check that you see it detect the correct `sha` for `fallbak-unknown` and `fallback-wolfram-alpha`, the only two versioned skills in the 18.02 branch right now
 - Ensure other skills still update like normal

### DIFF
--- a/msm/msm
+++ b/msm/msm
@@ -18,6 +18,9 @@ RED='\033[0;31m'
 NOCOLOR='\033[0m'
 GREEN='\033[0;32m'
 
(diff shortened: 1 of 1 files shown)


forslund:
Seems to work as advertised. Merging

PR #1563, opened 2018-04-27
keep message context

JarbasAl:
## Description
This uses the message.reply method in all places it was missing to propagate message context information

I make heavy use of this in https://chat.mycroft.ai/community/pl/azu61hg1dpygjxy8cdnst9sqne

## How to test
Check nothing breaks, notice that complete_intent_failure and other messages includes message context

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -589,7 +589,7 @@ class MycroftSkill(object):
                 if handler_info:
                     # Indicate that the skill handler is starting if requested
                     msg_type = handler_info + '.start'
-                    self.emitter.emit(Message(msg_type, skill_data))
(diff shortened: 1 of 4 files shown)


forslund:
Could you fix the pep8 issues?

At a glance I can see no issues here, this should be as it was intended to work. I'll test it out more tomorrow.

JarbasAl:
pep8 bot missed them... it's not that great to have to check travis to find them out...

forslund:
I can recommend using https://github.com/cbrueffer/pep8-git-hook, checking PEP8 before allowing a commit.

To match mycroft's checks the following should be in the ignore list: `"E121", "E122", "E123", "E124", "E125", "E126", "E127", "E128", "E131", "E501"`

PR #1568, opened 2018-05-03
The BIG move to python3

forslund:
## Description
Boy has this been a long time coming...
Switches mycroft from python2 to python3. And hard, Mycroft will not be python2 compatible after this.

- Replace virtualenv with venv (developers will need to re-run dev-setup)
- Lots of fixes to make things work in python3
- Remove the 2-3 compatibility code for a cleaner codebase

CLA [Yes]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: python
 before_install:
  - sudo apt-get update -qq
- - sudo apt-get install -qq mpg123 portaudio19-dev libglib2.0-dev swig bison libtool autoconf libglib2.0-dev libicu-dev libfann-dev
+ - sudo apt-get install -qq mpg123 portaudio19-dev libglib2.0-dev swig bison libtool autoconf libglib2.0-dev libicu-dev libfann-dev realpath
 python:
-  - "2.7"
+  - "3.4"
 # don't rebuild pocketsphinx for every build
 cache: pocketsphinx-python
 # command to install dependencies
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ before_install:
  - sudo apt-get install -qq mpg123 portaudio19-dev libglib2.0-dev swig bison libtool autoconf libglib2.0-dev libicu-dev libfann-dev realpath
 python:
   - "3.4"
+  - "3.5"
+  - "3.6"
 # don't rebuild pocketsphinx for every build
 cache: pocketsphinx-python
 # command to install dependencies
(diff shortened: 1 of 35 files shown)


mycroft/util/lang/parse_fr.py
```diff
@@ -334,7 +334,7 @@ def number_ordinal_fr(words, i):
             strOrd = "2e"
         elif word.endswith("ième"):
             val2 = None
-            word = word[:-5]
+            word = word[:-4]
```
clusterfudge: Is this a bug fix or an actual change in array indexing semantics?

test/unittests/skills/core.py
```diff
@@ -106,17 +107,12 @@ def test_load_regex_from_file_none(self):
         self.check_regex_from_file('invalid/none.rx')
 
     def test_load_regex_from_file_invalid(self):
-        try:
```
clusterfudge: This seems like a mix of stylistic/cleanup changes, not necessarily something that's required to cut ties with py2.7

mycroft/util/lang/parse_fr.py
clusterfudge: Ack, thanks for clarifying!

test/unittests/skills/core.py
forslund: Long time since I went though these but I believe there was some issue with assertEquals matching the exception string (the string had changed) so I changed it to check that a re.error was raised.

mycroft/util/lang/parse_fr.py
forslund: Should have made a more clear commit message in the first place.

.travis.yml
```diff
@@ -1,9 +1,9 @@
 language: python
 before_install:
  - sudo apt-get update -qq
- - sudo apt-get install -qq mpg123 portaudio19-dev libglib2.0-dev swig bison libtool autoconf libglib2.0-dev libicu-dev libfann-dev
+ - sudo apt-get install -qq mpg123 portaudio19-dev libglib2.0-dev swig bison libtool autoconf libglib2.0-dev libicu-dev libfann-dev realpath
 python:
-  - "2.7"
+  - "3.4"
```
renekliment: Could this be tested against multiple py3 versions? Like here: https://github.com/alexa-pi/AlexaPi/blob/master/.travis.yml

forslund: Much easier than anticipated :)

PR #1570, opened 2018-05-04
skiller.sh addition: settingsmeta.json.example creation

jrwarwick:
## Description
Minor addition to skiller.sh to generate an annotated settingsmeta.json.exmample file.

## How to test
Simply execute skiller.sh and observe a valid (but generic) settingsmeta.json.example also exists.

CLA [Y] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

### DIFF
--- a/skiller.sh
+++ b/skiller.sh
@@ -139,9 +139,9 @@ echo '{
 	"name": "'${capital_desc}'",
 	"skillMetadata": {
 		"sections": [
-		{
-		 "name": "Section Friendly Display Name for Home.Mycroft.AI",
-		 "fields": [
+			{
+				"name": "Section Friendly Display Name for Home.Mycroft.AI",
+				"fields": [
 					{
 						"name": "internal_python_variable_name",
 						"type": "text",
@@ -155,9 +155,9 @@ echo '{
 				"name": "Login --Another Section But This One Has Two Values--",
 				"fields": [
 					{
-			"type": "label",
-			"label": "Just a little bit of extra info for the user to understand following settings"
-			},
+						"type": "label",
+						"label": "Just a little bit of extra info for the user to understand following settings"
+					},
 					{
 						"name": "username",
 						"type": "text",
--- a/skiller.sh
+++ b/skiller.sh
@@ -135,49 +135,49 @@ def create_skill():
     return $class_name()
 """ > "$skill_dir/__init__.py"
 
-echo """{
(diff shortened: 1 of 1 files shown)


skiller.sh
```diff
@@ -135,6 +135,48 @@ def create_skill():
     return $class_name()
 """ > "$skill_dir/__init__.py"
 
+echo """{
+	\"name\": \"${capital_desc}\",
```
MatthewScholefield: For readability, you should be able to use a single quote on the outside (the part that is `"""` right now) so that you don't have to backslash the quotes. To add a variable replacement, surround it with another single end-quote and start-quote. Here's an example:
```bash
echo '{
    "name": "'${capital_desc}'",
    "skillMetadata": {
        ...
}'
```

skiller.sh
```diff
@@ -135,6 +135,48 @@ def create_skill():
...
+						\"value\": \"\"
+					}
+				]
+			}
+		]
+	}
+}
+""" > "$skill_dir/settingsmeta.json.example"
+
 echo -e "*.pyc\nsettings.json\n" > $skill_dir/.gitignore
```
MatthewScholefield: Do you think we should add `settingsmeta.json.example` here to the .gitignore so that by default it won't get committed?

skiller.sh
jrwarwick: Ah yes, that looks much better. I was trying to match style of the rest of the script, but this would be much cleaner.

skiller.sh
jrwarwick: Definitely.

skiller.sh
MatthewScholefield: As a side note, I just realized I don't think those triple quotes actually do anything (it's really just `""` next to `"something`).

skiller.sh
```diff
@@ -135,7 +135,49 @@ def create_skill():
...
+						"value": "",
+						"placeholder": "demo prompt in the input box"
+					}
+				]
+			},
+			{
+				"name": "Login --Another Section But This One Has Two Values--",
+				"fields": [
+					{
+			"type": "label",
```
MatthewScholefield: One last thing, some of the json looks like it got formatted funny. ([here's a corrected version](https://hastebin.com/romofevuji.json)). Once that's fixed, it looks good to go :+1:

PR #1585, opened 2018-05-11
Fix skill reloading when files in subdirectories change

MatthewScholefield:
## Description
This fixes skills not reloading when something like a `.voc` file is changed.

## How to test
 - Change a `.voc` file
 - Make sure the skill is reloaded


forslund:
Looks good! Merging

PR #1586, opened 2018-05-11
Make skill update interval configurable

MatthewScholefield:
## Description
This is especially helpful for debugging

## How to test
 - Change the `update_interval` value in the config
 - Make sure it abides by this new interval

### DIFF
--- a/mycroft/skills/main.py
+++ b/mycroft/skills/main.py
@@ -203,9 +203,12 @@ class SkillManager(Thread):
         self.enclosure = EnclosureAPI(ws)
 
         # Schedule install/update of default skill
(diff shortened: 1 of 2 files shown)


mycroft/skills/main.py
```diff
@@ -203,9 +203,12 @@ def __init__(self, ws):
         self.enclosure = EnclosureAPI(ws)
 
         # Schedule install/update of default skill
+        self.update_interval = Configuration.get()['skills']['update_interval']
+        self.update_interval = int(self.update_interval * MINUTES)
```
forslund: Should this be `self.update_interval * 60 * MINUTES` since the time is specified as hours in the config?

PR #1354, opened 2018-01-09
msm - Values from config

JarbasAl:
## Description

This makes all bash scripts try to read values from all config files, they check platform, skills directory, and if we are supposed to use virtual env

duplicates some work from #1255 

## How to test
change config in enclosure.platform, enclosure.use_virtualenvwrapper, skills.directory and test the scripts

## Contributor license agreement signed?
CLA [X ] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

PR #1592, opened 2018-05-14
Remove all references to /opt/mycroft/skills

MatthewScholefield:
## Description
This get's rid of references to `/opt/mycroft/skills` so that it can properly be configured. The only thing that should reference it is the config. This fixes #1354 and closes #1255.

## How to test
 - Pull branch
 - Make sure skills load properly

### DIFF
--- a/mycroft/configuration/config.py
+++ b/mycroft/configuration/config.py
@@ -149,7 +149,7 @@ class RemoteConf(LocalConf):
     def __init__(self, cache=None):
         super(RemoteConf, self).__init__(None)
 
-        cache = cache or '/opt/mycroft/web_config_cache.json'
+        cache = cache or '/var/tmp/mycroft_web_cache.json'
         from mycroft.api import is_paired
         if not is_paired():
             self.load_local(cache)
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -83,14 +83,17 @@
       "offset": -21600000
     }
   },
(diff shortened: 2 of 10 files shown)


forslund:
Hmm wouldn't it be better in the long run to remove all references to `/opt/mycroft` (so cached websettings and the event-save file also are included in this.

Having the skills directory be `join(mycroft_dir, 'skills')`

forslund:
@MatthewScholefield seems like recent changes has created some conflicts, do you mind rebasing this?

MatthewScholefield:
@forslund Just rebased.

MatthewScholefield:
@forslund Any idea why these tests are failing? They are working on my end

forslund:
Found an issue with the configuration test, had probably been there a while but the ordering of the tests masked it. I did a quick fix and rebased this branch so now it's passing again.

MatthewScholefield:
@forslund I've routed all references of `data_dir` through `expanduser`. As for `prepare-msm`, I'm not sure the best thing for that since it is a bash script run on startup that needs to access all configurations (since it would likely be overridden user level). The best solution I could think of is making it perform the same actions in a startup python script, but it'd have to load all of the configuration manager which I'm not sure if it would be too long to put at startup. I think for now it's reasonable to reference the two so that to change your data_dir you need to also change `prepare-msm`.

forslund:
You're right. That sounds like a good solution for now.

forslund:
Looks good, Merging

PR #1599, opened 2018-05-17
Retry skill installation every 5 minutes if default skill install fails

MatthewScholefield:
## Description
This prevents errors with having a spotty internet connection during default skill installation

## How to test
 - Make sure regular skill installation works fine and doesn't reschedule
 - Try installing default skills with internet down part time and see if it retries after 5 minutes


forslund:
Seems to be working as intended! Merging

PR #1608, opened 2018-05-23
Fix chromecast quitting app on mycroft shutdown.

forslund:
## Description
The default method for the audio services call stop, in the chromecast case this isn't desirable since any playing application would be shutdown (even if mycroft didn't initiate it).

This overrides default behaviour and will merely disconnect from the device at shutdown.

## How to test
Start youtube on a chromecast device on the network and  start mycroft. stop the audio service and check that the youtube app is still running.

CLA [Yes]


forslund:
I'm going to go ahead and merge this myself so it's included in this release. It's a rather minor change.

PR #1612, opened 2018-05-25
Feature/skill commands

forslund:
## Description
- Add messagebus handlers to activate and deactivate skills
- Unload skills when they are removed on disk
- Add the `:activate`, `:deactivate` and `:keep commands`

## How to test
- Delete a skill and check that it gets unloaded
- try the `:deactivate` / `:activate` commands and assert that they function as intended

CLA [Yes]


penrods:
LGTM!

PR #1618, opened 2018-05-31
Add info when skill fails to load

forslund:
## Description
When skill is not loaded and error is raised skill will be loaded to provide any traceback to the test.

## How to test
Edit a skill `__init__` method adding a  `raise Exception` line, run discovertest and check that in the info about the failed tests the traceback is shown.

CLA [Yes]


forslund:
Hmm, I had a similar issue when I had a stray `__init__.py` in the `/opt/mycroft/skills` directory. But for me basically no skills were loaded. for you it seems like some are loaded but most are not.

forslund:
I've updated this with a slightly cleaner solution. Capture the logging during loading and printing it if the skill isn't loaded when the tests are to be evaluated.

forslund:
@MatthewScholefield I've rebased this PR so it should work again now.

PR #1619, opened 2018-06-01
Feature/fix skill tester launch

MatthewScholefield:
## Description
This fixes launching `./start-mycroft skillstest` and makes it more usable by adding a `SKILL_DIR` environment variable so that rather than executing `source ...; python x/y/z/single_test.py` it can be executed through pytest via start-mycroft.

## How to test
 - Make sure `./start-mycroft.sh skillstest` works as intended
 - Make sure something like `SKILL_DIR=/opt/mycroft/skills/mycroft-speak.MycroftAI ./start-mycroft.sh skillstest` works

### DIFF
--- a/start-mycroft.sh
+++ b/start-mycroft.sh
@@ -175,11 +175,11 @@ case ${_opt} in
     ;;
   "unittest")
     source ${VIRTUALENV_ROOT}/bin/activate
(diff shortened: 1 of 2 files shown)


forslund:
Looks good! Merging

PR #1624, opened 2018-06-01
Cleanup root

MatthewScholefield:
## Description
This cleans up the root a little bit by moving github specific documents into a `.github` folder (which should be supported) and removing an unused script.

## How to test
Look at the changes
Make sure references are still valid in README


forslund:
A good idea!

Personally I think the code of conduct should be left in the root directory so it's visble and easy to find.

MatthewScholefield:
@forslund Updated

forslund:
Looking good! Merging

PR #1628, opened 2018-06-06
Feature/pronounciations

forslund:
## Description
Add better pronunciation for Spotify and mycroftai.
Also make the pronunciation check disregard capitalization.

CLA [Yes]

### DIFF
--- a/mycroft/tts/__init__.py
+++ b/mycroft/tts/__init__.py
@@ -297,8 +297,9 @@ class TTS(object):
         create_signal("isSpeaking")
         if self.phonetic_spelling:
             for word in re.findall(r"[\w']+", sentence):
-                if word in self.spellings:
-                    sentence = sentence.replace(word, self.spellings[word])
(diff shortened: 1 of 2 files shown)


mycroft/tts/__init__.py
```diff
@@ -297,8 +297,10 @@ def execute(self, sentence, ident=None):
         create_signal("isSpeaking")
         if self.phonetic_spelling:
             for word in re.findall(r"[\w']+", sentence):
-                if word in self.spellings:
-                    sentence = sentence.replace(word, self.spellings[word])
+                if word.lower() in self.spellings:
+                    sentence = sentence.replace(word,
+                                                self.spellings[word.lower()])
+                    LOG.info(sentence)
```
MatthewScholefield: If we keep the new log here, do we want to add extra into like `New phonetic sentence:`?

forslund: Sorry that's not supposed to be there. will remove.

PR #1629, opened 2018-06-06
Change wake word upload to https request

MatthewScholefield:
## Description
This changes the wake word upload to an https request and also strips out old alternative ways to enable wake word upload. The logic for uploading wake words is now:
 - Only if opt-in is enabled
 - With an additional "disable" setting to selectively prevent it

## How to test
 - Make sure after opt-in, you see `POST /precise/upload` after triggering the wake word (but not your account id)
 - Make sure if opt-in stays enabled, but you set `{"listener": {"wake_word_upload": {"disable": true}}}`, no wake words are uploaded
 - Make sure if `disable` is still set to false, but you opt out, nothing is uploaded

### DIFF
--- a/mycroft/client/speech/mic.py
+++ b/mycroft/client/speech/mic.py
@@ -13,26 +13,23 @@
 # limitations under the License.
 #
 import audioop
-import collections
-import datetime
-from hashlib import md5
-import shutil
-from tempfile import gettempdir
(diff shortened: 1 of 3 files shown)


forslund:
Tested and works fine. Today was an excellent day to test since the training server throws the occasional 504 error :)

Merging

PR #1630, opened 2018-06-07
Make use of the default config when testing.

forslund:
## Description
Always use the default config as base config to make tests less error prone when changes occur.

## How to test
Make sure Travis tests still works.

CLA [Yes]

### DIFF
--- a/test/util.py
+++ b/test/util.py
@@ -1,0 +1,9 @@
+from mycroft.configuration.config import LocalConf, DEFAULT_CONFIG
+from copy import deepcopy
+
+__config = LocalConf(DEFAULT_CONFIG)
+
+
+# Base config to use when mocking
+def base_config():
+    return deepcopy(__config)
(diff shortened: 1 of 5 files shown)


test/util.py
```diff
@@ -0,0 +1,9 @@
+from mycroft.configuration.config import LocalConf, DEFAULT_CONFIG
+from copy import copy
+
+__config = LocalConf(DEFAULT_CONFIG)
+
+
+# Base config to use when mocking
+def base_config():
+    return copy(__config)
```
MatthewScholefield: Should this be `deepcopy` so that merged attribute of sub elements don't mutate the original?

forslund: If you think so it probably should, my python knowledge breaks down around here :) Will fix

MatthewScholefield: Only difference is:
```python
a = {'1': {'2': 'original'}}
b = copy(a)
c = deepcopy(a)

a['1']['2'] = 'altered'

print(b['1']['2'])  # altered
print(c['1']['2'])  # original
```

forslund: Thanks.

PR #1634, opened 2018-06-08
German language support

f-e-l-i-x:
Adds parse_de and format_de for German language support (now with PEP-8 line wrap).

Can be tested via the corresponding unit tests test_parse_de and test_format_de.

CLA [X ]


mycroft/res/text/de-de/not connected to the internet.dialog
```diff
@@ -0,0 +1,5 @@
+Es sieht so aus, als wäre ich nicht mit dem Internet verbunden
+Es scheint als wäre ich nicht mit dem Internet verbunden
+Ich kann das Internet gerade nicht erreichen
+Es ist mir gerade nicht möglich das das Internet zu erreichen
+
```
forslund: Empty lines like this will also be considered an alternative when selecting the output line causing Mycroft to be silent when speaking the dialog and should therefore be removed.

mycroft/util/lang/format_de.py
```diff
@@ -0,0 +1,344 @@
...
+NUM_POWERS_OF_TEN = [
+    '', 'tausend', 'Million', 'Milliarde', 'Billion', 'Billiarde', 'Trillion',
+    'Trilliarde'
+]
+
+FRACTION_STRING_DE = {
+    2: 'halb',
+    3: 'drittel',
+    4: 'viertel',
+    5: u'fünftel',
```
forslund: Not an issue just a general comment: The unicode `u` shouldn't be needed anymore, since the switch to python3 all strings are unicode :)

f-e-l-i-x: Okay, I'll keep it in mind

f-e-l-i-x: the blank lines should all be gone now

forslund:
Looks good I think, my german isn't the best.

Danke

gras64:
I've noticed, but it is not yet a translation of units of measure such as m / s in meters per second or kg in kilograms. This is important for skills like Wikipidea.

f-e-l-i-x:
@gras64 if you give me a list of which words you want to have replaced with which, then I can try to implement it. Which words need to be replaced depends on the speech engine as well. For example, google speaks "kg" as "Kilogramm", but does not say "m/s" as "Meter pro Sekunde".

PR #1648, opened 2018-06-20
Update default config for skills

forslund:
## Description
- Remove unused entries for skills
- Correct the entry for the NewsSkill

## How to test
Make sure default skill still loads and that the affected skills still load

CLA [Yes]


penrods:
LGTM

Issue #1647, opened 2018-06-19
small numbers and Scientific Notation support in utility functions


JarbasAl:
a few considerations about extract_number and pronounce_number:

# names of small numbers support

like PR #1646 and #1642 do for big numbers, support for long and short scale (PR #1663 ) should be added for small numbers in order to include all languages

relevant information:

- https://en.wikipedia.org/wiki/Names_of_small_numbers
- https://en.wikipedia.org/wiki/Long_and_short_scales

# supporting Scientific notation

a lab setting is one of the places an AI assistant makes a lot of sense, this should be a default part of mycroft-core

extract_number ( PR #1652 ) and pronounce_number (PR #1653) should handle scientific notation, optionally if a flag is set

scientific notation pronouncing cases:

- (times/by) 10 to the (minus/negative/positive/plus/ ) (n/nth) (power / )
- (times/by) 10 to the power of (minus/negative/positive/plus/ ) n
- (times/by) 10 to the (minus/negative/positive/plus/ ) power of n
- (times/by) 10 (exponentiated/raised/elevated) to (minus/negative/positive/plus/ ) ( n / nth )

forslund:
Closing this since it's a lingua-franca issue.

PR #1652, opened 2018-06-21
extract number scientific notation support

JarbasAl:
## Description

extract_number_en now supports scientific notation

Partially solves #1647 

## How to test
check unittests

### DIFF
--- a/mycroft/util/lang/format_en.py
+++ b/mycroft/util/lang/format_en.py
@@ -58,11 +58,11 @@ FRACTION_STRING_EN = {
     5: 'fifth',
     6: 'sixth',
     7: 'seventh',
-    8: 'eigth',
+    8: 'eighth',
     9: 'ninth',
     10: 'tenth',
     11: 'eleventh',
-    12: 'twelveth',
+    12: 'twelfth',
     13: 'thirteenth',
     14: 'fourteenth',
     15: 'fifteenth',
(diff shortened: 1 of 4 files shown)


JarbasAI:
any idea why this passes python 3.4 and 3.5 but fails in python 3.6 ?

forslund:
From a glance it looks like a floating point issue

JarbasAl:
any idea how i can fix this? i wanted to include it in a new PR (all possible number pronunciations)

penrods:
The issue is that you are probably getting a result like:

    666000000.00000000000000001 != 6660000000.0

Using == or != on floats is never really safe or reliable.  Instead it should be more like this:

```python
def assertEqualFloat(self, test, expected, epsilon=None):
    epsilon = epsilon or 0.0000001
    if test > expected+epsilon or test < expected-epsilon:
        self.fail("Value {} was not equal to {} within {}".format(test, expected, epsilon)
```

Then change the tests to:
```python
    self.assertEqualFloat(extract_number("666 times 10 raised to the ninth"),
                             666e9)
```

forslund:
```
6.66*10⁹:					6660000000
extract_number Python3.5:	6660000000.0
extract_number Python3.6:	666000000.0
```

It seems the extract number comes up short on python 3.6

forslund:
Some more tests on python 3.6, it seems the ninth gets replaced with eight:
`ten eight power`

JarbasAl:
closing and remaking against https://github.com/MycroftAI/lingua-franca

PR #1654, opened 2018-06-22
govivace_stt

JarbasAl:
## Description
adds goVivace STT option

## How to test
- adquire key
- update config
- use mycroft by voice

### DIFF
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -219,7 +219,9 @@
     // },
     //"govivace": {
     //   "uri": "https://services.govivace.com:49149/telephony",
-    //   "token": "xxxxx"
+    //   "credential": {
+    //      "token": "xxxxx"
+    //      }
     //}
   },
 
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -219,7 +219,7 @@
     // },
     //"govivace": {
     //   "uri": "https://services.govivace.com:49149/telephony",
-    //   "token": "678956a4c12527fbafa7aa95cb1de0dc"
+    //   "token": "xxxxx"
     //}
   },
 
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -209,7 +209,7 @@
   // Override: REMOTE
   "stt": {
     // Engine.  Options: "mycroft", "google", "wit", "ibm", "kaldi", "bing",
-    //                   "houndify", "deepspeech_server"
+    //                   "houndify", "deepspeech_server", "govivace"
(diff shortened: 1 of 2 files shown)


mycroft/stt/__init__.py
```diff
@@ -201,6 +201,29 @@ def execute(self, audio, language=None):
...
+        self.default_uri = "https://services.govivace.com:49149/telephony"
+
+        if not self.lang.startswith("en") and not self.lang.startswith("es"):
+            LOG.error("GoVivace STT only supports english and spanish")
+            raise NotImplementedError
+
+    def execute(self, audio, language=None):
+        url = self.config.get("uri", self.default_uri) + "?key=" + self.token + \
+              "&action=find&format=8K_PCM16&validation_string="
+        response = put(url,
```
forslund: `put` should be `requests.put`

mycroft/stt/__init__.py
```diff
@@ -201,6 +201,29 @@ def execute(self, audio, language=None):
...
+                       data=audio.get_wav_data())
+        return self.get_response(response)
+
+    def get_response(self, response):
+        try:
+            return response.json()["result"]["hypotheses"][0]["transcript"]
+        except:
+            return None
```
forslund: This except catch all is not helpful, it eats any indication of what is wrong and returns `None` which raises an error in the listener instead.

mycroft/configuration/mycroft.conf
```diff
@@ -209,14 +209,18 @@
...
     "module": "mycroft"
     // "deepspeech_server": {
     //   "uri": "http://localhost:8080/stt"
     // },
     // "kaldi": {
     //   "uri": "http://localhost:8080/client/dynamic/recognize"
     // },
+    //"govivace": {
+    //   "uri": "https://services.govivace.com:49149/telephony",
+    //   "token": "xxxxx"
```
forslund: Should be
```json
    "credential": {
      "token": "xxxxxx"
      }
```

mycroft/stt/__init__.py
JarbasAl: kaldi does it, aren't all STT engines supposed to return None if speech was not recognized?

forslund:
I think I've found the issue. Mycroft records at a rate of 16kHz

```python
LOG.info('Audio: {}, {}'.format(audio.sample_width, audio.sample_rate))
```
07:25:06.729 - mycroft.stt:execute:218 - INFO - Audio: 2, 16000

So when passing the wav_data to the service you'll need to convert it to the expected 8kHz rate:

```python
response = put(url, data=audio.get_wav_data(convert_rate=8000))
```

forslund:
Nice! Seems to work fine now.

PR #1656, opened 2018-06-25
Feature add query mic status

AIIX:
## Description
Adds ability for enclosures to query mycroft over the messagebus about the status of the microphone weather the state is muted or unmuted.

## How to test
Send message type:"mycroft.mic.is_muted" from an enclosure to mycroft core, return bool value should be in mycroft.mic.muted data {mic_state:true/false} 

### DIFF
--- a/mycroft/client/speech/main.py
+++ b/mycroft/client/speech/main.py
@@ -103,7 +103,7 @@ def handle_mic_is_muted(event):
     """
         Query microphone mute status.
     """
-    data = {'mic_state':loop.is_muted()}
+    data = {'mic_state': loop.is_muted()}
     ws.emit(Message('mycroft.mic.muted', data))
 
 
--- a/mycroft/client/speech/main.py
+++ b/mycroft/client/speech/main.py
@@ -99,6 +99,14 @@ def handle_mic_unmute(event):
     loop.unmute()
 
 
+def handle_mic_is_muted(event):
+    """
+        Query microphone mute status.
+    """
+    data = {'mic_state':loop.is_muted()}
+    ws.emit(Message('mycroft.mic.muted', data))
+
+
 def handle_paired(event):
     IdentityManager.update(event.data)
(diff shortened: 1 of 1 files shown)


mycroft/client/speech/main.py
```diff
@@ -99,6 +99,14 @@ def handle_mic_unmute(event):
     loop.unmute()
 
 
+def handle_mic_is_muted(event):
+    """
+        Query microphone mute status.
+    """
+    data = {'mic_state': loop.is_muted()}
+    ws.emit(Message('mycroft.mic.muted', data))
```
forslund: I suggest we start using the .response method on the Message object. It hasn't been widely used yet but is basically intended for this. In this case it would become:

```python
ws.emit(event.response(data)
```

would generate a new message from the original message type adding `.response` so the response message would become `mycroft.mic.is_muted.response` and is automatically handled when using the

`ws.wait_for_response`

AIIX: Thanks for the review, made the changes to the commit, on testing response is now mycroft.mic.is_muted.response data {mic_state: true/false}

forslund:
Works really nicely!

I'm just worried of confusion with the naming.

`mycroft.mic.is_muted` -> `{mic_state: True}`

This can be interpreted in two different ways.

I'm thinking either just change the key in the response data
`mycroft.mic.is_muted` -> `{"status": True}`

or change the message type to `mycroft.mic.get_status` and have a muted entry in the response data:
`mycroft.mic.get_status` -> `{"muted": True}`

What do you think? am I over thinking this?

AIIX:
Have changed it to message type "mycroft.mic.get_status" and key to "muted", this naming also seems easier to understand over the messagebus so response now is type: mycroft.mic.get_status.response data:{muted: true/false}

forslund:
Looks excellent! Merging

PR #1668, opened 2018-06-30
Enhance the behavior of the Mark 1 button

penrods:
## Description
The Mark 1 button press can now be "consumed" when a skill handles
the Stop command.  When this happens, the button press will not
trigger listening mode.  An additional press would be needed to
trigger listening.

This introduces the "mycroft.stop.handled" messagebus message.  It
carries a data field called "by" which identifies who handled it.
Currently the values are "TTS" for when speaking ends or the name
of a skill which implements Stop and returns True from the call.

Also fixed a potential bug when the flag to clear queued visemes
was left set after a button press.

## How to test
Use the Mark 1 button to stop speaking, timers, etc.  The skill must return
True to consume it, so some skills might need altering to not accidentally
trigger listening.

CLA [X ] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

### DIFF
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -21,6 +21,7 @@ from mycroft.metrics import report_timing, Stopwatch
 from mycroft.tts import TTSFactory
 from mycroft.util import create_signal, check_for_signal
 from mycroft.util.log import LOG
+from mycroft.messagebus.message import Message
 
 ws = None  # TODO:18.02 - Rename to "messagebus"
 config = None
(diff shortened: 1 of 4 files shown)


forslund:
I'll hold merging this until the pandora skill is updated. If you get time do the try-except-finally change but that's no biggie.

penrods:
The ethanaward repo also got updated.  I think this is good to go now.

forslund:
Confirmed. Merging

PR #1675, opened 2018-07-06
Fix crash during initial skill download

penrods:
When running on a platform for which there was not a DEFAULT.platform
file in the mycroft-skills repo, the downloading of skills would
crash.  Now an informational message is shown and the DEFAULT
skills alone are loaded.

## Description
When running on a platform such as Ubuntu, the default skills weren't being downloaded.

## How to test
Edit the /etc/mycroft/mycroft.conf so that enclosure.platform is a value other than 'kde',
'mycroft_mark_1', or 'picroft'.


forslund:
Looks good, merging.

PR #1678, opened 2018-07-09
Ogg123

joshuacox:
This is reimplementation of #1649 which became divergent.

## Description
Adds a Ogg123Service and a play_ogg exactly like Mpg123Service and play_mp3

## How to test
I have a skill for a podcast which does not have an mp3 feed:
https://github.com/joshuacox/skill-GNUworldOrder

signed by @joshuacox

### DIFF
--- a/mycroft/audio/services/ogg123/__init__.py
+++ b/mycroft/audio/services/ogg123/__init__.py
@@ -1,0 +1,123 @@
+# Copyright 2017 Mycroft AI Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


penrods:
This is largely in Ake's realm, but in the interest of keeping this moving I'll pull it in.  I changed the Stop handler very slightly to reflect some recent changes there.

I'm minorly concerned because ogg123 isn't a commonly installed utility, but I'm not certain if it should really become a dependency for mycroft-core.  For now we can make the hackers drop to a command line.  In the future the skill will probably need to add the package-dependency using the newly proposed YAML format that supersedes the requirements.txt and requirements.sh.

PR #1679, opened 2018-07-09
Simpleaudioservice

joshuacox:
This is split from #1649 that attempts to implement the simple audio service.  They have been split as I have found this method to break as mimetypes.guess_type() does not seem to actually inspect files carefully and gives lots of false negatives.

In an attempt to expedite the ability to play ogg's, I have split this into another PR   Which may at some point deprecate both the play_mp3  and play_ogg methods with another superseding method that encompasses both, but for now those should remain as these methods will fail and require someone to search through logs to find out why.

### DIFF
--- a/mycroft/audio/services/ogg123/__init__.py
+++ b/mycroft/audio/services/ogg123/__init__.py
@@ -1,0 +1,123 @@
+# Copyright 2017 Mycroft AI Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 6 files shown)


joshuacox:
This should not be merged before #1678 !

forslund:
@joshuacox Thanks for this. I've been sketching on a way to just use the single "simple" service based on the above PR. What do you think of this solution: https://github.com/forslund/mycroft-core/commit/ffa52783095ab981119d5cd7a5b3c7cff7f4c8a1

It adds a mime detecting for remote media (still stupid guessing for local) and adds the possibility to provide type hints when adding tracks
```python
  audioservice.play((url, 'audio/ogg'), utterance)
```

penrods:
I believe this was all covered in the #1761 PR that got merged.  Thanks for your help with this!

PR #1681, opened 2018-07-11
Enhance extract_datetime(), add time utilities

penrods:
Many cases that were missed in the unittests for extract_datetime()
from the original source.  Restored those tests and made code
adjustments to support them all.

Also adding the mycroft.util.time module.  This supports:
* mycroft.util.time.default_timezone()
  Returns the user-configured timezone based on location
* mycroft.util.time.now_utc()
  Returns the time in UTC
* mycroft.util.time.now_local()
  Returns the time in the user's timezone
* mycroft.util.time.to_utc()
  Converts to UTC
* mycroft.util.time.to_local()
  Converts to user's timezone

NOTE: Several skills should be updated to use these now.

==== Fixed Issues ====
Several issues for skills regarding parsing of "today"

Note the new module:  mycroft.util.time

==== Localization Notes ====
Localized versions of extract_datetime() likely need to be
updated, as most were based on the original English implementation

## How to test
Run the test/unittests/util/test_parse.py unittest

### DIFF
--- a/mycroft/util/time.py
+++ b/mycroft/util/time.py
@@ -37,7 +37,7 @@ def default_timezone():
         code = config["location"]["timezone"]["code"]
 
         return gettz(code)
-    except:
+    except Exception:
         # Just go with system default timezone
         return tzlocal()
 
--- a/mycroft/util/time.py
+++ b/mycroft/util/time.py
@@ -1,0 +1,95 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright 2018 Mycroft AI Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from datetime import datetime
+from dateutil.tz import gettz, tzlocal
+
+
+def default_timezone():
+    """ Get the default timezone
+
+    Based on user location settings location.timezone.code or
+    the default system value if no setting exists.
+
+    Returns:
+        (datetime.tzinfo): Definition of the default timezone
+    """
+    try:
+        # Obtain from user's configurated settings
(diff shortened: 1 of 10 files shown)


forslund:
Looks pretty good! Having proper timezone info will make lots of things easier!

I'm missing the parsing of "now", which would be useful for cases like "do I have anything scheduled in an hour", "do I have anything scheduled right now".

I've found a small issue, not sure it's worth going after. (and it might very well be my input that's missing something)

when using a weekday and date like Friday the 20th it seem to regard `the 20th` as a time and not a date.

```python
>>> extract_datetime('asdf Friday the 20th')
[datetime.datetime(2018, 7, 13, 20, 0, tzinfo=tzfile('/usr/share/zoneinfo/Europe/Stockholm')), 'asdf']
>>> extract_datetime('asdf friday the 20')
[datetime.datetime(2018, 7, 13, 20, 0, tzinfo=tzfile('/usr/share/zoneinfo/Europe/Stockholm')), 'asdf']
>>> extract_datetime('asdf friday the twenty')
[datetime.datetime(2018, 7, 13, 0, 0, tzinfo=tzfile('/usr/share/zoneinfo/Europe/Stockholm')), 'asdf twenty']
>>> extract_datetime('asdf friday the twentieth')
[datetime.datetime(2018, 7, 13, 0, 0, tzinfo=tzfile('/usr/share/zoneinfo/Europe/Stockholm')), 'asdf twentieth']
```

mycroft/util/time.py
```diff
@@ -0,0 +1,95 @@
...
+    Args:
+        dt (datetime): A datetime (presumably in some local zone)
+    Returns:
+        (datetime): time converted to UTC
+    """
+    tzUTC = gettz("UTC")
+    if dt.tzinfo:
+        return dt.astimezone(tzUTC)
+    else:
+        return dt.replace(tzinfo=gettz("UTC")).astimezone(tzUTC)
```
forslund: Isn't it enough to just do the 

`dt.replace(tzinfo=gettz("UTC"))`

The `astimezone()` would convert from UTC to UTC right?

forslund: I understand, I just thought the `astimezone` was redundant in this case (I only did a quick test so I might be wrong) but I trust you that it's necessary.

penrods: Yeah, we could special-case to skip calling astimezone() in the UTC -> UTC case, but astimezone() already has the code to handle it.

forslund:
Merged through #1693, Closing

PR #1682, opened 2018-07-11
Add 'venv-activate.sh' script

penrods:
## Description

This script in the root of a Github checkout makes it easier for devs to discover
the virtual environment which Mycroft should run under.  Virtual
environments are new to many, so the commands aren't obvious even
if they are only one liners.

Should probably be mentioned in the Getting Started docs for
Github users.  This mirrors the 'mycroft-venv-activate' alias found
under Mark 1/Picroft.

## How to test
Github install, try running "bash venv-activate.sh" and then ". venv-activate.sh"

### DIFF
--- a/venv-activate.sh
+++ b/venv-activate.sh
@@ -24,5 +24,7 @@ if [ "$0" = "$BASH_SOURCE" ] ; then
     echo "ERROR: Invoke with 'source venv-activate.sh' or '. venv-activate.sh'"
 else
     echo "Entering mycroft-core virtual environment.  Run 'deactivate' to exit"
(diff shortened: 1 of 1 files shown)


forslund:
To make this callable from places outside of the mycroft-core root dir the following can be added at the beginning of the file:
```bash
CALLED=$_
SCRIPT_DIR="$( cd "$( dirname "${CALLED}" )" ; pwd -P )"
```
($_ only seem to be valid in the absolute beginning of a sourced file hence the CALLED=$_)

and then

```bash
source ${SCRIPT_DIR}/.venv/bin/activate
```

PR #1684, opened 2018-07-13
MAXCORES bug fix in dev_setup.sh 

jwallace00:
Replaced MAXCORES and SYSMEM with MINCORES=1 and check to ensure at least 1 core will be used to compile mimic

## Description
- Removes SYSMEM and MAXCORES variables
- Adds MINCORES=1
- Changes check for CORES to look for positive integers,   if fail CORES is set to ${MAXCORES}

## How to test
use dev_setup.sh to build mimic

Agreement request sent from Jeremy Wallace (wallace.jeremyc@gmail.com)

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -198,12 +198,11 @@ if ! pip install -r test-requirements.txt; then
   echo "Warning test requirements wasn't installed, Note: normal operation should still work fine..."
 fi
 
-SYSMEM=$(free|awk '/^Mem:/{print $2}')
-MAXCORES=$(($SYSMEM / 512000))
+MINCORES=1
 CORES=$(nproc)
 
-if [[ ${MAXCORES} -lt ${CORES} ]]; then
-  CORES=${MAXCORES}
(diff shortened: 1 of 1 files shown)


forslund:
I think the MAXCORES is there for a reason. The build will fail due to lack of memory if too many cores are used together with too little memory. The test is made with a rough estimation of memory needed during build and might be off by a bit.

Adding a Minimum cores of 1 to the script is a good addition though. since the calculation can obviously fail while gcc can actually compile with the resources available.

@el-tocino might want to weigh in on this since I think he originally made the calculation.

jwallace00:
All,  
I re-added the MAXCORES SYSMEM calculation,  and added a check to make sure it's set to at least $MINCORES before attempting the build.    This should preserve the original intent, but add a safety in case the SYSMEM calculation floors to 0.

forslund:
Sounds like a good solution. I'll try to test it out during the weekend

forslund:
Tested and works good, Merging. Thank you!

PR #1687, opened 2018-07-14
Clean up setup scripts

forslund:
## Description
As discussed with @MatthewScholefield this makes the setup.py script more standardized and removes the unused mycroft-skills-sdk.

- Remove setup scripts for mycroft-skills-sdk since it's not used anymore
- Rename mycroft-base-setup.in/MANIFEST.in to setup.py and MANIFEST.in
- Remove skill-container, since it hasn't been used or kept up to date since 17.08 and the cli commands to remove and activate skills is easier to work with

==== Environment Notes ====
Small update of the packaging script is needed due to this change

CLA [Yes]


MatthewScholefield:
It should be noted that the setup.py will still be broken on a fresh install because it tries to import a utility function from within `mycroft` which causes it to error on any non-standard imports that haven't been installed yet. Do you want to include that fix with this or do that in another PR?

forslund:
Right forgot about that. It should be included in this.

setup.py
```diff
@@ -0,0 +1,83 @@
...
+                elif 'CORE_VERSION_MINOR' in line:
+                    minor = line.split('=')[1].strip()
+                elif 'CORE_VERSION_BUILD' in line:
+                    build = line.split('=')[1].strip()
+
+                if ((major and minor and build) or
+                        '# END_VERSION_BLOCK' in line):
+                    break
+        version = '.'.join([major, minor, build])
+    except Exception as e:
```
MatthewScholefield: I'm not a fan of this `except Exception`. If something goes wrong (like we change the way we write the version), we might as well let the build process fail with that exception

setup.py
```diff
@@ -0,0 +1,83 @@
...
+        print('Couldn\'t determine version ({})'.format(repr(e)))
+
+    return version
+
+
+def required(requirements_file):
+    with open(os.path.join(BASEDIR, requirements_file), 'r') as f:
+        requirements = f.read().splitlines()
+        return [pkg for pkg in requirements if not pkg.startswith("--")]
```
MatthewScholefield: Will this include empty lines and should the check be `if not pkg.startswith("#")`?

setup.py
```diff
@@ -0,0 +1,83 @@
...
+
+setup(
+    name="mycroft-core",
+    version=get_version(),
+    license='Apache-2.0',
+    author='Mycroft A.I.',
+    author_email='devs@mycroft.ai',
+    url='https://github.com/MycroftAI/mycroft-core',
+    description='Mycroft Core',
+    install_requires=[required('requirements.txt')],
```
MatthewScholefield: This will create a double list (ie. `[['foo', 'bar']]`) and should probably be `install_requires=required('requirements.txt')`.

setup.py
```diff
@@ -0,0 +1,83 @@
...
+    packages=find_all_packages("mycroft"),
```
MatthewScholefield: The packages flag should work fine with module names (ie. `mycroft.util`). Because of this, would it work to just use `find_packages(include=['mycroft*'])` which gives the module names directly?

setup.py
```diff
@@ -0,0 +1,83 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from setuptools import setup, find_packages
+import os.path
+
+BASEDIR = os.path.abspath(os.path.dirname(__file__))
+
+
+def get_version():
```
MatthewScholefield: Since these functions used to be in `mycroft.util.setup_base`, are we able to remove everything in that file now.

setup.py
forslund: Yeah, that seems to work fine as well.

setup.py
forslund: I'll remove the references to it and remove it then. I was a bit hesitant since it was part of the distribution

MatthewScholefield: I've only seen it used twice in `mycroft/metrics/__init__.py` which can be replaced by `CORE_VERSION_STR`.

mycroft/metrics/__init__.py
```diff
@@ -121,7 +121,7 @@ def __init__(self):
         self._timers = {}
         self._levels = {}
         self._attributes = {}
-        self.attr("version", get_version())
+        self.attr("version", CORE_VERSION_STR)
```
MatthewScholefield: [Line 145](https://github.com/MycroftAI/mycroft-core/pull/1687/files#diff-3de6fddb63358003a713be6d85a76490L145) as well.

forslund: Thanks

PR #1688, opened 2018-07-14
Bugfix/unmunge

forslund:
## Description
The unmunge of keywords would in rare cases miss keys due to the fact that the dict is modified while being iterated. This breaks out the keys into a list before iterating through them to ensure that all original keys are checked.

## How to test
Make sure intents works as they have previously.

CLA [Yes]


penrods:
Good catch!  LGTM

PR #1692, opened 2018-07-17
Make dialog loading more strict

forslund:
## Description
While creating dialog files temporary files may be loaded and confuse
skill creators.

This checks that the file isn't hidden (starts with `.`) and the extention matches `.dialog`.

## How to test
Make sure dialogs are loaded as expected

CLA [Yes]

### DIFF
--- a/mycroft/dialog/__init__.py
+++ b/mycroft/dialog/__init__.py
@@ -17,6 +17,7 @@ from io import open
 
 import os
 import re
+from pathlib import Path
 
 from mycroft.util import resolve_resource_file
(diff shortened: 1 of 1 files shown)


MatthewScholefield:
Could we switch the piece of code to use `glob` since we are modifying it? (It would simplify it quite a bit). Also, is it a problem if we load something like `.confirm.dialog` if the only way to use it would be via `self.speak_dialog('.confirm')`?

forslund:
I updated it using pathlib's Path()

PR #1696, opened 2018-07-19
Fix updating padatious intents after intent reload

forslund:
## Description
Training is called from wait_and_train with None as message, this commit adds propper handling for this case.

## How to test
Install for example the cocktails skill and make sure the skill is usable afterwards.

CLA [Yes]


forslund:
Self merging, if this explodes it's on me!

PR #1697, opened 2018-07-19
Replace python with python3 in start-mycroft.sh script

juliengk:
## Description

## How to test


MatthewScholefield:
Thanks for contributing to mycroft-core! Within a virtual environment, `python` is the same as `python3`. Is there a particular reason you think it'd be better to change which one is used?

juliengk:
I have been testing it in Docker without virtualenv and had no python but only python3.

penrods:
Looking at https://www.python.org/dev/peps/pep-0394/, the recommendation is that "python" points to python2 by default for the near future.  They also suggest that python2 and python3 exist as appropriate for specific versions.  There appears to be a little variance in these suggestions (under Arch they made python point to Python 3 by default), but I think "python3" is consistently used unless I hear otherwise.

Given that we only operate under Python 3, I am OK with this proposal unless someone can point to a system where python3 does NOT work as defined in PEP394.

penrods:
That is true, @MatthewScholefield, but these folks are talking about the Docker environment.  Given that Docker is an app container, they want to skip the venv stage.  That makes sense to me and I think we can support that with a few adjustments to the Docker image.  I'm recommending they add a mycroft.conf to the image, then we can look for a 'platform' setting.

juliengk:
sorry close by mistake

PR #1700, opened 2018-07-22
Replace static /opt/mycraft in prepare-msm.sh

Whisprin:
Check the access rights of the right folder

## Description
Use the bash variable mycroft_root_dir when checking the access rights of the mycroft root directory instead of hardcoded /opt/mycroft

## How to test
Change mycroft_root_dir to any folder other than /opt/mycroft, before this change the script complaint if the static folder /opt/mycroft doesn't exist. Now it's checking the right dir


forslund:
Looks like a good fix, thanks. We'll look into merging when the CLA is received on our end!

PR #1703, opened 2018-07-23
Check if only python3 is installed in start-mycroft.sh script

juliengk:
Related to PR #1697

### DIFF
--- a/start-mycroft.sh
+++ b/start-mycroft.sh
@@ -90,7 +90,7 @@ function launch-process() {
 
     # Launch process in background, sending log to scripts/log/mycroft-*.log
     echo "Starting $1"
-    python ${_script} $_params
+    python3 ${_script} $_params
 }
 
 function launch-background() {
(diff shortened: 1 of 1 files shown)


MatthewScholefield:
Sorry about the confusion. If it makes it easier for a specific use case, I'm fine changing it all to `python3` which is cleaner than conditionally switching it. As a side note, how are you getting `start-mycroft` to work on docker without it failing when it starts the virtualenv?

penrods:
I agree -- I'd rather not do it as a conditional.  I'm fine with just adopting "python3" as there is no way we will be going back to python2 support.  Now that I've read PEP394 (https://www.python.org/dev/peps/pep-0394/) I'm good with that plan.

juliengk:
@MatthewScholefield I have modified the start-mycroft.sh and dev_setup.sh scripts. I will add the modifications in the next PR.

PR #1706, opened 2018-07-26
Immediately speak warning while Padatious trains

penrods:
At startup, Padatious was blocking in the fallback while the
training occurred.  As a result, any attempts to use Mycroft
during that period would queue up rather than giving the
user feedback.  Now it immediately returns False, allowing
user notification to occur elsewhere.

## How to test
Say anything during skill startup

### DIFF
--- a/mycroft/skills/padatious_service.py
+++ b/mycroft/skills/padatious_service.py
@@ -109,7 +109,6 @@ class PadatiousService(FallbackSkill):
     def handle_fallback(self, message):
         if not self.finished_training_event.is_set():
             LOG.debug('Waiting for Padatious training to finish...')
-            # self.finished_training_event.wait()
             return False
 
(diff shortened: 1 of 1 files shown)


mycroft/skills/padatious_service.py
```diff
@@ -64,7 +64,7 @@ def train(self, message=None):
             single_thread = message.data.get('single_thread', False)
         self.finished_training_event.clear()
 
-        LOG.info('Training...')
+        LOG.info('Training...'+str(single_thread))
```
MatthewScholefield: What's this part for?

penrods: I don't recall why I cared, to be honest.  I added it during debugging.  I can drop if you feel it is useless.

PR #1707, opened 2018-07-27
Add enclosure to the start/stop dev scripts

penrods:
The start-mycroft.sh and stop-mycroft.sh scripts now automatically
run the enclosure client if the platform is defined as
"mycroft_mark_1".

Also corrected a copy/pasted comment and added a clarifying message during the stop when killing a process.

## How to test
Start on a Mark 1 with a Github install.  The "enclosure" process should also start.

### DIFF
--- a/stop-mycroft.sh
+++ b/stop-mycroft.sh
@@ -33,6 +33,7 @@ function help() {
   echo "  audio     stop the audio playback service"
   echo "  skills    stop the skill service"
   echo "  voice     stop voice capture service"
+  echo "  enclosure stop mark_1 enclosure service"
   echo
   echo "Examples:"
   echo "  ${script}"
@@ -42,7 +43,7 @@ function help() {
 }
 
 function process-running() {
-    if [[ $( ps aux | grep "[p]ython .*${1}/main.py" ) ]] ; then
+    if [[ $( ps aux | grep "python3\? .*${1}/main.py" ) ]] ; then
         return 0
     else
         return 1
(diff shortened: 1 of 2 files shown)


stop-mycroft.sh
```diff
@@ -87,6 +89,13 @@ case ${OPT} in
     end-process skills
     end-process audio
     end-process speech
+    
+    # determine platform type
+    mycroft_platform=$( jq ".enclosure.platform" < /etc/mycroft/mycroft.conf )
```
MatthewScholefield: Use `jq -r` to get rid of the quotes in the output

stop-mycroft.sh
```diff
@@ -42,7 +43,7 @@ function help() {
 }
 
 function process-running() {
-    if [[ $( ps aux | grep "[p]ython .*${1}/main.py" ) ]] ; then
+    if [[ $( ps aux | grep "[p]ython[2|3]\? .*${1}/main.py" ) ]] ; then
```
MatthewScholefield: The regex should be `[23]` (even though it won't fail by allowing it to match `|`)

stop-mycroft.sh
```diff
@@ -87,6 +89,13 @@ case ${OPT} in
     end-process skills
     end-process audio
     end-process speech
+
+    # determine platform type
+    mycroft_platform=$( jq -r ".enclosure.platform" < /etc/mycroft/mycroft.conf )
```
MatthewScholefield: This errors on a desktop because the file doesn't exist. Can we extract this into a `get_platform` function that checks if the file exists?

PR #1711, opened 2018-08-02
Standardize

juliengk:
CLA [x]

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -133,9 +133,9 @@ install_deps
 git config commit.template .gitmessage
 
 # Check whether to build mimic (it takes a really long time!)
-build_mimic='y'
+build_mimic="y"
 if [[ ${opt_skipmimic} == true ]] ; then
-    build_mimic='n'
+    build_mimic="n"
 else
     # first, look for a build of mimic in the folder
     has_mimic=""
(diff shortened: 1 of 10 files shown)


MatthewScholefield:
Thanks for this; it looks a lot more consistent now. Two style changes that I think are more common in bash:
 - `echo $(echo hi)` rather than `echo $( echo hi )`
 - `if [ "$val" == "hi" ]; then` rather than `if ["$val" == "hi" ] ; then`

Any opinion on them? (@juliengk, @forslund, @penrods)

forslund:
I'm ok with those rules @MatthewScholefield. At some point I think it may have been @el-tocino that suggested we'd use the google shell style guide https://google.github.io/styleguide/shell.xml

Is that something we should try to follow?

juliengk:
* No space after $(: if we look at that [page](https://www.gnu.org/software/bash/manual/html_node/Command-Substitution.html#Command-Substitution), there are no spaces at all.
* No space after if [] : https://www.gnu.org/software/bash/manual/html_node/Conditional-Constructs.html#Conditional-Constructs

I could remove spaces for those 2 cases

penrods:
Are these all manual edits, or do you have some sort of bash lint tool you used?  It is nice to clean this up, as there is a ton of variance in sh script styles and oddball rules if you were attempting to follow Posix vs bash, etc.  I see tools like https://www.shellcheck.net/, but that doesn't enforce spacing, consistent use of newline/then vs " ; then" etc.

I looked through all of these changes and am going to go ahead an merge this as-is.  We can do another PR with more cleanup later if we pick a specific style.  Ideally, I want one that has tools that reinforce it -- many of the changes that are in this code are from where different devs added to the scripts.

PR #1712, opened 2018-08-03
start-mycroft.sh improvements (change log dir and exit 1 when out of date)

penrods:
Previously, the location of the log files changed depending
on whether you were running on a Mark1/Picroft or under a
"Github" install.  Now they are always under the same directory
at /var/log/mycroft-*.log

This also updates the CLI to pull from that location always.

Additionally:
* Removed the nonfunctional 'wifi' option from start-mycroft.sh
* Made the validation for dev_setup.sh exit instead of just show
  a warning message in start-mycroft.sh
* Added code to allow dev_setup.sh to be run from different
  directories successfully

## How to test
Install, run dev_setup.sh and run via start-mycroft.sh


mycroft/client/text/main.py
```diff
@@ -1259,17 +1259,10 @@ def simple_cli():
...
-# Monitor relative logs (for Github installs)
-start_log_monitor(localLogPath + "/mycroft-skills.log")
-start_log_monitor(localLogPath + "/mycroft-voice.log")
-
-# Also monitor system logs (for package installs)
+# Monitor system logs
 start_log_monitor("/var/log/mycroft-skills.log")
-start_log_monitor("/var/log/mycroft-speech-client.log")
+start_log_monitor("/var/log/mycroft-voice.log")        # when start-mycroft.sh
+start_log_monitor("/var/log/mycroft-speech-client.log")  # when Debian package
```
MatthewScholefield: Should we change all references of `voice` to `speech-client` so that they are consistent, especially since they will now be in the same directory?

start-mycroft.sh
```diff
@@ -1,218 +1,218 @@
...
-        ;;
-    "enclosure")
-        launch-background ${_opt}
-        ;;
-
-    *)
-        help
-        ;;
-esac
+#!/usr/bin/env bash
```
MatthewScholefield: Can you change the line endings back to unix (or setup git to do it automatically) so the diff shows relevant changes?

dev_setup.sh
```diff
@@ -41,6 +41,9 @@
 # exit on any error
 set -Ee
 
+TOP=$( cd $( dirname $0 ) && pwd -L )
+cd $(dirname $0)
```
MatthewScholefield: We can swap these to make it less redundant:
```bash
cd $(dirname $0)
TOP=$(pwd -L)
```

MatthewScholefield:
One thing to note that I found while researching the topic is that having a log directory is cleaner, permission-wise for things like log rotation. I think we should change to placing logs in something like `/var/log/mycroft/skills.log`, but it could be in a future PR. @penrods What do you think?

el-tocino:
The mycroft subdirectory is a suitably good idea as well and reduces permissions issues in /var/log. Either/or, for me, but consistent for 'em all would be nice.

mycroft/client/text/main.py
penrods: I thought about doing that.  Made the start-mycroft script a little more complex -- that is the one exception.  This was a little lazy but felt like it was a less error prone change.  Or I might just be justifying my laziness.

start-mycroft.sh
penrods: Whoops, that was unintentional!  I'll push that...

mycroft/client/text/main.py
penrods: Looking at the /var/log/mycroft/*.log thought below, that would be a reason to stick with this approach.  I can switch to /var/log//mycroft/voice.log here and still support the package install for now.  Later we can change the packages to use the mycroft subdirectory.  I think that is all safe and definitely does simplify the permission issues.

dev_setup.sh
penrods: Done!

penrods:
In the future we can modify the packaging to also log to the /var/logs/mycroft/*.log, so that should make things easier.  We'll also need to update the support-skill, I think.

penrods:
Whew, lots of comments!  Most have merit, but I don't want to touch all of that in this commit -- we can tackle those in an independent PR.  I try to keep PRs that hit multiple files focused on only the related changes.

KathyReid:
No worries, I am rusty with `bash` so wanted to be thorough. 
TL;DR the `/var/log` changes look good. Ship them!

MatthewScholefield:
@KathyReid Would you mind creating an issue with that content in it? Thanks.
This looks good to me so I'll pull it in. Necessary changes:
 - [ ] skill-support
 - [ ] mycroft-packaging

penrods:
Cool.  The packaging doesn't _have_ to be changed yet, this will still be backwards compatible with that until it is updated (then we can clean the log monitor from the CLI code).

KathyReid:
@MatthewScholefield sure no worries at all :+1:

PR #1713, opened 2018-08-03
Allow Padatious to override Adapt

penrods:
Allow a Padatious intent to override Adapt when it is VERY
certain that the utterance is directed at it.  (95% confidence
or greater.)  Right now that only occurs if the intent match
for the given phrase is perfect.

This solves this kind of issue:
* Adapt:  Matching on "Set" and "Alarm"
* Padatious: Handling "is an alarm set"

## Description

## How to test
Can be tested by creating the example mentioned above:
* Adapt:  Matching on "Set.voc" and "Alarm.voc" (containing those words)
* Padatious: Handling "alarm.status" that contains "is an alarm set"

The Padatious intent should match on "is an alarm set" but not on "is a alarm set"

### DIFF
--- a/mycroft/skills/intent_service.py
+++ b/mycroft/skills/intent_service.py
@@ -23,6 +23,7 @@ from mycroft.skills.core import open_intent_envelope
 from mycroft.util.log import LOG
 from mycroft.util.parse import normalize
 from mycroft.metrics import report_timing, Stopwatch
+from mycroft.skills.padatious_service import PadatiousService
 
 
 class AdaptIntent(IntentBuilder):
@@ -244,7 +245,6 @@ class IntentService(object):
 
         NOTE: This only applies to those with Opt In.
         """
-        LOG.debug('Sending metric if opt_in is enabled')
         ident = context['ident'] if context else None
         if intent:
             # Recreate skill name from skill id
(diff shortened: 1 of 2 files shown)


mycroft/skills/intent_service.py
```diff
@@ -287,15 +287,18 @@ def handle_utterance(self, message):
...
 
             if converse:
                 # Report that converse handled the intent and return
                 ident = message.context['ident'] if message.context else None
                 report_timing(ident, 'intent_service', stopwatch,
                               {'intent_type': 'converse'})
                 return
-            elif intent:
-                # Send the message to the intent handler
+            elif intent and padatious_intent and padatious_intent.conf < 0.95:
```
MatthewScholefield: What if there's no Padatious intent but there is an adapt intent? (Logic should probably be `if intent and not (padatious_intent and padatious_intent.conf < 0.95):`)

penrods: Ah, yup!  Changing.

penrods: Except the confidence level comparison needs to be inverted, too:
```python
elif intent and not (padatious_intent and padatious_intent.conf >= 0.95):
```

penrods:
Yep, definitely a hack at the moment!

PR #1715, opened 2018-08-03
Remove git locks when starting mycroft

MatthewScholefield:
## Description
This is a preventative measure to fix errors with git lock files not being removed when devices are unplugged or processes are killed

## How to test
 - Make sure this doesn't cause a regression with downloading default skills

### DIFF
--- a/mycroft/skills/main.py
+++ b/mycroft/skills/main.py
@@ -438,12 +438,16 @@ class SkillManager(Thread):
                         continue
             self._load_or_reload_skill(skill.path)
 
+    def remove_git_locks(self):
+        """If git gets killed from an abrupt shutdown it leaves lock files"""
+        for i in glob(join(self.msm.skills_dir, '*/.git/index.lock')):
+            LOG.warning('Found and removed git lock file: ' + i)
+            os.remove(i)
+
     def run(self):
         """ Load skills and update periodically from disk and internet """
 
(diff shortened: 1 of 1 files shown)


mycroft/skills/main.py
```diff
@@ -438,11 +438,17 @@ def load_priority(self):
...
+    def remove_git_locks(self):
+        for i in glob(join(self.msm.skills_dir, '*/.git/index.lock')):
+            LOG.warning('Found and removed git lock file: ' + i)
+            os.remove(i)
+
     def run(self):
         """ Load skills and update periodically from disk and internet """
 
         # Load priority skills first, in order (very first time this will
         # occur before MSM has run)
```
penrods: Let's clean this up -- the comment got left behind and makes absolutely no sense anymore.  The code was originally this:
```python
        # Load priority skills first, in order (very first time this will
        # occur before MSM has run)
        self.load_skill_list(PRIORITY_SKILLS)
```

So, that comment is no confusing, not helpful.  Axe it.

Additionally, I think we need to explain the reasoning for the code that removes .git lock files.   In the future, somebody is going to look at this and ask: "Why would someone do this? .git/index.lock is there for a reason!"

So add a comment that describes the reasoning behind the decision to override the locking mechanism.

MatthewScholefield: Fixed!

PR #1716, opened 2018-08-06
Better logging around why writing .PHO file fails.

ndarilek:
I need to know why this is failing, and the logs here give no indication of what file is being written or why it is failing. This commit hopefully fixes that (I say "hopefully" because I can't easily run it on the failing platform.)

I'd appreciate this making it into 18.2.13 so I can debug this more thoroughly in the next release. It doesn't appear to be a permissions issue, but as of now I just don't get enough info from this error to diagnose things further.


forslund:
Works excellent and shows the issue clearly! Merging

PR #1719, opened 2018-08-07
add skill vocab to self.is_match

forslund:
## Description
- Look for vocab in skill dir as well as the res/text dir
- Cache the result to avoid hitting the disk each time the method is called.

## How to test
Verify with a skill that uses ask_yes_no for example

CLA [Yes]


penrods:
I like it, but renamed from "is_match()" to "voc_match()" after taking a straw poll in the office of the purpose of the function.

PR #1722, opened 2018-08-09
Use full path in function check-dependencies

fortwally:
## Description
In the start-mycroft.sh file the check-dependencies function uses local file references for the .installed
file. Add a path to the .installed references.

## How to test
ssh into a picroft system using the "Lighting preview". From the pi directory "source mycroft-core/start-mycroft.sh" you should be dropped from the picroft system with the message to "Please update dependencies by running ./dev_setup.sh again."

### DIFF
--- a/start-mycroft.sh
+++ b/start-mycroft.sh
@@ -144,7 +144,7 @@ function launch-all() {
 }
 
 function check-dependencies() {
-  if [ ! -f .installed ] || ! md5sum -c &> /dev/null < .installed ; then
+  if [ ! -f ~/mycroft-core/.installed ] || ! md5sum -c &> /dev/null < ~/mycroft-core/.installed ; then
     echo "Please update dependencies by running ./dev_setup.sh again."
     if command -v notify-send >/dev/null ; then
       notify-send "Mycroft Dependencies Outdated" "Run ./dev_setup.sh again"


MatthewScholefield:
That script shouldn't be sourced but rather run like a normal script as `./start-mycroft.sh` or in your case, probably `./mycroft-precise/start-mycroft.sh`.  We can't hardcore this path because this is the same script used with a git install on a computer that might not be in `~/mycroft-core`.

Let me know if `./mycroft-core/start-mycroft.sh` works.

fortwally:
I am running the 2018-07-31 stretch picroft version.  I don't know where the pi auto_run.sh file comes from. I think @steve-mycroft was trying to make the picroft a git build.
From my reading source and . are the same.

MatthewScholefield:
@fortwally I would suggest that you try it exactly as I wrote and test if it works anyways. `.` is only the same as `source` if you put a space after it. Here's a list of the two ways of launching a shell script:

**Run script in _current_ session**:
 - `source path/to/script.sh`
 - `. path/to/script.sh`
 - `. ./path/to/script.sh`

**Run script in _new sub-_ session**:
 - `./path/to/script.sh` # This executes it using the shebang (`#!/usr/bin/env bash`)
 - `bash path/to/script.sh`
 - `bash ./path/to/script.sh`

Any of the bottom three should work fine.

forslund:
Possibly the `${DIR}` variable should be used if the start-mycroft.sh script is called from a different directory. `${DIR}/.installed`

MatthewScholefield:
@forslund Oh yeah, I didn't realize the `cd` at the top was run in a subshell. We could also pull that out so that we are always running commands in the root folder.

fortwally:
Maybe with the different configurations the path to mycroft-core should be determined during the install and saved in the mycroft.conf file.

MatthewScholefield:
@fortwally No, I don't think there's no need to store that. When a bash script is run properly, it's possible to determine where the script that is being executed is run from. We already determine this location in the `$DIR` variable. So, to fix this you should be able to do one of two things:
 - Change `.installed` to `"$DIR/.installed"` like forslund mentioned
 - Change [`DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )`](https://github.com/MycroftAI/mycroft-core/blob/dev/start-mycroft.sh#L21) to `cd -P "$(dirname "$SOURCE")"` followed by `DIR=$(pwd)`

penrods:
See:  https://github.com/MycroftAI/mycroft-core/pull/1744

I believe that PR will supersede this one.

penrods:
I'm closing this now that #1744 is merged.

PR #1727, opened 2018-08-12
Fix camel case splitting

juliengk:
Signed-off-by: Julien Kassar <github@kassisol.com>

## Description
It returns the passed argument `PublicIPSkill` to `Public IP Skill` instead of `Public IPSkill`

Found the function to split correctly camel case in this [post](https://stackoverflow.com/questions/29916065/how-to-do-camelcase-split-in-python)

CLA [Yes]

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -39,7 +39,7 @@ from mycroft.metrics import report_metric, report_timing, Stopwatch
 from mycroft.skills.settings import SkillSettings
 from mycroft.skills.skill_data import (load_vocabulary, load_regex, to_alnum,
                                        munge_regex, munge_intent_parser)
-from mycroft.util import resolve_resource_file
+from mycroft.util import camel_case_split, resolve_resource_file
 from mycroft.util.log import LOG
 
 MainModule = '__init__'
@@ -648,7 +648,7 @@ class MycroftSkill(object):
 
             except Exception as e:
                 # Convert "MyFancySkill" to "My Fancy Skill" for speaking
-                handler_name = re.sub(r"([a-z])([A-Z])", r"\1 \2", self.name)
+                handler_name = camel_case_split(self.name)
                 msg_data = {'skill': handler_name}
                 msg = dialog.get('skill.error', self.lang, msg_data)
                 self.speak(msg)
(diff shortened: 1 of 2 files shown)


juliengk:
I tried the suggested simpler regex. It does work for a string like `PublicIPSkill` but not for `ABCWordSkill`. The result for the latter is `ABCWord Skill`.

forslund:
Is the python version on the Mark-1 updated to >=3.5 so type hinting is supported?

MatthewScholefield:
@forslund Python 3.4 supports type hints. It's only assignment type hints that 3.4 doesn't support (ie. `a: int = 12`)

forslund:
Ah gotcha

juliengk:
that adds a leading space :)

MatthewScholefield:
@juliengk Dammit xD, then it'd be:
```python
' '.join(re.sub(r'([a-z]?)(?:([A-Z]+)?([A-Z][a-z])|([A-Z]+))', r'\1 \2 \3\4', identifier).split())
```

juliengk:
@MatthewScholefield  how about no regex?

```python
def camel_case_split(identifier: str) -> str:
    result = ''

    lenStr = len(identifier)-1

    i = 0
    previous = None
    for c in identifier:
        current = c.isupper()
        next = True if i < lenStr and identifier[i+1].isupper() else False

        if (i > 0) and ((previous != current and current) or (current and not next)):
            result += ' '
        result += c

        previous = current
        i += 1

    return result
```

MatthewScholefield:
@juliengk `camel_cast_split('VALUE')` would be `VALU E` with that implementation :)

juliengk:
ah! i haven't tried single word with that. I tried `ABCWordSkill`, `abcWordSkill` and `PublicIPSkill`

juliengk:
```python
def camel_case_split(identifier: str) -> str:
    result = ''

    lenStr = len(identifier)-1

    i = 0
    previous = None
    for c in identifier:
        current = c.isupper()
        next = True if i < lenStr and identifier[i+1].isupper() else False

        if (i > 0) and ((previous != current and current) or (current and not next and lenStr-i > 1)):
            result += ' '
        result += c

        previous = current
        i += 1

    return result
```

juliengk:
@MatthewScholefield Let me know whichever of the solutions you prefer, I would make the change. Thanks.

MatthewScholefield:
@juliengk It's fine how it is :)

PR #1732, opened 2018-08-13
Remove Docker check from `source-venv` to be consistent with dev_setup.sh, which always initializes a virtualenv.

### LINKED ISSUE #1730: Docker build is failing
I made a fix for part of this in the Docker repository, but it looks like the ultimate fix is a bit more complex, and I wanted to ask for a quick sanity check before completing it. From dev_setup.sh:

```bash
# Start the virtual environment
...
```

From start-mycroft.sh:

```
function source-venv() {
    # Enter Python virtual environment, unless under Docker
    if [ ! -f "/.dockerenv" ] ; then
        source ${VIRTUALENV_ROOT}/bin/activate
    fi
}
```

So it looks like dev_setup.sh, which is run from Docker, indiscriminately creates the virtualenv, while the start script only enters it when in Docker.

Is that analysis correct? If so, which solution is the right one? Should we always create a virtualenv, or should we selectively not create/enter one when in Docker? I'd opt for no Docker-specific tweaks in the main repo unless absolutely necessary, so removing the Docker check from `source-venv` or maybe even `source-venv` entirely and just running `source`. But maybe there's a reason for that?

In any case, Docker for development environments is nice, so I'd like to get this working again. :) The official build on Docker Hub is several months out-of-date, and it looks like it's automated, so I assume if we fix this then the process will start again and we'll have updated builds without effort on your part. Thanks!

ndarilek:
Not sure if this is the correct fix, but it seems to work.

Closes #1730

### DIFF
--- a/start-mycroft.sh
+++ b/start-mycroft.sh
@@ -75,10 +75,7 @@ function name-to-script-path() {
 }
 
 function source-venv() {
-    # Enter Python virtual environment, unless under Docker
-    if [ ! -f "/.dockerenv" ] ; then
-        source ${VIRTUALENV_ROOT}/bin/activate
-    fi
+    source ${VIRTUALENV_ROOT}/bin/activate
 }
 
 first_time=true


juliengk:
how about to add the same condition if within a docker container in the dev_setup.sh script?

ndarilek:
That may be the way forward. I just don't know what the logic behind 
that check is. Maybe it isn't strictly necessary to install into a 
virtualenv in Docker, but if someone introduces a new branch point and 
doesn't make the same check then things will fail. So, unless needed, my 
inclination is to remove the unnecessary check.

FWIW, I'm testing this locally and it seems to work fine--in terms of 
local skill development, that is.

penrods:
From my understanding of how Docker is used, the "virtual environment" mechanism doesn't make sense at all -- a Docket container is never mixed in with anything else, right?  If that is the case, then the changes of #1710 seem appropriate.  Is this correct from you perspectives?

If so, is the real issue is with being able to run dev_setup.sh within the Docker environment?  If so, adding the same kind of check within dev_setup.sh seems to make sense to me.

ndarilek:
I wouldn't say it *never* makes sense to avoid the virtualenv. You 
might, for instance, keep Mycroft in a Docker container but ephemeral 
data like skills outside. In that case, the virtualenv makes sense at 
least for skills, since if the container ever goes away but your skills 
are still present, Mycroft won't reinstall dependencies when noticing 
the skills since it is assumed they already are. My attempt at a Hassio 
container keeps the virtualenv outside the container, since keeping it 
inside would make upgrades very slow, and would make it impossible to 
wipe out the virtualenv and restart if something in it ever got corrupt.

Maybe the check should be generic, virtualenv vs. no virtualenv, and 
folks can pick which makes sense for their environment, vs. making 
assumptions about a given environment. So the Dockerfile instruction 
might call:

```bash
$ ./dev_setup.sh --no-virtualenv
```

rather than hard-coding the script to make assumptions that aren't 
always obvious.

FWIW, this was a frustration of mine when Mycroft bundled a couple 
assumptions under the "your platform is picroft" designation. I'd have 
much rather it let me opt into specific behaviors, and just opt into all 
of those when running on picroft. Instead, I had to look through the 
source for all the instances where the platform was checked, and see if 
those behaviors were ones I wanted or didn't.

Thanks, apologies for rambling. :)

forslund:
In my opinion the start-mycroft.sh and dev_setup.sh shouldn't really be used in a docker environment. And I think the dev_setup.sh should install mycroft in exactly one way.

The dev_setup.sh is intended to create a standard easy to use setup for developers and the start-mycroft.sh script is intended to start this setup.

In a docker setting I think it's better to install mycroft using pip or the setup.sh script, this can be done either into the system, the users site-packages or in a virtualenv depending on what makes sense for the setup. (unless the docker wants to create a development environment in witch case the dev_setup.sh makes sense and also the default virtualenv)

ndarilek:
I like this, mainly because it encourages straying off the happy path 
and, hopefully, finding issues with that sooner. :)

krisgesling:
Hi there, it seems like this came to a resolution of not to merge. So closing the PR.

PR #1738, opened 2018-08-15
Fix CLI lockup; simple 'speak' support; help display

penrods:
Several minor and simple CLI fixes:
* The CLI was locking up with at Lock() deadlock when rebuilding the
  filtered log.
* The simple_cli() wasn't registering handle_speak() on the websocket,
  even though the handler was setup for it.  The utterances are already
  clear from the Input: line or from the logs being printed, so no need
  to have a handle_utterance() registered in the simple_cli() case.
* Help wasn't refreshing the screen, so was invisible.
* An error could be thrown if Ctrl+Z was hit to suspend the 'fg' to
  return to foreground before any key was pressed.

## Description

## How to test


forslund:
Looks good!

PR #1747, opened 2018-08-15
Add feature to install beta skills

MatthewScholefield:
## Description
This adds two methods to `SkillManager` that load and save skill data to and from `~/.mycroft/skills.json`. This data has attributes about the skills that are installed, including whether or not they are installed as beta meaning that they will be upgraded automatically when the skill developer pushes a change (before it gets reviewed in the mycroft-skills repo).

## How to test
 - Run `git pull && git checkout feature/install-beta` in `/opt/mycroft/skills/mycroft-installer.mycroftai` to get the [beta change](https://github.com/MycroftAI/skill-installer/pull/31) in the installer skill
 - Pull this branch of mycroft-core (`git pull && git checkout feature/install-beta`)
 - Try installing skills as beta. Query examples:
    - `install mycroft timer beta` - upgrade or install the beta version
    - `install mycroft timer stable` - downgrade to stable or install the stable version
    - `install mycroft timer` - same as above

### DIFF
--- a/mycroft/skills/main.py
+++ b/mycroft/skills/main.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 #
 import gc
+import json
 import os
 import sys
 import time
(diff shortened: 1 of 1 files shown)


mycroft/skills/main.py
```diff
@@ -290,8 +307,12 @@ def download_skills(self, speak=False):
         default_names = set(chain(default_groups['default'], platform_groups))
         default_skill_errored = False
 
+        skills_data = self.load_skills_data()
+
         def install_or_update(skill):
             """Install missing defaults and update existing skills"""
+            if skills_data.get(skill.name, {}).get('beta'):
+                skill.sha = None
```
forslund: I assume this is what makes it fetch the head of the 18.02 branch? Maybe add a comment since it's non-obvious

MatthewScholefield: I realized I could just change it to `skill.sha = 'HEAD'`. Let me know if you think a comment would still be nice.

forslund: Nah, that's readable enough. will retest

PR #1751, opened 2018-08-17
 fix for broken communication with Arduino,  added missing skill.error.dialog for language de-de

domcross:
## Description
1) fix for broken communication with Arduino:
After Mid July update Mark-I faceplate (eyes and led display controlled by Arduino) no longer worked
see: https://community.mycroft.ai/t/mark-i-no-longer-works-after-update/3929/10?u=dominik
2) added missing skill.error.dialog for language de-de
when Mycroft lang is set to "de-de" and an internal skill error occurs Mycroft says "skill.error" as no skill.error.dialog for german exists

## How to test
apply patch, restart all mycroft-services

### DIFF
--- a/mycroft/client/enclosure/__init__.py
+++ b/mycroft/client/enclosure/__init__.py
@@ -69,7 +69,13 @@ class EnclosureReader(Thread):
             try:
                 data = self.serial.readline()[:-2]
                 if data:
-                    self.process(data.decode())
+                    try:
+                        data_str = data.decode()
+                    except UnicodeError as e:
+                        data_str = data.decode('utf-8', errors='replace')
+                        LOG.warning('Invalid characters in response from '
+                                    ' enclosure: {}'.format(repr(e)))
+                    self.process(data_str)
             except Exception as e:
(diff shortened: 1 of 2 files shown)


mycroft/client/enclosure/__init__.py
```diff
@@ -70,7 +70,8 @@ def read(self):
             try:
                 data = self.serial.readline()[:-2]
                 if data:
-                    self.process(data.decode())
+                    # TODO: catch and log 0xff in position 0
+                    self.process(data.decode('utf-8', errors='replace'))
```
forslund: Thanks!

Can you add the todo item as well: maybe something like
```python
try:
    data_str = data.decode()
except UnicodeError as e:
    data_str = data.decode('utf-8', errors='replace')
    LOG.warning('Invalid characters in response from enclosure: {}'.format(repr(e)))
self.process(data_str)
```

forslund:
I went ahead and fixed the final small issue. Thanks for submitting this patch, will be merging now!

PR #1757, opened 2018-08-21
Replace emitter/ws with bus

forslund:
## Description
Replaces emitter and ws variables/members with bus since these are used to interact with Mycroft messagebus. the fact that they are websocket/event emitters are of litte interest in the grander scheme of things.

## How to test
Make sure it looks somewhat nice and is functional.

CLA [Yes]

### DIFF
--- a/mycroft/audio/__main__.py
+++ b/mycroft/audio/__main__.py
@@ -31,14 +31,14 @@ def main():
     """ Main function. Run when file is invoked. """
     reset_sigint_handler()
     check_for_signal("isSpeaking")
-    ws = WebsocketClient()
-    Configuration.init(ws)
-    speech.init(ws)
+    bus = WebsocketClient()  # Connect to the Mycroft Messagebus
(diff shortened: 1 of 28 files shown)


AIIX:
oh..that broke a grand ton of display stuff on my skills between 18.02 to 18.08, applying changes to my skills now, for anyone else who comes across this the change seems basically from usage of: "self.enclosure.ws.emit(Message())" to "self.enclosure.bus.emit(Message())"

forslund:
Sorry about that @AIIX. Where are you using code like that, from skills? if that's the case you should use `self.bus` directly (we also kept the old `self.emitter` around for backwards compatibility)

forslund:
Good eyes. Corrected this in #1760

JarbasAl:
mpg123 service missing a lot of self. also, theres plenty of "bus" around that throw errors, audio service will fail to play because of this

PR #1760, opened 2018-08-22
Separate enclosure api from enclosure client

forslund:
## Description
Moves the display_manager and enclosure api out of the enclosure client so the enclosure can be completely replaced. The EnclosureWeather is merged into the EnclosureMouth since Weather uses the mouth area and only has a single handler.

Will break a lot of skills. timer, spotify, pandora, alarm, mark-1 and possible a couple more.

As an optional follow up the display_manager functions could be wrapped in a class. At the moment that doesn't provide anything but for the mark-2 it may be positive. But in that case it can be added at a later date.

Creds to @JarbasAI, some of this has been based on his work.

CLA [Yes]


forslund:
Wrapping the displaymanager in a class and including that as part of the EnclosureApi might be better, (see https://github.com/forslund/mycroft-core/tree/refactor/enclosure-2)

this allows skills to use `self.enclosure.display_manager` instead of importing the displaymanager separately.

forslund:
I've convinced myself that the display manager is better as a class even if it hasn't got much state at the moment. when expanding the architecture for the mark-2 the display manager can more easily be extended to do more complex things this way.

JarbasAl:
what about moving the enclosure to mycroft/enclosure instead of mycroft/client/enclosure ?

i think it makes more sense, specially if we want to include enclosures for stuff like google kit or respeaker

PR #1767, opened 2018-08-28
Move Mark-1 specific image code from api to mark-1 enclosure

forslund:
## Description
Move the image conversion from the enclosure api to the enclosure. Instead of converting and sending the display code to the enclosure the image path is sent to the enclosure and is converted there instead.

Also fix a bug preventing active skills from automatically being unregistered

## How to test
Make sure the mycroft-timer skill works as expected.

CLA [Yes]


penrods:
Timer works much better with this.  👍

PR #1781, opened 2018-09-05
Skip mimic build by default when already built

MatthewScholefield:
Description
-----------

Skip mimic build by default when it's already been built since it's much more common to skip it than build it. Now, -fm forces mimic to build in contrast to the previous -sm forcing it to skip

How to test
-----------
Make sure `./dev_setup.sh` and `./dev_setup.sh -fm` both work, without prompting to rebuild mimic

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -50,7 +50,7 @@ function show_help() {
     echo
     echo "Options:"
     echo "    -r, --allow-root  Allow to be run as root (e.g. sudo)"
-    echo "    -sm               Skip building mimic"
+    echo "    -fm               Force mimic build"
     echo "    -h, --help        Show this message"
     echo
     echo "This will prepare your environment for running the mycroft-core"
@@ -58,7 +58,7 @@ function show_help() {
     echo "not as root/sudo."
 }
 
-opt_skipmimic=false
+opt_forcemimicbuild=false
 opt_allowroot=false
 
 for var in "$@" ; do
@@ -71,8 +71,8 @@ for var in "$@" ; do
         opt_allowroot=true
     fi
 
-    if [[ ${var} == "-sm" ]] ; then
-        opt_skipmimic=true
+    if [[ ${var} == "-fm" ]] ; then
+        opt_forcemimicbuild=true
     fi
 done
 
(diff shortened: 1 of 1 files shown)


dev_setup.sh
```diff
@@ -135,14 +135,14 @@ install_deps
...
-    build_mimic="n"
+build_mimic="n"
+if [[ ${opt_forcemimicbuild} == true ]] ; then
+    build_mimic="y"
 else
     # first, look for a build of mimic in the folder
     has_mimic=""
     if [[ -f ${TOP}/mimic/bin/mimic ]] ; then
-        has_mimic=$( ${TOP}/mimic/bin/mimic -lv | grep Voice ) || true
+        has_mimic=$( ${TOP}/mimic/bin/mimic -lv | gbuildrep Voice ) || true
```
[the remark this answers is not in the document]
MatthewScholefield: Yep :laughing: nice catch

el-tocino:
I saw @forslund also posted a pi mimic package recently that could be used instead of building?

penrods:
Ah, if it still prompts when mimic isn't found, then I'm behind that.  This fits in with changes I'm intending to ./start-mycroft.sh that will optionally perform a ```git pull``` and invoke dev_setup.sh if necessary.

(FYI, I also intend to ask several questions on first-run, most likely switching most users to 'master' and enabling the auto ```git pull```)

MatthewScholefield:
Yeah, when it's not found it does the same as the previous behavior and just build it with a warning message.

dev_setup.sh
```diff
@@ -152,9 +152,8 @@ else
         fi
     fi
 
-    if ! [ "$has_mimic" == "" ] ; then
-        echo "Mimic is installed. Press 'y' to rebuild mimic, any other key to skip."
-        read -n1 build_mimic
+    if [ "$has_mimic" != "" ]; then
+        build_mimic="n"
```
forslund: Should this be reversed since `build_mimic` is `n` by default? So it builds mimic if not present?

```bash
    if [ "$has_mimic" == "" ]; then
        build_mimic="y"
    fi

MatthewScholefield: Oh, yes, you're absolutely right!

forslund:
Looks good now, merging

PR #1782, opened 2018-09-05
Allow local model file in precise

MatthewScholefield:
Description
-----------
This PR allows Precise to use a custom model file instead of checking for an updated one at boot

How to test
-----------

Place [a downloaded .pb model](https://github.com/MycroftAI/precise-data/blob/models/hey-mycroft.tar.gz) (Extract from tar) in a random location and point to it in the config. Make sure it works as expected.


forslund:
Tested and seem to be working as intended

PR #1784, opened 2018-09-07
Unify the command line experience across platforms

penrods:
Interacting with Mycroft was slightly different on different platforms
(Mark 1, Github, Picroft), which resulted in confusion and the beginnings of a
documentation/support nightmare.  This moves several of the common commands
into a mycroft-core/bin folder instead of being build in packaging or part
of the Picroft package only.

This required a small changes to the common scripts, in addition to adding
the new scripts.  Here are details:
* dev_setup.sh
  - Stubbed out a setup wizard (TODO)
  - Jumped to latest version of pip (18.0)
  - Set execution flags for new scripts
* start-mycroft.sh
  - Allow auto-execution of dev_setup.sh
  - De-sourced the stop-mycroft.sh calls (not needed)
  - Add quotes to better handle paths with spaces
* stop-mycroft.sh
  - Return exit code to show if a service was stopped or killed
  - Extend the time to wait for a gently shutdown from 2 secs to 5
  - Changed to say "messagebus.service" instead of just "service" to be more clear (the name "service" means nothing to users, but "bus does")
* venv-activate.sh
  - Added help
  - Added -q or --quiet mode
  - Added toggled availability of 'mycroft-venv-activate' and 'mycroft-venv-deactivate' aliases.  Only one is available at any given time.

New scripts:
* mycroft-cli-client: start up the CLI
* mycroft-help: shows hints on using mycroft
* mycroft-mic-test: runs the basic record/playback, playing nicely with a running Mycroft by stopping and restarting services as appropriate
* mycroft-msm: run Mycroft Skill Manager within the venv
* mycroft-msk: run Mycroft Skill Kit within the venv
* mycroft-pip: manage the venv via pip
* mycroft-say-to: send utterances, like a user spoke them
* mycroft-speak:  speaks the given phrase using Mycroft
* mycroft-skill-testrunner: invoke the single-skill test runner

All of these scripts are easy to discover by typing "mycroft-" then
hitting the TAB key.

==== INSTALLER NOTES ====
* We no longer need to create mycroft-cli-client and mycroft-pip in the Debian packages.
* mycroft-core/bin should be added to the path


start-mycroft.sh
```diff
@@ -144,13 +144,21 @@ function launch-all() {
 }
 
 function check-dependencies() {
-  if [ ! -f .installed ] || ! md5sum -c &> /dev/null < .installed ; then
-    echo "Please update dependencies by running ./dev_setup.sh again."
-    if command -v notify-send >/dev/null ; then
-      notify-send "Mycroft Dependencies Outdated" "Run ./dev_setup.sh again"
+    if [ ! -f .installed ] || ! md5sum -c &> /dev/null < .installed ; then
+        # Critical files have changed, dev_setup.sh should be run again
+        auto_update=$( jq -r ".auto_update" < .dev_opts.json )
```
forslund: Maybe add 2> /dev/null for the time being since this shows file not found error messages right now.

start-mycroft.sh
```diff
@@ -100,17 +100,17 @@ function launch-process() {
...
-        ${DIR}/scripts/prepare-msm.sh
+        "${DIR}/scripts/prepare-msm.sh"
         source-venv
         first_time=false
     fi
 
     name-to-script-path ${1}
 
-    # Check if already running
+    # Start (or restart if already running)
```
forslund: I believe the old comment might be good to keep since `pgrep -f "python3 -m ${_module}" > /dev/null` isn't trivial to understand.

bin/mycroft-help
```diff
@@ -0,0 +1,42 @@
...
+cd -P "$( dirname "$SOURCE" )"/..
+DIR="$( pwd )"
+
+echo -e "\e[36mMycroft\e[0m is your open source voice assistant.  Full source"
+echo "can be found at: ${DIR}"
+echo
+echo "Mycroft-specific commands you can use from the Linux command prompt:"
+echo "  mycroft-cli-client       command line client, useful for debugging"
+echo "  mycroft-msm              Mycroft Skills Manager, to manage your Skills"
+echo "  mycroft-msk              Mycroft Skills Kit, create and share Skills"
```
forslund: It seems like these two wrappers don't exist yet

dev_setup.sh
```diff
@@ -233,6 +250,12 @@ fi
 # set permissions for common scripts
 chmod +x start-mycroft.sh
 chmod +x stop-mycroft.sh
+chmod +x bin/mycroft-cli-client
+chmod +x bin/mycroft-help
+chmod +x bin/mycroft-mic-test
+chmod +x bin/mycroft-pip
+chmod +x bin/mycroft-skill-testrunner
+chmod +x bin/mycroft-speak
```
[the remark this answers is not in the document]
penrods: Hmm, this isn't possible on Windows -- I'm not sure if I trust that mechanism as long as there are boneheads like me around.

dev_setup.sh
```diff
@@ -165,7 +179,10 @@ fi
 source "${VIRTUALENV_ROOT}/bin/activate"
 cd "${TOP}"
 
-easy_install pip==9.0.1 # force version of pip
+# Force version of pip for reproducability, but there ins nothing special
```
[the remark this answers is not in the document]
penrods: Got it!

start-mycroft.sh
penrods: Fixed

bin/mycroft-help
penrods: Added

start-mycroft.sh
penrods: Clarified

dev_setup.sh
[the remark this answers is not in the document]
forslund: The problem currently that the files in bin gets marked as modified as soon as the dev_setup.sh is run and changes their exeutable flag. You need at least update the rights before checking in.

You can see how it's done in windows here: https://stackoverflow.com/questions/21691202/how-to-create-file-execute-mode-permissions-in-git-on-windows

forslund:
Too me this looks good now, is it still work in progress as the tag indicates?

penrods:
I think it is at a clean and safe place to merge.

PR #1789, opened 2018-09-10
Workaround for adapt issue with context + one_of

forslund:
## Description
Adapt doesn't populate the entry from the one_of correctly from context. To work around the issue intent structure is scanned for empty keys and tries to populate them from entities in __tags__

## How to test
Check out the PR, verify that normal operation of mycroft isn't affected.

CLA [Yes]


penrods:
Not thrilled with having to do this, but can live with it until the Adapt bug (https://github.com/MycroftAI/adapt/issues/86) is dealt with.

PR #1797, opened 2018-09-14
Fix incorrect escape sequences found by skill test

forslund:
## Description
One occurrence in `dialog/__init__.py` and one in `skill_tester.py`. To handle this the strings were marked as raw-strings.

## How to test
Check that dialogs are rendered as intended.

CLA [Yes]


penrods:
LGTM

PR #1798, opened 2018-09-14
Update Arch (pacman) install for Python 3

penrods:
Installation under Arch Linux wasn't working for many because the dependencies
hadn't been updated for Python 3.  Mycroft would only work if all of the needed
components were already installed on the user's system.

Thanks to community member Chowbok for the help straightening this out!

## How to test
Install under Arch Linux

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -104,15 +104,18 @@ function install_deps() {
     echo "Installing packages..."
     if found_exe sudo ; then
         SUDO=sudo
+    else
+        echo "This script requires \"sudo\" to install system packages. Please install it, then re-run this script."
+        exit 1
     fi
 
     if found_exe zypper ; then
-        $SUDO zypper install -y git python glibc-devel linux-glibc-devel python-devel python2-virtualenv python2-gobject-devel python-virtualenvwrapper libtool libffi-devel libopenssl-devel autoconf automake bison swig glib2-devel portaudio-devel mpg123 flac curl libicu-devel pkg-config pkg-config libjpeg-devel libfann-devel python-curses
+        $SUDO zypper install -y git python glibc-devel linux-glibc-devel python-devel python2-virtualenv python2-gobject-devel python-virtualenvwrapper libtool libffi-devel libopenssl-devel autoconf automake bison swig glib2-devel portaudio-devel mpg123 flac curl libicu-devel pkg-config libjpeg-devel libfann-devel python-curses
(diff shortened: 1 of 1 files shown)


kjcole:
My mileage varied. ext/_yaml.c failed. Specifically:

```
gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -flto -fuse-linker-\
plugin -ffat-lto-objects -flto-partition=none -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -march=x86-64 -mtune=generic -O2 -pipe -fstack-prote\
ctor-strong -fno-plt -fPIC -I/usr/share/mycroft-core/.venv/include -I/usr/include/python3.7m -c ext/_yaml.c -o build/temp.linux-x86_64-3.7/ext/_yaml.o

```
[mycroft-core.log](https://github.com/MycroftAI/mycroft-core/files/2384998/mycroft-core.log)

chowbok:
I'm guessing this has to do with the out-of-date Arch libyaml package. The current version is 0.2.1 and the package has 0.1.7. The pyyaml build works fine if libyaml isn't installed at all, but if you have it, it sucks in the system include files and barfs.

I'll see if I can get the maintainer to update libyaml.

chowbok:
kjcole's issue doesn't really have much to do with my patch; probably should be a new bug.

chowbok:
If anyone's interested, I built [an updated libyaml package](https://www.unknown.nu/temp/libyaml-0.2.1-1-x86_64.pkg.tar.xz) for Arch if you want to try it. Or you can try just temporarily removing libyaml long enough to build Mycroft.

penrods:
What do you two think, @chowbok and @kjcole ?  Shall we merge this as-is?

chowbok:
> What do you two think, @chowbok and @kjcole ? Shall we merge this as-is?

Yes. Without this patch, Arch installs won't work at all (because they run some of the skills under Python 2 and all kinds of weird errors ensue). With it, it works fine unless that old libyaml package is installed. 

I guess I could put in logic for the script to exit if the old libyaml is there, although it seems like a better fix would be to configure the built-in pyyaml build to ignore the system libyaml install if it's below a certain version. Not sure how difficult that would be, though.

penrods:
I'm good with it as-is.  Merging!

PR #1801, opened 2018-09-17
Add locking when accessing the IdentityManager

forslund:
## Description
Adds the mycroft.util.combo_lock ComboLock class for interprocess/Thread
lock.

Loading updated to be more reliable:
- Flush and sync file
- wait 1.2 seconds before load

Split the logic from the locking so the lock can be avoided when calling
update from save or load from get.

## How to test
This one requires a bit of code review. Please check that the locking logic is good and if there's deadlock conditions I haven't thought of.

CLA [Yes]


forslund:
Since this hasn't been widely tested yet let's hold of (a potential) merge until after the release on thursday.

PR #1803, opened 2018-09-19
Fix ambiguous time handling

forslund:
## Description

Before noon an ambigous time (like *feed the fish at 10 o'clock*) skipped a day forward. This updates the logic to handle a bit better.

Adds a test for the ambigous time

## How to test
Test with alarm skill and/or reminder skill and check that the times are what's expected.

CLA [Yes]


forslund:
Replaced by #1804, closing

PR #1804, opened 2018-09-21
Feature/extract datetime update

forslund:
## Description
- Fixes issue with ambiguous time (replaces PR #1803)
- Removes some unused code
- Adds an optional parameter default_time allowing the skill creator to determine the time to set if no time was found in the input string

## How to test
Verify that extract_datetime is still functional.
 
CLA [Yes]

### DIFF
--- a/mycroft/util/lang/parse_de.py
+++ b/mycroft/util/lang/parse_de.py
@@ -426,8 +426,8 @@ def extract_datetime_de(string, currentDate, default_time):
     hrOffset = 0
     minOffset = 0
     secOffset = 0
-    hrAbs = 0
-    minAbs = 0
+    hrAbs = None
+    minAbs = None
 
     for idx, word in enumerate(words):
         if word == "":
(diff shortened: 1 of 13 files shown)


penrods:
This is looking good.  How do you feel about updating the language variant versions?  I'm concerned that we get out of sync on these complex functions and I am trying to figure out how we manage these sorts of changes going forward.  If we can't update them completely to match, I think I'd rather leave the language specific code alone so a future "language expert" dev can just look at all of the changes in the base English code.

forslund:
Swedish and german should definitely be doable, I'll look at the latin based languages as well and see if I can decipher enough to fix it.

forslund:
If not we can maybe set them as possible hacktober-tasks

forslund:
Updated for all languages.

JarbasAl:
i will handle all the portuguese parsing methods

penrods:
Awesome!  ,Check out what we merged and shout if you see any issues.  Building out the test suite while confirming operation will pay dividends later.

Portuguese tests:  https://github.com/MycroftAI/mycroft-core/blob/dev/test/unittests/util/test_parse_pt.py

English tests:  https://github.com/MycroftAI/mycroft-core/blob/dev/test/unittests/util/test_parse.py

PR #1813, opened 2018-09-26
Add ability to schedule event in seconds

penrods:
The MycroftSkill.schedule_event() method now accepts an integer in addition to
a datetime for the 'when' parameter.  The integer represents the number of
seconds in the future to fire off the event.  E.g.
```python
   self.schedule_event(some_handler, 7)
```
Will invoke some_handler() seven seconds from now.

Also unified language used in event docstrings.

## How to test
Add this to a skill:

```python
   def initialize(self):
       self.schedule_event(self.test, 15)  # should speak 15 seconds after (re)load

   def test(self, message):
       self.speak("Success!")
```

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -1049,7 +1049,7 @@ class MycroftSkill(object):
                 self.bus.emit(Message("mycroft.stop.handled",
                                       {"by": "skill:"+str(self.skill_id)}))
             timer.cancel()
-        except:
+        except:  # noqa
             timer.cancel()
             LOG.error("Failed to stop skill: {}".format(self.name),
                       exc_info=True)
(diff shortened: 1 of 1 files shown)


mycroft/skills/core.py
```diff
@@ -1049,7 +1049,7 @@ def __stop_timeout():
                 self.bus.emit(Message("mycroft.stop.handled",
                                       {"by": "skill:"+str(self.skill_id)}))
             timer.cancel()
-        except:
+        except:  # noqa
```
forslund: Maybe use `except Exception:` instead of bare except. This will catch KeyboardInterrupt and other things.

penrods: I'm fine with that on this one.  I'll leave it alone in the except on line 1093.

penrods: Done!

forslund:
Working nicely. Merging

PR #1815, opened 2018-09-26
feature: allow adapt context across skills

JarbasAl:
## Description

self.set_context does not let you share contexts across skills because of the intent name munged in the context

This PR adds a new method to set a context in all skills

This allows for better interactions between skills 

Someone asked about this in mattermost recently, i also need it in my location tracker skill to provide a default location

EDIT: 

added methods to also remove context

fixed a bug with remove context where skill id wasn't munged, 

`context = to_alnum(self.skill_id) + context`

## How to test

in skill A use self.set_cross_skill_context, check that you can trigger an intent from skill B requiring that context

from skill C remove the context with self.remove_cross_skill_context, check that you can not trigger skill B, doing this you are also testing the bug fix for the standard remove_context, it is now correctly munging skill id into the keyword name

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -865,7 +865,7 @@ class MycroftSkill(object):
                                                       'registered.')
         return False
 
-    def set_context(self, context, word=''):
+    def set_context(self, context, word='', origin=None):
         """
             Add context to intent service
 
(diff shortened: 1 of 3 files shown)


JarbasAl:
should i add an equivalent method to remove the contexts?

forslund:
Looks pretty good. And well spotted on the remove_context.

I will test and check how the context scoring is affected by this method.

forslund:
I did a couple of tests and it's a bit of an issue that the context stack gets flooded.

I did this test-addition which makes the cross-context entries only add 1 to the depth even if there are hundreds of entries.

This adds an "origin" which is "" for normal contexts and skill_id of the origintating skill. This is not a 100% good solution if a skill sets multiple cross skill contexes in a row they are all treated as the same level. Maybe origin should be a timestamp or the skill id + context.

What do you think?

PR #1817, opened 2018-09-26
Strip trailing blank lines

forslund:
## Description
In ascii mode many editors leaves a trailing \n at the end of file. This recently caused an issue with the installer-skill's german translation.

This will also remove trailing empty lines from the end of dialog files. If a blank line is intended add a single space and it will be included.

## How to test
Check that skills that use the feature such as installer still works.

CLA [Yes]


forslund:
Might need some consideration, but I've not seen anywhere where this would be bad

PR #1824, opened 2018-09-29
Create nice_unit and expand_unit helper methods

### LINKED ISSUE #1802: Feature Request: Expand units for speaking
Just like the currently available methods [`nice_number`](https://github.com/MycroftAI/mycroft-core/blob/bd6f9ffde1baf00606ba0eb0a12382a24ec6abad/mycroft/util/format.py#L193) for converting an integer to a nicely speakable version, there could be a method `nice_unit` or `expand_unit` for converting string containing units to a speakable string.

i.e. "The power plant produced 10Wh today" => "The power plant produced ten watt hours today"
"The bathroom temperature is 22 C" => "The bathroom temperature is twenty two degrees celsius"

There exist tools for unit extraction from unstructured text like [quantulum](https://github.com/marcolagi/quantulum) (or its python3 equivalent with further development [quantulum3](https://github.com/nielstron/quantulum3)).
These tools can be used (or supply a method, see `inline_parse_and_expand` at quantulum3), to convert the unit into a string. 

It can be a very helpful method, used to expand wikipedia articles, wolframalpha answers or home assistant replies.

nielstron:
## Description
Resolves #1802 

## How to test
Run the methods nice_unit and expand_unit in mycroft/util/format.py on some example strings like "It's cold outside. Almost -10 C!"

This likely needs to wait until quantulum3 is python 2.7 compatible (for which there is an open pull request, only waiting on inflects python 2 compatibility)

CLA [ Yes ]


JarbasAl:
i like this, but im concerned with language support, relevant issue https://github.com/nielstron/quantulum3/issues/55

but i see the units are stored in json files, i think a good first approach would be to have those json per language, maybe like for date time https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/res/text/en-us/date_time.json

not sure if that should be at mycroft level, or in quantulum itself

requiring numpy, scipy and sklearn just for this is also overkill, but those seem to be optional :+1:

nielstron:
@JarbasAI language support is now being tackled, the only remaining problem is some stubbornness of unittest in python 3.4...

I'd prefer keeping this an external library thing so that the processing tools can be used for different projects too. It is quite an extensive project and I see usages for it at a lot of places.

The classifier for disambiguation is optional yet strongly encouraged, as it has quite a high probability of successfully determining which unit was actually meant. In fact I'm surprised that numpy and scipy aren't already install requirements of the mycroft core.

JarbasAl:
i think for now returning None or the original string for missing languages would be enough, i agree language support should go into the main package!

JarbasAl:
i believe this should now be submitted against https://github.com/MycroftAI/lingua-franca

forslund:
This is a lingua franca issue now so I'm going to close this PR. I do think the language team would appreciate this pr on that repo instead.

PR #1825, opened 2018-09-30
Add protection for naive skill authors

penrods:
It is fairly common for new skill authors to attempt actions in the __init__()
method which are not legal yet, as the Skill has not been fully connected to
the Mycroft system.  This adds an @property protection layer for the two most common
issues:
* Accessing MycroftSkill.bus
* Accessing MycroftSkill.enclosure

Now those are properties instead of variables and provide appropriate warnings
when used before they exist.

Also enhancing the handling of error logs in the CLI to highlight problems such
as this:
* Color "- ERROR -" log messages in red
* Retaining leading characters from log messages, improving readability in formatted messages

## How to test
Create a skill that uses the bus in the __init__(), e.g.:
```python
class PersonalSkill(MycroftSkill):

    def __init__(self):
        super().__init__(name="PersonalSkill")
        self.bus.emit("Don't do it!")   # access bus directly
        self.speak("Don't do i t!")     # access bus inside speak()
        self.enclosure.eyes_blink()     # access enclosure
```

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -20,6 +20,7 @@ import time
 import csv
 import inspect
 import os
+import traceback
 from inspect import signature
 from datetime import datetime, timedelta
 
@@ -216,7 +217,8 @@ class MycroftSkill(object):
         self._dir = dirname(abspath(sys.modules[self.__module__].__file__))
         self.settings = SkillSettings(self._dir, self.name)
 
-        self.bus = None
+        self._bus = None
+        self._enclosure = None
         self.bind(bus)
         self.config_core = Configuration.get()
         self.config = self.config_core.get(self.name) or {}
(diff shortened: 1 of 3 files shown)


mycroft/skills/core.py
```diff
@@ -231,13 +233,41 @@ def __init__(self, name=None, bus=None):
...
+                      " __init__() to initialize() to correct this.")
+            tb = "Traceback:\n"
+            for line in traceback.format_stack()[:-1]:
+                if line.strip():
+                    tb += line
+            LOG.error(tb)
+            raise Exception("Accessed MycroftSkill.enclosure in __init__")
+
+    @property
+    def bus(self):
```
forslund: Should we add setters for these as well?

```python
@bus.setter
def bus(self, value):
    self._bus = value
```

mycroft/skills/core.py
```diff
@@ -231,13 +233,41 @@ def __init__(self, name=None, bus=None):
...
+        if self._enclosure:
+            return self._enclosure
+        else:
+            LOG.error("ERROR:  Skill not fully initialized.  Move code from " +
+                      " __init__() to initialize() to correct this.")
+            tb = "Traceback:\n"
+            for line in traceback.format_stack()[:-1]:
+                if line.strip():
+                    tb += line
+            LOG.error(tb)
```
forslund: What is the benefit of this over using `LOG.exception()` to log the traceback?

mycroft/skills/core.py
penrods: I assume we don't really want people doing a straight "set", they should call MycroftSkill.bind().

mycroft/skills/core.py
penrods: This isn't happening inside an exception -- I'm just dumping the stack trace to show the skill author where they goofed up.  I also strip out a level of the call stack so it is more obvious.

I suppose I could throw an exception instead, but this really enforces that they should never be accessing self.bus or self.enclosure until AFTER the __init__() has completed.  This will always dump text into the logs even if they do gracefully handle the self.bus being None.

forslund: Ok, then I get the intent of this.

mycroft/skills/core.py
forslund: Fair point. The problem as I see it is when it's used as in padatious_service where ._bus is set and then .bus is used. which looks like a typo when just looking at that file.

forslund:
Looks good, I'll merge this as it is

PR #1827, opened 2018-09-30
Fix date format for today/tomorrow/yesterday

penrods:
The test for today/tomorrow/yesterday was only looking at the day,
not the month and year.  So on July 14, 2018 it would claim that
the date June 14 2018 and July 14 2020 are all "today".

## How to test
"Create an alarm for 30 days from now" would say the alarm was set for today.  Now
it should list the future date.

### DIFF
--- a/mycroft/util/format.py
+++ b/mycroft/util/format.py
@@ -148,7 +148,7 @@ class DateTimeFormat:
                 format_str = 'tomorrow'
             elif now.date() == dt.date():
                 format_str = 'today'
-            elif yesterday.date() = dt.date():
+            elif yesterday.date() == dt.date():
                 format_str = 'yesterday'
 
         return self.lang_config[lang]['date_format'][format_str].format(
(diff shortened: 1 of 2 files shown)


mycroft/util/format.py
```diff
@@ -142,11 +142,16 @@ def date_format(self, dt, lang, now):
                 if dt.month == now.month and dt.day > now.day:
                     format_str = 'date_full_no_year_month'
 
-            if (now + datetime.timedelta(1)).day == dt.day:
+            tomorrow = now + datetime.timedelta(1)
```
forslund: For readability maybe specify datetime.timedelta(days=1)

mycroft/util/format.py
```diff
@@ -142,11 +142,16 @@ def date_format(self, dt, lang, now):
...
+            yesterday = now - datetime.timedelta(1)
+            if (tomorrow.day == dt.day and tomorrow.month == dt.month and
```
forslund: I think this can be simplified by using
```python
if tomorrow.date() == dt.date():
```

The same goes for the ones below

penrods: Yep, much cleaner

mycroft/util/format.py
penrods: Done

PR #1828, opened 2018-10-01
Make the combo lock handle multiple users

forslund:
## Description
On the Mark-1 the locking would fail since the first time the lock is
accessed the process is run under root (script checking the enclosure
version). This caused the locking to fail since the normal mycroft
processes got an access error (lock-file owned by root)

This change creates lock-files and sets the permissions to `0o777`
if the file doesn't exist so it's accessible for all users

## How to test
Install the change on a mark-1, reboot and verify that all the processes starts correctly.

CLA [Yes]

### DIFF
--- a/mycroft/util/combo_lock.py
+++ b/mycroft/util/combo_lock.py
@@ -1,10 +1,22 @@
 from threading import Lock
 from fasteners.process_lock import InterProcessLock
+from os.path import exists
+from os import chmod
 
 
 class ComboLock():
-    """ A combined process and thread lock. """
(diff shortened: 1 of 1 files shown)


forslund:
Hold on on this one a bit longer, I'm rechecking some things. (It suddently stopped working again)

forslund:
False alarm, it was my half done pulse changes that made the skill loading halt when getting to the volume control

penrods:
I'm assuming Codacy is stuck on something else.  I'm ignoring the "Pending".

Issue #1556, opened 2018-04-25
Unlocalized functions should warn if they are called with an unsupported language


zeehio:
Some functions in https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/util/parse.py (extract_number, extraxt_datetime...) need to be localized. When they are used in an unsupported language they return something (usually the input) without raising a warning.

To make everyone aware that these functions have to be localized, they should raise a warning or log something if the function is being used in an unsupported language. Otherwise some skill translations will not work and we will not know where the error comes from.

forslund:
Fixed by #1831

PR #1831, opened 2018-10-03
Language error checking

silvia-odwyer:
## Description
I added error messages to functions that required a language from the set supported by MycroftAI. So if a user uses a language that isn't supported, the error message will be returned. 
This fixes issue #1556 

## How to test
Use an unsupported language for the `normalize`, `extract_number`, and `extract_datetime` functions found in `parse.py`.

Yes, I have signed the CLA.

### DIFF
--- a/mycroft/util/parse.py
+++ b/mycroft/util/parse.py
@@ -103,8 +103,9 @@ def extract_number(text, short_scale=True, ordinals=False, lang="en-us"):
     elif lang_lower.startswith("de"):
         return extractnumber_de(text)
     # TODO: extractnumber_xx for other languages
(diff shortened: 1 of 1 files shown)


forslund:
Hi and thanks for this submission. I would like to change this a bit. Instead of returning the error as string keep the original return and instead add a log message:
```python
    LOG.warning('FUNCTION does not support something something)`
```

Also the coding guidelines we enforce uses 79 character line length, thus this is not passing the automatic tests.

python allows you to split strings on multiple lines:
```python
    LOG.warning('Error: Language not recognized! Please make sure your language is'
                ' one of the following: en, es, pt, it, fr, sv, de.')
```

forslund:
Thank you very much. I'll fix the remaining pip issues and merge.

PR #1837, opened 2018-10-08
centos support

### LINKED ISSUE #1836: Centos 7 support
*I am trying to install Mycroft from the repo (what ever that newest version is)
*I am trying to install this on Centos 7 (which I know is not supported yet)

My basic process for understanding is listed as follows on a fresh minimal install.

1. sudo yum install epel-release
2. sudo yum update -y
3. sudo yum upgrade -y
4. sudo yum install nano wget git python-pip python-devel (This is not all of them)
5. sudo yum groupinstall 'development tools'

The main issue here is that dev_setup.sh will not work because it cannot use apt and it needs yum. However I was able to get pip to install the entire requirement.txt (with some work). I can go through the process again and correct every issue I ran into originally so we can get this working. However I am not very good with scripts so I would just need help fixing that so that it can easily be install on centos.

JarbasAl:
## Description
fixes #1836 

## How to test
run dev setup in centOs, or other distro that uses yum

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -113,7 +113,14 @@ function install_deps() {
         $SUDO zypper install -y git python3 python3-devel libtool libffi-devel libopenssl-devel autoconf automake bison swig portaudio-devel mpg123 flac curl libicu-devel pkg-config libjpeg-devel libfann-devel python3-curses pulseaudio
         $SUDO zypper install -y -t pattern devel_C_C++
     elif found_exe yum ; then
-        $SUDO yum install -y git python34 python34-devel libtool libffi-devel openssl-devel autoconf automake bison swig portaudio-devel mpg123 flac curl libicu-devel python34-pkgconfig libjpeg-devel fann-devel python34-libs pulseaudio
+        $SUDO yum install epel-release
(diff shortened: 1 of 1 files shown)


JarbasAl:
hunting down some packages... seems python3 is not available in yum

JarbasAl:
thats not the correct package, it seems theres no equivalent according to https://repology.org/metapackage/libfann/versions

i am now compiling from source, doing final test in a fresh vm but everything seems to be working

forslund:
Hmm, the manifest matches the debian package pretty well...

What issue do you have with it?

In any case you should probably use the fixed release  instead of HEAD, the tag releases/tag/2.2.0 or just use the tar.gz https://github.com/libfann/fann/archive/2.2.0.tar.gz

JarbasAl:
pip simply fails to install fann if using the provided package, from source it works

forslund:
It could be the multiarch issue with the fann pip package that's acting up again.

I still think the release tag or release tar.gz should be used

JarbasAl:
i have no way to test this, if someone wants to test and propose changes i can add them, but until some real user gives the working changes i don't want to touch this since it is currently working (and just using the other package is confirmed not to work)

forslund:
If you don't want to use the tag, at least fix the SHA to the current head commit.

forslund:
Thanks @JarbasAI and @assasen11. Merging now

PR #1852, opened 2018-10-16
Add check separating centOS and Fedora

### LINKED ISSUE #1851: dev_setup script issues
I am running the latest Git version of Mycroft dev on Fedora 28. When running dev_setup.sh I get the error `no match for argument: epel-release`.

Looking in the script, on line 115 it has information on how to install on CentOS and RHEL (I assume), and then on 136 has how to install on Fedora. The issue lies in that `hash yum` in Fedora 28 (and earlier versions) does not return and error and so the script thinks I'm not running a later version of Fedora.

~~Maybe we should make line 115 read `if found_exe yum and not found_exe dnf ; then` to resolve this issue.~~ This did not work, but moving the dnf line did.

Also, the script did work on the Fedora 29 beta, but I think that's because the Fedora devs removed all `yum` calls.

forslund:
## Description
Use /etc/os-release scanning for the ID present in os-release. Which is
supported by most modern linux distributions.

Should resolve #1851 

## How to test
Test that the dev_setup.sh script doesn't error on Fedora

CLA [Yes]

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -96,6 +96,11 @@ fi
     #   echo '{"use_branch":"dev", "auto_update": false}' > .dev_opts.json
 # fi
 
+
+function os_is() {
+    [[ $(grep "^ID=" /etc/os-release | awk -F'=' '/^ID/ {print $2}' | sed 's/\"//g') == $1 ]]
+}
+
 function found_exe() {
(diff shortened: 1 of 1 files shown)


forslund:
Tested on Fedora by @testman42 and on centos by cschantz on mattermost

PR #1855, opened 2018-10-17
add support for -prefix to dev_setup.sh

### LINKED ISSUE #1854: Mycroft installs skills in /opt and logs in /var
I'm using Debian and would like to use Mycroft in a separate directory, not in the system directory.

```
+ echo 'Creating /var/log/mycroft/ directory'
Creating /var/log/mycroft/ directory
+ [[ ! -d /var/log/mycroft/ ]]
+ sudo mkdir /var/log/mycroft/
[sudo] contrasenya per a ftyers: 
```

And:

```
$ bash start-mycroft.sh debug
Starting all mycroft-core services
Initializing...
Create /opt/mycroft/skills
[sudo] contrasenya per a ftyers: 
Starting background service bus
CAUTION: The Mycroft bus is an open websocket with no built-in security
         measures.  You are responsible for protecting the local port
         8181 with a firewall as appropriate.
```

It shouldn't be hard to have these be relative paths, e.g. created in a `$prefix` with `--prefix` or something like that.

ftyers:
## Description

Fixes  #1854

## How to test

You can run `dev_setup.sh` with `-prefix /path/to/your/prefix`.

Requested an agreement (PS. there is a typo on the page "contributer").


dev_setup.sh
```diff
@@ -74,8 +78,20 @@ for var in "$@" ; do
...
+
+    prev=${var}
 done
 
+echo ${prefix}
+pfound=$(grep "^\.prefix$" .gitignore | wc -l)
+if [[ $pfound != 0 ]]; then 
+	echo ".prefix" >> .gitignore; 
+fi
+
```
forslund: .gitignore doesn't need to be updated by the dev_setup script. You can just add .prefix to the .gitignore file and commit that one.

This might even cause some issues since the .gitignore is under version control and after the dev_setup script the repo would read as dirty.

mycroft/configuration/mycroft.conf
```diff
@@ -20,7 +20,7 @@
   // Language used for speech-to-text and text-to-speech.
   // Code is a BCP-47 identifier (https://tools.ietf.org/html/bcp47), lowercased
   // TODO: save unmodified, lowercase upon demand
-  "lang": "en-us",
+  "lang": "ca-es",
```
forslund: I think you've unintentionally committed changes in the mycroft.conf file

forslund:
Hi @ftyers Since there's been no updates on this for some time I'll close it.

Hope everything is well with you.

PR #1856, opened 2018-10-19
Feature/rhel

forslund:
## Description
Adding support for Redhat Enterprise Linux. This also includes PR #1852 since the same mechanism is used.

This has been developed together with **cschantz** from the chat who has tested it.

CLA [Yes]


penrods:
LGTM!

PR #1857, opened 2018-10-23
Bugfix/audioservice invalid utterance

forslund:
## Description
The CommonPlaybackSkill may send None as an utterance when invoking the audioservice.play() method. This fix makes sure that None is handled in a sane way.

Additionally the CPS_play method has been altered to not add an utterance if the skill tries to add one.

## How to test
Try invoking the NPR_news skill with "what's the news" and make sure it works.

CLA [Yes]


forslund:
Codacy seems to have hung. I'm going ahead with the merge

PR #1866, opened 2018-11-07
extract date time improvement/bugfix

JarbasAl:
## Description

improves datetime, it now supports extracting "now"

fixes a couple edge cases where day offset was wrong, "in a minute", "in an hour"

## How to test
check the unittests


forslund:
Looks good. I can't seem to recreate the incorrect date when using *in a hour* on dev. is an issue at some specific time of day?

Can't see anything wrong so I'll be merging anyway but would be interesting to know the exact circumstances.

JarbasAl:
it only happened after 12 i think, when you changed for the next day, the problem was the singular form, the code only accounted for plural (minutes and hours)

PR #1867, opened 2018-11-07
feature/extract multiple numbers from text

JarbasAl:
## Description

extract_number assumes only 1 number is present in text, this adds a extract_numbers method to support cases like "test one two three"

normalize only accounts for numbers up to 20 (word to digit), method is improved to handle all extracted numbers by above method (needs improvements in further PRs)

## How to test

check unit tests


mycroft/util/lang/parse_en.py
```diff
@@ -1039,15 +1040,21 @@ def normalize_en(text, remove_articles):
...
-                       "thirteen", "fourteen", "fifteen", "sixteen",
-                       "seventeen", "eighteen", "nineteen", "twenty"]
-
-        if word in textNumbers:
-            word = str(textNumbers.index(word))
-
         normalized += " " + word
 
+    # needs to be here to avoid import error
+    from mycroft.util.parse import extract_numbers
```
forslund: I'm not a fan of doing these kinds of workarounds for cyclic imports.

What do you think of creating a base method for extract_numbers in `lang/parse_common.py` which takes the extract_number function as an argument and a just a thin wrapper function in `util/parse.py`?

JarbasAl: sounds good!

JarbasAl:
added support for more languages, but since i don't speak them i would appreciate someone to write unittests

it is using the "extract number" logic already implemented for each language recursively, so it should be correct

forslund:
Is the intent to only match numbers directly following eachother?

"so two beers for two bears" should return `[2]` and not `[2, 2]`?

JarbasAl:
it should extract every number, "1 dog, seven pigs, macdonald had a farm, 3 times 5 macarena" should also work

forslund:
The example I gave above didn't work for me.

JarbasAl:
you are right, i need to account for duplicate numbers, what is happening is that when it find the first "two" it replaces all instances in the string instead of only the consumed one

will account for this

JarbasAl:
moved main logic into parse_common to make it easier for supporting other languages

improved normalization of numbers, now "two twenty two" correctly becomes "2 22" instead of "2 twenty 2"

JarbasAl:
geez, how come pycharms auto format is creating new pep8 issues instead of solving them...

forslund:
Looks good. I can't figure out how to clean up the commit history so I'll just squash it all into one commit and merge it.

forslund:
Ok a conflict seem to have occurred when I merged the datetime update. Can you look at it?

JarbasAl:
rebased and added some more relevant unittests, wanted to leave some TODOs in there to make some limitations clear

follow up PR for pronounce_numbers (ordinals and long scale support) should fix some of those

follow up PR for issue https://github.com/MycroftAI/mycroft-core/issues/1875 will fix remaining edge cases

PR #1873, opened 2018-11-16
bugfix/utterance remainder, only replace complete words

JarbasAl:
## Description

“Turn the room light on in 30 seconds” will return a remainder of “Turn in 30 secds”

utterance remainder now is parsed word by word instead of just replacing the keywords in the string

[bug originally reported here](https://community.mycroft.ai/t/intent-builder-vocab-files-extracting-from-remainder-portion-of-utterance/5115)

## How to test

make an intent with following required keywords

state.voc -> on / off
action.voc -> turn / switch
object.voc -> light / bulb / plug

say “Turn the room light on in 30 seconds” 

check utterance remainder, notice "on" will no longer be wrongly replaced in "seconds"

### DIFF
--- a/mycroft/messagebus/message.py
+++ b/mycroft/messagebus/message.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 #
 import json
+import re
 from mycroft.util.parse import normalize
 
 
@@ -160,5 +161,6 @@ class Message(object):
         utt = normalize(self.data.get("utterance", ""))
         if utt and "__tags__" in self.data:
             for token in self.data["__tags__"]:
-                utt = utt.replace(token.get("key", ""), "")
+                # Substitute only whole words matching the token
+                utt = re.sub(r'\b' + token.get("key", "") + r"\b", "", utt)
         return normalize(utt)


forslund:
Can we do something slightly simpler:
```python
        utt = normalize(self.data.get("utterance", ""))
        if utt and "__tags__" in self.data:
            for token in self.data["__tags__"]:
                if utt == token.get("key", ""):
                    utt = ""
                    break
                utt = utt.replace(token.get("key", "") + " ", "")
                utt = utt.replace(" " + token.get("key", ""), "")
        return normalize(utt)
```

or using a simple regex:
```python
        utt = normalize(self.data.get("utterance", ""))
        if utt and "__tags__" in self.data:
            for token in self.data["__tags__"]:
                # Substitute only whole words matching the token
                utt = re.sub(r'\b' + token.get("key", "") + r"\b", "", utt)
        return normalize(utt)
```

JarbasAl:
regex could work, the "simple" way would not account for words that end or start with the substring, "ongoing" would become "going" etc

forslund:
True true. the regex is my favorite anyways :)

forslund:
Looks excellent! I took the liberty of rebasing and squashing your commits.

Issue #1865, opened 2018-11-06
It should be possible to set the (grammatical) gender of Mycroft


ftyers:
For some languages (Portuguese, Spanish) it will be necessary to set the gender of Mycroft according to the voice (female, male) and allow this to be referenced in the dialogues. For example, in Portuguese
the word for thankyou is *obrigado* (male speaker) and *obrigada* (female speaker). Or "Qué tal Mycroft?" ... "Estoy cansado" (male speaker) / "Estoy cansada" (female speaker)

For other languages, it would be convenient to know the gender of the person speaking to Mycroft to be able to choose the correct greeting. E.g. Icelandic *sæll* for addressing a male, and *sæl* for addressing a female. This could be either estimated at utterance time using the speech signal, or alternatively set on a per-user basis.

ftyers:
There are lots of fun typological things to take into account :) I'll try and continue posting it here as I think about it. For the second one, a short-term solution would be to allow a gender to be associated to a particular user. 

In the end, the more linguistically extensible the platform is, the easier it will be to adapt. Would it be useful to you guys to have a document that discusses language typology and dialogue systems?

KathyReid:
One of the elements to consider here is also the choice and agency of the speaker. Some speakers may have a male-sounding voice, but choose to express their gender as female, and vice versa. Some speakers may not associated with one of the binary genders (male/female). So to be inclusive and respectful, we also need to consider these use cases and how to handle them in a way that provides a positive user experience. 

_Side note: I'm reading "Made by Humans" by Ellen Broad at the moment, it's such a mind opener_
https://www.goodreads.com/book/show/39806467-made-by-humans

ftyers:
@KathyReid yes, but this should be programmable, we shouldn't have "male morphology with female voice" or "male morphology with male voice" as the only options.

KathyReid:
> In the end, the more linguistically extensible the platform is, the easier it will be to adapt. Would it be useful to you guys to have a document that discusses language typology and dialogue systems?

Yes! Absolutely! Thank you for volunteering!

JarbasAl:
as a portuguese speaker the thank you example really annoys me, i proposed a solution for that, the second part is trickier

i played with detecting gender from utterance audio using gmmhmms , maybe it was the training data but results were mixed, mostly correct but i think we need a strong baseline before integrating something like this in the pipeline

associating per user could work, and would account for gender identity preferences, but we don't yet have speaker identification in the pipeline either

[...]

PR #1874, opened 2018-11-16
grammatical gender support

JarbasAl:
## Description

adds a solution for first half of #1865 , mostly an idea up for discussion but usable already

in some languages the speaker gender matters when generating dialog, this PR adds a new property to the base skill class "grammatical_gender"

gender can be configured and does not necessarily need to match the voice, this means it is inclusive and also allows male sounding voices that identify as female or the other way around

if provided in config, the selected gender will be used, else it will be extrapolated from the TTS config if a gender parameter exists there

this PR assumes skills that want to account for gender will have a dialog structure like:

- file_name.dialog <- default, gender neutral
- file_name.gender.dialog <- if gender is configured and file exists use this one instead

## How to test

- in a skill check self.grammatical_gender
- create my_dialog.female.dialog, my_dialog.male.dialog, my_dialog.dialog
- check that different dialog files are used according to gender

responsive voice TTS has a gender parameter if you want to test TTS gender, for "global" gender change your config and add

`"grammatical_gender": "male"`

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -474,10 +474,10 @@ class MycroftSkill(object):
         """
         gender = self.config.get("grammatical_gender")
         if not gender:
-            tts_config = self.config["tts"]
+            tts_config = self.config.get("tts", {})
             gender = tts_config.get("gender")
             if not gender:
-                tts_module = tts_config["module"]
+                tts_module = tts_config.get("module", "mimic")
                 gender = tts_config.get(tts_module, {}).get("gender")
         return gender
 
@@ -1098,8 +1098,8 @@ class MycroftSkill(object):
         """
         data = data or {}
         if self.grammatical_gender is not None:
-            if resolve_resource_file(key + "." + self.grammatical_gender +
-                                     ".dialog"):
+            gender_file = key + "." + self.grammatical_gender + ".dialog"
+            if resolve_resource_file(gender_file):
                 key = key + "." + self.grammatical_gender
         self.speak(self.dialog_renderer.render(key, data),
                    expect_response, wait)
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -462,6 +462,25 @@ class MycroftSkill(object):
             line = on_fail_fn(response)
             self.speak(line, expect_response=True)
 
+    @property
+    def grammatical_gender(self):
+        """
+        mycroft's gender
+
+        if not configured extrapolate from TTS config
+
+        Returns:
+              string:  'male', 'female' or None
+        """
+        gender = self.config.get("grammatical_gender")
+        if not gender:
+            tts_config = self.config["tts"]
+            gender = tts_config.get("gender")
+            if not gender:
+                tts_module = tts_config["module"]
+                gender = tts_config.get(tts_module, {}).get("gender")
+        return gender
+
     def ask_yesno(self, prompt, data=None):
         """ Read prompt and wait for a yes/no answer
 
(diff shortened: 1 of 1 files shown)


forslund:
I think this is a really good idea. There will be some challenges with the translation import/export for the dialog files (currently the files and lines in the translation need to match one to one with the one in the english translation)

A possible way to do this in the current state of the translate system is to encode male/female words in the dialogs themselves:

*dialog/pt-br/thank_you.dialog*:
`%male:obrigada%female:obrigado%netural:something%`

or something similar.

JarbasAl:
that was my first idea,however sometimes translation is complitcated, in portuguese pronouns and adjectives also need translation! and these won't necessarily be sequential words 

i also like the idea that personality could change, no need for binary gender, you can tell your mycroft that it's gender is dinosaur and replace all dialogs with roars

forslund:
in file could do all that. the blocks will be longer but you could add a whole sentence in each block (it'll be a bit clunky though) and the options could be made to take any number of weird genders such as dinosaur.

forslund:
One way to get your way working with the translation system could be to duplicate the dialog files from the something.dialog to *something.dialog*, *something.female.dialog*, *something.male.dialog* and *something.netural.dialog* (and *something.dinosaur.dialog* of course ;)

I'm not sure how much work it'll be to do a system like this on the import side of things though.

JarbasAl:
as a side note, sometimes its very hard to get a gender neutral formulation

for the thank you example in portuguese, to say it in a genderless way you are a bit screwed, the closest way is probably to say something like "gratitude" and avoid references to the speaker completely, it will sound strange.

this will happen for nearly every language where words themselves are gendered

adocampo:
Now I'm translating the skills into spanish, I faced the same difficulties trying to do some magic and force mycroft to speak neutrally and putting as many options in both male and female for the speaker to say.

In spanish, like other romance language, nouns, adjectives and pronouns need to be gendered to sound natural, so perhaps splitting the files into gender is the best idea.

Anyway, I see we'll need some modifications to let Mycroft now if the user is a man or a woman, and if Mycroft itself has a male or female voice and speak accordingly. I guess the easy way is to put somewhere in settings that information, but it would be awesome if Mycroft can recognize the speaker's gender thanks to speech analysis. I saw this project https://github.com/primaryobjects/voice-gender but I don't know what complications or latency would add to Mycroft's answers

JarbasAl:
@adocampo before getting that in the pipeline we need really good accuracy, you don't want it only getting the gender right sometimes, and as @KathyReid  would say, we must be considerate with people who identify with a different gender or just have deeper voices / higher pitches etc

but i do like that idea

zeehio:
Hi all, long time no see!

It is difficult to me to know if `thankyou.female.dialog` refers to a female user or to a female device. If this approach is used I would suggest `thankyou.female_device.dialog`. It's longer but less ambiguous.

I guess there are lots of contextual features to consider besides device and user gender. I can think of "formality level" (e.g. for Spanish: "usted", "vos", "tú"), "device gender" , "user gender", "number" (e.g: English, Spanish: "singular", "plural"), countability (e.g English: countable/uncountable nouns)...

Eventually we will need a way to handle that if we want natural conversation.

I have for instance a way of identifying the Mycroft user when he/she uses the telegram skill to communicate with Mycroft (my wife and I right now). There the user id is known and therefore the gender and everything could be known as well and passed on to the core if an API existed. Similarly if there was a Mycroft Android app where commands were given through the smartphone the user could be identified as well.

Side note: Apparently my 8 months old daughter is amazed by Mycroft. When I say "where is daddy?" she doesn't give a ****. If I say "hey Mycroft" she looks directly to it. 🙈

PR #1876, opened 2018-11-17
Adding hability to have an independent log per each skill

augustnmonteiro:
## Description
Adding a tool for the development of skills, I had this problem a couple of times, where I was developing a skill and it was hard to find the logs of that skills, this is configuration flag to enable logging by each skill so it makes the development of skill easier

## How to test
Enable the flag run the code and check if the files were created.

### DIFF
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -208,6 +208,8 @@
   // Messagebus types that will NOT be output to logs
   "ignore_logs": ["enclosure.mouth.viseme", "enclosure.mouth.display"],
 
+  "log_per_skill": true,
+
   // Settings related to remote sessions
   // Overrride: none
   "session": {
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -138,7 +138,8 @@ def load_skill(skill_descriptor, bus, skill_id, BLACKLISTED_SKILLS=None):
                 skill._register_decorated()
                 skill.initialize()
             except Exception as e:
-                # If an exception occurs, make sure to clean up the skill
+                log = LOG.create_logger(name)
+                log.debug(e)
                 skill.default_shutdown()
                 raise e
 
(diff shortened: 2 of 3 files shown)


KathyReid:
Thanks so much @augustnmonteiro! Obrigada ;-) 

One minor comment - can I suggest that the name of the skill log is 

`skill-NAME.log` instead of `skill-NAME`

The reason for this is that in the future we want to implement `logrotate` - and the easiest way to implement `logrotate` is to apply it to all log files - `*.log`. If the Skill logs are named with the `.log` suffix then they will be logrotated too.

forslund:
Hey Augusto!

The big issue with this will be it breaking the cli, since it reads the `skills.log` and `voice.log`. so replacing the skills.log with a variety of files might end up more harmful than useful.

augustnmonteiro:
@KathyReid that makes totally sense!

@forslund the normal skill log will have everything, this is more as a skill development tool, let's say you are developing a skill, you only want see the log inside the skill or if there was an error during the loading or during the execution, everything else isn't a resourceful information.

forslund:
Ok so the main skills log will also have a copy of all logs from the skill in addition to the one in the specific file. Then that's no problem.

We should really document all the ways there is to isolate a skill. The :find command and the :keep command are also made for this :) But I can definitely see the use case for this.

augustnmonteiro:
I see, the problem I see on the :keep is that it show the core logs and not only the skills one, another thing is that with the keep the other skills are unloaded so if there is a conflict between an user skill vocab and a core skill vocab the user won't know

forslund:
Agreed, they have a bit different use cases.

Can you fix the pep-8 issue?

augustnmonteiro:
Yep, I will that at night, fix the pep8 and use a constant on the directory instead of a string : D

forslund:
Also I've started seeing this error message:

forslund:
Also noticed that the timestamps seem to be missing in the skill log files

forslund:
Since there's been no work on this branch for quite some time I'm going to close this PR now.

PR #1879, opened 2018-11-18
Issue-1877 - fix ordinal followed by one

### LINKED ISSUE #1877: extract_number('the third one') returns 1
When a user is asked to select from a list of options eg "Would you like apples, bananas or oranges?", `extract_number` can be used to parse an ordinal response eg 'second' or 'third'. However if the user says 'the third one' this returns `1` instead of `3`.

## To Reproduce
`extract_number('the third one', ordinals=True)`

## Cause
parse_en.py correctly picks up `third` however on [line 268](https://github.com/MycroftAI/mycroft-core/blob/d1298744f1310ad7dcfe3b5c7326826d3d38a164/mycroft/util/lang/parse_en.py#L268) the current `val` is written to `prev_val` such that `one` overwrites `third`.

## Possible solution
Replacing [line 268](https://github.com/MycroftAI/mycroft-core/blob/d1298744f1310ad7dcfe3b5c7326826d3d38a164/mycroft/util/lang/parse_en.py#L268) with the following works:
```
if prev_val and val is 1 and ordinals:
    val = prev_val
else:
    prev_val = val
```

This passes the existing `test_parse` unit tests, and some additional assertions:
```
self.assertEqual(extract_number("the second one", ordinals=True), 2.0)
self.assertEqual(extract_number("the fourth one", ordinals=True), 4.0)
self.assertEqual(extract_number("the thirty sixth one", ordinals=True), 36.0)
```

Is there a cleaner way to achieve this, or other cases I'm not considering?

krisgesling:
## Description
Fixes #1877 - `extract_number('the third one', ordinals=True)` returns `1` instead of `3`

## How to test
Existing unit tests passing, and added new assertions.
From virtual environment:
`python -m test.unittests.util.test_parse`

CLA [Yes]

### DIFF
--- a/mycroft/util/lang/parse_en.py
+++ b/mycroft/util/lang/parse_en.py
@@ -138,6 +138,8 @@ def extractnumber_en(text, short_scale=True, ordinals=False):
         "million": 1000000,
         'millions': 1000000}
 
+    string_num_ordinal_en = {}
+
     for num in NUM_STRING_EN:
         num_string = NUM_STRING_EN[num]
         string_num_en[num_string] = num
@@ -147,10 +149,12 @@ def extractnumber_en(text, short_scale=True, ordinals=False):
         if short_scale:
             for num in SHORT_ORDINAL_STRING_EN:
                 num_string = SHORT_ORDINAL_STRING_EN[num]
+                string_num_ordinal_en[num_string] = num
                 string_num_en[num_string] = num
         else:
             for num in LONG_ORDINAL_STRING_EN:
                 num_string = LONG_ORDINAL_STRING_EN[num]
+                string_num_ordinal_en[num_string] = num
                 string_num_en[num_string] = num
 
     # negate next number (-2 = 0 - 2)
@@ -229,6 +233,11 @@ def extractnumber_en(text, short_scale=True, ordinals=False):
         if word in string_num_en:
             val = string_num_en[word]
 
+        # is the prev word an ordinal number and current word is one?
+        # second one, third one
+        if ordinals and prev_word in string_num_ordinal_en and val is 1:
+            val = prev_val
+
         # is the prev word a number and should we sum it?
         # twenty two, fifty six
         if prev_word in sums and word in string_num_en:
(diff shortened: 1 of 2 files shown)


JarbasAl:
i think in that case extracting all numbers makes more sense, i have a PR #1867  for that (which this change will break)

krisgesling:
Hi Jarbas, thanks for pointing out your PR. I'll have a look at what you've done and revise.

I will also make it more explicit eg:
```
# is the prev word an ordinal number and current word is 'one'?
# second one, third one
if word=='one' and prev_word in SHORT_ORDINAL_STRING_EN.values():
    val = prev_val
```

JarbasAl:
PR #1880 for pronoucing ordinal numbers may be useful [here](https://github.com/MycroftAI/mycroft-core/pull/1879/files#diff-3dfdc2a85f591a9c2217a2e90730805bR139), 

[line 76 in my PR](https://github.com/MycroftAI/mycroft-core/pull/1880/files#diff-5053bfe4c4cadfcd9a863d70c708477cR76) has a ORDINALS_EN you can import from mycroft.util.lang.format_en

forslund:
@JarbasAI @krisgesling what do you consider being the correct output here?

As jarbas commented using `extract_numbers("the third one")` previously returned `[3, 1]` but after applying this PR it returns `[3]`.

Personally I would argue this (`[3]`) is the correct output since "the third one" is basically the same as saying "the third object".

Thoughts or comments on this?

krisgesling:
Hi @forslund, this was my thinking also. 

I have been letting it settle in my mind, trying to come up with scenario's where an ordinal might be followed by `"one"`. Say if: 

1. a user is reading out a list numbers that includes fractions eg `extract_numbers("seven, one third, one")`, however in this instance the `ordinals` argument should not be set to `True`. Thus it still outputs correctly `[7, 1/3, 1]'.
2.  a user is prompted to select an option and how many they would like, they would likely respond "one [qty] of the third [object]" outputting `[1,3]`. This might get expressed as "the third, one" which would output just `[3]`, however it's a very unlikely case, and even if it did output `[3, 1]` there's no way to know if the user meant "1 of the 3rd object" or, "3 of the 1st object", so shouldn't be used.

I'm yet to come up with a realistic example where `ordinals=True` and the intention isn't "the third object".

forslund:
Ok, I've rebased it and changed updated the failing test @JarbasAI warned about. Merging now.

Thanks for this addition!

PR #1881, opened 2018-11-18
Catalan translation added

kayron8:
Catalan translation added


forslund:
Thanks! I'm going to look into what's the causes the date test to fail.

forslund:
Ok, I've checked it over and the problem is that there's no catalan function for `nice_time()` (required by `nice_date_time()`.

Since the date_time.json and date_time_test.json isn't translated it looks like you can simply delete those files to stop the test from trying to run against those languages.

`git rm mycroft/res/text/ca-es/date_time*`

and then commit and push to the branch.

forslund:
Gràcies! Looks good now! Merging

kayron8:
De res (your welcome) :) Thank you for taking time to locate the problem and how to fix it, it's very appreciated. Sorry for the inconvenience, I'm still a newbie on GitHub :S

PR #1884, opened 2018-11-20
fix converse

JarbasAl:
## Description

during converse request if a skill failed to load a KeyError would happen silently, this would cause latency (5 seconds until converse timeout) and converse method general failure

to reproduce:
- create a bad skill (needs to be loaded first so its checked first in the converse loop)
- trigger hello world (to activate converse)
- trigger hello world again and check latency and lack of converse.response bus message

## How to test
check everything is working as supposed and that you can't replicate behaviour with PR merged

### DIFF
--- a/mycroft/skills/intent_service.py
+++ b/mycroft/skills/intent_service.py
@@ -227,7 +227,8 @@ class IntentService(object):
 
     def handle_converse_error(self, message):
         skill_id = message.data["skill_id"]
-        self.remove_active_skill(skill_id)
+        if message.data["error"] == "skill id does not exist":
+            self.remove_active_skill(skill_id)
         if skill_id == self.converse_skill_id:
             self.converse_result = False
             self.waiting_for_converse = False
--- a/mycroft/skills/intent_service.py
+++ b/mycroft/skills/intent_service.py
@@ -223,6 +223,7 @@ class IntentService(object):
             t = time.time() - start_time
             time.sleep(0.1)
         self.waiting_for_converse = False
+        self.converse_skill_id = ""
         return self.converse_result
 
     def handle_converse_error(self, message):
--- a/mycroft/skills/intent_service.py
+++ b/mycroft/skills/intent_service.py
@@ -174,6 +174,7 @@ class IntentService(object):
         self.bus.on('clear_context', self.handle_clear_context)
         # Converse method
         self.bus.on('skill.converse.response', self.handle_converse_response)
+        self.bus.on('skill.converse.error', self.handle_converse_error)
         self.bus.on('mycroft.speech.recognition.unknown', self.reset_converse)
         self.bus.on('mycroft.skills.loaded', self.update_skill_name_dict)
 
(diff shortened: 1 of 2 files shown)


forslund:
I believe the ['instance'] check is good to have if a reload of the skill fails it shouldn't trigger the skills converse method. But maybe using the .get() to ensure that no key error occurs is the way to go?

JarbasAl:
the check is still there, it was actually a duplicate check that i removed

a couple lines bellow there is a try: except: block handling this

                try:
                    instance = self.loaded_skills[skill]["instance"]
                except BaseException:
                    LOG.error("converse requested but skill not loaded")
                    self.bus.emit(message.reply("skill.converse.response", {
                        "skill_id": 0, "result": False}))
                    return

forslund:
Yeah, but we don't want to hit the try-except and generate an error message. If the instance is None the entire section should be skipped for the skill.

JarbasAl:
i would maybe even add an extra LOG.error after the for loop, to inform that converse for a non existing skill was requested, that way we can know exactly what went wrong (unloaded skill, no skill, bad converse method)

forslund:
I can see that the current behavior is incorrect but it shouldn't hang for 5 seconds since it'll send the response after the for loop right? And possibly it's that line that's wrong returning skill id 0?

Getting additional error messages each time a request is posted until the skill has left the queue is in my opinion not right and may just clutter up the log. Since a load error message is already posted when the skill is loaded this doesn't add any information.

JarbasAl:
there was no exception throw if "instance" was missing in the if check, it would fail silently and never send the .response message, causing the 5 second hang

i agree it is a good idea to make the intent parser remove a failed converse from the active skill list when this happens

skill id has no impact in the outcome, returning 0 is basically an artifact/error code, my original implementation used a simple counter for skill_id instead of being derived from the skill path

JarbasAl:
what if instead of logging all that stuff we included the cause in the .response message?

it would improve converse interaction by messagebus (manual converse trigger)

forslund:
That's a pretty good idea.

Also seems a bit ugly that we delete the dictionary value `"instance"` instead of keeping it and just setting it to `None`

JarbasAl:
it could also be a new message, "skill.converse.error" , but that's kinda optional since checking for skill_id == 0, or even isinstance(skill_id, int), does the same

"skill.converse.error" feels right, but makes you need to set up a new message listener to also abort the waiting, seems easier to just check the skill id

forslund:
Can't we just add an error field to the .response message?

forslund:
I think it's easier to just use the .response message with an error field.

forslund:
This PR's scope just got a lot bigger :)

JarbasAl:
some more changes:
- declaring variables in __init__
- renaming variables to be more representative of what they do
- check skill_id on message handlers
- on converse error remove skill from active list (is this a good idea? what if in the next converse call the error does not happen and the skill had logic to recover?)

rationale behind this new changes is allowing converse to be used simultaneously in several skills by messagebus interaction

i find converse very useful for passive skills that peak at an utterance before the intent parsing stage and might choose to consume the utterance or not

i can remove the converse.error messages if you want, i was just thinking i might want to emit those messages for some messagebus vodoo

forslund:
Can we also remove the check for instance using an exception and replace it with an if statement?

forslund:
Removing from the active list might be a bad idea as you say. The converse can hit during a skill update and skills can be designed to recover their state...

JarbasAl:
i will make it remove only if skill does not exist

if its unloaded it might be in the middle of an update, if it throws an exception it might recover on the next round, but if it does not exist it should be removed

forslund:
Looks great, going to do some tests before pulling

forslund:
Tested works excellent. Merging

PR #1885, opened 2018-11-20
Feature/msm 0.6.2

forslund:
## Description
Update to the latest msm and start making use of the internal handling of the `skills_data.json`

## How to test
- check that `~/.mycroft/skills_data.json` file gets upgraded to the current version (the version field should be 1)
- ensure that the skills_data upload works ( in the cli run `:find skillJson` will find the post requests, make sure these indicate success, status 200)

CLA [Yes]

### DIFF
--- a/mycroft/skills/skill_manager.py
+++ b/mycroft/skills/skill_manager.py
@@ -175,6 +175,11 @@ class SkillManager(Thread):
     def schedule_now(self, message=None):
         self.next_download = time.time() - 1
 
+    @staticmethod
+    @property
+    def manifest_upload_allowed(self):
+        return Configuration.get()['skills'].get('upload_skill_manifest')
+
     @property
     def installed_skills_file(self):
         venv = dirname(dirname(sys.executable))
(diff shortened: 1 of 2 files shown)


forslund:
Noticed that on updating an old skills.json file the installed field will we "true" and not zeroed. Tiny update needed.

mycroft/skills/skill_manager.py
```diff
@@ -257,20 +247,23 @@ def install_or_update(skill):
                         nonlocal default_skill_errored
                         default_skill_errored = True
                     raise
-                installed_skills.add(skill.name)
+            installed_skills.add(skill.name)
 
         try:
-            self.msm.apply(install_or_update, self.msm.list())
+            with self.msm.lock, self.thread_lock:
+                self.msm.skills_data = self.msm.load_skills_data()
```
MatthewScholefield: Is this and the still needed now that it performs `sync_skills_data` in the constructor? Additionally, would it cause a deadlock? (I haven't checked to see if nested locks will deadlock with filelock)

PR #1888, opened 2018-11-26
add changes to work with gui

AIIX:
## Description: 
Changes made to feature/gui branch to work with mycroft-gui:
-- This shouldn't really be merged except for testing and needs way better implementation

#### In skills/core.py 
- find resource did not have the correct path to qml files required to be sent, temp fix was to create find_qml_resource to run os.walk on the correct path where the qml UI is expected to exist.
#### In client/enclosure/base.py:
- As per the protocol specs list of active skills is required to be sent before mycroft.session.set message call, the active skills list needs to implement correct position of active skills in the list and if existing avoid duplicates, refer to autotest in mycroft-gui https://github.com/MycroftAI/mycroft-gui/blob/master/autotests/servertest.cpp#L174
- gui_url was changed to gui_urls to provide a list of active pages in skill, required to be a list of pages even if there is just a single page. also most likely need to avoid adding duplicates of an existing page in the list

### DIFF
--- a/mycroft/client/enclosure/base.py
+++ b/mycroft/client/enclosure/base.py
@@ -242,19 +242,38 @@ class GUIConnection(object):
         self.callback_disconnect(self.id)
 
     def set(self, namespace, name, value):
+        skillIdList = []
+        # hack to send active skills list before
+        if namespace in skillIdList:
+            print("skill already in skillIdList")
+        else:
+            skillIdList.append({'skill_id': str(namespace)})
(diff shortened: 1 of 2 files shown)


notmart:
@penrods this is mostly a prototype (as we don't know yet the right place in the mycroft architecture for this) but it shows the need of server side bookeeping and client sync of both running skills and running QML files

penrods:
I believe the latest feature/gui branch supersedes this example.  If so, we can close it.

PR #1889, opened 2018-11-28
Run emitter using threadpool

forslund:
## Description
This moves the thread pool from the websocket client into the eventemitter allowing each registered function to run in a separate thread and not just each event.

This speeds up cases where there is a one to many call (multiple functions listening for the same message) such as the common play framework and the upcoming common query framework.

## How to test
Make sure Mycroft runs normally with this change

CLA [ Yes ]

### DIFF
--- a/mycroft/messagebus/client/threaded_event_emitter.py
+++ b/mycroft/messagebus/client/threaded_event_emitter.py
@@ -1,0 +1,43 @@
+from pyee import EventEmitter
+from multiprocessing.pool import ThreadPool
+from collections import defaultdict
+
+
+class ThreadedEventEmitter(EventEmitter):
+    """ Event Emitter using the threadpool to run event functions in
+        separate threads using a threadpool.
+    """
+    def __init__(self, threads=10):
+        super().__init__()
+        self.pool = ThreadPool(threads)
+        self.wrappers = defaultdict(list)
+
+    def on(self, event, f=None):
+        """ Wrap on with a threaded launcher. """
+        def wrapped(*args, **kwargs):
+            return self.pool.apply_async(f, args, kwargs)
+
+        w = super().on(event, wrapped)
+        # Store mapping from function to wrapped function
+        self.wrappers[event].append((f, wrapped))
+        return w
+
+    def once(self, event, f=None):
+        """ Wrap once with a threaded launcher. """
+        def wrapped(*args, **kwargs):
+            return self.pool.apply_async(f, args, kwargs)
+
+        wrapped = super().once(event, wrapped)
+        self.wrappers[event].append((f, wrapped))
+        return wrapped
+
+    def remove_listener(self, event_name, func):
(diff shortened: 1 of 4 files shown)


penrods:
Ping to reviewers!  Let's get this merged, it's a pretty big deal.

mycroft/messagebus/client/threaded_event_emitter.py
```diff
@@ -0,0 +1,19 @@
+from pyee import EventEmitter
+from multiprocessing.pool import ThreadPool
+
+
+class ThreadedEventEmitter(EventEmitter):
+    """ Event Emitter using the threadpool to run event functions in
+        separate threads using a threadpool.
+    """
+    def __init__(self, threads=None):
```
MatthewScholefield: We can just use `threads=10`

mycroft/messagebus/client/threaded_event_emitter.py
```diff
@@ -0,0 +1,19 @@
...
+        super().__init__()
+        self.pool = ThreadPool(threads or 10)
+
+    def emit(self, event, *args, **kwargs):
+        """ Override the normal emit launching the function using
+            threadpool.
+        """
+        for f in self._events[event]:
```
MatthewScholefield: Will this mess up once events? What if we call `EventEmitter.emit(self, *args, **kwargs)` via the threadpool?

forslund: I do not think that will get around the issue with the event emitter serializing the actual calls for messages with multiple registered handlers.

This can possibly be done in a simpler and cleaner way by overriding on and once instead using a wrapper to launch the handler in a thread...

JarbasAl: this messed the once signals, forgot i had merged this to test and it broke get_response for me, any speak message now starts listening

mycroft/messagebus/client/threaded_event_emitter.py
forslund: Fixed

mycroft/messagebus/client/threaded_event_emitter.py
forslund: I've updated the commit with a wrapper approach instead of overriding emit. This is working well for me so far.

forslund:
Well that certainly looks wrong. I'll look into it...

forslund:
I've been running the updated pyee locally so I didn't see the issue. the once mechanic seems to have been revamped a bit. We can either (like above) upgrade pyee or I can re-implement the once function so it'll register the function correctly.

forslund:
For reference: [forslund:feature/threaded-emitter-2](https://github.com/forslund/mycroft-core/tree/feature/threaded-emitter-2) is a workaround of the issue in pyee instead of the upgrade.

If the pyee upgrade feels dodgy this branch could be usable instead.

PR #1914, opened 2018-12-22
bugfix/issue-1913 fixed bug with new flag

Batcastle:
moved the `set -Ee` to run when a flag _IS NOT_ thrown on lines 77-80
Also improved print out in function show_help() to take a little less time and take up a little less space by using only one `echo` command, should work the exact same otherwise.

## Description
issue-1913, added flags so that `dev_setup.sh` will not exit on an error. Flags only take effect if explicitly called.

This flag is meant to be used it `apt` or some other package manager is throwing an error but otherwise is working perfectly fine.

## How to test
run `dev_setup.sh -n` and compare with normal usage of the file.

Yes

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -38,24 +38,24 @@
 # </endRant>
 ######################################################
 
-# exit on any error
-set -Ee
+
 
 cd $(dirname $0)
 TOP=$( pwd -L )
 
 function show_help() {
-    echo "dev_setup.sh: Mycroft development environment setup"
-    echo "Usage: dev_setup.sh [options]"
-    echo
-    echo "Options:"
-    echo "    -r, --allow-root  Allow to be run as root (e.g. sudo)"
-    echo "    -fm               Force mimic build"
-    echo "    -h, --help        Show this message"
-    echo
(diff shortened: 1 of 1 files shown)


dev_setup.sh
```diff
@@ -74,6 +74,10 @@ for var in "$@" ; do
     if [[ ${var} == "-fm" ]] ; then
         opt_forcemimicbuild=true
     fi
+    fi [[ ${var} != "-n" ]] || [[ ${var} != "--no-error" ]] ; then
```
forslund: This line should probably start with `if` and not `fi` :)

dev_setup.sh
```diff
@@ -74,6 +74,10 @@ for var in "$@" ; do
...
+    	# exit on any error
+	set -Ee
```
forslund: This is mainly me thinking loud: Would it make sense to have `set -Ee` at the top of the file and then on --no-error `set +Ee` to turn off the errors.

This might be good to easily detect errors in the startup or argument parsing code.

Batcastle:
@forslund I made the changes you suggested. Thank you for catching my typo!

forslund:
Tested it and it looks good. Merging

PR #1920, opened 2018-12-29
Add basic barge-in functionality

penrods:
The ability to "barge-in" has been lacking from Mycroft Core.  The Mark 1
microphone was unable to support this due to physical limitations, but the
Mark II and other implementations with more advanced mic tech which can
hear over themselves are able to continuously listen.

To enable this while retaining backwards compatibility with simpler mic.py
systems, there are now two mycroft.conf values:

{
   "listener": {
        "mute_during_output" : true,
        "duck_while_listening" : 0.3
   }
}

The above values are defaults, and implementers will likely override them
using the /etc/mycroft/mycroft.conf file when appropriate.

The duck_while_listening setting is currently handled in Mycroft's
skill-volume.  The mute_during_output is handled within mycroft-core itself.

## How to test
Add ```"mute_during_ouptut" : false``` to /etc/mycroft/mycroft.conf.  System should allow listening even while speaking.  

### DIFF
--- a/mycroft/client/speech/__main__.py
+++ b/mycroft/client/speech/__main__.py
@@ -29,6 +29,7 @@ from mycroft.util.log import LOG
 bus = None  # Mycroft messagebus connection
 lock = Lock()
 loop = None
+config = None
 
 
 def handle_record_begin():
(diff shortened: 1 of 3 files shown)


forslund:
Seems to work just fine. Merging.

PR #1924, opened 2018-12-30
Improve setup and start-mycroft scripts

penrods:
Improved the dev_setup.sh in several ways:
* Add 'wizard' that walks new users through setup
  - Option to put the mycroft-core/bin folder in the path
  - Option to change to 'master' instead of 'dev' branch.  This is
    usually desirable for users who only run Mycroft, not developers.
  - Option to auto-update when running Mycroft
  - Add soft link <mycroft-core>/skills to the /opt/mycroft/skills
    directory

Improve the start-mycroft.sh script
* Unify some duplicated code under init-once() function
* Add support for "auto-update", which attempts a git pull at startup
* Add require-process() function which doesn't restart already
  running services
* Add require-process for messagebus and skills services when running
  the CLI
* Tweak the --help message

This significantly changes the setup and normal operation

## How to test
Run dev_setup.sh

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-
+#
 # Copyright 2017 Mycroft AI Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


dev_setup.sh
```diff
@@ -87,23 +63,160 @@ done
...
+        read -N1 -s key
+        case $key in
+        [Yy])
+            echo -e "${HIGHLIGHT} $key - using 'master' branch ${RESET}"
+            branch=master
+            break
+            ;;
+        [Nn])
+            echo -e "${HIGHLIGHT} $key - using 'dev' branch ${RESET}"
+            branch=dev
```
forslund: This is not a good idea, if running on a non-dev branch it will try to switch back to dev when running dev_setup. Option to Skip?

dev_setup.sh
```diff
@@ -87,23 +63,160 @@ done
...
+        read -N1 -s key
+        case $key in
+        [Yy])
+            echo -e "${HIGHLIGHT} $key - Adding mycroft-core/bin to your PATH ${RESET}"
+            echo '' >> ~/.profile
+            echo '# set path so it includes Mycroft utilities' >> ~/.profile
+            echo 'if [ -d "'${TOP}'/bin" ] ; then' >> ~/.profile
+            echo '    PATH="'${TOP}'/bin:$PATH"' >> ~/.profile
+            echo 'fi' >> ~/.profile
+            echo -e "Type ${CYAN}mycroft-help${RESET} to see available commands."
```
forslund: Should we prepare a `.mycroft_profile` file that's sourced in the .profile instead? Would probably be better in case of future additions and if a new mycroft installation is done the `.mycroft_profile` file can just be overwritten.

I don't think we should place the mycroft-core/bin folder as the first lookup folder.

dev_setup.sh
```diff
@@ -87,23 +63,160 @@ done
...
+    echo "first run of dev_setup we will ask you a few questions to help setup"
+    echo "your environment."
+    echo
+    sleep 0.5
+    echo "Do you want to run on 'master' or on the 'dev' branch?  Unless you are"
+    echo "a developer modifying mycroft-core itself, you should run on the"
+    echo "'master' branch.  It is updated bi-weekly with a stable release."
+    echo "  Y)es, run on the 'master' branch"
+    echo "  N)o, I want to run against the unstable 'dev' branch"
+    echo -e -n "Choice [${CYAN}Y${RESET}/${CYAN}N${RESET}]: "
```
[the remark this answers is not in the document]
penrods: I switched to using multi line strings.  I'm not sure if I really like that style, but maybe I'm  just biased because of the highlighting limitations of nano vs your superior vim highlighting ;)

dev_setup.sh
penrods: I modified it to only change branches if you pick 'master'.  I also tweaked the wording to use "stable" and "unstable".

dev_setup.sh
penrods: Good idea.  I implemented this, along with a comment in the .mycroft_profile indicating that the file should not be customized as it might be replaced.

dev_setup.sh
```diff
@@ -87,23 +63,160 @@ done
...
+    while true; do
+        read -N1 -s key
+        case $key in
+        [Yy])
+            echo -e "${HIGHLIGHT} $key - Adding mycroft-core/bin to your PATH ${RESET}"
+            echo '' >> ~/.profile
+            echo '# set path so it includes Mycroft utilities' >> ~/.profile
+            echo 'if [ -d "'${TOP}'/bin" ] ; then' >> ~/.profile
+            echo '    PATH="'${TOP}'/bin:$PATH"' >> ~/.profile
```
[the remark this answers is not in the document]
penrods: Looks like a bogus warning -- the test is misunderstanding what is going on in this admittedly complex use case.

Issue #1929, opened 2019-01-03
ModuleNotFound :requests in 18.8.9


KathyReid:
## Description of error

When attempting to run `./start-mycroft *` after doing a Git-based install on Linux platforms (docker, Debian and Ubuntu reported), the error; 

`ModuleNotFound :requests` is shown. 

_placeholder_, more diagnostic information required

KathyReid:
Closed by #1930

PR #1930, opened 2019-01-03
dev_setup: Exit if pip installation fails.

forslund:
## Description
Running the get-pip-script will return a status Ok even if the command fails due to network error causing an install without pip. The system pip will then be used to install packages resulting in a faulty venv.

This adds a simple check that pip exists in the virtualenv after the setup and if not fails the build. This could be a possible cause for #1929 

CLA [ Yes ]

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -287,6 +287,8 @@ function install_venv() {
     # about this version.  Update whenever a new version is released and
     # verified functional.
     curl https://bootstrap.pypa.io/3.3/get-pip.py | "${VIRTUALENV_ROOT}/bin/python" - 'pip==18.0.0'
+    # Function status depending on if pip exists
+    [ -x "${VIRTUALENV_ROOT}/bin/pip" ]
 }
 
 install_deps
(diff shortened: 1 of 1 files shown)


KathyReid:
This sounds *exactly* like the issues we've seen over the last two days with `requests` not being loaded.

dev_setup.sh
```diff
@@ -287,6 +287,7 @@ function install_venv() {
     # about this version.  Update whenever a new version is released and
     # verified functional.
     curl https://bootstrap.pypa.io/3.3/get-pip.py | "${VIRTUALENV_ROOT}/bin/python" - 'pip==18.0.0'
+    ls "${VIRTUALENV_ROOT}/bin/pip" > /dev/null 2>&1
```
MatthewScholefield: Would it be clearer to use the bash file flag?:
```bash
[ -f "$VIRTUALENV_ROOT/bin/pip" ]
```

forslund: Indeed it would. Didn't realize that could be used outside of if statements. Good to know. will fix

forslund: Update has been pushed.

PR #1932, opened 2019-01-07
Upgrade pyyaml to address CVE-2017-18342

KathyReid:
CVE-2017-18342 
https://nvd.nist.gov/vuln/detail/CVE-2017-18342
high severity
Vulnerable versions: < 4.2b1
Patched version: 4.2b1
In PyYAML before 4.1, the yaml.load() API could execute arbitrary code. In other words, yaml.safe_load is not used.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ SpeechRecognition==3.8.1
 tornado==4.5.3
 websocket-client==0.32.0
 requests-futures==0.9.5
-pyyaml==3.13
+pyyaml==4.2b1
 pyalsaaudio==0.8.2
 xmlrunner==1.7.7
 pyserial==3.0


KathyReid:
Not ready to merge this yet; need to do a test build to test it

forslund:
Not even sure if pyyaml is needed at all:

```
$ pipdeptree --reverse --packages pyyaml
PyYAML==3.13
```

So no module depends on it and we don't use yaml in mycroft-core afaik.

Maybe we can try to remove it completely?

KathyReid:
Possibly, but it's likely that it was added to `requirements.txt` for a reason. Let me see when it was first introduced into mainline.

forslund:
Seems to be for an issue with arch/manjaro?

KathyReid:
It was already listed in March 2016; 
https://github.com/MycroftAI/mycroft-core/blob/545acfd65d89ba63d5846c24ad81f900fae040be/requirements.txt

The Manjaro / Arch issue was different - Arch package was ahead of what we specified in `requirements.txt` which caused the Manjaro / Arch `./dev_setup` to bork. 

I think you are right and this could be removed with no ill effect. Let me close this and do a build with it removed. Removing a dep is better than keeping a dep up to date ;-)

forslund:
I did a build as well and seems to be working fine without it.

PR #1934, opened 2019-01-07
Remove pyyaml

MatthewScholefield:
## Description
This removes pyyaml since it is not used.

## How to test
Make sure build succeeds


forslund:
Talked with Kathy about this and it doesn't seem to be used at all and could maybe be removed?

forslund:
See discussion in https://github.com/MycroftAI/mycroft-core/pull/1932

PR #1935, opened 2019-01-08
Tweaks to start/stop scripts

penrods:
Several minor changes:
* Add mycroft-start and mycroft-stop, which can be used as commands anywhere
  if the user has added this to the PATH.  They are shortcuts to the
  start-mycroft.sh and stop-mycroft.sh scripts.
* Add to the mycroft-help command reference
* Add "reset" parameter for start-mycroft.sh, which forces a service restart.
  If a service is currently running, it will not restart by default anymore.
* Make the 'enclosure' service part of 'all' regardless of platform.  This
  makes sense not that it can handle a remote GUI connection.
* BUG: mycroft-stop would sometimes show odd messages if the skill process
  had active child processes.

### DIFF
--- a/bin/mycroft-help
+++ b/bin/mycroft-help
@@ -25,6 +25,8 @@ echo "Mycroft-specific commands you can use from the Linux command prompt:"
 echo "  mycroft-cli-client       command line client, useful for debugging"
 echo "  mycroft-msm              Mycroft Skills Manager, to manage your Skills"
 echo "  mycroft-msk              Mycroft Skills Kit, create and share Skills"
+echo "  mycroft-start            Launch/restart Mycroft services"
+echo "  mycroft-stop             Stop Mycroft services"
 echo
 echo "Scripting Utilities:"
 echo "  mycroft-speak <phr>      have Mycroft speak a phrase to the user"
--- a/bin/mycroft-start
+++ b/bin/mycroft-start
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 Mycroft AI Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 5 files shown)


forslund:
In the description you refer to "reset" as the command to restart the services. Seems to be referred to as "restart" in the actual code (which I think is better).

Am I right to assume this is just a typo in the PR description?

bin/mycroft-start
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
forslund: Set the execution flag on this file

bin/mycroft-stop
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
forslund: Execution flag for this one as well

bin/mycroft-help
```diff
@@ -25,6 +25,8 @@ echo "Mycroft-specific commands you can use from the Linux command prompt:"
 echo "  mycroft-cli-client       command line client, useful for debugging"
 echo "  mycroft-msm              Mycroft Skills Manager, to manage your Skills"
 echo "  mycroft-msk              Mycroft Skills Kit, create and share Skills"
+echo "  mycroft-start            Launch/restart Mycroft services"
```
forslund: I'm not sure about this help line. from reading it I would expect that entering

`mycroft-start`

would start Mycroft. but I could just be reading to much into it. When launching the command it's pretty obvious what's needed to run it.

bin/mycroft-start
penrods: Whoops, Windows stripped the X bit.  Will do.

bin/mycroft-stop
penrods: Ditto

bin/mycroft-help
penrods: I'm open to better phrasing.  I expected if someone types it they'll figure it out pretty easily.

PR #1936, opened 2019-01-08
Support selecting microphone by name

penrods:
Normally Mycroft will use the default PortAudio input device as the
microphone input for the user.  However in some cases there is reason
to specify a different input.

The "device_index" under the "listener" section in mycroft.conf has
always allowed a user to select the microphone explicitly.  But on
some systems the indices can change from reboot to reboot.  So this
adds the "device_name" setting.  If it exists (and the "device_index"
has not been specified explicitly), a regex match will be run against
the PortAudio device names.

When "device_name" is used, the voice.log will contain a listing of
the devices as they are tested.  This can be used to debug if a
name isn't matching as expected.

EXAMPLES:
/etc/mycroft/mycroft.conf
```json
{
    "listener": {
        "device_name": "aawsrc"
    }
}
```
Would find a match against "aawsrc" or "aawsrcplug".  To force a specific
match, you can use a regex such as "aawsrc$".

/etc/mycroft/mycroft.conf
```json
{
    "listener": {
        "device_index": 2
    }
}
```
The PortAudio device index specified will be used.

Names and indexes for PortAudio are difficult to guess.  Under Linux they are based on the ALSA asound.conf entry names, but might be different in other OSes.  The simplest way to
view them is either enter a value for "device_name" and look at the names
which appear in the log when starting Mycroft, or to run a simple program
such as:
```python
import pyaudio

pa = pyaudio.PyAudio()
for i in range(pa.get_device_count()):
    dev = pa.get_device_info_by_index(i)
    print((i, dev['name'], dev['maxInputChannels']))
```

### DIFF
--- a/mycroft/client/speech/listener.py
+++ b/mycroft/client/speech/listener.py
@@ -16,6 +16,8 @@ import time
 from threading import Thread
 import sys
 import speech_recognition as sr
+import pyaudio
+import re
 from pyee import EventEmitter
 from requests import RequestException, HTTPError
 from requests.exceptions import ConnectionError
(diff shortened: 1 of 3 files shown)


benklop:
It might also be worth noting that the PortAudio names are the same names given to customized PCMs in the /etc/asound.conf file, so you can make a device have any specific name you want using that config file. For that reason, I'm not sure it's worthwhile having this support regex - I would personally want this to select the PCM named "default" if there is one (this is an ALSA convention) and otherwise only match exactly.

penrods:
There was a little discussion on [Mattermost](https://chat.mycroft.ai/community/pl/wiasqzdna3y8mme5zmxfnpiaio) regarding this.  I've seen a system where the device names change from boot to boot, so allowing regex in the name selection is really necessary.

Also discussed adding a feature to the CLI to select the active mic.

forslund:
Also worth noting is that if no device_name is specified the "default" will be used. So this is mainly if the default isn't reliable for whatever reason.

forslund:
@benklop Let me check your case. The code isn't really changed just moved...

benklop:
@forslund my bad, it was actualyl an issue with my configuration.

forslund:
Cool, was getting worried something really strange was going on :)

PR #1944, opened 2019-01-09
feature/audio seek

JarbasAl:
## Description
adds seek functionality to audio service

only implemented in vlc audio backend

as a follow up the playback skill should also be updated

## How to test

use the new methods in audioservice

        audio = AudioService(bus)
        audio.play(track)
        audio.seek(10)  # skip 10 seconds
        audio.seek(-10)  # rewind 10 seconds
        audio.seek_forward(20)  # skip 20 seconds
        audio.seek_backward(5)  # rewind 5 seconds

CLA [affirmative] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

### DIFF
--- a/mycroft/audio/audioservice.py
+++ b/mycroft/audio/audioservice.py
@@ -187,6 +187,8 @@ class AudioService:
         self.bus.on('mycroft.audio.service.next', self._next)
         self.bus.on('mycroft.audio.service.prev', self._prev)
         self.bus.on('mycroft.audio.service.track_info', self._track_info)
+        self.bus.on('mycroft.audio.service.seek_forward', self._seek_forward)
+        self.bus.on('mycroft.audio.service.seek_backward', self._seek_backward)
         self.bus.on('recognizer_loop:audio_output_start', self._lower_volume)
         self.bus.on('recognizer_loop:record_begin', self._lower_volume)
         self.bus.on('recognizer_loop:audio_output_end', self._restore_volume)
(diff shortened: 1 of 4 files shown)


mycroft/skills/audioservice.py
```diff
@@ -121,6 +121,23 @@ def resume(self):
         """ Resume paused playback. """
         self.bus.emit(Message('mycroft.audio.service.resume'))
 
+    def seek(self, seconds=1):
+        """ seek X seconds """
```
forslund: Add the *Arguments:* section to these docstrings

mycroft/audio/audioservice.py
```diff
@@ -187,6 +187,8 @@ def load_services_callback(self):
         self.bus.on('mycroft.audio.service.next', self._next)
         self.bus.on('mycroft.audio.service.prev', self._prev)
         self.bus.on('mycroft.audio.service.track_info', self._track_info)
+        self.bus.on('mycroft.audio.service.seek.forward', self._seek_forward)
+        self.bus.on('mycroft.audio.service.seek.backward', self._seek_backward)
```
forslund: I think `mycroft.audio.service.seek_forward` and `mycroft.audio.service.seek_backward` will fit better with the others.

mycroft/audio/services/__init__.py
```diff
@@ -15,7 +15,7 @@
 from abc import ABCMeta, abstractmethod
 
 
-class AudioBackend():
+class AudioBackend(object):
```
forslund: No need to inherit from object in python3

JarbasAl: isn't that the standard however? we are doing it everywhere else

mycroft/audio/audioservice.py
JarbasAl: agreed, will make changes

mycroft/skills/audioservice.py
JarbasAl: :+1: forgot about that

mycroft/audio/services/__init__.py
forslund: Yeah I've noticed it was still hanging around. Added a PR to remove them.

JarbasAl: see https://github.com/MycroftAI/mycroft-core/pull/1945#issuecomment-453237439

either way i do not oppose this change, when #1945 is merged i'll update this

JarbasAl:
should probably add a check in vlc to ensure time is not negative

and also that it isn't higher than audio length, need to dig into vlc api for this one

forslund:
Tested and works really well. I went ahead and rebased it. Merging now.

Many thanks @JarbasAI !

PR #1945, opened 2019-01-10
Remove inheritance from object

forslund:
## Description
Inheriting from object isn't necessary in python3. Removing it to keep the code cleaner and more consistent.

## How to test
Make sure tests passes and mycroft still functions

CLA [ Yes ]

### DIFF
--- a/mycroft/api/__init__.py
+++ b/mycroft/api/__init__.py
@@ -39,8 +39,8 @@ class InternetDown(RequestException):
     pass
 
 
-class Api(object):
-    """ Generic object to wrap web APIs """
+class Api:
+    """ Generic class to wrap web APIs """
     params_to_etag = {}
     etag_to_response = {}
 
--- a/mycroft/audio-accuracy-test/audio_accuracy_test.py
+++ b/mycroft/audio-accuracy-test/audio_accuracy_test.py
@@ -29,7 +29,7 @@ def to_percent(val):
     return "{0:.2f}".format(100.0 * val) + "%"
 
 
-class FileStream(object):
+class FileStream:
     MIN_S_TO_DEBUG = 5.0
 
     # How long between printing debug info to screen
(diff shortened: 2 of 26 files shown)


mycroft/api/__init__.py
```diff
@@ -39,8 +39,8 @@ class InternetDown(RequestException):
     pass
 
 
-class Api(object):
-    """ Generic object to wrap web APIs """
+class Api():
```
MatthewScholefield: We might as well exclude the extra parentheses here as well (so it becomes `class Api:`).

JarbasAl:
i would like to keep the object in there, i know we moved to python 3 and python 2 support is not a priority, but no need to make it harder if someone wants to port it

i got used to write code compatible with python2 and 3, and while support shouldn't be mandatory i think compatibility is a good habit

JarbasAl:
this is still valid syntax in python 3, it's not "8-year old syntax", i could argue old style classes are 8-year old syntax still supported, but to be fair even the python3 docs use it 

i see it as supporting "bad" developer habits, just like enforcing pep8 is a good habit even if the code was valid otherwise
 
i totally get what your point about this not being a library though, and supporting python2 is not the main point i wanted to make

mycroft/api/__init__.py
forslund: Ah, right you are!

Should be fixed now.

forslund:
I certainly agree that you have a valid point @JarbasAI but I think keeping a "clean" (for some value of clean) code base is more important for this project. I believe this increases readability while no functionality is lost:

- Explicitly Inheriting from object doesn't add any relevant information.
- Classes that actually inherits from something of interest is easier to locate in the code.
- Easier to maintain consistency

penrods:
My $0.02 on this is that we did make the cut over to Python 3.  As I'm working on code, I'm eliminating Python 2-isms, such as using the simpler ```super().__init__()``` instead of ```super(ChildB, self).__init__()```.  It is just more readable.  Unless we are actively forcing ourselves to be 2.x compatible, there is no reason to be kinda-compatible.  A semi-compatible hybrid is just ugly and really less useful to both the 3.x (who have to suffer old syntax) and the 2.x (who pull in something, but can't actually get it to work).

PR #1957, opened 2019-01-17
text_client.py: Don't throw exception on missing conf

alistair23:
## Description
If we are missing the ".mycroft_cli.conf" file we print a message to the
user informing them that we are ignoring the missing file, but then
throw an exception. Remove the exception throwing and instead allow the
user to continue without the file, as we tell them we are doing.

## How to test
Run CLI without ".mycroft_cli.conf" file.

CLA [ Yes ]

### DIFF
--- a/mycroft/client/text/text_client.py
+++ b/mycroft/client/text/text_client.py
@@ -169,7 +169,6 @@ def load_settings():
             show_meter = config["show_meter"]
     except Exception as e:
         LOG.info("Ignoring failed load of settings file")
-        LOG.exception(e)
 
 
 def save_settings():


alistair23:
Ah, the serial console I was using doesn't support all of the fancy formatting so when the exception is printed it would break the display. I tried over an SSH connection and it doesn't exit.

I still think this patch is useful though as we don't really need a stack trace just because a file is missing and even over SSH it ends up being hard to read.

penrods:
I'm fine with killing that extended exception log.

forslund:
Me too.

PR #1958, opened 2019-01-17
dialog: Open template file as utf8 encoded

alistair23:
## Description
To avoid UnicodeDecodeError's when opening files tell Python that we
want to open the file as a utf8 encoded file.

## How to test
Load skills in mycroft

CLA [ Yes ]


forslund:
Thanks, looks good. I was under the impression that it defaulted to utf-8 but it seems to be locale dependent.

Will merge when the cla arrives

forslund:
CLA has been recieved, many thanks for this contribution. Merging.

PR #1961, opened 2019-01-19
Bugfix/speech stop

forslund:
## Description
Improve the TTS stop mechanic

- Make the mimic2 TTS more similar to the other TTS engines (I think this fixes another bug with local caching)
- Make sure pending requests are also stopped
- Make sure that any TTS operation that was running when the stop signal arrived will be cleared

## How to test
A good skill to test with is the [fairytalez skill](https://github.com/andlo/fairytalez-skill) by @andlo 

CLA [ Yes ]

### DIFF
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -55,6 +55,7 @@ def handle_speak(event):
     else:
         ident = 'unknown'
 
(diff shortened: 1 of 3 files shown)


andlo:
I have testet with fairytalez-skill, and it seems to be much better at stoppping than earlier :)

PR #1965, opened 2019-01-21
Add alias to bash profile for easier starting and stopping.

impearker:
## add echo command to allow for easier starting and stopping of mycroft
`echo "alias Mycroft-Start='~/mycroft-core/start-mycroft.sh all'" >> ~/.bashrc`
`echo "alias Mycroft-Debug='~/mycroft-core/start-mycroft.sh debug'" >> ~/.bashrc`
`echo "alias Mycroft-Stop='~/mycroft-core/stop-mycroft.sh'" >> ~/.bashrc`

## How to test
* Run `dev_setup.sh` to have the script append the aliases to the end of bash profile
* Reload bash profile

CLA [X]

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -421,5 +421,11 @@ if [[ ! -w /var/log/mycroft/ ]] ; then
     sudo chmod 777 /var/log/mycroft/
 fi
 
+# add bash aliases to start and stop mycroft
+echo "alias Mycroft-Start='~/mycroft-core/start-mycroft.sh all'" >> ~/.bashrc
+echo "alias Mycroft-Debug='~/mycroft-core/start-mycroft.sh debug'" >> ~.bashrc
+echo "alias Mycroft-Stop='~/mycroft-core/stop-mycroft.sh'" >> ~.bashrc
+
+
 #Store a fingerprint of setup
 md5sum requirements.txt test-requirements.txt dev_setup.sh > .installed


forslund:
Hi @pseams, in the recently we (and by we I mean @penrods) added a mycroft-start script that the dev-setup.py can install to your path. When doing this we're using a .profile_mycroft file that gets sourced by the .profile (which is similar to .bashrc)

We currently add a mycroft-start command which needs to be called with the argument all to work.

If you think the alias can still be useful in addition to this please add them to the `.profile_mycroft` if the user agrees to add convenience commands to his or her path [around here somewhere](https://github.com/MycroftAI/mycroft-core/blob/dev/dev_setup.sh#L179)

penrods:
I hate to stomp on this first contribution, but believe the ```Mycroft-Stop``` would behave exactly the same as ```mycroft-stop```.  The ```mycroft-start``` is slightly different in that it still requires a parameter, but if they don't provide on they are given options.

Mycroft-Debug is the one that would be most distinct.  I'm not opposed to the idea of that leading to a script under the mycroft-core/bin folder that turns around and calls ```mycroft-start debug```.  If you do that then users should have access to that if they say Y to the deb_setup wizard prompt for modifying the PATH.  (https://github.com/MycroftAI/mycroft-core/pull/1965/files#diff-18dbf5207379240f00af7b954cde81ffR169)

penrods:
Actually, if you want to make this really spiffy you can implement bash completion.  This would allow a user to type:  mycr<TAB>-sta<TAB> d<TAB> to get ```mycroft-start debug```, for instance.

See:  https://iridakos.com/tutorials/2018/03/01/bash-programmable-completion-tutorial.html

I believe this would be an addition to the PATH wizard step that could be as easy as adding a call to:
```bash
complete -W "debug all audio bus skills voice enclosure cli unittest"
```
This kind of stuff can be placed in a mycroft-core/scripts/mycroft-start-completion.bash that gets copied to ```/etc/bash_completion.d/```.

Same could be done for several other scripts, such as ```mycroft-stop```, ```mycroft-msk``` and ```mycroft-msm```.

That would be an awesome addition I'd be happy to pull in, also!

penrods:
@impearker Shall we close this PR and you can just open a PR for future work?

impearker:
Yeah, that works fine. Sorry for the delay, I've been busy lately.

PR #1971, opened 2019-01-25
Add some additional info to audiotest

forslund:
## Description
Audiotest now prints the device and samplerate used as well as the commandline used to playback the audio for easier debugging if things doesn't work as intended.

## How to test
Run the audiotest and check that the output matches the settings.

CLA [ Yes ]

### DIFF
--- a/mycroft/util/audio_test.py
+++ b/mycroft/util/audio_test.py
@@ -14,6 +14,7 @@
 #
 import argparse
 import os
+import pyaudio
 from contextlib import contextmanager
 
 from speech_recognition import Recognizer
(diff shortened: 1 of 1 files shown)


mycroft/util/audio_test.py
```diff
@@ -80,6 +81,19 @@ def main():
         help="Add extra output regarding the recording")
     args = parser.parse_args()
 
+    config = Configuration.get()
+    if "device_name" in config["listener"]:
+        dev = config["listener"]["device_name"]
+    elif "device_index" in config["listenter"]:
```
penrods: Typo: listenter

mycroft/util/audio_test.py
```diff
@@ -80,6 +81,19 @@ def main():
...
+        dev = "Device at index {}".format(config["listener"]["device_index"])
+    else:
+        dev = "Default device"
+    samplerate = config["listener"]["sample_rate"]
+    play_cmd = config["play_wav_cmdline"].replace("%1", "WAV_FILE")
+    print(" ============================= Info ==============================")
```
penrods: Nice!  Would also be nice to list all devices when someone is looking for a mic.  I'm going to add a -l option.

forslund: Sounds good

mycroft/util/audio_test.py
forslund: Nice catch

forslund:
Nice addition. Merging

PR #1975, opened 2019-02-01
Update padatious and padaos

forslund:
## Description
- Fixes exception when adding/removing intents while processing an utterance
- Terminates worker processes correctly (8 sub-processes of mycroft.skills gets left hanging around)

## How to test
Make sure padatious intents still works as expected

CLA [ Yes ]


forslund:
These should be trival changes. self-merging.

PR #1978, opened 2019-02-02
Move sudo check to beginning of script

forslund:
## Description
The sudo command check was performed only when packages were installed and not before the interactve portion leaving $SUDO unset when trying to create the /opt/mycroft folder.

## How to test
Remove the `.dev_ops.json` file and remove `/opt/mycroft` and make sure the dev_setup script can recreate the folder successfully.

CLA [ Yes ]


forslund:
Self merging this change since users are having trouble getting mycroft setup.

PR #1980, opened 2019-02-04
Document "save_utterance" option

penrods:
The Listener has an option to save the sound that was recognized as a wake-word.   This isn't new functionality, but was previously undocumented.

## How to test
None needed, comments only


forslund:
Seems like the config key is [record_utterances](https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/client/speech/mic.py#L191)

Issue #1986, opened 2019-02-12
Error on fresh picroft install - No such file or directory: '/home/pi/.mycroft/skills/TimeSkill/uuid'


andlo:
in the log shows:
```
20:43:26.341 - mycroft.skills.settings:_poll_skill_settings:389 - ERROR - Failed to fetch skill settings: FileNotFoundError(2, 'No such file or directory')
Traceback (most recent call last):
  File "/home/pi/mycroft-core/mycroft/skills/settings.py", line 382, in _poll_skill_settings
    self.initialize_remote_settings()
  File "/home/pi/mycroft-core/mycroft/skills/settings.py", line 156, in initialize_remote_settings
    self.save_skill_settings(skill_settings)
  File "/home/pi/mycroft-core/mycroft/skills/settings.py", line 224, in save_skill_settings
    self._save_uuid(skill_settings['uuid'])
  File "/home/pi/mycroft-core/mycroft/skills/settings.py", line 259, in _save_uuid
    with open(uuid_file, 'w') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/pi/.mycroft/skills/TimeSkill/uuid'
```

Ill suspect that the user alreddy has a timersetting stored on home.mycroft.ai and during initialazation the settings is written to the device. But the skill isnt yet installed and thus there isnt the directory to store the setting in...

andlo:
I think line 259 in settings.pu should be changed from 

```
        with open(uuid_file, 'w') as f:
            f.write(str(uuid))
 ```
to 
```
  try:
            with open(uuid_file, 'w') as f:
                f.write(str(uuid))
        except Exception:
            LOG.exception('Failed to save settings to file ' + uuid_file)
``` 
which would fix that skills initilization breaks if a directory is missing.

forslund:
This is captured by the exception on [L389](https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/skills/settings.py#L389) (And the trace is logged).

I'm not sure how allowing code execution to proceed after the uuid-file write failed will work. I'll ping @MatthewScholefield who has more experience in the interactions of the skill settings with the server.

Possibly code that ensure that the path exists could be added as well?

andlo:
Ill agree. Either make sure the path exists or skip if it dosnt. Later when the skill is install the path will be made.

andlo:
Ill think this issue is handled so closing.

PR #1987, opened 2019-02-12
Fix#1986

andlo:
## Description
fix for #1986 


MatthewScholefield:
@forslund Would it be better to create the folder if it doesn't exist or just log a message like it does here?

Either way, it's better than throwing an exception so we can go ahead an pull it in. @andlo Can you fix this PEP8 issue?:
```
mycroft/skills/settings.py:266:5: E303 too many blank lines (2)
```
Thanks!

forslund:
I don't know the logic here enough to answer that.

I would assume it'd be best to create the folder if needed (I think that's done for the hash file). I'm not quite sure of the logic in here if the failure is expected, The exception generated here is currently caught at [L389](https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/skills/settings.py#L389) basically skipping all steps after the failed uuid file write. So there _may_ (and I don't really know) be consequences if the exception is caught and the following code is executed without an uuid file.

forslund:
Hi Andlo, thanks for this. @penrods went ahead and added the creation of the skills directory before trying to write the uuid-file. Do you think we should add this check as well?

andlo:
Not nessasaraly - depends on what happens if creating the directory fails....

forslund:
The same as it did previously: Get caught by this except statement: https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/skills/settings.py#L393

andlo:
Then I think @penroids solution will fix the issue. Ill close this then.

PR #1995, opened 2019-02-14
Clean up some exception handling in settings code

penrods:
The settings code worked, but was noisy and generally messy about
a few exceptional but common situations:
* When the .mycroft/skills/<SkillName> folder didn't already exist
* When network timeouts and such occcurred

I also slipped in a couple trivial code cleanups for an unused variable
and a log message.

## Description

## How to test
This mainly happens when running a fresh install of Mycroft

### DIFF
--- a/mycroft/skills/settings.py
+++ b/mycroft/skills/settings.py
@@ -60,8 +60,10 @@
 
 import json
 import hashlib
+import os
 from threading import Timer
 from os.path import isfile, join, expanduser
+from requests.exceptions import RequestException
 
 from mycroft.api import DeviceApi, is_paired
 from mycroft.util.log import LOG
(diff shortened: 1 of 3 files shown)


mycroft/skills/settings.py
```diff
@@ -140,7 +142,10 @@ def initialize_remote_settings(self):
 
         self._device_identity = self.api.identity.uuid
         self._api_path = "/" + self._device_identity + "/skill"
-        self._user_identity = self.api.get()['user']['uuid']
+        try:
+            self._user_identity = self.api.get()['user']['uuid']
+        except RequestException:
+            return
```
forslund: Will this change the behavior and stop the polling timer? If so is that good or bad?

penrods: I don't believe this changes any behavior, as the code above was created using simple "returns" in those cases.  And in my testing observations this was just throwing unhandled exceptions, which obviously didn't do anything smartly either.

penrods: Actually, you are right!  That _poll_skill_settings() code handles this poorly and needs some love.  This has probably been failing silently for some time.

PR #1998, opened 2019-02-15
Feature/update requirements

forslund:
## Description
Update some of the core requirements to newer version.

- Adapt
- pyee
- websockets

Note: pyee seems to need the SkillSettings to be "hashable" to work for members of it. This adds a sort of fake hash based on the id and name, if a cleaner approach can be suggested I'd be happy to update the PR.

## How to test
Make sure mycroft functions normally.

CLA [ Yes ]


penrods:
Did you intend to include the files besides requirements.txt?

forslund:
Yes, the changes in setting.py is necessary for pyee and the changes in messagebus/client/ws.py is needed for the updated websocket as far as I recall.

PR #1999, opened 2019-02-15
Add (|) syntax to dialog files

MatthewScholefield:
## Description
This PR allows dialog files to use the `(this|that)` syntax that has previously only been available in `.intent` files.

**Note:** This requires making the following changes to all `.voc` files:
```
a|b|c|d
```
->
```
(a|b|c|d)
```

## How to test
 - Make sure dialog files for previous skills work
 - Test a skill dialog with something like `This is (a (foo|bar) (test|))` and make sure it outputs a random choice of the options
 - Test adding skill vocab for something like `shut (it|) up`

### DIFF
--- a/mycroft/dialog/__init__.py
+++ b/mycroft/dialog/__init__.py
@@ -19,6 +19,7 @@ from pathlib import Path
 from os.path import join
 
 from mycroft.util import resolve_resource_file
+from mycroft.util.format import expand_options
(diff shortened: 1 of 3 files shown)


forslund:
Very neat addition. Works as expected. Merging

forslund:
Let's party like it's PR #1999

PR #2004, opened 2019-02-20
Tweak English extract_datetime() parsing

penrods:
Ambiguous times previously used a generally unexpected rule about
when to jump 12 hours when parsing times.  Now it follows the rule:
If a time is spoken without am/pm indicator, assume the next time
that hasn't passed was intended.

For example:
  "at 7 o'clock"
Would mean 7:00am if spoken at 6:59am, but would mean 7:00pm
if spoken at 7:01am.

NOTE: This behavior is distinctly different for 19.02.


forslund:
Looks good. Merging

PR #2009, opened 2019-02-22
Fix behavior of MycroftSkills.voc_match()

penrods:
Previously the voc_match() method had several problems:
* Blank lines in the .voc file would cause it to match all strings
* Comments weren't ignored
* Substrings matched so "book" would match "ok", for instance

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -721,9 +721,22 @@ class MycroftSkill:
                 self.voc_match_cache[cache_key] = f.read().splitlines()
 
         # Check for match
-        if utt and any(i.strip() in utt
-                       for i in self.voc_match_cache[cache_key]):
-            return True
+        if utt:
+            words = utt.split()
+            for i in self.voc_match_cache[cache_key]:
+                term = i.strip()
+                if not term or term[0] is "#":  # ignore blanks and comments
+                    continue
+
+                if len(term.split() == 1):
+                    if term in words:
+                        return True  # Found a single word match!
+                else:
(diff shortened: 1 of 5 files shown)


mycroft/skills/core.py
```diff
@@ -721,9 +721,22 @@ def voc_match(self, utt, voc_filename, lang=None):
...
 
         # Check for match
-        if utt and any(i.strip() in utt
-                       for i in self.voc_match_cache[cache_key]):
-            return True
+        if utt:
+            words = utt.split()
+            for i in self.voc_match_cache[cache_key]:
+                term = i.strip()
+                if not term or term[0] is "#":  # ignore blanks and comments
```
forslund: Let's do this when reading lines into the cache instead... Also we need to run the expand parenthesis on this as well

mycroft/skills/core.py
```diff
@@ -721,9 +721,22 @@ def voc_match(self, utt, voc_filename, lang=None):
...
+                    continue
+
+                if len(term.split() == 1):
```
forslund: I also wonder if a regex match would be cleaner using the word boundry
```python

if utt and any([re.match(r'\b' + term + '\b', utt) for i in self.voc_match_cache[cache_key]):
    return True
```

forslund:
Seems ok, the change that's needed is expanding parentheses.

forslund:
Ok, I've been doing some additional random testing and seems good. Together with the unittests I threw in this should be safe to merge.

PR #2010, opened 2019-02-22
Fix bug in the mic-level meter

penrods:
The mic-level meter that appears in the Mycroft CLI could stop updating in
rare cases when a file I/O operation failed.  It would also fail to show
a meter ever if the mic_level file hadn't been created before the CLI was
started.


forslund:
Looks good! Merging

PR #2011, opened 2019-02-22
Improve debug logging

penrods:
Several changes to the logging using in all processes:
* Disable logging of bus messages by default (massively cleans up the logs)
* Add "mycroft.debug.log" bus message.  It supports the data={"bus": True/False} as well as
  data={"log": LEVEL}, where level is a Python logging level.

CLI now supports several commands:
* ```:log level XXX``` where XXX is the name of a standard Python level (e.g. 'debug')
* ```:log bus on``` or ```:log bus off```
* Removed the requirement for "log" in ```:clear log```.  ```:clear``` works now.


forslund:
I don't get any log messages at all with this branch

PR #2012, opened 2019-02-22
Protect against failure to launch WAV/MP3

penrods:
The playback of WAV/MP3 could throw exceptions if something went
wrong, which seemed inappropriate.  Now it just generates an error
message in the log.


forslund:
I do think it's better if it generates an exception allowing the caller to handle the specific issue.

forslund:
Fixed the traceback output. Now it seems to be working as originally intended

PR #2014, opened 2019-02-25
Correct extract_duration doc and failure value

penrods:
The documentation indicated an array instead of a tuple, plus
the default return value didn't match the documented return type.


forslund:
Good catch, Merging

PR #2015, opened 2019-02-26
Feature/common io t

ChristopherRogers1991:
## Description
This add a CommonIotSkill base class, that will be used by skill that work with the CommonIot framework.

This and the two skills linked below are still works in progress. Combined they implement the core, initial CommonIot framework.

## How to test
Load https://github.com/MycroftAI/iot-test-skill and https://github.com/MycroftAI/skill-iot-control and issue commands like "turn on the lights," "toggle the kitchen lights," and "bedroom off."

### DIFF
--- a/mycroft/skills/common_iot_skill.py
+++ b/mycroft/skills/common_iot_skill.py
@@ -1,4 +1,4 @@
-# Copyright 2018 Mycroft AI Inc.
+# Copyright 2019 Mycroft AI Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -22,10 +22,8 @@ from abc import ABC, abstractmethod
 from enum import Enum
 from mycroft import MycroftSkill
 from mycroft.messagebus.message import Message
-from mycroft.util.log import getLogger
 
 
-LOGGER = getLogger(__name__)
 ENTITY = "ENTITY"
 SCENE = "SCENE"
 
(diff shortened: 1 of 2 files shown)


mycroft/skills/common_iot_skill.py
```diff
@@ -0,0 +1,340 @@
+# Copyright 2018 Mycroft AI Inc.
```
forslund: 2019?

mycroft/skills/common_iot_skill.py
```diff
@@ -0,0 +1,340 @@
...
+# BE WARNED THAT THE CLASSES, FUNCTIONS, ETC MAY CHANGE WITHOUT WARNING.
+
+from abc import ABC, abstractmethod
+from enum import Enum
+from mycroft import MycroftSkill
+from mycroft.messagebus.message import Message
+from mycroft.util.log import getLogger
+
+
+LOGGER = getLogger(__name__)
```
forslund: We usually do 

```python
```
and then just use

```python
LOG.info('This is useful information')
```

ChristopherRogers1991:
Sorry, I assumed everyone in the org would have access to those repos by default. They are very rough and not ready for public consumption yet, but I did intend for the core team to have access. I've added you (and the Mycroft core team) to the repos, so you should have access now.

I'll address the other nits a little later this evening.

forslund:
Thanks now I can access. Seems to work fine as far as I can see.

forslund:
If @penrods doesn't have any other objections I'll merge it tomorrow morning after you've got a chance to address those tiny things.

mycroft/skills/common_iot_skill.py
ChristopherRogers1991: Fixed.

mycroft/skills/common_iot_skill.py
ChristopherRogers1991: Removed this. The logger was only being used for some debugging yesterday, and I just forgot to remove it.

PR #2016, opened 2019-02-26
Wait for ntp sync complete message

MatthewScholefield:
## Description
We can easily wait for the message, even if it doesn't exist. I figured we can introduce this as long as there are no regressions, and then when we update the admin service for boot changes, we can also add this message. [Relevant mycroft-wifi-setup commit in PR](https://github.com/MycroftAI/mycroft-wifi-setup/pull/33/commits/5d9f71df52330f2699a7a44bc9084545b785d502).

## How to test
Start a device and make sure it still goes through the ntp sync like normal.


forslund:
Looks great. Merging.

PR #2017, opened 2019-02-26
Try updating the system on boot when not paired

MatthewScholefield:
## Description
This sends a new system.update.check message when the device is booted and not paired
Note: It is not sent every boot because the command could upgrade across major versions which the user could not want.

[Relevant mycroft-wifi-setup PR](https://github.com/MycroftAI/mycroft-wifi-setup/pull/33).

## How to test
 - Test that on desktop, the boot process is unaltered
 - Test that on a normal mark 1 the boot process us unaltered
 - [Optionally] Test that on an unpaired mark 1 that a system.update message is sent

### DIFF
--- a/mycroft/skills/__main__.py
+++ b/mycroft/skills/__main__.py
@@ -77,6 +77,19 @@ def _starting_up():
     check_connection()
 
 
+def try_update_system(platform):
+    bus.emit(Message('system.update'))
+    msg = Message('system.update', {
+        'paired': is_paired(),
+        'platform': platform
+    })
+    resp = bus.wait_for_response(msg, 'system.update.processing')
+
+    if resp and (resp.data or {}).get('processing', True):
+        bus.wait_for_response(Message('system.update.waiting'),
+                              'system.update.complete', 1000)
+
+
 def check_connection():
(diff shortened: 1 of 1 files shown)


mycroft/skills/__main__.py
```diff
@@ -77,6 +77,19 @@ def _starting_up():
...
 
+def try_update_system(platform):
+    bus.emit(Message('system.update'))
+    msg = Message('system.update', {
+        'paired': is_paired(),
+        'platform': platform
+    })
+    resp = bus.wait_for_response(msg, 'system.update.processing')
+
+    if resp.data.get('processing', True):
```
forslund: Should a check be made here if `resp` is None since a timeout will return None?

MatthewScholefield: Nice catch! Updated.

forslund: Yeah something like that will be needed, otherwise unpaired desktops will hang on start

forslund:
This looks quite good, seem to work well with the current mycroft setup and should work well with the wifi-setup updates. Merged

PR #2018, opened 2019-02-26
Add default values to Message attributes

MatthewScholefield:
## Description
This ensures no attributes are None and simplifies some checks within the function
Otherwise, if, for instance, there were no 'data' field in a message, when deserializing, the caller would have to check for None

## How to test
Make sure the following works correctly:
```python
class FooMessage:
    def __init__(self, msg_type):
        self.msg_type = msg_type

    def serialize(self):
        return {'type': self.msg_type}  # Doesn't include 'data'
bus.on('test.message', lambda msg: print(msg.data.get('something', 'HELLO'))
bus.emit(FooMessage('test.message'))
```

### DIFF
--- a/mycroft/messagebus/message.py
+++ b/mycroft/messagebus/message.py
@@ -36,10 +36,9 @@ class Message:
         Message objects will be used to send information back and fourth
         bettween processes of mycroft service, voice, skill and cli
         """
-        data = data or {}
         self.type = type
-        self.data = data
-        self.context = context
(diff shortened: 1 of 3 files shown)


forslund:
Seems like a good change, there seem to be some places where values are expected to be None which needs to be updated. Ex [intent_service.py L280](https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/skills/intent_service.py#L280)

PR #2023, opened 2019-02-27
date and time - for tonight, weekdays

Ruthvicp:
## Description
Not picking the time in case of tonight/weekdays when am/pm is not specified
examples 

- "set an alarm at 7 tonight"

- "set an alarm at 9 on weekdays"

## How to test

Test cases 

- set alarm at 7:30 on weekdays
- set an alarm tonight at 9

More test cases have been added to  test_extractdatetime_en in unittests\util\test_parse.py

### DIFF
--- a/test/unittests/util/test_parse.py
+++ b/test/unittests/util/test_parse.py
@@ -455,6 +455,19 @@ class TestNormalize(unittest.TestCase):
                     "2017-07-08 10:00:00", "remind me to call mom")
         testExtract("remind me to call mom at 10am next saturday",
                     "2017-07-08 10:00:00", "remind me to call mom")
+        # Below two tests, ensure that time is picked even if no am/pm is specified
+        # in case of weekdays/tonight
+        testExtract("set alarm for 9 on weekdays",
+                    "2017-06-27 21:00:00", "set alarm weekdays")
+        testExtract("for 8 tonight",
+                    "2017-06-27 20:00:00", "")
+        testExtract("for 8:30pm tonight",
+                    "2017-06-27 20:3:00", "")
+        # Tests a specified time with : & without am/pm
+        testExtract("remind me about the game tonight at 11:30",
+                    "2017-06-27 23:30:00", "remind me about game")
+        testExtract("set alarm at 7:30 on weekdays",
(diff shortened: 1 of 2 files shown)


test/unittests/util/test_parse.py
```diff
@@ -455,6 +455,19 @@ def testExtract(text, expected_date, expected_leftover):
...
+                    "2017-06-27 21:00:00", "set alarm weekdays")
+        testExtract("for 8 tonight",
+                    "2017-06-27 20:00:00", "")
+        testExtract("for 8:30pm tonight",
+                    "2017-06-27 20:30:00", "")
+        # Tests a specified time with : & without am/pm
+        testExtract("remind me about the game tonight at 11:30",
+                    "2017-06-27 23:30:00", "remind me about game")
+        testExtract("set alarm at 7:30 on weekdays",
+                    "2017-06-27 19:30:00", "set alarm on weekdays")
```
forslund: Just an observation here: I would probably expect this to set an alarm in the morning.

I'm used to working with a 24-hour time system so this might just be my cultural training.

Ruthvicp: The unit test case considers reference time as date = datetime(2017, 6, 27, 13, 4) , [here](https://github.com/MycroftAI/mycroft-core/blob/64666590947464b454260d9c11ca489fa8d5a181/test/unittests/util/test_parse.py#L181) 

so when am/pm is not specified, the time picked would be the next available time.
For example : If current time is 6:30pm 

- Set an alarm at 7:30 on weekdays will extract time as 7:30 pm from today

- Set an alarm at 6 on weekdays will extract 6:00 am from tomorrow

- If am/pm is specified then it extracts as per the request

penrods: This usage is a little ambiguous, which is part of why the alarm skill provides verbal feedback (e.g. "starting in 2 hours") so the user will realize the 12 hour misunderstanding if it happens.

The same extraction code would be used for "remind me to exercise at 6 on weekdays", which could mean either AM or PM depending on the person.

I'm OK with this for now, we can revisit it if it feels wrong in more use cases.

PR #2026, opened 2019-02-28
Tweak CommonPlay scoring

penrods:
The scoring for CommonPlay would favor skills that responded with overly long matches.  E.g. "Huey Lewis and the News" would be considered a better fit than just "News" for the request "Play the News"

Also added the self.spoken_name for future use by the CommonPlay framework.  It will likely be pronounced in something like ```"Playing XXX using "+self.spoken_name```


forslund:
Auto-bots agree, good PR. Merging

PR #2030, opened 2019-02-28
Hide deprecated warning in certain cases

penrods:
The deprecation warning was firing off at the load of every skill when
the decorators are being iterated through by internal code (which
includes the self.config with an @property).  Add check for these
special cases to not show a warning.

## How to test
Reboot or ```sudo service mycroft-skills restart```.  Watch the logs sparkle!


forslund:
Tested and working as expected.

forslund:
Codacy is blocking for ever...Merging

PR #2035, opened 2019-03-04
Ignore the 'skills' symlink

penrods:
The .gitignore wasn't ignoring the symlink that would be created in
dev_setup.sh to point to /opt/mycroft/skills.  The gitignore treats
what looks like an absolute path "/skills" as relative to the
repo.

## How to test
A ```git status``` before this change lists the 'skills' directory as an untracked file.  With this change it shouldn't appear in that last.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@ dev.env
 *.swp
 *~
 mimic
-./skills
+/skills
 pocketsphinx-python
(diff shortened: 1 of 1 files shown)


forslund:
Looks good, ignores skills in the root folder but added files under the /mycroft/skills folder shows up. Merging.

PR #2045, opened 2019-03-08
Fix simple audio backend pausing by voice

forslund:
## Description
The restore_volume would previously unpause the service even if the user just asked for pausing playback.

This separates the paused flag from the pausing/resuming action so lowering/restoring volume doesn't affect the logical pausing state and can check it to determine if lowering/restoring is needed/wanted.
## How to test
Play the news, ask Mycroft to "pause playback" by voice and make sure it's paused afterwards. Resume playback.

CLA [ Yes ]


penrods:
LGTM

PR #2047, opened 2019-03-09
Made start_listening louder

domcross:
## Description
The new start_listening.wav (introduced with release 19.02.0) is a little soft compared to the old one. I have raised the level a bit (approx. 10-12 dB) while still leaving enough headroom, so there shouldn't be any distortion.

## How to test
Say "hey mycroft". Listening sound should be loud and clear.

CLA [X] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)


forslund:
Hi @domcross, I think part of the idea with the new listening sound was to make it slightly more quiet. I'll ping @penrods to make a decision here.

domcross:
In my opinion a clearly audible listening sound is still important for "low cpu-powered" devices (Picroft, Mark1, Respeaker CoreV2) as there is a short lag between the wake word detection and the recording start of the utterance. This requires the speaker make a pause after the wake word otherwise the beginning of the utterance is not recorded. 

This may not be an issue on Desktop machines and Mark2 which have better performing cpu where the utterance recording is starting immediately after wake word is detected and the speaker can keep a natural "speaking flow".

penrods:
Let me play with this and think about options.  As Ake mentioned, the new softer sound was intentional.

forslund:
Possibly we could add 3 versions of the sound at different volumes and make it a setting like the on/off is.

forslund:
Hi, was the latest commit pushed to the wrong branch?

domcross:
Oops, my fault. What to do now?

forslund:
We can separate out the first commits if we'll decide to merge this PR. If you want to keep working on the new common de things you might want to create a new branch for it.

From the command line:
`git checkout -b my-good-and-descriptive-branch-name`

PR #2054, opened 2019-03-14
Fix mixing of int/str in extract_datetime_* code

penrods:
Found several mismatches of str/int in parsing code that got copied to other
implementations.  Fixed them as a whole.

## How to test
Run unit tests


forslund:
We'll soon (I hope) switch over to lingua franca so I'm closing this wip branch.

PR #2055, opened 2019-03-15
Mimic2 fallback overlaps sentences

Ruthvicp:
## Description
Mimic2 fallback to mimic overlaps sentences because multiple Mimic objects are being created for each sentence. This happens in 2 cases
1. When connection to mimic2 is lost while speaking multiple sentences 
2. Select mimic2 as your TTS engine and reset the device. Utterances overlap on startup

## How to test

- Set Mimic2 as you TTS engine
- Reset your device and listen to utterances on startup

If you do not want to reset your device
- Hit the top button & select 'WIFI' 
- Do not setup the wifi and ask general questions to Mark1 (ex: what is the time, weather..)
- Look for the var/logs/mycroft/auido.log to verify the connection error & fallback
- The utterance "Use your mobile device or computer to connect to the wifi network 'MYCROFT'. Then
    enter the password 1 2 3 4 5 6 7 8" should **not overlap**


mycroft/tts/__init__.py
```diff
@@ -172,9 +174,11 @@ def __init__(self, lang, config, validator, audio_ext='wav',
         self.filename = '/tmp/tts.wav'
         self.enclosure = None
         random.seed()
-        self.queue = Queue()
-        self.playback = PlaybackThread(self.queue)
-        self.playback.start()
+        if not self.queue:
```
penrods: This isn't doing what you think it is doing.  ```self.test``` is NOT the same as ```TTS.queue```.  You can try the following

```python
class X:
    test = None
    def __init__(self):
        print("Initializing")
        if not X.test:
            print("Setting test for the first time")
            X.test = 1
        print("__init__ complete")
    def do_test(self):
        print("Test is "+str(X.test))
        X.test += 1
		
		
		
class Foo:
    test = None
    def __init__(self):
        print("Initializing")
        if not self.test:
            print("Setting test for the first time")
            self.test = 1
        print("__init__ complete")
    def do_test(self):
        print("Test is "+str(self.test))
        self.test += 1
```

Then when you create two instances of X you'll see that the value is independent.


So, change all of the ```self.queue``` to ```TTS.queue```

mycroft/tts/__init__.py
```diff
@@ -172,9 +174,11 @@ def __init__(self, lang, config, validator, audio_ext='wav',
...
+            self.queue = Queue()
+        if not self.playback:
```
penrods: Same deal -- this should be TTS.playback instead of self.playback

PR #2059, opened 2019-03-21
Adjust and document Padatious loose fallback priority

penrods:
The loose (conf > 0.5) Padatious match was previously occurring as Fallback
priority 99. The AIML fallback at priority 90 would consume lots of
utterances, interferring with many skills.  Now Padatious runs at priority
89.

Additionally, added documentation of the intent and fallback system, including
guidelines for priorities.

## Description

## How to test
* Speak a near match to an existing skill's Padatious intent.  E.g. "how long to el next alarm?"
* Enable AIML at in the "Chatbot Fallback" skill setting at Home, checking "ENABLE CHATBOT FALLBACK"
* Speak a near match to an existing skill's Padatious intent.  E.g. "how long to el next alarm?"

With this change, the second utterance should also trigger the Alarm skill instead of getting a sarcastic answer from AIML (like "More than 17 inches" or "More than 3 years")

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -1637,10 +1637,26 @@ class MycroftSkill:
 #######################################################################
 class FallbackSkill(MycroftSkill):
     """
-        FallbackSkill is used to declare a fallback to be called when
-        no skill is matching an intent. The fallbackSkill implements a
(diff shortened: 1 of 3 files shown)


forslund:
Documentation for the Win! Merging

PR #2060, opened 2019-03-21
Make precise sensitivity configurable

MatthewScholefield:
## Description
This adds the following new configuration options relevant to precise wake words:
```json
{"hotwords": {"MY-WAKE-WORD-THAT-USES-PRECISE": {"sensitivity": 0.5, "trigger_level": 3}}}
```
A description of each is as follows:
 - `sensitivity`: Network output necessary to be considered an activation
 - `trigger_level`: Number of consecutive chunks that are activated to cause a wakeword detection

**Note**: As it is right now (might change soon), the threshold is very nonlinear so changing from 0.5 to 0.6 won't do much, but changing from 0.5 to 0.02 or 0.01 should make a noticeable difference.

## How to test
 - Try setting a precise wakeword's sensitivity setting from 0.5 to `0.01 and 0.99 to see if it makes a difference.
 - Set the trigger_level to a value like 8 and it should be harder to trigger and trigger with a slight delay.

### DIFF
--- a/mycroft/client/speech/hotword_factory.py
+++ b/mycroft/client/speech/hotword_factory.py
@@ -153,9 +153,13 @@ class PreciseHotword(HotWordEngine):
         def on_activation():
             self.has_found = True
 
+        trigger_level = self.config.get('trigger_level', 3)
+        sensitivity = self.config.get('sensitivity', 0.5)
+
         self.runner = PreciseRunner(
             PreciseEngine(precise_exe, self.precise_model),
-            stream=self.stream, on_activation=on_activation
+            trigger_level, sensitivity,
+            stream=self.stream, on_activation=on_activation,
         )
         self.runner.start()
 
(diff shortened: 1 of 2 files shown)


forslund:
Everything in mycroft-core seems to be working as expected with this change. But I can't really detect much difference with the wakeword. Should I need to delete and re-download precise?

MatthewScholefield:
No, you shouldn't need to redownload anything. I just realized I was stupid and put `wakewords` as the config key value when it should actually be `hotwords`. If that doesn't make a difference, try changing it to more extreme values like `0.995` and `0.0001`. For reference, here's my config:
```json
{
  "hotwords": {
    "hey mycroft": {
      "module": "precise",
      "sensitivity": 0.995
    }
  }
}
```

forslund:
Yeah I verified my settings and they were OK.

I can't detect much of a difference between sensitivity 0.01 and 0.99, but 0.999 triggers continuously so it does something.

I guess any issue with the sensitivity setting is outside of mycroft-core so this is ok to merge by me.

PR #2062, opened 2019-03-22
Add translations from the Timer skill

penrods:
Moved translated text from the Timer skill into core.  This began
the mycroft-core translation effort for Russian and Turkish!

## How to test
Just text resources, but you could try in the languages if you speak them!


forslund:
I'm wondering here about the single minute/hour/day/etc words, they're all translations of "one minute/hour/day/etc" which may be tied very close to the use case of the timer. Is this intended for the use case you planned for these in core?

PR #2065, opened 2019-03-25
properly handling "tonight"

Ruthvicp:
## Description
```
 set an alarm for 9 o'clock tonight 
```
Creates an alarm for 9 am (If current time is before 9 am)

## How to test
Added unit tests for special cases. You can try the below examples on Mark1
(Note: Try these before the current time is 9 AM)
- set an alarm for tonight at 9:30
- set an alarm at 9 for tonight
- set an alarm at 10:30 tonight
- set an alarm at 10 o'clock tonight


JarbasAl:
shouldn't the "tonight" be removed from the utterance remainder aswell?

Ruthvicp:
> shouldn't the "tonight" be removed from the utterance remainder aswell?
> 
> i would expect the remainder for "set an alarm for tonight 9:30" to be just "set an alarm"

Yes, that was the previous implementation where it removes 'tonight' from the utterance. Because of this it picks the next immediate available time. 
For ex :  
Current time : 8AM
Utterance : set an alarm at 9 tonight
It picks the next available 9 which is 9AM. 

I also implemented (which works) by removing 'tonight' from the utterance but that would require additional variable to check

PR #2067, opened 2019-03-27
Normalize utterances before intent handling

penrods:
Previously Padatious intent matches were performed on non-normalized text, meaning
that things like "what's the weather" wouldn't match a Padatious intent but
"what is the weather" would.

Also added better debugging info for intent matching to the log.

## How to test
```When's the next alarm set for?``` was failing, but ```When is the next alarm set for?``` worked.  After this change they both work.

### DIFF
--- a/mycroft/skills/intent_service.py
+++ b/mycroft/skills/intent_service.py
@@ -441,7 +441,8 @@ class IntentService:
                 if intent:
                     take_best(intent, utt)
 
-                # Also test the normalized version
+                # Also test the normalized version, but set the utternace to
+                # the raw version so skill has access to original STT
                 norm_intent = next(self.engine.determine_intent(
                     norm_utt[idx], 100,
                     include_tags=True,
--- a/mycroft/skills/intent_service.py
+++ b/mycroft/skills/intent_service.py
@@ -316,35 +316,69 @@ class IntentService:
             # Get language of the utterance
             lang = message.data.get('lang', "en-us")
             set_active_lang(lang)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: mycroft/skills/intent_service.py [hunk 1]
```diff
@@ -316,7 +316,11 @@ def handle_utterance(self, message):
             # Get language of the utterance
             lang = message.data.get('lang', "en-us")
             set_active_lang(lang)
-            utterances = message.data.get('utterances', '')
+
+            # normalize() changes "it's a boy" to "it is a boy", etc.
+            utterances = [normalize(u.lower(), remove_articles=False)
+                          for u in message.data.get('utterances', '')]
```
forslund: This one should for consistency default to the empty list `[]`

JarbasAl:
not sure i like this much... i would expect unnormalized utterances in fallbacks, i'm not a fan of removing articles, if we are hitting a third party service (say a chatbot) this might cause trouble

this might also impact existing padatious intents, i personally recently started augmenting padatious when loading the files, adding new examples of normalized utterances to the intents, this also impacts confidence (in my limited experience for the better)

as long as we have access to the raw utterance i don't mind it

mycroft/skills/intent_service.py
penrods: Yup!

penrods:
I think my second approach strikes a balance.  It attempts to intent match both the raw and the normalized version of the utterance.  The raw is used unless the normalized() is needed to get a higher confidence.

PR #2070, opened 2019-03-27
Set default log level to INFO

forslund:
## Description
Decrease the default log level to INFO. This also adds a fix to allow changing log levels again and enabling bus logging.

## How to test
Start mycroft and check the cli for log messages. They should be greatly reduced.

CLA [ Yes ]

### DIFF
--- a/mycroft/util/__init__.py
+++ b/mycroft/util/__init__.py
@@ -490,6 +490,7 @@ def create_echo_function(name, whitelist=None):
                     LOG.level = lvl
                     LOG(name).info("Changing log level to: {}".format(lvl))
                     try:
+                        logging.getLogger().setLevel(lvl)
                         logging.getLogger('urllib3').setLevel(lvl)
                     except Exception:
                         pass  # We don't really care about if this fails...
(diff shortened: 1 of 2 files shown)


penrods:
Are you able to switch log levels with this?  I modified the skill-date-time with the lines:

```python
    def get_local_datetime(self, location, dtUTC=None):
        self.log.debug("DEBUG MESSAGE!")
        self.log.error("ERROR MESSAGE!")
        self.log.info("INFO MESSAGE!")
```
With this change, inside the CLI I only see the ERROR MESSAGE and INFO MESSAGE when I ask "what time is it" -- no DEBUG MESSAGE.  (So far so good)

Then if I use ```:log level debug``` in the CLI and try "what time is it" again I still don't see the debug message.

forslund:
That's a limit with the original implementation of the log level selection I commented on previously. The self.log.* in skills will not be affected only the logs using `LOG`

I'll add that functionality as well

forslund:
Now it should be able to change the log level of the MycroftSkill.log

PR #2071, opened 2019-03-27
separate the cache & avoid voice overlap

Ruthvicp:
## Description
Mimic2 fallback to mimic overlaps sentences because multiple Mimic objects are being created for each sentence. This happens in 2 cases

1. When connection to mimic2 is lost while speaking multiple sentences
2. Select mimic2 as your TTS engine and reset the device. Utterances overlap on startup

## How to test

- Set Mimic2 as you TTS engine
- Reset your device and listen to utterances on startup

If you do not want to reset your device

- Hit the top button & select 'WIFI'
- Do not setup the wifi and ask general questions to Mark1 (ex: what is the time, weather..)
- Look for the var/logs/mycroft/auido.log to verify the connection error & fallback
- The utterance "Use your mobile device or computer to connect to the wifi network 'MYCROFT'. Then
- enter the password 1 2 3 4 5 6 7 8" should not overlap

CLA [ Yes] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

### DIFF
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -30,6 +30,7 @@ config = None
 tts = None
 tts_hash = None
 lock = Lock()
+mimic_fallback_obj = None
 
 _last_stop_signal = 0
 
@@ -130,11 +131,15 @@ def mute_and_speak(utterance, ident):
 
 
 def mimic_fallback_tts(utterance, ident):
+    global mimic_fallback_obj
     # fallback if connection is lost
     config = Configuration.get()
     tts_config = config.get('tts', {}).get("mimic", {})
     lang = config.get("lang", "en-us")
-    tts = Mimic(lang, tts_config)
+    if not mimic_fallback_obj:
+        mimic_fallback_obj = Mimic(lang, tts_config)
+    tts = mimic_fallback_obj
+    LOG.debug("Mimic fallback, utterance : " + str(utterance))
     tts.init(bus)
     tts.execute(utterance, ident)
 
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -184,3 +184,6 @@ def shutdown():
     if tts:
         tts.playback.stop()
         tts.playback.join()
+    if mimic_fallback_obj:
+        mimic_fallback_obj.playback.stop()
+        mimic_fallback_obj.playback.join()
--- a/mycroft/tts/__init__.py
+++ b/mycroft/tts/__init__.py
@@ -177,6 +177,7 @@ class TTS:
         self.playback.start()
         self.clear_cache()
         self.spellings = self.load_spellings()
+        self.tts_name = type(self).__name__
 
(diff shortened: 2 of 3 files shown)


mycroft/audio/speech.py
```diff
@@ -118,23 +119,30 @@ def mute_and_speak(utterance, ident):
         tts = TTSFactory.create()
         tts.init(bus)
         tts_hash = hash(str(config.get('tts', '')))
+        tts.set_fall_back(False)
```
forslund: This method doesn't seem to exist

mycroft/audio/speech.py
```diff
@@ -118,23 +119,30 @@ def mute_and_speak(utterance, ident):
...
 
     LOG.info("Speak: " + utterance)
     try:
         tts.execute(utterance, ident)
     except RemoteTTSTimeoutException as e:
         LOG.error(e)
+        # Clear the previous cache
+        tts.clear_cache()
```
forslund: Is the subfolder needed if the cache is cleared

mycroft/tts/__init__.py
```diff
@@ -209,7 +209,7 @@ def end_audio(self):
 
         self.bus.emit(Message("recognizer_loop:audio_output_end"))
         # Clean the cache as needed
-        cache_dir = mycroft.util.get_cache_directory("tts")
+        cache_dir = mycroft.util.get_cache_directory("tts/" + self.voice)
```
forslund: `self.voice` can be `None` which would be an issue here can `type(self).__name__` to get the name of the class.

The cache is cleared when a new TTS is created I believe so changing voice should be handled

mycroft/tts/__init__.py
```diff
@@ -341,11 +342,12 @@ def clear_cache(self):
...
         if not os.path.exists(mycroft.util.get_cache_directory('tts')):
             return
-        for f in os.listdir(mycroft.util.get_cache_directory("tts")):
-            file_path = os.path.join(mycroft.util.get_cache_directory("tts"),
-                                     f)
-            if os.path.isfile(file_path):
-                os.unlink(file_path)
+        for d in os.listdir(mycroft.util.get_cache_directory("tts")):
+            dir_path = os.path.join(mycroft.util.get_cache_directory("tts"), d)
+            for f in os.listdir(dir_path):
```
forslund: Should check if it's a file or directory I got error here when the cache directory was not empty of normal files. (as it will be after an update)

mycroft/audio/speech.py
Ruthvicp: Yeah this is not mandatory, just to clear the old data

mycroft/audio/speech.py
Ruthvicp: yes, will remove that

forslund:
Travis is acting up. Merging

PR #2074, opened 2019-04-01
Feature/skillsmeta gid

forslund:
## Description
Skill gid and pretty name

CLA [ Yes ]

### DIFF
--- a/mycroft/skills/settings.py
+++ b/mycroft/skills/settings.py
@@ -130,12 +130,9 @@ def build_global_id(directory, config):
 
     s = SkillEntry.from_folder(directory, msm)
     if s.meta_info != {}:
-        name = s.meta_info['name']  # Name from skill's metadata
-        branch = s.msm.repo.branch  # Branch from the currently configured
-        return "{}-{}".format(name, branch)
+        return s.meta_info['skill_gid']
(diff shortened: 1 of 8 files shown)


mycroft/skills/settings.py
```diff
@@ -195,18 +270,37 @@ def __setitem__(self, key, value):
...
                 return None
         else:
-            return None
+            data = copy.copy(BLANK_META)
+
+        # Add Information extracted from the skills-meta.json entry for the
+        # skill.
+        skill_gid, title = build_global_id(self._directory, self.config)
+        data['skill_gid'] = skill_gid
+        data['title'] = (title or data.get('name') or
```
penrods: @chrisveilleux did we agree to "title" or "display_name" on this?  This is the settingmeta.json

@forslund let's also remove leading/trailing skill from the camel_case_split.  E..g "PandoraMusicSkill" becomes "Pandora Music".  I think this is better 99% of the time, and they 1% can explicitly define a "name" to force "Skill" to be there if they really need it..

forslund: I've updated to remove leading / trailing "Skill" and renamed the entry to `display_name`

penrods:
LGTM!

PR #2078, opened 2019-04-05
Remove boilerplate info from blank settingsmeta

forslund:
## Description
Removes the settingsMetadata from the skillsmeta to upload. We might want to hold off on this one until everything settles with the servers.

## How to test
Verify that settingsmeta can be uploaded and settings downloaded.

CLA [ Yes ]


forslund:
This is not quite right. I'll need to redo it completely and really thing about it now that hashes are back.

PR #2088, opened 2019-04-08
Feature/audioservice list

forslund:
## Description
The AudioService object now has the new method available_backends() returning a dict of all available audio backends as keys containing 'supported_uris' listing supported uri types (file:// http:// etc.), 'remote' (bool) True if remote), 'default' (bool) true if the default audio backend.

## How to test
Ensure the AudioService.available_backends() returns a valid response

CLA [ Yes ]

### DIFF
--- a/mycroft/audio/audioservice.py
+++ b/mycroft/audio/audioservice.py
@@ -194,6 +194,7 @@ class AudioService:
         self.bus.on('mycroft.audio.service.next', self._next)
         self.bus.on('mycroft.audio.service.prev', self._prev)
         self.bus.on('mycroft.audio.service.track_info', self._track_info)
+        self.bus.on('mycroft.audio.service.list_backends', self._list_backends)
         self.bus.on('mycroft.audio.service.seek_forward', self._seek_forward)
         self.bus.on('mycroft.audio.service.seek_backward', self._seek_backward)
         self.bus.on('recognizer_loop:audio_output_start', self._lower_volume)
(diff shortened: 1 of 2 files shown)


mycroft/audio/audioservice.py
```diff
@@ -496,4 +509,3 @@ def shutdown(self):
         self.bus.remove('recognizer_loop:audio_output_end',
                         self._restore_volume)
         self.bus.remove('recognizer_loop:record_end', self._restore_volume)
-        self.bus.remove('mycroft.stop', self._stop)
```
MatthewScholefield: Any reason for this change?

forslund: Yes, it's removed but it's never added so it's sort of redundant (I put it in a separate commit to differentiate it from the other change)

MatthewScholefield:
Where would I see the message reply? I'm doing `tail -f /var/log/mycroft/*.log` and all I see when I do `python3 -m mycroft.messagebussend mycroft.audio.service.list_backends` is a single message about chromecasts:
```
10:40:45.562 - chromecast__init__:supported_uris:67 - INFO - Chromecasts found: Chromecast(...)
```

forslund:
I do think the logging of the messagebus is broken. @penrods was looking into that.

forslund:
Try adding this to an intent handler:
```python
        print(AudioService(self.bus).available_backends())
```

PR #2089, opened 2019-04-09
Remove "stop_threshold"

penrods:
The setting "stop_threshold" is not used (and doesn't look like it ever was).  Removing
from mycroft.conf.

## How to test
Search for usage of "stop_threshold".  Not found in mycroft-core or any skills on a unit.

### DIFF
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -107,8 +107,6 @@
     "upload_skill_manifest": true,
     // Directory to look for user skills
     "directory": "~/.mycroft/skills",
-    // TODO: Old unused kludge, remove from code
-    "stop_threshold": 2.0,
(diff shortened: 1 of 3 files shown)


penrods:
Nice catch!  Your removal from the unit-test looks good to me.

JarbasAl:
was this some planned feature that was never implemented? what was this threshold meant for in the first place?

KathyReid:
> was this some planned feature that was never implemented? what was this threshold meant for in the first place?

@JarbasAI not sure, before my time I'm afraid.

penrods:
This is from WAY back, I don't think it had any specific meaning.  There were just "threshold" values for everything.

forslund:
I rebased the additions by @KathyReid into a single commit to keep the commit history clean. Merging now.

PR #2090, opened 2019-04-09
Fix bug with paused audio

penrods:
The simple audio service can be paused.  The audio is "paused" but placing the
audio playback mechanism (usually aplay or mpg123) into the background, kinda
like hitting Ctrl+Z.  While in the background that process can no longer react
to the SIGTERM signal.  So paused audio has to be brought to the foreground
using SIGCONT, and _then_ sent SIGTERM.

Also added a belt to go with these suspenders and perform a process.kill() if
the audio process doesn't go away after 1 second.

## How to test
Before this PR:
* "Hey Mycroft, sing a song"
* "Hey Mycroft, stop" (or press the button a a Mark 1)
* ```ps aux | grep mpg123```.  You'll see a suspended process
* "Hey Mycroft" doesn't make a ping sound anymore

After this PR the mpg123 should be gone after the Stop.

### DIFF
--- a/mycroft/audio/services/simple/__init__.py
+++ b/mycroft/audio/services/simple/__init__.py
@@ -46,7 +46,6 @@ class SimpleAudioService(AudioBackend):
     """
 
     def __init__(self, config, bus, name='simple'):
-
         super().__init__(config, bus)
         self.config = config
         self.process = None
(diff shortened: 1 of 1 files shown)


forslund:
This seems to be good now.

PR #2096, opened 2019-04-14
CommonIOT/acknowledge

ChristopherRogers1991:
## Description
Adds a function to MycroftSkill to acknowledge successful requests non-verbally. This is intended to be used in the commonIoT control skill, to acknowledge that a request was successfully passed to an appropriate skill.

Note that this also add a few key THING types for working with thermostats. I figured it would be easier to include that here than have a separate PR for the three lines in `common_iot_skill.py`, but am happy to split that out into it's own PR at your preference.

## How to test
Pull down the `improvement/acknowledge` branch of `skill-iot-control`, and run an IoT command (see  https://github.com/MycroftAI/skill-iot-control/pull/2, which will be merged after this PR goes through).

CLA [x ] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)


mycroft/skills/core.py
```diff
@@ -1648,6 +1650,26 @@ def cancel_all_repeating_events(self):
...
+        file = resolve_resource_file(
+            self.config_core.get('sounds').get('acknowledge'))
+
+        if not file:
+            LOG.warning("Could not find 'acknowledge' audio file!")
+            return
+
+        process = play_audio_file(file)
+        if process:
+            process.wait()
```
forslund: Should this be blocking by default? maybe a parameter `wait` defaulted to False?

mycroft/skills/core.py
```diff
@@ -1648,6 +1650,26 @@ def cancel_all_repeating_events(self):
...
             self.cancel_scheduled_event(e)
 
+    def acknowledge(self):
+        """ Acknowledge a successful request.
+
+        This method plays a sound to acknowledge a request that does not
+        require a verbal response. This is intended to provide simple feedback
+        to the user that their request was handled successfully.
+        """
+        file = resolve_resource_file(
```
clusterfudge: nit: I would discourage use of "file" as a variable name. While `file` is no longer a keyword in python3, most syntax highlighters are going to make it look risky (github's included!). Thanks!

ChristopherRogers1991: Good catch. I agree using file is not ideal. I copied from https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/client/speech/mic.py#L513. I'll change it in both places.

mycroft/skills/core.py
ChristopherRogers1991: I was referencing https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/client/speech/mic.py#L517 which uses `wait`. I suppose there it's more necessary because we want to wait for the sound to play before we start listening to the user (I would assume we run the risk of the sound being recorded otherwise, and impeding the STT). Here it's probably less necessary/less useful, since it's a separate process and there's nothing to stop the wake word/STT/other processes anyway. I'll remove it.

ChristopherRogers1991:
@forslund, regarding the sound and ACKNOWLEDGEMENTS.md`. I got the sound from @penrods. Not sure if it's his creation, or if he obtained it from somewhere. Ultimately I do expect it will be good to get the design folks to create a sound for this.

penrods:
> @forslund, regarding the sound and ACKNOWLEDGEMENTS.md`. I got the sound from @penrods. Not sure if it's his creation, or if he obtained it from somewhere. Ultimately I do expect it will be good to get the design folks to create a sound for this.

This was a sound I used in Christopher.  All sounds there were either excepts or modifications of totally free sounds from FreeSound.org, or sounds I created otherwise.  I can dig in my records to be certain, but it should be clear to use here without attribution.

I also expect this will only be a placeholder sound until Design comes up with something better.

forslund:
@penrods I'll merge this as it is now but if possible try to find the source for it. even if it's not required it's always good to list the sources of things like this.

PR #2098, opened 2019-04-15
Change default listener in config

MatthewScholefield:
## Description
The listener has been previously set via other configurations like /etc/mycroft.conf on the Mark 1. This sets it to Precise by default on all platforms.

## How to test
 - Ensure it starts on a desktop (look for `precise-something` executable in process monitor
 - Ensure Pocketsphinx starts as a fallback after making the `precise.dist_url` field invalid with something like:

```json
{"precise": {"dist_url": "https://github.com/MycroftAI/INVALID/{arch}/latest"}}
```

### DIFF
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -194,7 +194,7 @@
   // Hotword configurations
   "hotwords": {
     "hey mycroft": {
-        "module": "pocketsphinx",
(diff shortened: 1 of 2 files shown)


forslund:
I updated the tests they seem to be pocketsphinx based. Maybe at somepoint some precise specific unittests to match should be added.

Merging

PR #2106, opened 2019-04-23
feature/enable sending qml ui over network

AIIX:
## Description
Enables sending qml ui from remote mycroft-core instance over network 

## How to test
Add to mycroft.conf (any level):
{
    "remote": true,
    "remote-server": "https://instance-where-mycroft-core-is-hosted-.com"
}

Additional Configuration Server Side Running Apache:
`ln -s /opt /var/www/html/opt`

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -234,11 +234,12 @@ class SkillGUI:
         self.page = None    # the active GUI page (e.g. QML template) to show
         self.skill = skill
         self.on_gui_changed_callback = None
-        self.config_core = Configuration.get()
+        self.config = Configuration.get()
 
     @property
     def remote_url(self):
-        return self.config_core.get('remote-server')
+        """ Returns configuration value for url of remote-server. """
+        return self.config.get('remote-server')
 
     def build_message_type(self, event):
         """ Builds a message matching the output from the enclosure. """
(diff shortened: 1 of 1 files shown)


mycroft/skills/core.py
```diff
@@ -234,6 +234,11 @@ def __init__(self, skill):
         self.page = None    # the active GUI page (e.g. QML template) to show
         self.skill = skill
         self.on_gui_changed_callback = None
+        self.config_core = Configuration.get()
```
forslund: let's just call this member `config`. I know it doesn't match the one in the MycroftSkill class but that will be changed from self.config_core to self.config as part of 19.08

mycroft/skills/core.py
```diff
@@ -234,6 +234,11 @@ def __init__(self, skill):
...
+        
+    @property
+    def remote_url(self):
```
forslund: Add simple docstring here.

AIIX:
Yes, the generated path would be: http://my-mycroft.com/opt/mycroft/skills/mycroft-weather.mycroftai/ui/weather.qml, the simplest way i did it was linking /opt/mycroft/skills to the default html folder, access level of directory could be changed to only be able to view the skill folder from apache but the path would remain the same as above, or with a bit of tweaking path could be different. 

Fixing rest of the issues.

forslund:
Works nicely as far as I can tell. Merging!

In the future we might want to add a root_directory config as well so a server root can be in /opt/mycroft or something like that.

PR #2113, opened 2019-05-04
Allow settingsmeta to be a yaml file

ChristopherRogers1991:
## Description
Allows settingsmeta to be a yaml file (I find it easier to work with than json).

## How to test
Use the following with the HomeAssistant skill:

```
identifier: HomeAssistantSkill
name: HomeAssistant
skillMetadata:
  sections:
    - name: Login
      fields:
        - label: Host adress
          name: host
          type: text
          value: ''

        - label: Long-Lived Access Tokens
          name: token
          type: password
          value: ''

        - label: Port number
          name: portnum
          type: Number
          value: 8123

    - name: Options
      fields:
        - label: Use SSL
          name: ssl
          type: checkbox
          value: false

        - label: Verify SSL Certificate
          name: verify
          type: checkbox
          value: true

        - label: Enable conversation component as fallback
          name: enable_fallback
          type: checkbox
          value: true

        - label: Brightness Step
          name: brightness_step
          type: Number
          value: 20

        - label: Temperature Step
          name: temperature_step
          type: Number
          value: 2
```
CLA
 - [x] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

### DIFF
--- a/mycroft/skills/settings.py
+++ b/mycroft/skills/settings.py
@@ -94,7 +94,7 @@ class SkillSettings(dict):
         self.name = name
         # set file paths
         self._settings_path = join(directory, 'settings.json')
-        self._meta_path = self._get_meta_path(directory)
+        self._meta_path = _get_meta_path(directory)
         self.is_alive = True
         self.loaded_hash = hash(json.dumps(self, sort_keys=True))
         self._complete_intialization = False
@@ -112,16 +112,6 @@ class SkillSettings(dict):
     def __hash__(self):
         """ Simple object unique hash. """
         return hash(str(id(self)) + self.name)
-
-    @staticmethod
-    def _get_meta_path(base_directory):
-        json_path = join(base_directory, 'settingsmeta.json')
-        yaml_path = join(base_directory, 'settingsmeta.yaml')
-        if isfile(json_path):
-            return json_path
-        if isfile(yaml_path):
-            return yaml_path
-        return None
 
     def run_poll(self, _=None):
         """Immediately poll the web for new skill settings"""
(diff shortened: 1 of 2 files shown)


mycroft/skills/settings.py
```diff
@@ -104,13 +106,22 @@ def __init__(self, directory, name):
...
         # if settingsmeta exist
-        if isfile(self._meta_path):
+        if self._meta_path:
             self._poll_skill_settings()
 
     def __hash__(self):
         """ Simple object unique hash. """
         return hash(str(id(self)) + self.name)
 
+    def _get_meta_path(self, base_directory):
```
forslund: This could be moved outside of the class to a function

JarbasAl: i think this "belongs" there, makes code more readable and consistent. o you see that being used elsewhere?

i would just make it a staticmethod

ChristopherRogers1991: I like the static method approach as I also think this has very limited value outside this class, and think it's worth keeping it relatively close to where it's being used. I'll push up a commit with that change made. @forslund, let me know if you disagree with that approach though, and we can discuss further.

forslund: **I'll start with the important parts:**
If a static method feels better for you guys make it so. You have (at least) as much Python experience as me so I shall not dictate code style too much.

_Then I add a rant (feel free to skip reading this):_
I personally hate static methods :) (If you want something that's a basically a function write a function)... But my roots are in C so I like functions.

I don't see it being used outside of the module but I think the module namespace is as good a container as the class for this and since it's not a part of the core functionality of the class it just feels like it's in the way. Since it's only intended to be used from within the settings.py file I feel a function is better since it won't be imported anywhere else along with the Settings class (unless specifically.

Note: Above statements are exaggerated for comic effect but is basically how I feel. I do think static methods can be useful.

ChristopherRogers1991: With only one class in the file, top level or static is more or less the same, I suppose. I don't feel particularly strongly, so I've moved it to the bottom of the file, as a module level function. @forslund, let me know if there's anything else you'd like me to tweak.

forslund:
Looks great now @ChristopherRogers1991 and works great. Merging.

If you got time I think we should update the documentation to suggest yaml as the preferred format.

PR #2115, opened 2019-05-06
Feature/pre_loaded_cache

Ruthvicp:
## Description
Build the cache for default responses of mycroft TTS (mimic2)

## How to test
1. Reboot Mark1 and check for the cache in /opt/mycroft/preloaded_cache/Mimic2 (This is done for the first time)
2. Check if the same cache is present in /tmp/mycroft/cache/tts/Mimic2 
3. Reboot the device - on reboot it first clears the cache in /tmp/mycroft/cache and copies the pre_loaded cache from step 1.
4. The functionality should remain the same on device reset as well

CLA [Yes](Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

### DIFF
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -278,7 +278,8 @@
     },
     "mimic2": {
       "lang": "en-us",
-      "url": "https://mimic-api.mycroft.ai/synthesize?text="
+      "url": "https://mimic-api.mycroft.ai/synthesize?text=",
+      "preloaded_cache": "/opt/mycroft/preloaded_cache/Mimic2"
     },
     "espeak": {
       "lang": "english-us",
@@ -305,5 +306,6 @@
     },
     "default-backend": "local"
   },
+
   "debug": false
 }
--- a/mycroft/tts/cache_handler.py
+++ b/mycroft/tts/cache_handler.py
@@ -1,0 +1,180 @@
+# Copyright 2019 Mycroft AI Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Cache handler - reads all the .dialog files (The default
+mycroft responses) and does a tts inference.
+It then saves the .wav files to mark1 device
+
+"""
+
+import base64
+import glob
+import os
+import re
+import shutil
+import hashlib
+import json
+import mycroft.util as util
+from urllib import parse
+from requests_futures.sessions import FuturesSession
+from mycroft.util.log import LOG
+
+
+REGEX_SPL_CHARS = re.compile(r'[@#$%^*()<>/\|}{~:]')
+MIMIC2_URL = 'https://mimic-api.mycroft.ai/synthesize?text='
+
+# For now we only get the cache for mimic2-kusal
+TTS = 'Mimic2'
+
+# Check for more default dialogs
+res_path = os.path.abspath(os.path.join(os.path.abspath(__file__), '..',
+                                        '..', 'res', 'text', 'en-us'))
+wifi_setup_path = '/usr/local/mycroft/mycroft-wifi-setup/dialog/en-us'
+cache_dialog_path = [res_path, wifi_setup_path]
+
+
+def generate_cache_text(cache_audio_dir, cache_text_file):
+    """
+    This prepares a text file with all the sentences
+    from *.dialog files present in
+    mycroft/res/text/en-us and mycroft-wifi setup skill
+    Args:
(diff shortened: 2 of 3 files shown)


mycroft/audio/cache_handler.py
```diff
@@ -0,0 +1,130 @@
...
+from requests_futures.sessions import FuturesSession
+from mycroft.util.log import LOG
+
+
+REGEX_SPL_CHARS = re.compile('[@#$%^*()<>/\|}{~:]')
+MIMIC2_URL = 'https://mimic-api.mycroft.ai/synthesize?text='
+
+# Check for more default dialogs
+res_path = '/opt/venvs/mycroft-core/lib/python3.4/' \
+           'site-packages/mycroft/res/text/en-us/'
```
forslund: Can this be made more generic with something along the lines of

```python
res_path = normpath(join(dirname(abspath(__file__)),
                         '..', 'res', 'text', 'en-us'))
```

mycroft/audio/cache_handler.py
```diff
@@ -0,0 +1,130 @@
...
+wifi_setup_path = '/usr/local/mycroft/mycroft-wifi-setup/dialog/en-us/'
+cache_dialog_path = [res_path, wifi_setup_path]
+# Path where cache is stored and not cleared on reboot/TTS change
+cache_audio_dir = '/opt/mycroft/preloaded_cache'
```
forslund: This directory should be configurable. In a packaging situation we'd probably like to place it somewhere in /usr/share while on a desktop / github-install /opt/mycroft would be best.

mycroft/audio/cache_handler.py
```diff
@@ -0,0 +1,130 @@
+# Cache handler - reads all the .dialog files (The default
+# mycroft responses) and does a tts inference.
+# It then saves the .wav files to mark1 device
```
forslund: Add license header and maybe make this a docstring

mycroft/audio/__main__.py
```diff
@@ -40,6 +41,13 @@ def main():
     audio = AudioService(bus)  # Connect audio service instance to message bus
     create_daemon(bus.run_forever)
 
+    try:
+        LOG.info("Getting Pre-loaded cache")
+        cache_handler.main()
```
forslund: Since this only applies to Mimic2 maybe this should be moved to be called as the mimic2 tts is being created?

mycroft/audio/cache_handler.py
```diff
@@ -0,0 +1,130 @@
...
+        LOG.info("Completed downloading cache for mimic2")
+    else:
+        LOG.info("Pre-loaded cache already exists")
+
+
+def move_cache():
+    source = cache_audio_dir + '/Mimic2'
+    if os.path.exists(source):
+        # tmp directory where tts cache is stored
+        dest = '/tmp/mycroft/cache/tts/Mimic2'
```
forslund: This should be taken from `get_cache_directory()` (mycroft.util)

mycroft/audio/cache_handler.py
```diff
@@ -0,0 +1,130 @@
...
+                    except Exception:
+                        # For now skip that .pho file and continue
+                        # todo: Add proper log statements
+                        pass
+        LOG.info("Completed downloading cache for mimic2")
+    else:
+        LOG.info("Pre-loaded cache already exists")
+
+
+def move_cache():
```
forslund: Since it's a copy maybe naming it copy_cache()?

Also add docstring

mycroft/audio/cache_handler.py
```diff
@@ -0,0 +1,130 @@
...
+        cache_audio_files = cache_audio_dir + '/Mimic2'
+        os.mkdir(cache_audio_files)
+        session = FuturesSession()
+        with open(cache_text_file, 'r') as fp:
+            all_dialogs = fp.readlines()
+            for each_dialog in all_dialogs:
+                each_dialog = each_dialog.strip()
+                key = str(hashlib.md5(
+                    each_dialog.encode('utf-8', 'ignore')).hexdigest())
+                wav_file = cache_audio_files + '/' + key + '.wav'
```
forslund: `os.path.join()` might be better instead of using `+ '/' +` we don't support other systems as of yet but we shouldn't make it harder.

mycroft/audio/cache_handler.py
```diff
@@ -0,0 +1,130 @@
...
+import re
+import shutil
+import hashlib
+import json
+from urllib import parse
+from requests_futures.sessions import FuturesSession
+from mycroft.util.log import LOG
+
+
+REGEX_SPL_CHARS = re.compile('[@#$%^*()<>/\|}{~:]')
```
forslund: You should use `r'[@#$%^*()<>/\|}{~:]'` (raw string literal) to make sure python doesn't interpret some of this as escaped characters

mycroft/audio/__main__.py
```diff
@@ -40,6 +41,13 @@ def main():
...
+        LOG.info("Successfully downloaded Pre-loaded cache")
+    except:
+        LOG.info("Error in getting the pre-loaded cache")
```
forslund: Give some more info here:

```python
    except Exception as e:
        LOG.error(Could not get the pre-loaded cache ({})".format(repr(e)))
```

Ruthvicp:
Removed the caller from audio and now it is called from mimic2. Also corrected the location in mycroft.conf.

forslund:
Did some minor fixes, will merge when the Autobots are done checking.

PR #2118, opened 2019-05-10
Interaction target

forslund:
## Description
This restore the accidentally deleted branch for allowing clients to specify destinations for the messages. The idea that a client will assign a destination in the message context, for the speech client it will be the "audio" system but web clients or similar things can set an id per connection so the reply is only picked up and output by the correct client. This can also be used for the Android client to return not output the message on the base unit.

## How to test
Make sure mycroft responds to spoken queries. then edit the `mycroft/client/speech/__main__.py` and set the destination to "asdf" and make sure that no audio answer is received.

CLA [ Yes ]

### DIFF
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -50,6 +50,13 @@ def handle_speak(event):
     Configuration.init(bus)
     global _last_stop_signal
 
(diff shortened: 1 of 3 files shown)


forslund:
Closing in favor for #2461

PR #2121, opened 2019-05-14
Feature/System_QML

AIIX:
## Description
Adds support for show_text() and show_image() system generic delegates, all "SYSTEM_" starting filenames are served by core rather than a skill.
Adds proportional delegate based show_text() with proper page wrapping
Adds basic image delegate to show image with titles and captions or without either of the text elements

## How to test
Add following to any skill:
self.gui.show_text("Lorem ipsum dolor sit amet", "standard Lorem passage") // can be displayed with or without title depending on provided values
self.gui.show_image("random_image.jpg", "random caption", "random title") // can be displayed with or without title and caption depending on provided values

CLA [ yes ] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)


mycroft/skills/core.py
```diff
@@ -406,7 +409,7 @@ def show_text(self, text, title=None):
         self.clear()
         self["text"] = text
         self["title"] = title
-        self.show_page("SYSTEM_TEXTFRAME")
+        self.show_page("SYSTEM_TextFrame.qml")
 
     def show_image(self, url, caption=None, title=None):
```
forslund: I think you forgot to add the fill as a parameter here.

```python
def show_image(self, url, caption=None, title=None, fill=None):
```

and then below
```python
        self["fill"] = fill or 'PreserveAspectFit'
```

to give a decent default action.

AIIX: fixed the missing parameter, the fill is by default set to stretch in qml so if there is none it will fallback to default

forslund:
Thanks @aiix, works great! Merging

PR #2124, opened 2019-05-15
Support chopping old audio via precise upgrade

MatthewScholefield:
# Description
This supports dropping audio buffers when the cpu falls behind so that the latency doesn't infinitely increase.

## How to test
 - Make sure precise still activates on a Desktop / Mark 1


forslund:
Seems to work as far as I can determine. Merging.

PR #2128, opened 2019-05-20
add the missing url parameter

### LINKED ISSUE #2127: Missing url Parameter for MaryTTS constructor
In mycroft latest version v19.2.7. is a parameter url missing for the marytts class constructor (like ibm tts, which has also a RemoteTTS baseclass).

missing parameter url:
https://github.com/MycroftAI/mycroft-core/blob/b0157b46c804bc46533fdcf5a9169a40508a9225/mycroft/tts/mary_tts.py#L31

for example ibm tts:
https://github.com/MycroftAI/mycroft-core/blob/b0157b46c804bc46533fdcf5a9169a40508a9225/mycroft/tts/ibm_tts.py#L25

A following error is the wrong parameter syntax to the base class RemoteTTS
https://github.com/MycroftAI/mycroft-core/blob/b0157b46c804bc46533fdcf5a9169a40508a9225/mycroft/tts/mary_tts.py#L32

better would be a init method like this:
`
    def __init__(self, lang, config, url="http://localhost:59125"):
        super(MaryTTS, self).__init__(lang, config, url, '/process', MaryTTSValidator(self))
`

Sorry for the bad code syntax - i tried my best but it doesn't work :(

motom001:
## Description
fixes #2127

## How to test
MaryTTS log show the correct request for /version with startup of mycroft

### DIFF
--- a/mycroft/tts/mary_tts.py
+++ b/mycroft/tts/mary_tts.py
@@ -28,8 +28,8 @@ class MaryTTS(RemoteTTS):
         'OUTPUT_TYPE': 'AUDIO'
     }
 
-    def __init__(self, lang, config):
-        super(MaryTTS, self).__init__(lang, config, '/process',
+    def __init__(self, lang, config, url="http://localhost:59125"):
+        super(MaryTTS, self).__init__(lang, config, url, '/process',
                                       MaryTTSValidator(self))
 
     def build_request_params(self, sentence):


forslund:
Hi thanks for this. I do think the url should be fetched from the config, so the code should be something along the lines of

```python
    def __init__(self, lang, config):
        super(MaryTTS, self).__init__(lang, config, config['url'], '/process',
                                      MaryTTSValidator(self))
```

Possibly the url you're using should be set in the default config as well? Is it the default when running your own instance of MaryTTS?

Also before we can merge your PR we need you to sign the CLA (https://mycroft.ai/cla). This is to protect yourself, Mycroft and the community. This agreement makes it crystal clear that along with your code you are offering a license to use it within the confines of this project. You retain
ownership of the code, this is just a license. Once more thanks for contributing

motom001:
You are right with your note to "config['url']" but this means also, that the second tts from ibm is false:
https://github.com/MycroftAI/mycroft-core/blob/b0157b46c804bc46533fdcf5a9169a40508a9225/mycroft/tts/ibm_tts.py#L26

I took this init as example and add the url to the marytts init method. But this was wrong - also the ibm tts init method should be modify. Now I see in the TTSFactory, that there will be never a parameter url for ibm or mary tts:
https://github.com/MycroftAI/mycroft-core/blob/db06962ca257da5c39db4ce5ee2f3e91ab50db3e/mycroft/tts/__init__.py#L490

I will close this PR and will create a new one, when I correct both tts files.

PR #2130, opened 2019-05-22
Fix copy/paste error in chmod command

penrods:
The command that is supposed to re-enable the mycroft-core hourly cron job was
using "-x" instead of "+x".

## How to test
Run mycroft-use to switch to unstable then back.  The look for the 'x' attribute on the
file /etc/cron.hourly/mycroft-core.  Previously it would be missing, but after this change
it will be restored upon switching back to stable.


forslund:
Good Catch, seems to work nicely.

PR #2132, opened 2019-05-22
Don't create skill settings for Padatious Service

forslund:
## Description
Padatious is registered as a FallbackSkill and was pushing settings to
the backend.

This change allows Padatious to flag to the MycroftSkill class
that no settings should be created. This is done via the new optional
MycroftSkill parameter use_settings.

## How to test
Assert that Padatious intents are still working.

CLA [ Yes ]

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -111,6 +111,8 @@ def load_skill(skill_descriptor, bus, skill_id, BLACKLISTED_SKILLS=None):
         skill_descriptor: descriptor of skill to load
         bus:              Mycroft messagebus connection
         skill_id:         id number for skill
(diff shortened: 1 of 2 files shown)


krisgesling:
Just to confirm, any Skill can be prevented from displaying on the Skill settings page using this flag?

forslund:
Yes, that's true. Not sure if it's a good thing or a bad thing to have available. It would still be listed in the skills manifest so it'll show up on the Marketplace.

penrods:
I'm OK with this for now, although it doesn't feel perfect. Is the main point of this to prevent the appearance of "system-y" things on the web UI skill list?  Or is there actually a problem with Padatious having a SkillSettings object?

PR #2141, opened 2019-05-29
Restored ability to store wakewords locally

penrods:
The code that handles the local save of wake words configuration -
"record_wake_words" - was removed some time ago.  This restores that
capability.

## How to test

Add the mycroft.conf
```json
{
     "record_wake_words": true
}
```
Then restart Mycroft.  It should begin saving .wav files under the
/tmp/mycroft_wake_words directory.

### DIFF
--- a/mycroft/client/speech/mic.py
+++ b/mycroft/client/speech/mic.py
@@ -471,13 +471,13 @@ class ResponsiveRecognizer(speech_recognition.Recognizer):
                     if self.save_wake_words:
                         # Save wake word locally
                         audio = self._create_audio_data(byte_data, source)
-                        metadata = self._compile_metadata()
+                        meta = self._compile_metadata()
                         if not isdir(self.saved_wake_words_dir):
                             os.mkdir(self.saved_wake_words_dir)
                         module = self.wake_word_recognizer.__class__.__name__
 
                         fn = join(self.saved_wake_words_dir,
-                                  '.'.join([v for v in metadata.values()])
+                                  '_'.join([meta[k] for k in sorted(meta)])
                                   + '.wav')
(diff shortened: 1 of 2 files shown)


mycroft/client/speech/mic.py
```diff
@@ -356,14 +362,16 @@ def _upload_wake_word(self, audio):
...
 
-        metadata = {
+        return {
             'name': self.wake_word_name.replace(' ', '-'),
             'engine': md5(ww_module.encode('utf-8')).hexdigest(),
             'time': str(int(1000 * get_time())),
             'sessionId': SessionManager.get().session_id,
             'accountId': self.account_id,
             'model': str(model_hash)
         }
```
forslund: I wonder if you want a simple list here. This currently results in the filename `name.engine.time.sessionId.accountId.model.wav` (join of a dict seem to be joining the keys).

Alternatively you could use the `.values()` but I'm not sure if that will always preserve the order so maybe make it an ordered dict?

penrods: Works better if I hit "Save" in my editor...

I'd already changed that to use values().  I don't honestly care that much about order when reviewing a local copy, so I don't think it matters much

forslund: I personally care (otherwise I wouldn't have mentioned it). It's good style to guarantee consistency and if anyone makes a script for doing any type of analysis on the recorded wakewords or something it'd be good if it can easily parse out the name.

penrods: Fair enough!

forslund:
The grumpy swede is pleased with the consistency. Merging

penrods:
"Bork a bork a bork bork a...."
:)

PR #2149, opened 2019-06-08
Feature/streamingstt

JPEWdev:
## Description
Adds support for live streaming speech to text backends.

## How to test
A DeepSpeech Streaming Speech to Text implementation has been provided. It can be used with https://github.com/JPEWdev/deep-dregs to verify that live streaming STT works.

I'm also planning on adding support for an STT backend that directly uses DeepSpeech's streaming STT support (without a server), but it's not ready yet.

CLA Pending

### DIFF
--- a/mycroft/client/speech/listener.py
+++ b/mycroft/client/speech/listener.py
@@ -32,6 +32,25 @@ from mycroft.util.log import LOG
 from mycroft.util import find_input_device
 from queue import Queue, Empty
 
+AUDIO_DATA = 0
+STREAM_START = 1
+STREAM_DATA = 2
+STREAM_STOP = 3
+
+
+class AudioStreamHandler(object):
+    def __init__(self, queue):
+        self.queue = queue
+
+    def stream_start(self):
+        self.queue.put((STREAM_START, None))
+
+    def stream_chunk(self, chunk):
+        self.queue.put((STREAM_DATA, chunk))
+
+    def stream_stop(self):
+        self.queue.put((STREAM_STOP, None))
+
 
 class AudioProducer(Thread):
     """
@@ -40,7 +59,7 @@ class AudioProducer(Thread):
     mic for potential speech chunks and pushes them onto the queue.
     """
 
-    def __init__(self, state, queue, mic, recognizer, emitter):
+    def __init__(self, state, queue, mic, recognizer, emitter, stream_handler):
         super(AudioProducer, self).__init__()
         self.daemon = True
         self.state = state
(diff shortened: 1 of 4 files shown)


forslund:
Hi, many thanks for this! I think this looks really interesting.

@davidwagnerkc has been working on a similar thing for streaming STT's, I think you two should sync and see that all needed features are present and if the separate works can be combined.

Also can you check on the failing unit tests?

davidwagnerkc:
Hey @JPEWdev great work here! Thanks for your contribution 👏 💯

I fixed the PEP8 complaint from Travis so we should be good to merge as soon as you can sign the CLA. 

I planned on coming in today to put the final touches on my implementation of streaming STT, but saw your PR over the weekend and I liked yours better 😄Great solution to the somewhat tricky problem of squeezing in streaming capability into the existing AudioProducer/Consumer paradigm without creating a giant mess.

I was able to get up and running with your DeepSpeech server and shortly after we merge this PR I will put one up for adding Google Streaming STT. 

Thanks!

JPEWdev:
Excellent, thanks. I thought the line limit was 80 chars, not 79 :)

I just signed the CLA.

The processing requirements for live streaming with DeepSpeech are surprisingly small, I actually suspect you could run the whole stack on a Raspberry Pi now; I might try that.

I also would like to add a direct DeepSpeech backend so that you don't even need the server (just have Mycroft directly access the python library)

davidwagnerkc:
Thanks for doing that quickly. I'll merge after Travis completes again, forgot I spotted a missed rename of `audio` to `message`. 

PEP8 is 79, reserving the 80th spot as the special newline spot 😄 

Running DS on the Pi would be awesome, be sure to let us know how that goes.

forslund:
As @davidwagnerkc said, awesome work. As a follow up it would be great if you could add docstrings for the AudioStreamHandler, the StreamThread and the new methods in STT. The docs hasn't historically been very good but we're trying to improve that.

@davidwagnerkc is that about the NL true? I always thought it was to allow the +- signs when reading diffs?

davidwagnerkc:
@forslund I've always thought of it as 79 + NL, but the diff thing makes more sense. Of course you got me on a Googling adventure and landed on this Reddit thread where the guy apparently got this answer out of Guido: "Because an 80 char wide Emacs window starts wrapping at 79." Interesting!

forslund:
Hah, and now I know. I feel quite enriched :)

kjcole:
emacs, for the win! I'm am of the Church. ;-)

JarbasAl:
awesome work, it would be cool to also add support for kaldi in a follow up PR

kaldi can run on a raspi real time if streaming (if you pass the full audio it takes ~20 secs), accuracy ain't great, but it's an offline STT option, way better than pocketsphinx!

you can see some experiments i did with kaldi here https://github.com/JarbasAl/kaldi_spotter

PR #2154, opened 2019-06-10
Improvement/schedule fractional seconds

ChristopherRogers1991:
## Description
This updates `schedule_event` to allow the `when` argument to be a `float`, which allows for scheduling events at fractions of seconds (e.g. `.5` seconds in the future).

## How to test
I updated the `test_add_scheduled_event` test in `test_core` to use an int and a float (previously only used a datetime).

CLA
 - [x] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

### DIFF
--- a/mycroft/skills/core.py
+++ b/mycroft/skills/core.py
@@ -1563,8 +1563,8 @@ class MycroftSkill:
 
             Args:
                 handler:               method to be called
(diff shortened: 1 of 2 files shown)


forslund:
Looks good, I squashed the pycodestyle correction into the previous commit to keep a clean commit history.

PR #2156, opened 2019-06-10
Remove abstractmethod for MycroftSkill.stop()

penrods:
The MycroftSkill.stop() method does not need to be implemented by the majority of skills, no reason to require its implementation with an abstractmethod decorator.

## How to test
No real behavioral impact.


forslund:
Indeed, I don't even think this has been working as intended because I've run several skills without a stop method.

Merging.

PR #2158, opened 2019-06-11
Add mycroft.util.time.to_system()

penrods:
Pulling method originally implemented in the default Mycroft skill-alarm, but useful by many for
converting a different timezone to whatever the native timezone is on the host machine.

## How to test
Run the skill-alarm, it should pull in this implementation instead of its own.


forslund:
Tested and works as expected. Merging.

PR #2160, opened 2019-06-11
Feature: Google Streaming STT

davidwagnerkc:
## Description
Adds Google Cloud STT and refactors Streaming STT classes so that additional providers can be added easily. 

## How to test
Can test streaming with both DeepSpeech and Google by editing the `mycroft.conf` file. 

### DIFF
--- a/mycroft/stt/__init__.py
+++ b/mycroft/stt/__init__.py
@@ -232,14 +232,27 @@ class StreamingSTT(STT):
     @abstractmethod
     def create_streaming_thread(self):
         pass
+
+class DeepSpeechStreamThread(StreamThread):
+    def __init__(self, queue, language, url):
         if not language.startswith("en"):
             raise ValueError("Deepspeech is currently english only")
+        super().__init__(queue, language)
+        self.url = url
+
+    def handle_audio_stream(self, audio):
+        self.response = post(self.url, data=audio, stream=True)
+        return self.response.text if self.response else None
+
+class DeepSpeechStreamServerSTT(StreamingSTT):
+    def create_streaming_thread(self):
         self.queue = Queue()
-        self.stream = StreamThread(self.config.get("stream_uri"), self.queue)
-        self.stream.start()
-
-
-class GoogleStreamThread(Thread):
+        return DeepSpeechStreamThread(
+            self.queue,
+            self.lang,
+            self.config.get('stream_uri')
+        )
+    
     def __init__(self, queue, lang, client, streaming_config):
         super().__init__()
         LOG.info('WAGNER Thread init')
--- a/mycroft/stt/__init__.py
+++ b/mycroft/stt/__init__.py
@@ -178,6 +178,7 @@ class StreamThread(Thread):
         super().__init__()
         self.language = language
         self.queue = queue
+        self.text = None
 
     def _get_data(self):
         while True:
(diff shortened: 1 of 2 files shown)


davidwagnerkc:
@JarbasAI @JPEWdev This is untested, WIP code 😄Just thought I'd tag you early in case you had any thoughts. 

You both mentioned adding local streaming STT options and so while adding Google streaming STT I refactored a bit to make @JPEWdev 's new streaming classes work as abstract base classes.

To add new streaming STT requires subclassing `StreamThread` and implementing the `handle_audio_stream` function as well as subclassing `StreamingSTT` and implementing the `create_streaming_thread` function. Something like this:

```
class LocalKaldiThread(StreamThread):
    def __init__(self):
        # Pass in what is necessary to handle audio stream
        # from STT object

    def handle_audio_stream(self, audio):
        # Consume the audio generator stream with Kaldi
        # and produce text result

class LocalKaldiStreamingSTT(StreamingSTT):
    def __init__(self):
        # Get Kaldi config, instantiate Kaldi object, etc...

    def create_streaming_thread(self):
        # return a new LocalKaldiStreamThread object
```

JarbasAl:
i will look into this and add a kaldi option, nice work!

JPEWdev:
FWIW, I don't think that mandating a thread for every streaming backend is necessary. I only did t that way for the deepspeech server because it uses HTTP chunk transfers, and using a thread and generator was the most expedient way to send that.

There are many other ways of streaming that would not require a thread, for example a session based HTTP server, or the local deepspeech backend I'm currently working on that directly calls into deepspeech without need for a server

TL;DR I'm fine with API class to make it easy to add a thread for streaming STT, but I think it is a mistake to make it mandatory

davidwagnerkc:
@JPEWdev Think we are on the same page. By saying "To add new streaming STT requires" I only meant it in the sense of fulfilling the requirements of the abstract class, not as a mandatory requirement for all future implementations. Open to new classes for new methods of streaming and/or refactoring the `StreamingSTT` class, which I am sure will happen. I like how you implemented it though and think it could make streaming support fairly easy for a number of use cases. 

Apparently forget to hit the send button on this one yestready 😆 Taking a look at your new PR now.

devs-mycroft:
Hello, @davidwagnerkc, thank you for helping with the Mycroft project! We welcome everyone
for Everyone.

you!

davidwagnerkc:
Yeah I had never heard of HTTP chunk protocol, learning something new everyday! Google uses gRPC with protocol buffers which was a fun one to learn about last week. Of course the Google client library encapsulates all the details of that so caller does not have to worry about it.

I'll play around with removing the thread today and maybe update the PR (_@JarbasAI Don't get too far on that Kaldi addition! : ) Thoughts on thread vs no thread?_ ) The Google client takes an `Iterable` of `StreamingRecognitionRequests` so I could probably use the `_get_data()` generator directly inside of the `stream_data()` function and consume it there. Could be nice if we could get rid of the need for an additional `StreamThread` class for each new STT option. I haven't worked with generators or async problems generally too much so appreciate the feedback. Worked with threads more so they make more sense to me.

davidwagnerkc:
@JPEWdev For now I think we are going to keep the thread, if only for your DS server and my Google Cloud addition. I think if you do the gRPC calls yourself the `stream_start`, `stream_data`, `stream_stop` interface would work well without a thread, but as far as I can tell a thread works better for the API Google provided (very possible I am wrong on that, but I need to move on to other tasks for now)

Totally fine if other additions like your local implementation do not use the `StreamingSTT` class or if we determine we need a refactor at a later date. Appreciate your feedback :+1:

mycroft/stt/__init__.py
```diff
@@ -179,11 +170,16 @@ def execute(self, audio, language=None):
 
 
 class StreamThread(Thread):
-    def __init__(self, url, queue):
+    """
+        ABC class to be used with StreamingSTT class implementations.
+    """
+    __metaclass__ = ABCMeta
```
forslund: Should we maybe move to the new-style meta class declaration?

```python
class StreamThread(Thread, metaclass=ABCMeta):
```

forslund:
Nevermind, seems to be all over the place. I'll make a PR fixing it globally later. Merging

PR #2161, opened 2019-06-12
Add local DeepSpeech STT

JPEWdev:
## Description
Adds a Speech to Text backend that directly interfaces with the
deepspeech streaming API.

## How to test
Edit mycroft.conf to use a local deepspeech server (a DeepSpeech model will need to be downloaded). For example:
```
{
  "stt": {
    "module": "deepspeech",
    "deepspeech": {
      "model": "/home/jwatt/DeepSpeech/models/output_graph.pbmm",
      "alphabet": "/home/jwatt/DeepSpeech/models/alphabet.txt",
      "lm": "/home/jwatt/DeepSpeech/models/lm.binary",
      "trie": "/home/jwatt/DeepSpeech/models/trie"
    }
}
```
Using a memory-mapped model is highly recommended.

Yes

### DIFF
--- a/mycroft/stt/__init__.py
+++ b/mycroft/stt/__init__.py
@@ -19,6 +19,9 @@ from requests import post, put, exceptions
 from speech_recognition import Recognizer
 from queue import Queue
 from threading import Thread
+import deepspeech
+import numpy
+import time
 
 from mycroft.api import STTApi
 from mycroft.configuration import Configuration
@@ -240,6 +243,53 @@ class DeepSpeechStreamServerSTT(DeepSpeechServerSTT):
         self.stream.start()
 
 
+class DeepSpeechSTT(STT):
+    def __init__(self):
+        super().__init__()
+
+        start_time = time.perf_counter()
+        LOG.info("Loading DeepSpeech model...")
+
+        model = self.config['model']
+        alphabet = self.config['alphabet']
+        num_context = self.config.get('num_context', 9)
+        beam_width = self.config.get('beam_width', 512)
+        num_features = self.config.get('num_features', 26)
+        lm = self.config.get('lm')
+        trie = self.config.get('trie')
+
+        self.model = deepspeech.Model(model, num_features, num_context,
+                                      alphabet, beam_width)
+
+        if lm is not None and trie is not None:
+            lm_weight = self.config.get('lm_weight', 1.5)
+            vwcw = self.config.get('valid_word_count_weight', 2.25)
+
+            self.model.enableDecoderWithLM(alphabet, lm, trie, lm_weight, vwcw)
+
+        LOG.info("Loaded DeepSpeech model in %0.3fs" % (time.perf_counter() -
+                                                        start_time))
+        self.stream_ctx = None
+        self.can_stream = True
+
+    def execute(self, audio, language=None):
+        text = self.model.finishStream(self.stream_ctx)
+        self.stream_ctx = None
+        LOG.info(text)
+        return text
+
+    def stream_start(self, language=None):
+        self.stream_ctx = self.model.setupStream()
+
+    def stream_data(self, data):
+        self.model.feedAudioContent(
+            self.stream_ctx, numpy.frombuffer(data, numpy.int16))
+
+    def stream_stop(self):
+        LOG.info("stop")
+        self.stream_ctx = None
+
+
 class KaldiSTT(STT):
     def __init__(self):
         super(KaldiSTT, self).__init__()
@@ -309,6 +359,7 @@ class STTFactory:
         "houndify": HoundifySTT,
         "deepspeech_server": DeepSpeechServerSTT,
         "deepspeech_stream_server": DeepSpeechStreamServerSTT,
+        "deepspeech": DeepSpeechSTT,
         "mycroft_deepspeech": MycroftDeepSpeechSTT
     }
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,6 +22,8 @@ pulsectl==17.7.4
 google-api-python-client==1.6.4
 fasteners==0.14.1
 PyYAML==3.13
+deepspeech==0.4.1
+numpy==1.16.4
 
 msm==0.7.6
 msk==0.3.13


forslund:
Hi @JPEWdev Thanks for this. However I'm of the opinion that the STT shouldn't be part of mycroft-core like this. I think it's better if a server or STT is setup outside of core and a stream or socket is used to send the data over. This helps with packaging (package size and build-time, possible conflicts between module versions) and keeps mycroft-core more of a "hub" software.

What is the main benefits with a direct implementation in core like this compared with a local deepspeech server which is streamed to?

This is definitely open for discussion, as stated above this is my opinion and @pendrods, @davidwagnerkc and @Ruthvicp may have other thoughts on the matter.

I can see a couple of different approaches as alternatives:
- A skill that sets up a deepspeech server and changes the config to use the Streaming deepspeech STT, model and parameters could be selected using skill settings

- A plugin system where a python module (from github or pypi) can be defined as a STT (I showed a simple version of this for TTS last year), it would functionally work like this one but the requirements wouldn't need to be included in the mycroft-core package but can easily be installed.

forslund:
Update: The build time is only increased by 50% to ~30 Minutes so that's not too bad and the additional dependencies only increases the package size with only 10 MB so that's not too big of a deal. I however do think it's better to keep deepspeech separate from mycroft-core from an organizational point of view.

Again, this is only my opinion and this is no big issue practically since we can keep the models outside of the core package

forslund:
Hmm, did a proper testing and it seems there's something missing when packaged for the mark-1:
```
    import deepspeech
  File "/opt/venvs/mycroft-core/lib/python3.4/site-packages/deepspeech/__init__.py", line 4, in <module>
    from deepspeech.impl import AudioToInputVector as audioToInputVector
  File "/opt/venvs/mycroft-core/lib/python3.4/site-packages/deepspeech/impl.py", line 17, in <module>
    _impl = swig_import_helper()
  File "/opt/venvs/mycroft-core/lib/python3.4/site-packages/deepspeech/impl.py", line 16, in swig_import_helper
    return importlib.import_module('_impl')
  File "/opt/venvs/mycroft-core/lib/python3.4/importlib/__init__.py", line 109, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
ImportError: No module named '_impl'
```

davidwagnerkc:
@forslund Definitely do not want to add more dependencies to core for every STT option. I actually added a package in my Google streaming STT PR to the `requirements.txt`, but was planning on removing it since it is just a handful of users who might set that up and they are going to be capable enough to install a package. I was thinking with some form of lazy loading the imports or even try/except, but your suggestions are good options as well. 

Long term though I think if we can make packaging work it would be cool to included a local STT option like DeepSpeech that would work out of the box for users. Maybe not something for this PR though. 

Those errors are with SWIG which is provides Python bindings to C++ code. More likely some sort of mismatch of the package to the platform arch.

davidwagnerkc:
@JPEWdev You are running this on desktop right? How is the setup working so far? 

Did you see the new 5.0 release? https://github.com/mozilla/DeepSpeech/releases/tag/v0.5.0
I haven't play with a mmap model yet as you suggest, but look forward to playing with that and TFLite model.

JPEWdev:
It works really well on my laptop (the same one I used to do the performance analysis of the server). My laptop easily handles the stream in real time, and Mycroft doesn't appear to do anything CPU intense while recording an utterance, so it works out well.

I didn't see the 0.5 release... It was only 21 hours ago ;) I'll try it first chance I get

As far as packaging deepspeech with Mycroft.... I'm not sure the best route. I suspect that the target here would be deeply embedded systems without internet (I'm going to try on a raspberry pi here when I get a chance), or people who don't want to be bothered to set up a deepspeech server. In those cases, it would be nice if it could work "out of the box". I don't know the best way to facilitate that

JPEWdev:
Also I should note that you still have to download the deepspeech model which is several GB, so it might be difficult to get a 100% "out of the box" experience anyway since I doubt you want to package that with Mycroft ;)

forslund:
So one possibility is to move the import of deepspeech and numpy into the `__init__` of the DeepspeechSTT class. That way it would only be imported if it is selected. We could also add a deepspeech_setup.sh script to install the requirements and modify the config.

The plugin thing I played with for tts can be seen here https://github.com/forslund/mycroft-core/commit/716ca26922352e4b21fbc30186597a78ea084a8b. And a plugin module would look like this: https://github.com/forslund/mycroft-tts-plugin-gtts

JPEWdev:
Ya, I originally had that. I'll change it back

forslund:
I did a test implementation using the plugin approach in the [feature/tts-stt-plugin](https://github.com/MycroftAI/mycroft-core/tree/feature/tts-stt-plugin) branch which can be tested with a [plugin](https://github.com/forslund/mycroft-stt-plugin-deepspeech) based on the Deepspeech stt by @JPEWdev on desktop devices.
Install the plugin using pip into the mycroft venv:
`./bin/mycroft-pip install git+https://github.com/forslund/mycroft-stt-plugin-deepspeech`

Edit the config like the description above but the module is now called `mycroft_stt_plugin_deepspeech`

```json
{
  "stt": {
    "module": "mycroft_stt_plugin_deepspeech",
    "mycroft_stt_plugin_deepspeech": {
      "model": "/home/jwatt/DeepSpeech/models/output_graph.pbmm",
      "alphabet": "/home/jwatt/DeepSpeech/models/alphabet.txt",
      "lm": "/home/jwatt/DeepSpeech/models/lm.binary",
      "trie": "/home/jwatt/DeepSpeech/models/trie"
    }
}

krisgesling:
Hi there, we're planning to add a [plugin system for audio backends, STT and TTS services](https://github.com/MycroftAI/mycroft-core/pull/2594). 

This system will allow users (or developers configuring Mycroft for use in other projects) to pull in these services as they're needed rather than having them all included by default on every Mycroft installation. The aim is to keep mycroft-core lighter but still provide the same level of features and customization through the provision of these plugins.

So I'm going to close this PR but only because it won't need to be included in mycroft-core. We'll be transitioning all of these out to be independent plugins tracked via #2701. 

Thanks

Issue #1839, opened 2018-10-08
dev setup recommends wrong manual install


JarbasAl:
if package manager is missing dev setup tells you to install python2 packages, needs update for python3

```
echo "${green}Could not find package manager"
        echo "${green}Make sure to manually install:${blue} git python 2 python-setuptools python-virtualenv pygobject virtualenvwrapper libtool libffi openssl autoconf bison swig glib2.0 portaudio19 mpg123 flac curl fann g++"
        
```

forslund:
dev_setup.sh has been updated with (more) correct packges

PR #2165, opened 2019-06-17
Feature/better os detection

### LINKED ISSUE #2064: Picroft Wakeup word and Temperature Error
PIcroft is taking wakeup word again and again even though I am Not Saying it. And It starts Loops of it. And It also Is'nt saying Correct Temperature of place even though i have set the location through web. And When I try to add New skill through web It is Showing Connection Error than Key Error and Etc. on a Loop.

forslund:
## Description
Use more of [/etc/os-release](http://0pointer.de/blog/projects/os-release.html) to detect debian like systems and Fedora versions.

Resolves #2064 #1839 

## How to test
Ensure dev_setup.sh still invokes apt.

CLA [ Yes ]

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -279,17 +279,17 @@ function install_deps() {
     echo "Installing packages..."
     if found_exe zypper ; then
         # OpenSUSE
-        echo ${YELLOW} Installing packages for OpenSUSE...${RESET}
+        echo ${GREEN} Installing packages for OpenSUSE...${RESET}
         $SUDO zypper install -y git python3 python3-devel libtool libffi-devel libopenssl-devel autoconf automake bison swig portaudio-devel mpg123 flac curl libicu-devel pkg-config libjpeg-devel libfann-devel python3-curses pulseaudio
         $SUDO zypper install -y -t pattern devel_C_C++
     elif found_exe yum && os_is centos ; then
         # CentOS
-        echo ${YELLOW} Installing packages for Centos...${RESET}
+        echo ${GREEN} Installing packages for Centos...${RESET}
         $SUDO yum install epel-release
         redhat_common_install
     elif found_exe yum && os_is rhel ; then
         # Redhat Enterprise Linux
-        echo ${YELLOW} Installing packages for Red Hat...${RESET}
+        echo ${GREEN} Installing packages for Red Hat...${RESET}
         $SUDO yum install -y wget
         wget https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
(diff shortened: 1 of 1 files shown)


dev_setup.sh
```diff
@@ -301,13 +314,10 @@ function install_deps() {
...
             rm -rf fann
         )
-    elif found_exe dnf ; then
-        # Fedora
-        $SUDO dnf install -y git python3 python3-devel python3-pip python3-setuptools python3-virtualenv pygobject3-devel libtool libffi-devel openssl-devel autoconf bison swig glib2-devel portaudio-devel mpg123 mpg123-plugins-pulseaudio screen curl pkgconfig libicu-devel automake libjpeg-turbo-devel fann-devel gcc-c++ redhat-rpm-config jq
     else
     	echo
         echo "${GREEN}Could not find package manager"
-        echo "${GREEN}Make sure to manually install:${BLUE} git python 2 python-setuptools python-virtualenv pygobject virtualenvwrapper libtool libffi openssl autoconf bison swig glib2.0 portaudio19 mpg123 flac curl fann g++"
+        echo "${GREEN}Make sure to manually install:${BLUE} git python3 python-setuptools python-venv pygobject libtool libffi libjpg openssl autoconf bison swig glib2.0 portaudio19 mpg123 flac curl fann g++ jq"
```
penrods: Green seems like an odd color to use here.  I usually consider green to be an "OK" and nothing needs to be done.  Yellow means you might need to do something as a warning.

How about making this message YELLOW instead of GREEN.  Then change the above messages that show which either BLUE or GREEN? (as noted above)

dev_setup.sh
```diff
@@ -275,24 +279,33 @@ function install_deps() {
     echo "Installing packages..."
     if found_exe zypper ; then
         # OpenSUSE
+        echo ${YELLOW} Installing packages for OpenSUSE...${RESET}
         $SUDO zypper install -y git python3 python3-devel libtool libffi-devel libopenssl-devel autoconf automake bison swig portaudio-devel mpg123 flac curl libicu-devel pkg-config libjpeg-devel libfann-devel python3-curses pulseaudio
         $SUDO zypper install -y -t pattern devel_C_C++
     elif found_exe yum && os_is centos ; then
         # CentOS
+        echo ${YELLOW} Installing packages for Centos...${RESET}
```
penrods: Maybe BLUE or GREEN?  See comment below.

forslund: Updated to green, the instructions if no package manager was found is updated to Yellow while highlighting the packages in Blue as previous.

PR #2173, opened 2019-06-26
Fix debian detection

forslund:
## Description
Debian is not like Debian, Debian is Debian. Check os_is as well to determine debianess.

CLA [ yes ]


forslund:
Doing a quick fix to allow installation on debian systems. Tested on ubuntu and debian now. Self-merging

PR #2177, opened 2019-06-29
Remove incorrect settings meta upload disable flag

forslund:
## Description
This allows core to retry sending settings meta after a failed attempt. The flag stopping upload must have been added in mistake when implementing the hold-off exception. I can't see any direct use for that flag today, though when I implemented it I must have had some thought behind it. (can't figure out what thought though)

CLA [ Yes ]


forslund:
Self-merging

PR #2180, opened 2019-06-30
Make is_paired fail early if no uuid in identity

forslund:
## Description
Refactors code so no request is sent to backend if the local identity
file is empty / has no uuid. Since if this part is missing core already
knows that it's not paired correctly.

## How to test
Check that an unpaired device launches the pairing sequence as intended

CLA [ Yes ]

### DIFF
--- a/mycroft/api/__init__.py
+++ b/mycroft/api/__init__.py
@@ -482,7 +482,7 @@ def has_been_paired():
 
 
 def is_paired(ignore_errors=True):
-    """ Determine if this device is actively paired with a web backend
+    """Determine if this device is actively paired with a web backend
 
     Determines if the installation of Mycroft has been paired by the user
     with the backend system, and if that pairing is still active.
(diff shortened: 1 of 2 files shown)


mycroft/api/__init__.py
```diff
@@ -497,19 +497,33 @@ def is_paired(ignore_errors=True):
         # The Mark 1 does perform a restart on RESET.
         return True
 
+    api = DeviceApi()
```
chrisveilleux: this does not appear to be needed here.

mycroft/api/__init__.py
```diff
@@ -497,19 +497,33 @@ def is_paired(ignore_errors=True):
...
+    _paired_cache = api.identity.uuid and check_remote_pairing(ignore_errors)
+
+    return _paired_cache
+
+
+def check_remote_pairing(ignore_errors):
+    """ Check that a basic backend endpoint accepts our pairing.
```
chrisveilleux: I will get you to follow PEP257 yet! :)

mycroft/api/__init__.py
```diff
@@ -497,19 +497,33 @@ def is_paired(ignore_errors=True):
...
+
+    Arguments:
+        ignore_errors (bool): True if errors should be ignored when
+
+    Returns: True if pairing is checks out, otherwise False.
```
chrisveilleux: usually we have a newline after `Returns: `.  consistency is our friend!

mycroft/api/__init__.py
```diff
@@ -497,19 +497,33 @@ def is_paired(ignore_errors=True):
...
         if e.response.status_code == 401:
             return False
     except Exception as e:
         LOG.warning('Could not get device info: ' + repr(e))
-    if ignore_errors:
-        return False
-    if connected():
-        raise BackendDown
-    raise InternetDown
+        if ignore_errors:
```
chrisveilleux: I am not sure that moving these lines inside the exception block is a good idea.  if the backend throws a 404 or 500 error, that would trigger the HTTPError exception and these lines of code would not be executed.

test/unittests/api/test_api.py
```diff
@@ -351,7 +351,7 @@ def test_is_paired_true(self, mock_request, mock_identity_get):
     @mock.patch('mycroft.api._paired_cache', False)
     @mock.patch('mycroft.api.IdentityManager.get')
     @mock.patch('mycroft.api.requests.request')
-    def test_is_paired_false(self, mock_request, mock_identity_get):
+    def test_is_paired_false_local(self, mock_request, mock_identity_get):
```
chrisveilleux: does this test pass?  I thought you had to put an argument in the method definition for each patch.  same for the other test.  also, if you find yourself repeating a bunch of mocks like this, you can define them in setup and kill them in teardown.

mycroft/api/__init__.py
forslund: It's used on the line bellow?

chrisveilleux: must not have been wearing my glasses when I made that comment! :(

test/unittests/api/test_api.py
forslund: It does pass, I can't quite remember but I think it's because the _paired_cache one is set to False directly in the decorator.

I'll do some refactoring moving common mocks into a setup/teardown...Might need to restructure a bit to get it sane though

mycroft/api/__init__.py
forslund: Good catch

forslund: I've updated the section a bit now. Not sure if it makes things worse or better :)

PR #2182, opened 2019-07-01
Remove 'update' field from skills data

forslund:
## Description
Filter out update field from skills data when sending skills manifest
since backend will reject fields with unexpected data

CLA [ Yes ]


forslund:
Self-merging

PR #2184, opened 2019-07-02
'next {day}' to be 3-9 days in future

krisgesling:
## Description
The "next" instance of a day or weekend is considered to be no earlier than 48 hours in the future. This means that the next day will never be outside of 3-9 days in the future.
- On Friday, "next Monday" would be in 3 days. 
- On Saturday, "next Monday" would be in 9 days.

## How to test
Unit tests updated, or ask Mycroft:
- what is the date next Wednesday/Friday/Sunday

- [x] CLA

### DIFF
--- a/mycroft/util/parse.py
+++ b/mycroft/util/parse.py
@@ -14,15 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
-"""
-The mycroft.util.parse module provides various parsing functions for
-things like numbers, times, durations etc.
-
-The focus of these parsing functions is to extract data from natural speech
-and to allow localization.
-"""
-
 from difflib import SequenceMatcher
 from mycroft.util.time import now_local
 from mycroft.util.lang import get_primary_lang_code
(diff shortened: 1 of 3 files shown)


mycroft/util/parse.py
```diff
@@ -214,6 +205,10 @@ def extract_datetime(text, anchorDate=None, lang=None, default_time=None):
     common ways that humans express dates and times, including relative dates
     like "5 days from today", "tomorrow', and "Tuesday".
 
+    The "next" instance of a day or weekend is considered to be no earlier than
+    48 hours in the future. On Friday, "next Monday" would be in 3 days.
+    On Saturday, "next Monday" would be in 9 days.
+
```
forslund: Should this be added to the general functions description / specification. This is as far as I can tell something that's different in different cultures / languages and might be better left unspecified in the general function

mycroft/util/parse.py
```diff
@@ -14,15 +14,6 @@
...
 #
-
-"""
-The mycroft.util.parse module provides various parsing functions for
-things like numbers, times, durations etc.
-
-The focus of these parsing functions is to extract data from natural speech
-and to allow localization.
-"""
-
```
forslund: Why is this module level docstring removed?

krisgesling: Looks like I didn't fully rebase somehow, thanks for catching it

JarbasAl:
i like this, however it is always ambiguous, even native speakers often ask for clarification, each language implementation would have to judge if this definition make sense, in portuguese it does

forslund:
The majority has spoken. Merging

PR #2187, opened 2019-07-04
Fix parsing dates in utterances that include non-month 'may'

krisgesling:
## Description
If an utterance includes the term "may" it is considered a date. However this may not be the month of May. 

Have proposed one possible solution but there are likely to be other cases I'm not considering.

## How to test
Unit tests included

- [x] CLA

### DIFF
--- a/mycroft/util/lang/parse_en.py
+++ b/mycroft/util/lang/parse_en.py
@@ -929,7 +929,9 @@ def extract_datetime_en(string, dateNow, default_time):
                     hasYear = False
 
             # if no date indicators found, it may not be the month of May
-            elif word == 'may' and wordNext == 'i':
+            # may "i/we" ...
+            # "... may be"
+            elif word == 'may' and wordNext in ['i', 'we', 'be']:
                 datestr = ""
 
         # parse 5 days from tomorrow, 10 weeks from next thursday,
--- a/mycroft/util/lang/parse_en.py
+++ b/mycroft/util/lang/parse_en.py
@@ -927,6 +927,11 @@ def extract_datetime_en(string, dateNow, default_time):
                     hasYear = True
                 else:
                     hasYear = False
+
+            # if no date indicators found, it may not be the month of May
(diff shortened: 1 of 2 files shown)


mycroft/util/lang/parse_en.py
```diff
@@ -927,6 +927,11 @@ def date_found():
                     hasYear = True
                 else:
                     hasYear = False
+
+            # if no date indicators found, it may not be the month of May
+            elif word == 'may' and wordNext == 'i':
```
forslund: Should we make this a list so this becomes

```
elif word == 'may' and wordNext in ['i', 'we', 'be']
```

covering *may we...*, *... as it may be* as well as the i case.

krisgesling:
Yeah nice, I may also add a check on prevWord for "I may", "we may" type phrases.

forslund:
Good idea.

Not related to this PR this whole thing looks a bit dodgy. I'm getting a ValueError parsing the string if there's a month but no day number.

"let's plan to do that in june"...

krisgesling:
Yeah that's how I got here in the first place. Was getting a ValueError on "May I know the time"

I was also considering what should happen if it is genuinely a month reference, but no day was specified like your example. 

"How long until June" > 1 June 2020
"Let's do that in June" > reminder on 1 June useful, if it's a deadline probably 30 June
"It has been like that since June" > presumably 30 June to be safe

My feeling is that the best option is to always provide the 1st of the month and it's up to the developer to decide if they want to manually manipulate that?

forslund:
Yes, for now I think it should always be the first of the month similarly to "tomorrow" giving you midnight the following date.

I was pondering we could add a precision field as well in the return...If only a month was found precision would be a Timedelta of 1 month. If a date was found the precision would be a timedelta of 1 day and so on...

But that is definitely outside of the scope for this PR :)

forslund:
Added some of the discussed changes. Merging this for now and we can iterate upon it.

PR #2192, opened 2019-07-06
Fetch CLI log path from config if available

forslund:
## Description
Make log location configurable for CLI.

## How to test
Set the config parameter "log_dir" to something and ensure that the CLI doesn't show anything.

CLA [ Yes ]

### DIFF
--- a/mycroft/client/text/__main__.py
+++ b/mycroft/client/text/__main__.py
@@ -21,6 +21,7 @@ from .text_client import (
         load_settings, save_settings, simple_cli, gui_main,
         start_log_monitor, start_mic_monitor, connect_to_mycroft
     )
+from mycroft.configuration import Configuration
 
 sys.stdout = io.StringIO()
 sys.stderr = io.StringIO()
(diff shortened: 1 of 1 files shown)


mycroft/client/text/__main__.py
```diff
@@ -38,11 +39,17 @@ def custom_except_hook(exctype, value, traceback):
...
+    config = Configuration.get()
+    if 'log_dir' in config:
+        log_dir = os.path.expanduser(config['log_dir'])
+        start_log_monitor(os.path.join(log_dir, 'skills.log'))
+        start_log_monitor(os.path.join(log_dir, 'voice.log'))
+    else:
+        start_log_monitor("/var/log/mycroft/skills.log")
+        start_log_monitor("/var/log/mycroft/voice.log")
+        # logs when using Debian package   TODO: Unify all
+        start_log_monitor("/var/log/mycroft-skills.log")
```
penrods: Does the Debian package write to /var/log/mycroft-skills.log anymore?  I don't have a Mark 1 up at the moment to check., but I'm pretty sure it doesn't.   If not, let's remove support for those old logs.

forslund: You're right, those aren't needed anymore. Mark-1 switched a long time ago.

forslund: PR updated

PR #2197, opened 2019-07-08
Skills process startup speedup

forslund:
## Description
Try to speed up the skills process startup.
- Run msm update after skills are loaded with a reduced thread count
- Fetch skill settings in parallel with rest of the skill loading
- Ensure that skill manifest is always uploaded as early as possible

This depends on a pending change in msm to allow reducing the number of threads used by the msm.apply() method.

I'm also testing how it behaves when loading skills in parallel using a thread pool.

## How to test
Assert the skills are loaded and updated as normal, with any luck some of the blocking calls have been removed from the critical path and a speedup can be seen. 

CLA [ Yes ]

### DIFF
--- a/mycroft/skills/__main__.py
+++ b/mycroft/skills/__main__.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 #
 import time
+import datetime as dt
 from threading import Timer
 import mycroft.lock
 from mycroft import dialog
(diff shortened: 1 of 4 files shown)


forslund:
Msm has been released with the max_thread support so this can now be tested. Suggestions for other improvements are welcome.

PR #2201, opened 2019-07-11
Explain the config system used for the log level.

forslund:
## Description

@davidwagnerkc was a bit confused about the log_level setting (totally justified confusion), this adds some more documenation around the logger.

- Add module docstring
- Add comment in the init class method about setting up the log level
- Add mycroft.util.log to the readthedocs documentation
- Add comment in default log

## How to test
Read the docstrings and make sure they make sense.

CLA [ Yes ]

### DIFF
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -237,7 +237,11 @@
   },
 
   // Level of logs to store, one of  "CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG"
-  "log_level": "DEBUG",
+  // NOTE: This configuration setting is special and can only be changed in the
+  // SYSTEM or USER configuration file, it will not be read if defined in the
+  // DEFAULT (here) or in the REMOTE mycroft config.
+  // If not defined, the default log level is INFO.
+  //"log_level": "INFO",
 
   // Messagebus types that will NOT be output to logs
   "ignore_logs": ["enclosure.mouth.viseme", "enclosure.mouth.display"],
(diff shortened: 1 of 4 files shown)


mycroft/configuration/mycroft.conf
```diff
@@ -237,7 +237,9 @@
   },
 
   // Level of logs to store, one of  "CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG"
-  "log_level": "DEBUG",
+  // Must be set in the system or user config, this or remote will not affekt
```
penrods: Two things:
affekt -> affect

If this is not specified in any .conf file, what will the level be?  Looking at the code below it appears the default is INFO, but that isn't totally clear.  How about this text:

// Level of logs to store, one of  "CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG"
// NOTE:  This configuration setting is special and can only be changed in the
// SYSTEM or USER configuration file, it will not be read if defined in the
// DEFAULT (here) or in the REMOTE mycroft.conf settings.
// If not defined, the default log level is INFO.

mycroft/util/log.py
```diff
@@ -62,6 +77,13 @@ class LOG:
 
     @classmethod
     def init(cls):
+        """ Initializes the class, sets the default log level and creates
+        the required handlers.
+        """
+
+        # Check configs manually, the mycroft configuration system can't be
```
penrods: Capitalize "Mycroft"

forslund:
Updated!

PR #2221, opened 2019-07-19
Split mycroft.skills.core

forslund:
## Description
- MycroftSkill and related core functions to mycroft_skill.py
- FallbackSkill to fallback_skill.py
- Add important classes decorators and functions to __init__.py

core.py retains the same members as previously by means of imports to
retain backwards compatibility. When most of the available skills starts
using this new structure properly.

This makes some of the code easier to view. the mycroft_skill.py still weighs in at ~1400 LOC and we should work at splitting that out in more logical components. I have a suggestion as WIP using a couple of member classes, but the initial split-up is a good first step.

## How to test
Make sure that existing skills load and executes like before.

CLA [ Yes ]


penrods:
I definitely like this reorg!

mycroft/skills/fallback_skill.py
```diff
@@ -0,0 +1,158 @@
...
+        Multiple fallbacks can exist at the same priority, but no order is
+        guaranteed.
+
+        A Fallback can either observe or consume an utterance. A consumed
+        utterance will not be see by any other Fallback handlers.
+    """
+    fallback_handlers = {}
+
+    def __init__(self, name=None, bus=None, use_settings=True):
+        MycroftSkill.__init__(self, name, bus, use_settings)
```
chrisveilleux: just use super() here?

mycroft/skills/fallback_skill.py
```diff
@@ -0,0 +1,158 @@
...
+                self.make_active()
+                return True
+            return False
+
+        self.instance_fallback_handlers.append(wrapper)
+        self._register_fallback(wrapper, priority)
+
+    @classmethod
+    def remove_fallback(cls, handler_to_del):
+        """
```
chrisveilleux: PEP257?

mycroft/enclosure/gui.py
```diff
@@ -0,0 +1,248 @@
...
+
+        # Convert pages to full reference
+        page_urls = []
+        for name in page_names:
+            if name.startswith("SYSTEM"):
+                page = resolve_resource_file(join('ui', name))
+            else:
+                page = self.skill.find_resource(name, 'ui')
+            if page:
+                if self.config.get('remote') is True:
```
chrisveilleux: shouldn't need the `is True` here

mycroft/enclosure/gui.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def show_html(self, html, resource_url=None, override_idle=None):
+        """ Display an HTML page in the GUI
+
+        Arguments:
+            html (str): HTML text to display
+            resource_url (str): Pointer to HTML resources
+        """
+        self.clear()
+        self["html"] = html
```
chrisveilleux: I have never seen the self['name'] notation before.  what does this do?  is it different than self.name?

mycroft/enclosure/gui.py
```diff
@@ -0,0 +1,248 @@
+from os.path import join
```
chrisveilleux: license?

mycroft/skills/__init__.py
```diff
@@ -11,3 +11,27 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
```
chrisveilleux: if we are going to refactor, should we consider moving code out of __init__.py?

mycroft/skills/fallback_skill.py
```diff
@@ -0,0 +1,158 @@
+# Copyright 2017 Mycroft AI Inc.
```
chrisveilleux: 2019

mycroft/enclosure/gui.py
forslund: This class implements the `__setitem__()` magic method to access a member dictionary and send over the bus. using self['name'] will trigger the aforementioned method.

mycroft/skills/__init__.py
forslund: This `__init__.py` doesn't contain any code? Any more would definitely be outside of the scope of this PR since it focus on splitting core.py into logical pieces preparing for futurue refactoring on the MycroftSkill class.

mycroft/enclosure/gui.py
forslund: Good catch, always forgetting this :)

forslund:
My original intent was just to move things around into new locations but I can do some additional refactoring as you've suggested.

mycroft/skills/fallback_skill.py
forslund: Addressed.

mycroft/enclosure/gui.py
chrisveilleux: Interesting.  Not sure I am a fan of this implementation, but that is a problem for another day.

PR #2224, opened 2019-07-22
Remove time based forced quick update

forslund:
## Description
Now skills will always load before any update is performed. A quick
update will only be triggered if a default skill is missing, and then
only after skills have been loaded.

## How to test
Make sure skills starts.

CLA [ Yes ]


davidwagnerkc:
@chrisveilleux Before all your changes, this was added to speed up boot process, then in testing it didn't seem like it sped anything up but maybe opposite (there were a lot of other factors so hard to tell) so this PR was to remove the quick update.

Is this relevant anymore? Should we close this?

forslund:
This change was included in the work by chris so closing it.

PR #2228, opened 2019-07-23
tts_hash to an actual hash

davidwagnerkc:
## Description
tts_hash was initially set to the text of the config rather than hash so always loaded twice on boot (~750ms for create() for Mimic2)

## How to test
Add log to speech.py mute_and_speak L123 and make sure a reload does not occur at first boot


forslund:
Works exactly as advertised! Merging.

PR #2235, opened 2019-07-26
Handle removing unregistered intents

forslund:
## Description
This small change adds a check that the intent has been registered
before removing it.

If an unregistered intent was removed padatious would silently throw an
exception due to a list operation error. But when run synchronously from
the skill tester this silent exception was actually loud causing skills
to fail to load.

## How to test
Run the skill tester on the latest beta of the npr-news skill.

CLA [ Yes ]


krisgesling:
Looks good to me. Merging.

PR #2236, opened 2019-07-26
 Allow startup if update attempt occurs

forslund:
## Description
This changes the flow to allow proceeding to do proper startup after a system update attempt.

update_attempt will block waiting for the update to complete and if an actual update occurs it (the .deb update script) will shutdown the running skills process if an update is actually performed.

I also added a short voice confirmation that we're checking for updates.

## How to test
Check that an unpaired mark-1 starts up ok

CLA [ Yes ]

### DIFF
--- a/mycroft/skills/__main__.py
+++ b/mycroft/skills/__main__.py
@@ -79,18 +79,20 @@ class DevicePrimer(object):
         """Internet dependent updates of various aspects of the device."""
         self._get_pairing_status()
         self._update_system_clock()
-        update_attempted = self._update_system()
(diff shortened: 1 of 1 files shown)


forslund:
Ping @davidwagnerkc and @chrisveilleux. Can you check on this, it's sort of a critical issue.

I removed the dialog on update attempt I had originally added. Since it'd speak for the Mark-2 which can't do an update yet...

davidwagnerkc:
Tested on Mark I. Blocked during update then started up as expected.

PR #2258, opened 2019-08-16
Fix TypeError for number values in settings

krisgesling:
In some instances, the backend returns number setting values as an integer rather than a string.

This seems to be the quickest fix for it, but worth investigating further from the backend.

## How to test
This has been seen in the wild but has been difficult to replicate. The circumstances under which it happens are still unclear. I was able to semi-consistently have a number returned as int from the [Severe Weather Skill](https://github.com/domcross/severe-weather-information-skill) when editing an unrelated setting. To test, remove the 3 character country code (`[A-Z]{3} - `) from the service options of this skill .

### DIFF
--- a/mycroft/skills/settings.py
+++ b/mycroft/skills/settings.py
@@ -489,7 +489,7 @@ class SkillSettings(dict):
                     value = field.get('value')
 
                     if to_platform == 'core':
-                        if "." in value:
+                        if "." in str(value):
                             sections[i]['fields'][j]['value'] = float(value)
                         else:
                             sections[i]['fields'][j]['value'] = int(value)


krisgesling:
@chrisveilleux - just FYI - I ran into the same TypeError reported by malevolent in the chat today

chrisveilleux:
> @chrisveilleux - just FYI - I ran into the same TypeError reported by malevolent in the chat today

Yeah, I am of the opinion that the logic in core needs to be changed. I am working on a change right  now that plays more nicely with the backend.  This will do as a temp fix until I can get my new thing working.

forslund:
@chrisveilleux Just remember that the backend needs to be backwards compatible and should send these numbers as strings so older versions of core will still work.

PR #2265, opened 2019-08-21
Changes to MSM logic to improve performance.

chrisveilleux:
## Description
Half of the time it takes to boot the device is consumed by interactions with MSM. Skill-related data just doesn't change that frequently. Use caching of MSM data to speed cut boot time in half. There is a related PR for MSM that has the same branch name.

## How to test
At its core, this is a refactor. The logic hasn't changed, how frequently it executes has. Boot up a device and ensure all still works as expected. Install, remove and update a skill to ensure those actions reset the cached values.

CLA [Yes]

### DIFF
--- a/mycroft/skills/msm_wrapper.py
+++ b/mycroft/skills/msm_wrapper.py
@@ -12,42 +12,97 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+"""Common logic for instantiating the Mycroft Skills Manager.
 
-import os
-from os.path import join, expanduser, exists
+The Mycroft Skills Manager (MSM) does a lot of interactions with git.  The
+more skills that are installed on a device, the longer these interactions
+take.  This is especially true at boot time when MSM is instantiated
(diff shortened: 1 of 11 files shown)


mycroft/util/memoize.py
```diff
@@ -0,0 +1,40 @@
+import collections
+import functools
+
+from .log import LOG
+
+
+# The following memoized class is copied from the PythonDecoratorLibrary at:
```
davidwagnerkc: Difference between built in lru_cache built in?

https://docs.python.org/dev/library/functools.html#functools.lru_cache

We use it in padatious_service

forslund:
Did some minor changes and fixups and will now merge. Let me know of any mess-ups and I'll try to compensate for them.

- Replaced the memoization lru_cache as suggested by @davidwagnerkc 
- Replaced the property code with a small automatic function
- Fixed the skill updater initialization
- Fixed the failing tests

So without further ado...Merging

PR #2269, opened 2019-08-23
Check for AVX instruction set during install for Precise compatibility

mathmauney:
## Description
Check for AVX instruction set during install and warns the user that the Precise Engine will not be supported.

## How to test
Run installer on a machine that does not support AVX

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -137,6 +137,22 @@ This script is designed to make working with Mycroft easy.  During this
 first run of dev_setup we will ask you a few questions to help setup
 your environment.'
     sleep 0.5
+    if ! grep -q avx /proc/cpuinfo; then
+      echo "
+The Precise Wake Word Engine requires the AVX instruction set, which is
+not supported on your CPU. Do you want to fall back to the PocketSphinx
+engine? Advanced users can build the precise engine with an older
+version of TensorFlow (v1.13) if desired and change use_precise to true
+in mycroft.conf.
+  Y)es, I want to use the PocketSphinx engine or my own.
+  N)o, stop the installation."
+      if get_YN ; then
+        sed -i "s/\"use_precise\": true/\"use_precise\": false/" config/
+      else
+        echo -e "$HIGHLIGHT N - quit the installation $RESET"
+        exit 1
+      fi
+    fi
     echo "
 Do you want to run on 'master' or against a dev branch?  Unless you are
 a developer modifying mycroft-core itself, you should run on the
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -153,7 +153,7 @@ in mycroft.conf.
           fi
           $SUDO cp $TOP/mycroft/configuration/mycroft.conf /etc/mycroft/
         fi
-        sed -i "s/\"use_precise\": true/\"use_precise\": false/" /etc/mycroft/mycroft.conf
+        $SUDO sed -i "s/\"use_precise\": true/\"use_precise\": false/" /etc/mycroft/mycroft.conf
       else
         echo -e "$HIGHLIGHT N - quit the installation $RESET"
         exit 1
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -147,7 +147,13 @@ in mycroft.conf.
   Y)es, I want to use the PocketSphinx engine or my own.
   N)o, stop the installation."
       if get_YN ; then
-        sed -i "s/\"use_precise\": true/\"use_precise\": false/" config/
+        if [[ ! -f /etc/mycroft/mycroft.conf ]]; then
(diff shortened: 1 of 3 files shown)


devs-mycroft:
Hello, @mathmauney, thank you for helping with the Mycroft project! We welcome everyone
for Everyone.

you!

mathmauney:
Currently waiting on my CLA form.

Sorry about the line too long error, I disabled those in my PEP

dev_setup.sh
```diff
@@ -137,6 +137,22 @@ This script is designed to make working with Mycroft easy.  During this
 first run of dev_setup we will ask you a few questions to help setup
 your environment.'
     sleep 0.5
+    if ! grep -q avx /proc/cpuinfo; then
```
forslund: Does this condition apply to the ARM platform as well? Since we have a precise build for that platform it might need to be excluded.

mathmauney: Didn't know about the ARM exception. Looks like that can be easily checked with  $(uname -m) == 'arm'* so I've added that to the logic

forslund: Not easy to know :)

That looks like a good solution, great work with that.

mycroft/client/speech/hotword_factory.py
```diff
@@ -326,6 +332,10 @@ def initialize():
                     hotword, module
                 ))
                 instance = None
+            except PreciseUnavailable:
+                LOG.warning('Settings prevent Precise Engine use,
+                            falling back to default.')
```
forslund: Hmm this multiline string doesn't seem to be valid. I suggest using

```python
        LOG.warning('Settings prevent Precise Engine use, '
                    'falling back to default.')
```

mathmauney: Huh, thought that was what I changed it to already. Should be fixed now.

mathmauney:
CLA has been submitted

dev_setup.sh
krisgesling: Hey, we had some chat about this [on the Forums](https://community.mycroft.ai/t/running-mycroft-on-an-arm-cpu-without-avx/7056/4?u=gez-mycroft). According to @domcross the ARM equivalent is called SIMD (Single Instruction, Multiple Data) and is available since ARMv7 (which includes the Cortex A53 used by the RaspberryPi and the Cortex A7 from the OrangePi).
Is it worth checking for SIMD on ARM?

forslund: The arm equivalents seem to have a bunch of different names. on the Pi it's *neon* and on the H64 it's called *asimd* so it might not be trivial to do a correct general check.

krisgesling: Ah, good to know, ignore me then :) 
I think there's much less presumption that it will work on random ARM hardware anyway.

dev_setup.sh
```diff
@@ -137,6 +137,22 @@ This script is designed to make working with Mycroft easy.  During this
...
+      echo "
+The Precise Wake Word Engine requires the AVX instruction set, which is
+not supported on your CPU. Do you want to fall back to the PocketSphinx
+engine? Advanced users can build the precise engine with an older
+version of TensorFlow (v1.13) if desired and change use_precise to true
+in mycroft.conf.
+  Y)es, I want to use the PocketSphinx engine or my own.
+  N)o, stop the installation."
+      if get_YN ; then
+        sed -i "s/\"use_precise\": true/\"use_precise\": false/" config/
```
forslund: Is this sed line correct?
I get
` sed: can't read config/: No such file or directory`

Not sure which config you intend to edit here but you should probably make the adjustments in the `~/.mycroft/mycroft.conf` config

forslund: Indeed might be better

dev_setup.sh
```diff
@@ -147,7 +147,13 @@ in mycroft.conf.
...
   N)o, stop the installation."
       if get_YN ; then
-        sed -i "s/\"use_precise\": true/\"use_precise\": false/" config/
+        if [[ ! -f /etc/mycroft/mycroft.conf ]]; then
+          if [[ ! -e /etc/mycroft/ ]]; then
+            $SUDO mkdir /etc/mycroft
+          fi
+          $SUDO cp $TOP/mycroft/configuration/mycroft.conf /etc/mycroft/
+        fi
+        sed -i "s/\"use_precise\": true/\"use_precise\": false/" /etc/mycroft/mycroft.conf
```
forslund: sudo here as well maybe?

krisgesling (commented):
Hey there, sorry this dropped off my radar for a bit. It's a great addition though.

What do you think about the following?

krisgesling:
Closing in favor of #2910

PR #2271, opened 2019-08-24
Fix stacktrace on use of deprecated config member

forslund:
## Description
The stacktrace when the deprecated config member was used would fail.

CLA [ Yes ]


forslund:
Self-merging after checking with linter.

PR #2281, opened 2019-08-30
fix MycroftSkill registered_intents instance attribute

chrisveilleux:
## Description
self.registered_intents was moved to the IntentServiceInterface class but the instance attributes were not changed to reflect this.  Also tidied up a few minor issues raised by linter.

## How to test
Validate code that loops over registered intents.

CLA [yes]


forslund:
Thanks for fixing! I added a test case and did some further cleanups. Merging

PR #2283, opened 2019-09-01
Replace config properties to simpler instances

forslund:
## Description
The config doesn't need to be a property

CLA [ Yes ]


davidwagnerkc:
https://github.com/MycroftAI/mycroft-core/pull/2237#discussion_r311661902

@chrisveilleux Not necessary for 19.08. But sounds like from discussion in previous PR we were going to change this to an instance attribute. 👍 for this PR?

PR #2284, opened 2019-09-01
Update skill-tester for 19.08 and add basic support for wait_for_response

forslund:
## Description
- Allow skill tester to handle `wait_for_response` in the most basic way
- Fix moved references (LOG)
- Basic refactoring making some of the bigger methods less complex.

## How to test
Check that current skill tester works:
https://jenkins.mycroft.team/job/19.08-skill-tester/8/console
(failure here is valid due to current state of the weather skill)

CLA [ Yes ]

### DIFF
--- a/test/integrationtests/skills/skill_tester.py
+++ b/test/integrationtests/skills/skill_tester.py
@@ -45,11 +45,15 @@ from mycroft.skills.core import MycroftSkill, FallbackSkill
 from mycroft.skills.settings import SkillSettings
 from mycroft.skills.skill_loader import SkillLoader
 from mycroft.configuration import Configuration
-from mycroft.util.format import expand_options
 
 from logging import StreamHandler
 from io import StringIO
 from contextlib import contextmanager
+
+from .colors import color
+from .rules import (intent_type_check, play_query_check, question_check,
+                    expected_data_check, expected_dialog_check,
+                    changed_context_check)
 
 MainModule = '__init__'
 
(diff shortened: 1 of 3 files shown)


test/integrationtests/skills/skill_tester.py
```diff
@@ -327,6 +304,92 @@ def run(self, loader):
             s.get_response = orig_get_response
             s.settings = original_settings
 
+    def play_query(self, s, test_case):
```
krisgesling: Would it make sense for these methods to be `emit_play_query` or `send_play_query`?

test/integrationtests/skills/skill_tester.py
```diff
@@ -519,9 +529,7 @@ def __init__(self, test_case, skill=None):
         _x = ['and']
         if 'utterance' in test_case and 'intent_type' in test_case:
```
krisgesling: This will highlight my own ignorance, but would it be a valid test at all if 'utterance' doesn't exist?

test/integrationtests/skills/skill_tester.py
forslund: It definitely would, I will update those names and some of the other

test/integrationtests/skills/skill_tester.py
forslund: intents can only be tested if there is an utterance. The tests for questions and common play handlers don't require an utterance in the same way.

test/integrationtests/skills/skill_tester.py
forslund: Names updated!

PR #2286, opened 2019-09-02
Bugfix/mycroft skill

forslund:
## Description
Fix a couple of regressions
- scheduled event handlers without "message" parameter would never run.
- Padatious intents wouldn't be enabled / disabled correctly.

CLA [ Yes ]


forslund:
Branch rebased and should now look sane.

forslund:
Thanks for testing!

PR #2287, opened 2019-09-03
Refactor/skill settings split

chrisveilleux:
## Description
Complete rethink of how we deal with settings and settingsmeta in core

## How to test
Make changes to settingsmeta.json for a skill.  Confirm that the changes are accurately reflected on the backend.  Make changes to skill settings at account.mycroft.ai.  Confirm that changes are accurately reflected in settings.json

CLA [yes]

### DIFF
--- a/mycroft/api/__init__.py
+++ b/mycroft/api/__init__.py
@@ -380,30 +380,13 @@ class DeviceApi(Api):
         """ Upload skill metadata.
 
         Arguments:
-            settings_meta (dict): settings_meta typecasted to suite the backend
+            settings_meta (dict): skill info and settings in JSON format
         """
         return self.request({
             "method": "PUT",
-            "path": "/" + self.identity.uuid + "/skill",
+            "path": "/" + self.identity.uuid + "/settingsMeta",
             "json": settings_meta
         })
-
-    def delete_skill_metadata(self, skill_gid):
-        """ Delete the current skill metadata from backend
-
-            TODO: Real implementation when method exists on backend
-        Args:
-            skill_gid (str): skill_gid identifying the skill
-        """
-        try:
-            LOG.debug("Deleting remote metadata for {}".format(skill_gid))
-            self.request({
-                "method": "DELETE",
-                "path": ("/" + self.identity.uuid + "/skill" +
-                         "/{}".format(skill_gid))
-            })
-        except Exception as e:
-            LOG.error("{} cannot delete metadata because this".format(e))
 
     def upload_skills_data(self, data):
         """ Upload skills.json file. This file contains a manifest of installed
--- a/mycroft/api/__init__.py
+++ b/mycroft/api/__init__.py
@@ -12,14 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-from copy import copy
-
-import json
-import requests
-from requests import HTTPError, RequestException
 import os
 import time
-from threading import Lock
+from copy import copy
+
+import requests
+from requests import HTTPError, RequestException
 
 from mycroft.configuration import Configuration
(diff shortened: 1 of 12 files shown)


forslund:
What's breaking here, The callback registration and the .store() methods? Is there a way of preserving these so we don't break the API without depreciation warning?

The save_settings should only occur if the settings were changed.

Let's move as much of the logic for this outside of the mycroft_skill.

Also let's drop the refactoring commits so it's easier to review.

chrisveilleux:
Implemented the change you suggested to make the skill API right again.

I don't think I can move any more of the logic outside of MycroftSkill.  The only thing I really added was a way to deal with settings changes when they occur.

Added logic to only call save_settings if settings change

mycroft/skills/settings.py
```diff
@@ -12,571 +12,383 @@
...
+            if settings_changed:
+                message = Message(
+                    'mycroft.skills.settings.changed',
+                    data={skill_gid: remote_settings}
+                )
+                self.bus.emit(message)
+
+
+# TODO: remove in 20.02
+class Settings:
```
forslund: Add __getattr__ to keep .get and .update, .keys, etc. methods

```
    def __getattr__(self, attr):
        if attr not in ['store', 'set_changed_callback']:
            return getattr(self._settings, attr)
        else:
            return super().getattr(attr)
```

mycroft/skills/settings.py
```diff
@@ -12,571 +12,383 @@
...
-    return skill.skill_gid, skill.meta_info.get('display_name')
+    return skill_settings
 
 
-def display_name(name):
-    """ Splits camelcase and removes leading/trailing Skill. """
-    name = re.sub(r'(^[Ss]kill|[Ss]kill$)', '', name)
-    return camel_case_split(name)
+def save_settings(skill_dir, skill_settings):
+    """Save skill settings to file."""
```
forslund: Needs to handle the Settings class specially...

Something along the lines of
```python
def save_settings(skill_dir, skill_settings):                                   
    """Save skill settings to file."""                                          
    settings_path = Path(skill_dir).joinpath('settings.json')                   
                                                                                
    if isinstance(skill_settings, Settings):                                    
        settings = skill_settings.to_dict()                                     
    elif isinstance(skill_settings, dict):                                      
        settings = skill_settings                                               
    else:                                                                       
        raise ValueError('skill_settings should be a dict, not '                
                         '{}'.format(type(skill_settings)))                     
    with open(str(settings_path), 'w') as settings_file:                        
        json.dump(settings, settings_file)
```

mycroft/api/__init__.py
```diff
@@ -362,49 +358,25 @@ def get_oauth_token(self, dev_cred):
...
-                    time.monotonic() > DeviceApi._skill_settings[0] + 30):
-                DeviceApi._skill_settings = (
-                    time.monotonic(),
-                    self.request({
-                        "method": "GET",
-                        "path": "/" + self.identity.uuid + "/skill"
-                        })
-                )
-            return DeviceApi._skill_settings[1]
+    def get_remote_skill_settings(self):
```
forslund: I think the "remote" part here is redundant since it's in the DeviceApi it's implied that it's a remote thing.

mycroft/skills/settings.py
chrisveilleux: Why is this necessary?  When I call this method from the settings class, I pass the _settings attribute, which is a dictionary.

forslund: In the future we intend to just pass the MycroftSkill.settings member so we should start doing that directly. Which will show people how to use the function as intended.

Also accessing a _attribute from outside the class should really not be done

forslund:
Backend has been updated. Merging this now.

PR #2290, opened 2019-09-06
Add mycroft-config tool to manage config files

penrods:
The new mycroft-config utility simplifies management of the the
various configuration files that control Mycroft.  Commands include:
edit, show, get, set, reload.

* ```mycroft-config edit (default|remote|system|user)```
Open the requested file in an editor (nano by default), performing
JSON validation warnings to minimize accidental edit errors.

* ```mycroft-config show (default|remote|system|user)```
Dump the selected file with nice highlighting using jq

* ```mycroft-config get <var>```
Load the mycroft.conf stack, outputting the effective value(s).  The
<var> is specified using jq-style specifiers.  For example:
   mycroft-config get enclosure.platform
   mycroft-config get location
   mycroft-config get

* ```mycroft-config set <var> <value>```
Set the given variable in the "user" configuration file.  Examples:
   mycroft-config set lang "es-es"
   mycroft-config set tts.mimic.voice "ap"

* ```mycroft-config reload```
Sends a signal on the messagebus to tell services to reload configs.
This automatically occurs after a 'set' or 'edit' command.

### DIFF
--- a/bin/mycroft-config
+++ b/bin/mycroft-config
@@ -136,6 +136,8 @@ function edit_config() {
             ;;
         default | remote)
             # View-only
+            echo "The default config shouldn't be changed, opening in View mode"
+            sleep 2
             $VIEWER $TEMP/mycroft.json
             ;;
         esac
--- a/bin/mycroft-config
+++ b/bin/mycroft-config
@@ -49,7 +49,11 @@ function help() {
 
 VIEWER="nano --syntax=json --view"
 if [ -z "$EDITOR" ] ; then
-    EDITOR="nano --syntax=json --tempfile"
+    if [ $( which sensible-editor ) ] ; then
+        EDITOR="sensible-editor"
+    else
+        EDITOR="nano --syntax=json --tempfile"
+    fi
 fi
 
 if [ -z "$TEMP" ] ; then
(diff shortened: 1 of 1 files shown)


JarbasAl:
this should reduce support requests by 70%!

most of the time "mycroft is broken" it's a bad comma messing json, probably in core config file....

bin/mycroft-config
```diff
@@ -0,0 +1,238 @@
...
+    echo "  ${script} set test.subvalue \"foo\" "
+
+    exit 1
+}
+
+################################################################
+# Setup stuff based on the environment
+
+VIEWER="nano --syntax=json --view"
+if [ -z "$EDITOR" ] ; then
```
forslund: should the sensible-editor link be checked as an alternative?

bin/mycroft-config
```diff
@@ -0,0 +1,238 @@
...
+        *)
+            echo "ERROR: Unknown name '${1}'."
+            echo "       Must be one of: default, remote, system, or user"
+            exit 1
+    esac
+}
+
+################################################################
+
+function edit-config() {
```
forslund: Let's use underscores instead of hyphens to keep constency and portability

penrods: Done

bin/mycroft-config
penrods: Done

forslund:
Rebased and added a warning for when opening in view mode instead of the expected edit mode. Merging now.

PR #2297, opened 2019-09-10
Cycle through dialog files to avoid repeating the same line

ChanceNCounter:
## Description
Attempts to use a different line from .dialog files each time a skill is invoked. If a dialog file has more than 2 entries, Mycroft will avoid speaking the same line twice in a row. Longer dialog files will not repeat for at least 3 runs.

Changes are to MustacheDialogRenderer.render(), so the little cache of "phrases I won't say next" is per _skill_, not per intent nor per dialog file. However, the decision to cache a phrase (or not) _is_ based on the length of the dialog file, so that the feature will never leave an intent without anything to say.

This could be refactored to keep track on a per-intent basis, or per dialog file, but I think it would be a waste of memory for not much gain. The cache inherently enforces a minimum number of _skill_ invokations before Mycroft can repeat the same phrase, even if it isn't guaranteed to draw another phrase from the same, specific dialog file in the interim.

## How to test
Manual testing with your favorite sandbox skill is easy. I started out testing with Dismissal because it had exactly 4 lines of dialog. At the end of each run, have an intent:

self.log.info(self.dialog_renderer.recent_phrases)

and it will log the array of phrases it's blocking from repetition.

Comment lines out of the associated .dialog file, a few at a time, and observe what happens to the array when the .dialog file is 1, 2, 3, or 4+ lines in length.

## Notes
Might wanna squash the commits. I know that's usually bad practice, but the second one's just noise, apart from its commit message.

credit to @forslund for making the very-pythonic line 105 out of my overdocumented, C-style first attempt.

CLA [ Yes ]

### DIFF
--- a/mycroft/dialog/__init__.py
+++ b/mycroft/dialog/__init__.py
@@ -35,6 +35,10 @@ class MustacheDialogRenderer:
 
     def __init__(self):
         self.templates = {}
+        self.recent_phrases = []
+
+        # TODO magic numbers are bad!
+        self.max_recent_phrases = 3
 
     def load_template_file(self, template_name, filename):
         """
@@ -67,6 +71,8 @@ class MustacheDialogRenderer:
         Given a template name, pick a template and render it using the context.
         If no matching template exists use template_name as template.
 
+        Tries not to let Mycroft say exactly the same thing twice in a row.
+
         Args:
             template_name (str): the name of a template group.
             context (dict): dictionary representing values to be rendered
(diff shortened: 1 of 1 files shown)


devs-mycroft:
Hello, @ChanceNCounter, thank you for helping with the Mycroft project! We welcome everyone
for Everyone.

you!

ChanceNCounter:
The fact that this thing works seems too good to be true, so I'm expecting somebody will tell me either why it sucks or how it breaks everything =P

But it Works On My Computer, and I'm having a lot of fun.

LinusS1:
Do you know if this impacts the speed of Mycroft finding a response?

ChanceNCounter:
I haven't profiled it, but I can tell you that any performance hit is entirely unnoticeable on my machine. I suppose it might be a problem if somebody created an exceptionally long dialog file, like, dozens or hundreds of lines. That might be something to test.

Either way, TTS still seems to be the primary bottleneck.

ChanceNCounter:
Update: I've just done some manual testing on a dialog file with ~3200 lines. It still seems to pick a response more or less instantaneously (I can't perceive the time it takes.) TTS, though, yeesh.

There are some clips of those tests, with the skill printing to the cli, in the mycroft-core chat channel.

forslund:
Looks good. Just squashed your last commit into the previous one, will merge now.

Many thanks for doing this, it sounds much better!

PR #2299, opened 2019-09-11
Catch exceptions when reloading skills

forslund:
## Description
I've experienced a couple of halts of the skill manager due to FileNotFoundErrors when reloading skills.

This will explicitly
- Explicitly catch FileNotFoundError
- Do a general catch and log for other exceptions

## How to test
Make sure skill service is still loading and reloading as expected

CLA [ Yes ]


davidwagnerkc:
Tested :+1:

PR #2305, opened 2019-09-13
fix issue with false internet connection test

chrisveilleux:
## Description
The ability to reach a name server was used to determine internet connectivity.  This occasionally returned false positives.  Changed to check ability to get to google.com

## How to test
This issue shows up in boot only occasionally it is not easy to reproduce.

CLA [yes]

### DIFF
--- a/mycroft/util/__init__.py
+++ b/mycroft/util/__init__.py
@@ -14,26 +14,22 @@
 #
 from __future__ import absolute_import
 
-from copy import deepcopy
+import json
+import logging
+import os
 import re
-import socket
+import signal as sig
 import subprocess
-import pyaudio
-
-from os.path import join, expanduser, splitext
-
+import tempfile
+from copy import deepcopy
+from stat import S_ISREG, ST_MTIME, ST_MODE, ST_SIZE
 from threading import Thread
 from time import sleep
+from urllib.request import urlopen
+from urllib.error import URLError
 
-import json
-import os.path
-import os
+import pyaudio
 import psutil
-from stat import S_ISREG, ST_MTIME, ST_MODE, ST_SIZE
-import requests
-import logging
-
-import signal as sig
 
 import mycroft.audio
 import mycroft.configuration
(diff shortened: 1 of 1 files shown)


forslund:
Just connecting to google isn't a strong test. To handle capture portals the actual data received need to be verified.

If there is a problem with the dns test causing false positives (internet connected but no DNS? yet) I suggest we just use ncis test where the retrieved test data is compared to a known entity.

Also please don't do refactoring in the same commit as logic changes.

davidwagnerkc:
Just using the ncsi test sounds reasonable to me. @chrisveilleux ?

davidwagnerkc:
@forslund @chrisveilleux 

What about replacing 8.8.8.8 check with something like this. 

It is faster and more efficient to only request headers rather than content with GET (https://stackoverflow.com/a/29854274/2395133)

and it checks for valid header response content. From what I have read no header fields are strictly required by the standard, but reasonable that one of these most common fields will be returned (all of them were returned in request I made)

```
res = requests.head('http://www.google.com', timeout=2)
connected = any(field in res.headers for field in ('Server', 'Date', 'Content-Type', 'Content-Length'))
```

forslund:
The problem would still be capture portals, they either use dns manipulation or directly intercepts HTTP packages.

The dns check circumvents those issues by using direct socket connections and uses an IP. the ncis test validates the data received to make sure the target uri is what's expected.

Using the dns_connection check together with the headers only check should be relatively fast and even more reliable... Except for networks blocking external dns connections... (which is why the ncis check was added).

```python
    (is_connected_dns() and is_connected_google_header()) or is_connected_ncis()` 
```
maybe?

chrisveilleux:
Why are we concerned about captive portals?  Could you give me a use case @forslund?

It occurs to me that what we really need is access to our API.  Even if there are skills that require the internet, you aren't going to get very far in your Mycroft experience without hitting our API.  Is there any reason we can't change this to hit a new API endpoint that returns a string value we can test?  Is there another reason we need to be able to test internet connectivity independently of connectivity to our API?

forslund:
What's the use of not falsely getting a connected state?

People trying out core on a network with a captive portal shouldn't get the impression that they're connected when in fact they're just connected to the captive portal. It was originally prepared for the handling of the captive portal case...which is still on the todo list...

I've had use for it several times when working on the road.

The endpoint was suggested a couple of years back and I'm not at all against it.

forslund:
Merging this with the addition of checking the connect to DNS as well. The time for the connection to DNS server is insignificant compared to the time the google check takes.

We need to revisit this when we have a dedicated endpoint to check this.

PR #2311, opened 2019-09-17
Catch all exceptions from upload device info

forslund:
## Description
Expands the HTTPError, Connection error capture to a general `Exception`

## How to test
Make sure skills process starts

CLA [ Yes ]


forslund:
Fairly simple change. Self-merging.

PR #2312, opened 2019-09-17
settingsMeta API call to be issued without a device ID in the URL.

chrisveilleux:
## Description
Fix a bug where the settingsMeta API call was failing on initial boot if the device was paired but the device identity attribute of the API class did not yet know the device's ID

## How to test
Boot an unpaired device. There should be no API calls with a URL of `v1/device//settingsMeta` that fail with a 405 error.

CLA [yes]


forslund:
Updated the test cases. The 405 is gone. Prepared PR #2313 which is a probable root cause for this issue.

forslund:
Oh, and Merging!

PR #2314, opened 2019-09-18
Restructure to get skill_gid after startup

forslund:
## Description
This restructures the code slightly allowing skills to get the skill_gid after pairing is complete.

## How to test
Pair a device fresh device and make sure no errors indicating that skill_gid is missing from the skill.

CLA [ Yes ]

### DIFF
--- a/mycroft/skills/settings.py
+++ b/mycroft/skills/settings.py
@@ -184,18 +184,21 @@ class SettingsMetaUploader:
         The device ID is known to this class.  To "finalize" the skill_gid,
         insert the device ID between the "@" and the "|"
         """
-        skills = {
-            skill.path: skill for skill in
-            self.msm.local_skills.values()
-        }
-        skill = skills[str(self.skill_directory)]
-        # If modified prepend the device uuid
-        self._skill_gid = skill.skill_gid.replace(
-            '@|',
-            '@{}|'.format(self.api.identity.uuid)
-        )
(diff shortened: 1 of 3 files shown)


davidwagnerkc:
Test on fresh boot. No skill_gid errors in the logs. :+1:

mycroft/skills/settings.py
```diff
@@ -184,18 +184,21 @@ def skill_gid(self):
...
-            skill.path: skill for skill in
-            self.msm.local_skills.values()
-        }
-        skill = skills[str(self.skill_directory)]
-        # If modified prepend the device uuid
-        self._skill_gid = skill.skill_gid.replace(
-            '@|',
-            '@{}|'.format(self.api.identity.uuid)
-        )
+        if self.api and self.api.identity and self.api.identity.uuid:
```
chrisveilleux: You shouldn't need to second condition here.  The API class gets an identity attribute in __init__.  This object will always have a DeviceIdentity object assigned to it.  It should never evaluate to False.

forslund: Good thinking, I'll simplify the condition to trust that the identity has a sane value.

forslund:
Updated according to feedback.

mycroft/skills/settings.py
davidwagnerkc: :+1:

PR #2316, opened 2019-09-18
Add a safety harness in skill_manager main loop

forslund:
## Description
This captures any unhandled expressions and will at least make sure
they're logged properly. (Exceptions in threads are mostly not printed
to stdout).

This sleeps slightly longer than normal (10 seconds) and then tries to resume
normal operation.

I've had one instance where skills reloading / updating stopped.

## How to test
Make sure the normal loop runs as expected and reloads are handled correctly.

CLA [ Yes ]

### DIFF
--- a/mycroft/skills/skill_manager.py
+++ b/mycroft/skills/skill_manager.py
@@ -143,11 +143,17 @@ class SkillManager(Thread):
         # Scan the file folder that contains Skills.  If a Skill is updated,
         # unload the existing version from memory and reload from the disk.
         while not self._stop_event.is_set():
-            self._reload_modified_skills()
-            self._load_new_skills()
-            self._unload_removed_skills()
-            self._update_skills()
-            sleep(2)  # Pause briefly before beginning next scan
+            try:
+                self._reload_modified_skills()
+                self._load_new_skills()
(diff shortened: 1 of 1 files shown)


mycroft/skills/skill_manager.py
```diff
@@ -143,11 +143,18 @@ def run(self):
...
+                self._reload_modified_skills()
+                self._load_new_skills()
+                self._unload_removed_skills()
+                self._update_skills()
+                sleep(2)  # Pause briefly before beginning next scan
+            except Exception as e:
+                LOG.exception('Something really unexpected has occured '
+                              'and the skill manager loop safety harness was '
+                              'hit.\n'
+                              'Exception: {}'.format(repr(e)))
```
chrisveilleux: The log.exception method will automatically log the exception traceback.  No need to include the last line of the log message.

forslund: You are technically correct, the best kind of correct!

forslund:
Updated according to feedback

mycroft/skills/skill_manager.py
davidwagnerkc: :+1:

PR #2320, opened 2019-09-19
Bugfixes/skills manifest

forslund:
## Description
- repost manifest when skill reload occur in case something has changed.
- Use a fresh msm instance when uploading skills manifest to make sure all skills are included.

CLA [ Yes ]

### DIFF
--- a/mycroft/api/__init__.py
+++ b/mycroft/api/__init__.py
@@ -14,7 +14,7 @@
 #
 import os
 import time
-from copy import copy
+from copy import copy, deepcopy
 
 import requests
 from requests import HTTPError, RequestException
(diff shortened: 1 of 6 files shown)


forslund:
@chrisveilleux I've updated some more. Seems to work as expected.

Installing release-test caused an error the first upload attempt, but the one one minute later (after manifest was uploaded) worked.

davidwagnerkc:
@forslund How can we validate this PR fixes the issue you are seeing? I pulled this PR down on a new image and it works fine. skill_gid errors are still in there and do not seem to go away. But this PR is more related to the gid in skills.json / in memory, right?

forslund:
The important thing with this is that It should not show settingsmeta upload failures forever and ever

PR #2323, opened 2019-09-24
Catch inspect error in logger

forslund:
## Description
Inspect would in certain cases fail to find the line number. This
catches the exception and provides a basic name for the logger.

## How to test
Make sure processes starts and logmessages are still shown.

CLA [ Yes ]

### DIFF
--- a/mycroft/util/log.py
+++ b/mycroft/util/log.py
@@ -126,18 +126,23 @@ class LOG:
             # [0] - _log()
             # [1] - debug(), info(), warning(), or error()
             # [2] - caller
-            stack = inspect.stack()
+            try:
+                stack = inspect.stack()
 
-            # Record:
-            # [0] - frame object
-            # [1] - filename
-            # [2] - line number
-            # [3] - function
-            # ...
-            record = stack[2]
-            mod = inspect.getmodule(record[0])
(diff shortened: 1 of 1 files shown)


mycroft/util/log.py
```diff
@@ -126,18 +126,23 @@ def _log(cls, func, *args, **kwargs):
...
+                # [2] - line number
+                # [3] - function
+                # ...
+                record = stack[2]
+                mod = inspect.getmodule(record[0])
+                module_name = mod.__name__ if mod else ''
+                name = module_name + ':' + record[3] + ':' + str(record[2])
+            except Exception:
+                # The location couldn't be determined
+                name = 'Mycroft'
```
davidwagnerkc: Can we do `name = name if name else 'Mycroft'` so we can potentially keep the custom name if supplied?

forslund: The inspect path does only occur when the custom name isn't supplied as I'm reading it?

'Mycroft' would only be used if the custom name is omitted and it tries to use inspect to generate the name and that inspect call fails.

Tested it out and the skills self.log.info comes out with skill names as expected.

davidwagnerkc: :+1: Thanks.

PR #2324, opened 2019-09-25
Replace custom logger with logger instance

forslund:
## Description
LOG is now a logger instance. module and line number was added to the
format string to preserve previous information

A create_logger function is added to the object for backwards
compatibility

## How to test
Make sure everything runs and logs are appearing as expected

CLA [ Yes ]


mycroft/util/log.py
```diff
@@ -51,99 +50,49 @@ def method(cls, *args, **kwargs):
...
+    config = {}
+    for conf in confs:
+        try:
+            merge_dict(config,
+                       load_commented_json(conf) if isfile(conf) else {})
+        except Exception as e:
+            print('couldn\'t load {}: {}'.format(conf, str(e)))
+    LOG.setLevel(logging.getLevelName(config.get('log_level', level)))
+
+    # Backwards compatibility, remove in rekease 20.02
```
JarbasAl: maybe add a TODO to the comment so we can easily grep the codebase and avoid missing this, seems to be the standard so far for stuff that will be deprecated in follow up versions

forslund: Good thinking. I'll add that

forslund: PR updated

mycroft/util/log.py
```diff
@@ -51,99 +50,49 @@ def method(cls, *args, **kwargs):
...
+    handler = logging.StreamHandler(sys.stdout)
+    formatter = logging.Formatter(log_message_format, style='{')
+    formatter.default_msec_format = '%s.%03d'
+    handler.setFormatter(formatter)
+    LOG.addHandler(handler)
+
+    # Check configs manually, the Mycroft configuration system can't be
+    # used since it uses the LOG system and would cause horrible cyclic
+    # dependencies.
+    confs = [SYSTEM_CONFIG, USER_CONFIG]
```
chrisveilleux: it's not clear to me what we are trying to do with the configs here

mycroft/util/log.py
```diff
@@ -51,99 +50,49 @@ def method(cls, *args, **kwargs):
...
-                name = 'Mycroft'
-
-        func(cls.create_logger(name), *args, **kwargs)
-
-
-LOG.init()
+LOG = logging.getLogger('Mycroft')
+
+
+def create_logger(name):
```
chrisveilleux: If we are going to do this, I would like to see a daily rotating file handler added (there is one built into python).  the current log files can get ungainly big.  this would mean removing shell script code that redirects STDOUT to the log file.  Another benefit to this is that the "log" file would only have log messages in it.  If we still want to capture the print statements and such, that should be written to a different file.  a quick cron job to delete older daily log files would be a good idea too.  it would take a while, but losing a device due to log file disk consumption would be bad.

chrisveilleux:
One other thing I forgot to mention.  Not a fan of letting skill developers name their own loggers in their skills.  I would like to see if there is a way in core to normalize logger names for skills.  For example, a logger named `Mycroft.Playback Control Skill` in the same file as a logger named `mycroft.skills.settings` is inconsistent.  Could we normalize he names passed at least?  So `Mycroft.Playback Control Skill` becomes `mycroft.playback_control_skill`?

ChanceNCounter:
> One other thing I forgot to mention in my review. Not a fan of letting skill developers name their own loggers in their skills. I would like to see if there is a way in core to normalize logger names for skills. For example, a logger named `Mycroft.Playback Control Skill` in the same file as a logger named `mycroft.skills.settings` is inconsistent. Could we normalize he names passed at least? So `Mycroft.Playback Control Skill` becomes `mycroft.playback_control_skill`?

What if we want to implement multiple loggers for the same skill? Say a skill spins up child processes, and wants to log them separately, or wants one file for call and response and another file for the system calls in the skill's "backend." Would prescribing logger names interfere with that?

mycroft/util/log.py
```diff
@@ -51,99 +50,49 @@ def method(cls, *args, **kwargs):
...
+    logger = LOG.getChild(name)
+    logger.addHandler(handler)
+    logger.propagate = False
+    logger.addHandler(handler)
+    return logger
+
+
+def setup_default_logger(level='INFO'):
+    """ Setup the default logger."""
+    log_message_format = ('{asctime} | {levelname:8} | {process:5} | '
```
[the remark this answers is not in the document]
forslund: To keep with the format of the default skill logger. where name is used to indicate the skill.

mycroft/util/log.py
```diff
@@ -51,99 +50,49 @@ def method(cls, *args, **kwargs):
...
-                name = module_name + ':' + record[3] + ':' + str(record[2])
-            except Exception:
-                # The location couldn't be determined
-                name = 'Mycroft'
-
-        func(cls.create_logger(name), *args, **kwargs)
-
-
-LOG.init()
+LOG = logging.getLogger('Mycroft')
```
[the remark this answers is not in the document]
forslund: We'll need to keep this as uppercase for compatibility, The reasoning for uppercase module level names was I believe to differentiate the object instances from classes and functions.

mycroft/util/log.py
forslund: No that's a system level implementation. service or running script that should handle. logging should be trivial so it can be redirected and modified depending on system implementation.

mycroft/util/log.py
forslund: I'll improve the docstring

forslund:
For the most part the skills loggers are named automatically assuming they use the skill built in `self.log` logger. But I agree with @ChanceNCounter, skill creators should remain free to create any logger they need to track their work.

Would be best if is they use `self.log.getChild()` which would put it under the skills logger but if they want to make their own logger all together that's fine with me.

mycroft/util/log.py
```diff
@@ -51,99 +50,49 @@ def method(cls, *args, **kwargs):
...
+
+def create_logger(name):
+    log_message_format = ('{asctime} | {levelname:8} | {process:5} | '
+                          '{name} | {message}')
+    handler = logging.StreamHandler(sys.stdout)
+    formatter = logging.Formatter(log_message_format, style='{')
+    formatter.default_msec_format = '%s.%03d'
+    handler.setFormatter(formatter)
+
+    logger = LOG.getChild(name)
```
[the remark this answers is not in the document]
forslund: I know this is suboptimal, I'll see if I can get the module log to be more explicit.

mycroft/util/log.py
chrisveilleux: I am not sure I understand the delineation you are trying to make here.  Skills, voice, audio, enclosure... these are all services and we have every right to control how the logging occurs for those services as we are the custodians of them. 

If we are going to allow skill developers to produce logs based on their personal preferences,  which seems to be what you are going for based on your comments, skill logs should be independent of core logs.  Each skill should have its own logger instance with messages that are not interleaved with other logs emitted by core.

chrisveilleux:
> > One other thing I forgot to mention in my review. Not a fan of letting skill developers name their own loggers in their skills. I would like to see if there is a way in core to normalize logger names for skills. For example, a logger named `Mycroft.Playback Control Skill` in the same file as a logger named `mycroft.skills.settings` is inconsistent. Could we normalize he names passed at least? So `Mycroft.Playback Control Skill` becomes `mycroft.playback_control_skill`?
> 
> What if we want to implement multiple loggers for the same skill? Say a skill spins up child processes, and wants to log them separately, or wants one file for call and response and another file for the system calls in the skill's "backend." Would prescribing logger names interfere with that?

As I mentioned in a previous comment, if we want skill developers to have control over their logging, that is fine.  However, those logs should not be intermingled with core logs, if that is the case.  Alternatively, if all logs for the skill service and skills are going to the same place, those logs should all behave the same way.

PR #2326, opened 2019-09-25
Remove ntp sync reboot

davidwagnerkc:
## Description
Reboot after NTP sync is no longer necessary with the changes to the skills process startup. 

## How to test
Boot up device with no network connection or identity file. Set date back: `sudo date -s "Sep 23 20:40:13 UTC 2019"` and continue to setup wifi, pair device and confirm it works as expected. 

### DIFF
--- a/mycroft/skills/__main__.py
+++ b/mycroft/skills/__main__.py
@@ -81,20 +81,17 @@ class DevicePrimer(object):
         """Internet dependent updates of various aspects of the device."""
         self._get_pairing_status()
         self._update_system_clock()
-        if self.time_skew_threshold_exceeded:
-            self._reboot()
-        else:
-            self._update_system()
-            # Above will block during update process and kill this instance if
-            # new software is installed
+        self._update_system()
+        # Above will block during update process and kill this instance if
+        # new software is installed
 
-            if self.backend_down:
-                self._notify_backend_down()
-            else:
-                self._display_skill_loading_notification()
-                self.bus.emit(Message('mycroft.internet.connected'))
-                self._ensure_device_is_paired()
-                self._update_device_attributes_on_backend()
+        if self.backend_down:
+            self._notify_backend_down()
+        else:
+            self._display_skill_loading_notification()
+            self.bus.emit(Message('mycroft.internet.connected'))
+            self._ensure_device_is_paired()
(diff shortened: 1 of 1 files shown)


mycroft/skills/__main__.py
```diff
@@ -81,20 +81,17 @@ def prepare_device(self):
         """Internet dependent updates of various aspects of the device."""
         self._get_pairing_status()
         self._update_system_clock()
-        if self.time_skew_threshold_exceeded:
-            self._reboot()
-        else:
-            self._update_system()
-            # Above will block during update process and kill this instance if
-            # new software is installed
+        self._update_system()
```
chrisveilleux: You got rid of the only usage of self.time_skew_threshold_exceeded.  Removal of that property should be included in this PR.

davidwagnerkc: Thanks. `_reboot()` was also no longer needed and remove. Re-tested and works just fine.

penrods:
I understand why you'd want to get rid of this, but keep an eye on it.  I don't recall the exact symptoms, but it involved an image which was months old being booted for the first time.  In the startup sequence, the sudden passage of those "months" in one clock tick had some squirrelly effects on the system.  This was a little ham-handed, but it was the most reliable solution at the time and seemed reasonable during that case.  Perhaps the one-hour threshold was too aggressive -- I was trying to make it deal reasonably with the unplug, move and replug scenario.

Not saying you should avoid doing this, but be sure to test the impact of a really long restart.  On the  Raspberry Pi they use a thing called [fake-hwclock](https://manpages.debian.org/jessie/fake-hwclock/fake-hwclock.8.en.html) to save the current time in a file at shutdown, then reset the clock on boot.  You can potentially edit that file to simulate a long dormancy -- look for /etc/fake-hwclock.data.

davidwagnerkc:
Thanks I was just testing with setting the date manually back a day. I can
try months and try it with that file rather than the date command as well.

davidwagnerkc:
Did more tests with changing clock back a year and also tested with fresh Mark 1 image that was from June, removing ntp reboot and adding in monotonic time for pairing code expiration (https://github.com/MycroftAI/skill-pairing/commit/2e42197125d5ca87e7d71a5174983135f6db6302). All test continue to look good. Merging.

PR #2327, opened 2019-09-26
Switch from the deprecated imp to importlib

forslund:
## Description
Replaces the current imp based skill loader with an importlib based solution. This also adds a small change allowing better use of submodules. Before submodules were never reloaded forcing the over-use of the `__init__.py` file in skills. The change makes the submodules reload together with the main module by removing the refs from `sys.modules`, allowing better structure of skills.

## How to test
Make sure skills load. Make a change to a skill and ensure that the skill is reloaded and that the change is accessible.
Add a log statement to the Spotify skill's spotify.py and check that it's included when the skill reloads.

CLA [ Yes ]

### DIFF
--- a/mycroft/skills/skill_loader.py
+++ b/mycroft/skills/skill_loader.py
@@ -14,7 +14,6 @@
 #
 """Periodically run by skill manager to load skills into memory."""
 import gc
-import imp
 import os
 import sys
 from time import time
@@ -27,6 +26,32 @@ from mycroft.util.log import LOG
 from .settings import SettingsMetaUploader
 
 SKILL_MAIN_MODULE = '__init__.py'
+
+
+def load_skill_module(path, skill_id):
+    """Load a skill module
+
+    This function handles the differences between python 3.4 and 3.5+ as well
+    as makes sure the module is inserted into the sys.modules dict.
+
+    Arguments:
+        path: Path to the skill main file (__init__.py)
+        skill_id: skill_id used as skill identifier in the module list
+    """
+    module_name = skill_id.replace('.', '_')
+    if sys.version_info >= (3, 5):
+        import importlib.util
+        spec = importlib.util.spec_from_file_location(module_name, path)
+        mod = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(mod)
+    else:
+        from importlib.machinery import SourceFileLoader
+        mod = SourceFileLoader(module_name, path).load_module()
+    if mod:
+        sys.modules[module_name] = mod
+        return mod
+    else:
+        return None
 
 
 def _get_last_modified_time(path):
@@ -139,7 +164,7 @@ class SkillLoader:
         """Call the shutdown method of the skill being reloaded."""
         try:
             self.instance.default_shutdown()
-        except Exception as e:
+        except Exception:
             log_msg = 'An error occurred while shutting down {}'
             LOG.exception(log_msg.format(self.instance.name))
         else:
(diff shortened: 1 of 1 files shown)


chrisveilleux:
If we are going to do this, I want it to include the ability to load the code from a file other than `__init__.py`.  Putting all your code in this file is a Python anti-pattern.  This PR should make it possible for an author to move their skill code from `__init__.py` to something like `skill.py`.  At the very least, it should make it possible for __init__.py to be able to import code from another file, like `skill.py` and then contain only the import and the `create_skill` function.

forslund:
Isn't that already possible?

The only thing that's needed is that `__init__.py` has the `create_skill()` function, either locally or imported from a module...

chrisveilleux:
I moving the skill class to a file called `skill.py` a while ago and importing that into `__init__.py`.  it did not work.  It is possible that I did it wrong.  It is also possible that the new version of importlib supports this whereas the older imp library did not.  

I can test against the code in this PR and see if there is different behavior.

forslund:
Ah, right imported submodules might not be reloaded...let me do some tests here

forslund:
Well the [spotify skill](https://github.com/forslund/spotify-skill/blob/19.08/__init__.py#L38) has the majority of the spotify logic in a separate file so it has worked before at least...

let me try with your case with the entire skill in a skill.py file...

chrisveilleux:
I would be happy if we just didn't hardcode the `__init__.py` as the only place skill code can come from.  Make it so that the code can be loaded from there or from another location, like `skill.py`.  We can support both for a time then remove support for `__inti__.py`.

forslund:
Do you mean like search all files for `create_skill()`?

Looking at a skill as a module it would require the `__init__.py` to show what's exported right?

chrisveilleux:
using `__init__.py` as a way to expose the API of a package is not a Python requirement.  Lots of folks just leave that file completely empty.  It is my preference the `__init__.py` be used in all of our code in this way.  So if we can get that to work with importlib, that would be ideal.  Ideally, we could import `create_skill()` from skill.py, load it in the skill loader and execute it.  I just don't know off the top of my head what is possible and what is not with importlib.  If we need to do something a bit different to support how we load skills the way we do, I am OK with that.

chrisveilleux:
SHIT.  Hit the close button accidentally.

forslund:
Gotcha.

In any case you're query highlighted an issue here. with this implementation relative imports no longer works in skills so I guess I need to look into that at least.

Will look at loading the skill from any file while I'm at it.

JarbasAl:
> Gotcha.
> 
> In any case you're query highlighted an issue here. with this implementation relative imports no longer works in skills so I guess I need to look into that at least.
> 
> Will look at loading the skill from any file while I'm at it.

relative imports haven't worked ever, i remember in the old times the usual suggestion in slack was to add the skill directory to the python path, i usually started from MYCROFT_ROOT_PATH but i remember seeing all sorts of ugly stuff with os.path and some approaches with skill variables

chrisveilleux:
In case there was any question, manipulating the Python path should be avoided at all costs.

ChanceNCounter:
> relative imports haven't worked ever, i remember in the old times the usual suggestion in slack was to add the skill directory to the python path, i usually started from MYCROFT_ROOT_PATH (remember when skill where part of core?) but i remember seeing all sorts of ugly stuff with os.path and some approaches with skill variables

I'm using relative imports in a skill I'm developing right now, and it seems to be working. Aside from this PR, is that going to break later? There seem to be conflicting experiences here.

forslund:
With the current `imp` implentation it works. I did some work last night and managed to get the relative imports working with importlib as well. It would be terrible if the skills were limited to a single file.

There is an issue with the relative imports not reloading when the `__init__.py` is reloaded (both in old imp and importlib implementations) but I think I've got a handle on how to tackle that issue.

forslund:
Pushed a WIP update supporting relative imports again and adding reloading for them as well...

forslund:
This has been updated to use importlib and a nice and easy fix to allow skill submodules to be reloaded.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2327))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2327))

mycroft/skills/skill_loader.py
```diff
@@ -29,6 +34,49 @@
 SKILL_MAIN_MODULE = '__init__.py'
 
 
+def remove_submodule_refs(module_name):
+    """Ensure submodules being reloaded by removing the refs from sys.modules.
```
krisgesling, line 38: #nitpick...

> Ensure submodule is reloaded by removing the refs from sys.modules.

mycroft/skills/skill_loader.py
```diff
@@ -29,6 +34,49 @@
...
+
+    Python import system puts a reference for each module in the sys.modules
+    dictionary to bypass loading if a module is already in memory. To make
+    sure skills are completely reloaded these references are deleted.
+
+    Arguments:
+        module_name: name of skill module.
+    """
+    submodules = []
+    LOG.info('Skill module'.format(module_name))
```
krisgesling, line 48: is this meant to be info level?

mycroft/skills/skill_loader.py
krisgesling, line 38: #nitpick...

> Ensure submodule is reloaded by removing the refs from sys.modules.

mycroft/skills/skill_loader.py
krisgesling, line 48: Is this intended to be an info level log message?

forslund: You're right that should be a debug level, will fix on my train ride home

mycroft/skills/skill_loader.py
forslund: Are you ok with

>Ensure submodules are reloaded by removing the refs from sys.modules.

krisgesling: Yep that reads great.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2327))

mycroft/skills/skill_loader.py
forslund: Comment updated

mycroft/skills/skill_loader.py
forslund: Changed to debug

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2327))

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2327))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2327))

Issue #2295, opened 2019-09-10
mycroft.format.nice_duration() output is awkward over long periods


ChanceNCounter:
For example, using nice_duration to check how long it's been since a datetime in early July of 2016 produced, "eleven sixty six days five hours one second."

If you just ask Mycroft that question, it kicks over to Wolfram, which produces the desired output. I realize that it's a nightmarish task trying to tackle days to years, given leap years and all, but I reckon it should at least move to hundreds: "eleven hundred sixty six days..." (or "one thousand one hundred sixty six days")

krisgesling:
Thanks for reporting this :)

If anyone is keen to tackle it and wants to handle leap years, I used the following in the Date-Time Skill recently:
```
def is_leap_year(self, year):
    return (year % 400 == 0) or ((year % 4 == 0) and (year % 100 != 0))
```

forslund:
There's also the `calendar.isleap` function in the standard library...

forslund:
The [humanize](https://pypi.org/project/humanize/) module might be something to look at as well...

```
>>> humanize.naturaldelta(timedelta(minutes=40, hours=360000))
'41 years'
```

ChanceNCounter:
If somebody _is_ going to refactor this function, it might be worth giving the thing resolution while they're in there. It's amusing and 100% in character for Mycroft to tell you how many seconds have passed since that day long ago, but coders might prefer to specify, `nice_duration_years()` or `nice_duration(dt1, dt2, <str>resolution)`

krisgesling:
Yeah I've been thinking about "resolution" for a number of these functions but don't want to make breaking changes. Seems like that could be less of an issue here if we only add an optional parameter and it still just returns a string, eg: 
`nice_duration(<int>seconds, <str>resolution)`

ChanceNCounter:
Because I'm not coding this myself, I'll make a thousand suggestions!

If these functions do get resolution, perhaps years could operate over trimesters, or consider how many months or etc. make the difference between "about 2 years," "almost 2 years," "a little over 2 years," or just "2 years"

ChanceNCounter:
I just started fussing with this in the interpreter, and I don't think it'll be as hard as I thought. If nobody else has started yet, I'll do it after dinner.

ChanceNCounter:
Fix merged.

PR #2330, opened 2019-09-27
Nice duration fix

ChanceNCounter:
**nice_duration heavily refactored:**

  - Original logic spun off into helper function
  - New function: nice_duration_dt
    - Operates on a pair of datetime.datetimes
    - Leap-year-aware
  - No longer produces strange output on large numbers
  - New optional parameters:
    - use_years (bool): optional
    - resolution (TimeResolution): optional, see below
    - clock (bool): optional, forces output to look like digital clock (displayed output only)
  - Attempts to display "pretty" output based on resolution:

```
>>> nice_duration(360, resolution=TimeResolution.SECONDS)
'six minutes'
>>> nice_duration(360, resolution=TimeResolution.SECONDS, speech=False)
'6:00'
>>> nice_duration(360, resolution=TimeResolution.MINUTES, speech=False)
'6m'
>>> nice_duration(360, resolution=TimeResolution.HOURS, speech=False)
'0h'
```
Companion enum:
mycroft.util.format.TimeResolution
offers YEARS, DAYS, HOURS, MINUTES, SECONDS, or MILLISECONDS

Will only return ms if MILLISECONDS is chosen. Default: SECONDS

==== Fixed Issues ====
#2295

====  Tech Notes ====
~~Will break anything calling nice_duration() using positional arguments.
I couldn't find anything that does that, so I don't think this will break
any code in the wild.~~

## How to test
Best to test in the interpreter. Don't forget to import `mycroft.util.format.TimeResolution`!
Other examples not in docstring:
```
>>> nice_duration(32655745.52, speech=False, resolution=TimeResolution.MILLISECONDS)
'1y 12d 23:02:25.52'
>>> nice_duration(32655745.52, speech=False, resolution=TimeResolution.MILLISECONDS, use_years=False)
'377d 23:02:25.52'
>>> nice_duration_dt(datetime(2019, 3, 12, 11, 15, 12), datetime(2019, 1, 1, 9, 7), resolution=TimeResolution.HOURS)
'seventy days two hours'
```

CLA [yes]

### DIFF
--- a/mycroft/util/format.py
+++ b/mycroft/util/format.py
@@ -468,7 +468,8 @@ def nice_year(dt, lang=None, bc=False):
 
 
 def _duration_handler(time1, lang=None, speech=True, *, time2=None,
-                      use_years=True, resolution=TimeResolution.SECONDS):
+                      use_years=True, clock=False,
+                      resolution=TimeResolution.SECONDS):
     """ Convert duration in seconds to a nice spoken timespan
         Used as a handler by nice_duration and nice_duration_dt
 
@@ -491,6 +492,7 @@ def _duration_handler(time1, lang=None, speech=True, *, time2=None,
         lang (str, optional): a BCP-47 language code, None for default
         speech (bool, opt): format output for speech (True) or display (False)
         use_years (bool, opt): rtn years and days if True, total days if False
+        clock (bool, opt): always format output like digital clock (see below)
         resolution (mycroft.util.format.TimeResolution, optional): lower bound
 
             mycroft.util.format.TimeResolution values:
@@ -502,6 +504,10 @@ def _duration_handler(time1, lang=None, speech=True, *, time2=None,
                 TimeResolution.MILLISECONDS
             NOTE: nice_duration will not produce milliseconds
             unless that resolution is passed.
+
+            NOTE: clock will produce digital clock-like output appropriate to
+            resolution. Has no effect on resolutions DAYS or YEARS. Only
+            applies to displayed output.
 
     Returns:
         str: timespan as a string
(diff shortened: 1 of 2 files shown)


ChanceNCounter:
I'm considering the merits of rounding minutes to hours when `resolution <= TimeResolution.HOURS`

mycroft/util/format.py
```diff
@@ -454,83 +468,256 @@ def nice_year(dt, lang=None, bc=False):
     return date_time_format.year_format(dt, full_code, bc)
 
 
-def nice_duration(duration, lang=None, speech=True):
+def duration_handler(time1, lang=None, speech=True, *, time2=None,
```
forslund: Let's mark this as an internal function by naming it `_duration_handler` then we can change the parameters and thing as we like without worrying about keeping the api stable.

ChanceNCounter: Pushed.

ChanceNCounter:
Hold off on merging. I've found a problem and I'm not finished diagnosing it.

ChanceNCounter:
That last commit should finally put a bow on it. I've got tests for most of the rest of util.format.py, but those can get their own PR.

forslund:
Thanks @ChanceNCounter, to you want to rebase and squash things into a logical order or should I just squash everything into a single commit?

ChanceNCounter:
@forslund interesting question. I think you might as well squash; I'll tag my branch, just in case, but I don't think there's any history here worth preserving.

ChanceNCounter:
One more round of edge cases coming. At least I'm finding them _before_ the merge.

forslund:
Squashed it down. Now merging.

Big thanks @ChanceNCounter !

PR #2334, opened 2019-09-28
gitignore vscode workspace and html coverage data

ChanceNCounter:
## Description
Adds local HTML output from coverage tests to `.gitignore`.

Also adds vscode workspace files. That's just for me, but I'm in here anyway, and surely I won't be the only workspace user in the long run :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 dev.env
 .dev_opts.json
 .idea
+*.code-workspace
 *.pyc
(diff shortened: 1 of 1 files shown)


ChanceNCounter:
If whoever reviews this would rather just edit the file inline, and close this, makes sense to me.

forslund:
Manually editing is hard! Merging this PR :)

Issue #2333, opened 2019-09-28
Extract "day after tomorrow" not working


moritzgloeckl:
Hi,

I'm running the latest docker image, with mycroft-core 18.8.4

In my skill's code I'm using `extract_datetime("day after tomorrow")` (for testing), this leads to the following error:

```
Traceback (most recent call last):
  File "/opt/mycroft/mycroft/skills/core.py", line 736, in wrapper
    handler(message)
  File "/opt/mycroft/skills/my-skill.mycroftai/__init__.py", line 25, in handle_read_tasks_intent
    LOG.error(extract_datetime("day after tomorrow"))
  File "/opt/mycroft/mycroft/util/parse.py", line 170, in extract_datetime
    return extract_datetime_en(text, anchorDate, default_time)
  File "/opt/mycroft/mycroft/util/lang/parse_en.py", line 397, in extract_datetime_en
    not wordPrev[0].isdigit()):
IndexError: string index out of range
```

Has this been fixed already or am I doing something wrong here?

Thanks.

ChanceNCounter:
I'm on it. Pushing a fix and PR now.

moritzgloeckl:
Hi @ChanceNCounter, thanks for fixing this so fast.

forslund:
The fix by @ChanceNCounter has been merged now. Closing this issue.

But @moritzgloeckl the docker project hasn't been maintained for some time so I'm not sure what's the best way for you to update. If you're experienced with docker and want to pick up the project we'd be happy for the help.

moritzgloeckl:
Hi @forslund 

Ok I did not know that it hasn't been maintained - it's working quite well for me, after some minor permission/venv issues. I am somewhat experienced (not a pro though) with docker, so I can take a look if you like.

The reason why I'm using docker is, because it's actually quite nice to develop skills with it. I start the container, attach a volume between my computer and the skills directory. I start the mycroft-cli in debug mode, and develop the skill from PyCharm on my computer. Once I hit save, it immediately reloads the skill and I can just type what I wanted to test, without having to speak to my pi constantly.

This discussion might go beyond this ticket, so we can discuss elsewhere on the chat/forum or something if you like. Then maybe we can chat about what needs to be done Docker-wise.

forslund:
I totally get it. feel free to reach out to me on the chat. The git repo for the docker is over here: https://github.com/MycroftAI/docker-mycroft/

PR #2335, opened 2019-09-28
Fixes index error in parse_en.extract_datetime()

ChanceNCounter:
==== Fixed Issues ====
#2333

====  Tech Notes ====
extract_datetime("day after tomorrow") attempted to check the previous
word. This resulted in an index error. Fixed by introducing a nested
condition. Seems to have solved it.

A similar bug is present in other parts of this file, such as line 843.

I have also discovered a number of other oddities, but they may be
outside the scope of this issue. At least Mycroft can now extract from
"day after tomorrow."

==== Localization Notes ====
I haven't checked parsers for any other languages.

## Description
Only fixes that specific bug. I can't stress that enough. This function will need a bunch more work.

## How to test
Unit tests pass, or just import `extract_datetime` and `extract_datetime("day after tomorrow")

CLA [ yes ]

### DIFF
--- a/mycroft/util/lang/parse_en.py
+++ b/mycroft/util/lang/parse_en.py
@@ -828,7 +828,7 @@ def extract_datetime_en(string, dateNow, default_time):
               wordNext == "after" and
               wordNextNext == "tomorrow" and
               not fromFlag and
-              not wordPrev[0].isdigit()):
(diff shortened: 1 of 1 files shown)


forslund:
Added a test vase for the failing scenario. Merging now.

Thanks for fixing this!

PR #2345, opened 2019-10-03
Feature/mic listen direct

forslund:
## Description
The listen trigger used a filesystem "signal" to indicate that the listener should trigger. This updates the "mycroft.mic.listen" signal to directly trigger the listening.

## How to test
Make sure the expect_response flag on speak commands (or get_response / ask_yesno) triggers the listening.

CLA [ Yes ]

### DIFF
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -41,8 +41,9 @@ def _start_listener(_):
 
 
 def handle_speak(event):
-    """
-        Handle "speak" message
+    """Handle "speak" message
+
+    Parse sentences and invoke text to speech service.
     """
     config = Configuration.get()
     Configuration.set_config_update_handlers(bus)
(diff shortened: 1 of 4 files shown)


mycroft/client/speech/mic.py
```diff
@@ -400,6 +405,10 @@ def _upload_wake_word(self, audio, metadata):
             }
         )
 
+    def trigger_listen(self):
```
davidwagnerkc: Probably good to have a log here to note when skipping the wake word loop is manually triggered. Here of in _skip_wake_word I guess.

mycroft/audio/speech.py
```diff
@@ -35,16 +35,15 @@
 _last_stop_signal = 0
 
 
-def _start_listener(message):
-    """
-        Force Mycroft to start listening (as if 'Hey Mycroft' was spoken)
-    """
-    create_signal('startListening')
+def _start_listener(_):
```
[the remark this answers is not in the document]
forslund: Yes it's used from [L66](https://github.com/MycroftAI/mycroft-core/pull/2345/files#diff-1d158d7b163e74b95987f0e65bda74b9R66)

The functionality is somewhat changed however since it emits the mycroft.mic.listen instead of setting the signal file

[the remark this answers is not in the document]
davidwagnerkc: :+1:

mycroft/client/speech/mic.py
forslund: Added debug level log message

davidwagnerkc: :+1:

PR #2353, opened 2019-10-04
Remove event before running event handler

forslund:
# Description
Allow events to reschedule themselves.
The reference was removed before the handler is executed to not remove anything added by the handler. This performes a slight restructuring clearing the event reference before the handler is called so any new reference set by the handler remains intact.

## How to test
Create a scheduled event that reschedules itself.

CLA [ Yes ]


krisgesling:
Nice! Working great

PR #2359, opened 2019-10-10
Make skill's logging visible in tester

forslund:
## Description
The skill logger was still set to the temporary logger, this resets it
to the intented logger name and handler

## How to test
The skill tester is running off of the branch. Check the output of the mycroft-skills skill tester that log info from the skill is shown.

CLA [ Yes ]

### DIFF
--- a/test/integrationtests/skills/skill_tester.py
+++ b/test/integrationtests/skills/skill_tester.py
@@ -45,6 +45,7 @@ from mycroft.skills.core import MycroftSkill, FallbackSkill
 from mycroft.skills.settings import Settings
 from mycroft.skills.skill_loader import SkillLoader
 from mycroft.configuration import Configuration
+from mycroft.util.log import LOG
 
 from logging import StreamHandler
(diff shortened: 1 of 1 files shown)


krisgesling:
Yep, working great.

Seems like you have to explicitly set log_level to DEBUG in your config, setting it using the command in the CLI doesn't have an effect. Is that expected?

forslund:
When running tests?

krisgesling:
Yeah, so on my local machine, if I want to see the debug level log messages I can use `log level debug` in the CLI. However if I then trigger the test runner I still only get info level messages.

To get debug level logs when running the skill-testrunner I need to explicitly set them using:
`mycroft-config set log_level DEBUG

forslund:
The log level set in the cli just affects the running session and doesn't affect the actual config. The skill tester doesn't connect to the actual messagebus at all so there's no communication with the CLI happening ever. So that is to be expected.

PR #2364, opened 2019-10-16
BUGFIX parsing of "this quarter of an hour"

penrods:
The test for "this" was occurring after the word had been wiped from the array.  It's a weird
and somewhat awkward edge case, but it could have resulted in the parse returning the date
of the following day rather than the current day.


forslund:
Thank you, could you also add a test case for the previously failing case?

PR #2366, opened 2019-10-18
Send listen flag for special picroft case as well

forslund:
## Description
Send listen flag for special picroft case as well

## How to test
Ensure that picroft starts listening automatically after a `get_response()` / speak() with `expect_response=True`

CLA [ Yes ]


krisgesling:
Working great.

PR #2370, opened 2019-10-24
Stabilize skill settings download

forslund:
## Description
This adds a queue system to capture the upload of settingsmeta while not paired and start the settings download so it doesn't start before pairing has been checked. A side effect is that skills load much faster into an usable state (though skill + latest settings downloaded from backend) are approximately the same time.

This handles the unpaired case better than the previous linear + start threads if pairing is missing.
It also handles the case where a skill is added through the command line msm application by making sure the skill manifest is updated and the skill cache is cleared.

Additionally removed skills will trigger an update of the home skills list directly.

We should give this a proper shakedown and test it thoroughly. I will complete the PR with a couple of test cases but it can definitely use an ocular review and some additional random testing in the meantime.

## How to test
Start up skills process, make sure it finishes, settingsmeta is uploaded and then downloaded without any errors that "skill_gid is not set" errors.

Install a skill from using the commandline and ensure that no errors occurs.

CLA [ Yes ]

### DIFF
--- a/mycroft/skills/skill_manager.py
+++ b/mycroft/skills/skill_manager.py
@@ -31,11 +31,12 @@ from .skill_updater import SkillUpdater
 SKILL_MAIN_MODULE = '__init__.py'
 
 
-class SettingsMetaQueue:
-    """Queue for sending settings metadata to backend.
+class UploadQueue:
+    """Queue for holding loaders with data that still needs to be uploaded.
 
-    This queue can be used during startup to capture all settingsmeta requests
-    and then processing can be triggered at a later stage.
+    This queue can be used during startup to capture all loaders
+    and then processing can be triggered at a later stage when the system is
+    connected to the backend.
 
     After all queued settingsmeta has been processed and the queue is empty
     the queue will set the self.started flag.
@@ -90,7 +91,7 @@ class SkillManager(Thread):
         self._stop_event = Event()
         self._connected_event = Event()
         self.config = Configuration.get()
-        self.upload_queue = SettingsMetaQueue()
+        self.upload_queue = UploadQueue()
 
         self.skill_loaders = {}
         self.enclosure = EnclosureAPI(bus)
@@ -155,6 +156,7 @@ class SkillManager(Thread):
 
     def _start_settings_update(self):
         LOG.info('Start settings update')
+        self.skill_updater.post_manifest(reload_skills_manifest=True)
         self.upload_queue.start()
         LOG.info('All settings meta has been processed or upload has started')
         self.settings_downloader.download()
@@ -162,7 +164,6 @@ class SkillManager(Thread):
 
     def handle_paired(self, _):
         """Trigger upload of skills manifest after pairing."""
-        self.skill_updater.post_manifest(reload_skills_manifest=True)
         self._start_settings_update()
 
     def load_priority(self):
@@ -192,9 +193,8 @@ class SkillManager(Thread):
         self._connected_event.wait()
         self._load_on_startup()
 
-        # Update sync backend and skills.
-        if is_paired():
-            self.skill_updater.post_manifest(reload_skills_manifest=True)
+        # Sync backend and skills.
+        if is_paired() and not self.upload_queue.started:
             self._start_settings_update()
 
         # Scan the file folder that contains Skills.  If a Skill is updated,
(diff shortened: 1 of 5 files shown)


mycroft/skills/skill_manager.py
```diff
@@ -30,6 +31,51 @@
 SKILL_MAIN_MODULE = '__init__.py'
 
 
+class SettingsMetaQueue:
```
chrisveilleux: I think it would make more sense to put this in settings.py.

mycroft/skills/skill_manager.py
```diff
@@ -138,8 +193,9 @@ def run(self):
         self._load_on_startup()
 
         # Update sync backend and skills.
-        self.skill_updater.post_manifest(reload_skills_manifest=True)
-        self.settings_downloader.download()
+        if is_paired():
+            self.skill_updater.post_manifest(reload_skills_manifest=True)
```
chrisveilleux: These two lines are duplicated in handle_paired().  Can we just call that method here?  Alternatively, write a new method both handle_paired and this method can call?

mycroft/skills/skill_manager.py
```diff
@@ -172,36 +233,36 @@ def _load_on_startup(self):
...
     def _reload_modified_skills(self):
         """Handle reload of recently changed skill(s)"""
-        reload_occured = False
         for skill_dir in self._get_skill_directories():
             try:
                 skill_loader = self.skill_loaders.get(skill_dir)
                 if skill_loader is not None and skill_loader.reload_needed():
                     skill_loader.reload()
-                    reload_occured = True
+                    if skill_loader.instance:
```
chrisveilleux: use skill_loader.loaded instead.

mycroft/skills/skill_manager.py
```diff
@@ -172,36 +233,36 @@ def _load_on_startup(self):
...
-        if reload_occured:
-            # If a reload occured a skill gid may have changed.
-            self.skill_updater.post_manifest(reload_skills_manifest=True)
-
     def _load_new_skills(self):
         """Handle load of skills installed since startup."""
         for skill_dir in self._get_skill_directories():
             if skill_dir not in self.skill_loaders:
-                self._load_skill(skill_dir)
+                sl = self._load_skill(skill_dir)
```
chrisveilleux: I want to stay away from over-abbreviated variable names like this one.  Could we call this "loader" or "skill_loader"?

test/integrationtests/skills/skill_tester.py
```diff
@@ -140,7 +140,7 @@ def load_skills(emitter, skills_root):
         from mycroft.util.log import LOG as skills_log
         buf = StringIO()
         with temporary_handler(skills_log, StreamHandler(buf)):
-            skill_loader = SkillLoader(emitter, path)
+            skill_loader = SkillLoader(emitter, path, None)
```
chrisveilleux: The skill loader only takes two arguments.  How does this work?

mycroft/skills/skill_manager.py
forslund: That's where it's originally was located but I feel like since it's nothing that needs to be shared and only is a helper class for the skill_manager it's better to keep it close to where it's actually used.

test/integrationtests/skills/skill_tester.py
forslund: Good catch, it probably doesn't. will fix

forslund: Fixed

mycroft/skills/skill_manager.py
forslund: Used the return value of the reload method directly instead.

mycroft/skills/skill_manager.py
chrisveilleux: I think it is more important to group code by function than by where it is used.  Helps discoverability, IMO

forslund: To me it doesn't make sense, it's more related to the skill manager than the settings system. In the settings.py it would just add confusion.

forslund: Maybe it should be renamed...

forslund: Renamed it UploadQueue

forslund:
Anything else at this point?

PR #2382, opened 2019-11-09
Fix mimic build on arm and aarch64

cbxbiker61:
## Description
Fixed mimic build on arm and aarch64.  These are memory constrained systems and this change allows the build to finish successfully on these systems.

## How to test
Install/build on arm and/or aarch64 that doesn't have a mimic binary available.

No, but I accept the terms of the Agreeement.

### DIFF
--- a/scripts/install-mimic.sh
+++ b/scripts/install-mimic.sh
@@ -21,6 +21,15 @@ MIMIC_DIR=mimic
 CORES=$1
 MIMIC_VERSION=1.2.0.2
 
+# This is important on memory constrained systems such as arm and aarch64
+if [[ $(uname -m) == arm* ]] || [[ $(uname -m) == aarch64 ]]; then
+  CORES=1
+fi
+# Avoid writing preprocessor output in /tmp
+# This will avoid compile fail and cut down build time
+# This is important on memory constrained systems such as arm and aarch64
+export CFLAGS="-pipe"
+
 # for ubuntu precise in travis, that does not provide pkg-config:
 pkg-config --exists icu-i18n || export CFLAGS="$CFLAGS -I/usr/include/x86_64-linux-gnu"
 pkg-config --exists icu-i18n || export LDFLAGS="$LDFLAGS -licui18n -licuuc -licudata"


devs-mycroft:
Hello, @cbxbiker61, thank you for helping with the Mycroft project! We welcome everyone
for Everyone.

you!

el-tocino:
The number of cores for building is derived from the amount of system memory available:
https://github.com/MycroftAI/mycroft-core/blob/dev/dev_setup.sh#L441

cbxbiker61:
el-tocino, It's fine that the number of cores is based on system memory available.  In the case of mimic that is too many cores and the build will fail.  hence the change to one core on arm and aarch64.  As well the addition of piping intermediate output.  Intermediate output is typically written to /tmp, which on most systems today is a ram filesystem, which will run the system out of memory and the build will fail.

cbxbiker61:
When I say the build will fail...that actually means the system will crash and reboot.  If mycroft is starting automatically, it will continue to boot/crash/reboot.

el-tocino:
What are you building on that's displaying that sort of behavior when run through the dev_setup?

cbxbiker61:
Primarily 64bit RasPi 3 or RasPi 4.  Either a 1G or 2G model will exhibit this particular issue.

The patch should make this problem go away for any non-std (doesn't have a binary mimic available for download) arm or aarch64 system.

forslund:
Sorry for the delay on this, could the original calulation be made better to result in a max core of 1 for these platforms?

vorpalhex:
I was able to recreate the issue described here (under #2694 ) on a Pi 4 with 2gb of ram, and can confirm the default single core build seems to fix the problem.

vorpalhex (approved):

krisgesling:
Hi there, if anyone is interested in digging into this. It would be good to see if we can improve the core calculation located here: https://github.com/MycroftAI/mycroft-core/blob/e756f96d9df61620a685b476d0ab570d9e8306ce/dev_setup.sh#L538

This would be a preferred solution over a patch for specific systems.

el-tocino:
A check to see if there's 2gb of RAM or less and settings cores=1 would likely cover this, though on larger memory pi4's not sure how they'd handle it. Perhaps instead of 512mb we divide by >=1.5gb?

el-tocino:
Looks like dividing by 2.1gb instead of 512mb would make the builds max at 3 cores on an 8gb pi, and 1 on all the rest. Probably the quickest solution.  

Secondarily, there should really be a package to download for mimic1 on aarch32/x86 instead of forcing it to be built every time.  That would also prevent a large chunk of these issues.

forslund:
I host mimic packages on the [mycroft-alternatives repo](https://github.com/forslund/mycroft-alternatives) Maybe those can be used. #2716 has now been merged so I'll close this PR.

Thanks for contributing, making us aware of the issue and proposing this solution!

PR #2386, opened 2019-11-14
add some more time idioms

maxbachmann:
## Description
- add some time idioms like 'viertel vier'/ 'quarter past three' for the nice_time function, which was a TODO
- improve code readability

## How to test
The changes to the german version are covered by unit tests.
For the english version there are no unit tests yet.

CLA [x] signed now

### DIFF
--- a/mycroft/util/lang/format_de.py
+++ b/mycroft/util/lang/format_de.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 #
 # Copyright 2017 Mycroft AI Inc.
 #
@@ -28,25 +27,25 @@ NUM_STRING_DE = {
     2: 'zwei',
     3: 'drei',
     4: 'vier',
-    5: u'fünf',
+    5: 'fünf',
     6: 'sechs',
     7: 'sieben',
     8: 'acht',
     9: 'neun',
     10: 'zehn',
     11: 'elf',
-    12: u'zwölf',
+    12: 'zwölf',
     13: 'dreizehn',
     14: 'vierzehn',
-    15: u'fünfzehn',
+    15: 'fünfzehn',
     16: 'sechzehn',
     17: 'siebzehn',
     18: 'achtzehn',
     19: 'neunzehn',
     20: 'zwanzig',
-    30: u'dreißig',
+    30: 'dreißig',
     40: 'vierzig',
-    50: u'fünfzig',
+    50: 'fünfzig',
     60: 'sechzig',
     70: 'siebzig',
     80: 'achtzig',
@@ -68,17 +67,17 @@ FRACTION_STRING_DE = {
     2: 'halb',
     3: 'drittel',
     4: 'viertel',
-    5: u'fünftel',
+    5: 'fünftel',
     6: 'sechstel',
     7: 'siebtel',
     8: 'achtel',
     9: 'neuntel',
     10: 'zehntel',
     11: 'elftel',
-    12: u'zwölftel',
+    12: 'zwölftel',
     13: 'dreizehntel',
     14: 'vierzehntel',
-    15: u'fünfzehntel',
+    15: 'fünfzehntel',
     16: 'sechzehntel',
     17: 'siebzehntel',
     18: 'achtzehntel',
@@ -250,7 +249,7 @@ def pronounce_ordinal_de(num):
     # this produces the base form, it will have to be adapted for genus,
     # casus, numerus
 
-    ordinals = ["nullte", "erste", "zweite", "dritte", "vierte", u"fünfte",
+    ordinals = ["nullte", "erste", "zweite", "dritte", "vierte", "fünfte",
                 "sechste", "siebte", "achte"]
 
     # only for whole positive numbers including zero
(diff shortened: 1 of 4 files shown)


devs-mycroft:
Hello, @maxbachmann, thank you for helping with the Mycroft project! We welcome everyone
for Everyone.

you!

forslund:
Thanks @maxbachmann It does look good. Could you add a couple of testcases for the english half past and quarter past (in `test/unittests/util/test_format.py`)

Maybe @krisgesling or @chrisveilleux who are native english speakers can do the final review on the changes so they're all ok. (I'm unsure about using 12 for midnight for example but I'm used to 24 hour clocks)

maxbachmann:
@forslund sure I can add testcases for the english part later.

on the 12 hour times I was unsure aswell.
I guess in german I would usually say:
  halb/dreiviertel/halb/viertel nach 12
but
 0 Uhr zwanzig (or even 10 vor halb eins)

maxbachmann:
For comparision: https://github.com/project-alice-powered-by-snips/ProjectAliceModules/blob/master/PublishedModules/Psychokiller1888/DateDayTimeYear/DateDayTimeYear.py is how I currently implemented this for ProjectAlice (without am/pm support right now)

krisgesling:
Haven't reviewed yet, but 12 is a normal way of talking about midnight including:
> 5 to 12
> half past 12
> 12 midnight
> 12 o'clock tonight
> quarter to 12

forslund:
Thanks for verifying @krisgesling, always good to have the input of someone who speaks the language everyday. I don't quite trust my own skills with the language.

@maxbachman I'll wait for the extra test cases, after that we can merge this.

Many thanks for contributing!

maxbachmann:
I added unittests however I am not sure on two things:

test/unittests/util/test_format_en.py
```diff
@@ -0,0 +1,370 @@
...
+                         "00:02")
+        self.assertEqual(nice_time(dt, speech=False,
+                                   use_24hour=True, use_ampm=True),
+                         "00:02")
+        self.assertEqual(nice_time(dt, use_24hour=True,
+                                   use_ampm=True),
+                         "zero zero zero two")
+        self.assertEqual(nice_time(dt, use_24hour=True,
+                                   use_ampm=False),
+                         "zero zero zero two")
```
maxbachmann: is the time really spoken this way? with the zero sounds awkward to me, but I am not a antive english speaker

test/unittests/util/test_format_en.py
```diff
@@ -0,0 +1,370 @@
...
+        self.assertEqual(pronounce_number(-21.234, places=1),
+                         "minus twenty one point two")
+        self.assertEqual(pronounce_number(-21.234, places=0),
+                         "minus twenty one")
+        self.assertEqual(pronounce_number(-21.234, places=3),
+                         "minus twenty one point two three four")
+        self.assertEqual(pronounce_number(-21.234, places=4),
+                         "minus twenty one point two three four")
+        self.assertEqual(pronounce_number(-21.234, places=5),
+                         "minus twenty one point two three four")
```
maxbachmann: why does it add `null` in german  when the precision is higher than the precision of the given number, while in english it does not add `zero` and `zero zero` in the end?

test/unittests/util/test_format_en.py
maxbachmann: Yup apparently it is google tells me xD

maxbachmann:
since I already cleaned up the unicode strings and file encoding for the files I updated, I did this globally (should this be in the same PR?)
Another thing I wondered is why does /doc/conf.py enforce python2? It thought the project dropped python2 support

test/unittests/util/test_format_en.py
ChanceNCounter: "minus twenty one" is a perfectly ordinary, colloquial way of phrasing it in English, but I should point out that (at least to an American) it comes across as extremely informal, as colloquial as it gets. It might be worth considering, "negative twenty one," which is the most common phrasing.

"Minus x," in my experience, is most frequently used when referring to weather. Two minus three is negative one, but a cold temperature would be, "minus one degree."

maxbachmann: thats not what I mean:
I mean why in german it is:
`minus einundzwanzig komma zwei drei vier null null`
and in english
`minus twenty one point two three four`
and not 
`minus twenty one point two three four zero zero`
when precision is set to 5

forslund:
It'd be better if it was in a separate PR but this is fine.

doc/conf.py is not used as a main file so it doesn't really matter but it seems like it was a change we missed when upgrading to python 3.

Can you join your English tests with the rest of the english tests in `test_format.py`?

maxbachmann:
@forslund oh I did not realise you already had english test cases. I added them to the correct file.

The python2 cleanup changes can be found in the https://github.com/MycroftAI/mycroft-core/pull/2396 now

Is the beahviour I described in my code comment correct? It appears to me, that it is wrong in one of the two languages

maxbachmann:
should the same PR be created for lingua_franca?

forslund:
Thank you.

Hmm, the number of places thing? from thinking about it briefly the english looks wrong...But reading the docstring in the format_en.py file says: 

`places(int): maximum decimal places to speak`

which makes the english version seem to follow the specification. On the other hand the entry function in `format.py` doesn't document that argument so technically anything goes but it should be specified to something...

We could raise a separate issue for that.

I think this is a decent state now. I'll do a final look over tomorrow morning and merge it.

maxbachmann:
I added an issue on the topic: https://github.com/MycroftAI/mycroft-core/issues/2397

forslund:
Changes looks good! Thanks for rebasing and providing a nice clean update! Merging

forslund:
And yes, I think you should create the same PR for lingua franca, the intention is to switch over to using that instead of the internal utils in the future.

PR #2389, opened 2019-11-19
Bugfix/audioservice mimetypes

forslund:
## Description
Fix issue when fetching mimetype using request header returns an error.

CLA [ Yes ]

### DIFF
--- a/mycroft/audio/services/simple/__init__.py
+++ b/mycroft/audio/services/simple/__init__.py
@@ -90,6 +90,8 @@ class SimpleAudioService(AudioBackend):
         else:  # Assume string
             track = self.tracks[self.index]
             mime = find_mime(track)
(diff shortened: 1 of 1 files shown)


mycroft/audio/services/simple/__init__.py
```diff
@@ -20,17 +20,22 @@
...
 import mimetypes
+import re
 from requests import Session
 
 
 def find_mime(path):
     mime = None
     if path.startswith('http'):
         response = Session().head(path, allow_redirects=True)
+        print(response.status_code)
```
penrods: Is this just for debugging?  Should you log instead?

forslund: It should be removed. Good catch

forslund: Commit updated.

PR #2392, opened 2019-11-22
Issue 2120 - Fallback to 'en-us' when resource loading fails

domcross:
## Description
When mycroft_skill.find_resource fails to load a resource for self.lang fall back to lang 'en-us' as most skills/resources are available in english by default.

==== Fixed Issues ====
MycroftAI#2120

## How to test
There are several skills missing some intent-files for 'de-de' (among them is the mycroft-weather skill). Without this PR loading of these skills will fail. When PR is applied all skills will load, but some WARNINGs will be logged.

CLA [X] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

### DIFF
--- a/mycroft/skills/mycroft_skill/mycroft_skill.py
+++ b/mycroft/skills/mycroft_skill/mycroft_skill.py
@@ -676,7 +676,7 @@ class MycroftSkill:
             string: The full path to the resource file or None if not found
         """
         result = self._find_resource(res_name, self.lang, res_dirname)
-        if not result:
+        if not result and self.lang != 'en-us':
             # when resource not found try fallback to en-us
             LOG.warning(
                 "Resource '{}' for lang '{}' not found: trying 'en-us'"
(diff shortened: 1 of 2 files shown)


mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -675,9 +675,22 @@ def find_resource(self, res_name, res_dirname=None):
         Returns:
             string: The full path to the resource file or None if not found
         """
+        result = self._find_resource(res_name, self.lang, res_dirname)
+        if not result:
```
forslund: Can you add `and self.lang != 'en-us'` to this if-statement so it doesn't try to get English resources twice if the file is missing.

forslund:
Tested some more and this doesn't handle dialogs but that might be outside the scope of the PR?

domcross:
Yes, dialogs are not handled by this. Main scope of this PR is that skill loading does not fail in case a intent file is missing. 
I had a quick look at the DialogRender class, so here we go ;-)

forslund:
I merged the commits needed to load the skills through #2411, I had some issues with the dialogs so we might need to go over those again. Thanks for contributing this and I hope the German experience improves (I know my Swedish experience will!)

If you like to continue the work on dialog and .voc files we can discuss the issues further. In the mean time I'm closing this PR and you/we can move the last commit over to a new branch based on the latest dev branch to continue the work.

Once more, many thanks.

PR #2393, opened 2019-11-22
Feature/geolocation api

chrisveilleux:
## Description
Add a call to the new Selene endpoint that returns geolocation data given a location string

## How to test
Execute a skill that requires geolocation data in one of its intents, like "what time is it in London"

CLA [Y]


forslund:
Looks good, couldn't even find a spelling error :) Merging

PR #2402, opened 2019-11-28
improve runtime

maxbachmann:
## Description
This PR improved the runtime of the pronounce_number_en function.
Depending on the arguments provided it is 2-5 times as fast, mainly by reducing the need to copy the dicts from common_data_en.py and type conversions.

## How to test
The behavior stays the same and is already covered by tests

CLA [ Yes ]

### DIFF
--- a/mycroft/util/lang/format_en.py
+++ b/mycroft/util/lang/format_en.py
@@ -72,7 +72,7 @@ def pronounce_number_en(num, places=2, short_scale=True, scientific=False):
     For example, '5.2' would return 'five point two'
 
     Args:
-        num(float or int): the number to pronounce (under 100)
+        num(float or int): the number to pronounce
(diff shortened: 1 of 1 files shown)


forslund:
This looks great,

not related to your work in this PR but I do think the docstring for the `num` is incorrect. It states

>num(float or int): the number to pronounce (under 100)

Would you mind handling this while you're touching this code. If you don't have the time I can do it later.

maxbachmann:
@forslund so this should probably be changed to
```
num(float or int): the number to pronounce
```
since it can be anything between -inf and +inf

forslund:
That sounds good :+1:

maxbachmann:
ok applied

forslund:
Many thanks, Merging now.

PR #2404, opened 2019-11-29
Remove cryptography dependency

forslund:
## Description
The dependency was previously added to try to suppress the SSL version
warning on the mark-1. This has been resolved via upgrading the mark-1
so the requirement is not of any importance right now.

CLA [ Yes ]


forslund:
This has been running on the Mark-1 unstable for the past couple of days and had a preliminary approval by @davidwagnerkc so I'm going to merge it.

PR #2405, opened 2019-11-30
improve test runtime

maxbachmann:
## Description
Improve runtime of test_ssml_support by 80%, since it is enough to update the supported ssml tags in between the tests.

## How to test
unit tests are still passing but faster

CLA [ Yes]


forslund:
Sorry for the delay. I was pondering if directly changing the internal values of the TTS object is really a good thing buuut the member isn't marked as no-tampering (leading _) so I guess this is alright to do. (In the future we might consider changing that and then the test will have to be updated)

As always, thaks for contributing :) Merging.

PR #2409, opened 2019-12-02
Restore Ctrl+C exit of simple CLI

penrods:
At some point during refactoring the ability to hit Ctrl+C to exit the basic CLI was lost.

## How to test
Run:
```mycroft-cli-client --simple```
Then hit Ctrl+C.  Before this change nothing happens (you just see the "^C" printed out).  After the change it exits as expected.

### DIFF
--- a/mycroft/client/text/__main__.py
+++ b/mycroft/client/text/__main__.py
@@ -13,13 +13,15 @@
 # limitations under the License.
 #
 import sys
+import signal
 import io
 import os.path
 import curses
 from mycroft.util import get_ipc_directory
 from .text_client import (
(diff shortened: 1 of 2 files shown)


forslund:
I think it's better to move the override of the ctrl+c handler for the normal cli to the __main__.py after the simple flag is checked. See this [quick mockup](https://github.com/forslund/mycroft-core/commit/f06b67ade66052354fbf3f555b609f8504f43057) for what I mean.

penrods:
Ah, yep!  That explains the current odd behavior in the simple CLI much better.  I'll adjust this PR.

forslund:
Looks all good, squashed it and will merge now!

Issue #2407, opened 2019-11-30
python version redhat


maxbachmann:
dev_setup.sh currently installs:
```
python34 python34-devel python34-pkgconfig python34-libs
```
on centos/redhat even though python 3.4 was already dropped from mycroft

maxbachmann:
I just tried it in a centos docker container. There it was enough to install
```bash
python3-devel
```
and the other python stuff was installed as dependency. (it installed python3.6 and python3.4 was not even available in the package manager)

forslund:
If you have a change that works, feel free to submit. If you want someone to try it *cschantz* on the mycroft chat was quite involved in testing the last version.

maxbachmann:
for testing on centos the docker container should be enough, however since the same install function is run on redhat, it would be good if someone could test it over there. I will open a PR and ask cschantz to test it.

PR #2412, opened 2019-12-05
use default python version on redhat

maxbachmann:
## Description
Fixes https://github.com/MycroftAI/mycroft-core/issues/2407
right now the python version is hardcoded to 3.4.
This PR replaces this with python3-devel, which includes all the other packages that are removed.
I only tested this in a centos7/8 docker container and not with redhat

## How to test
Test whether dev_setup.sh runs fine on centos and redhat

CLA [Yes ]

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -298,7 +298,7 @@ function os_is_like() {
 }
 
 function redhat_common_install() {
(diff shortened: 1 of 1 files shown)


forslund:
As far as I can tell this should work on redhat as well. I'll merge it, if issues with redhat occurs we'll handle it at that date :)

Thanks for fixing!

PR #2414, opened 2019-12-06
Fix --clean option of dev_setup.sh

forslund:
## Description
Removing the tmp files needed the recursive flag.

CLA [ yes ]


forslund:
Typo, selfmerging

PR #2415, opened 2019-12-06
Send mycroft.audio.queue_end

forslund:
## Description
The message is sent when there are no more tracks in the queue and
playback has stopped.

The intended use of this is for example restore the resting screen on the Mark-2 after a playlist has been completely played.

## How to test
Use the feature/restore-resting branch of the playback-control skill and make sure the log message is printed in the CLI when for example the news ends.

CLA [ Yes ]


JarbasAl:
this seems to be working good :+1:

PR #2422, opened 2019-12-09
Refactor mycroft.dialog

forslund:
## Description
- Replace the DialogLoader class with a simple function.
- move code from `__init__.py` to a dialog.py

Should this still exist as a mycroft.dialog module in the long term? It's quite a small but provides one of the core features of mycroft. With #2436 it's likely to grow in complexity so it may have some merit to keep it as it is.

## How to test
Make sure tests pass and that skills can speak responses still.

CLA [ Yes ]


forslund:
ping @chrisveilleux

PR #2424, opened 2019-12-12
Fix incorrect porcupine shutdown

forslund:
## Description
Fix porcupine issue flagged by flake8 after merge.


forslund:
Verified by Flake8. Self merging.

PR #2429, opened 2019-12-17
mark1/fix display big images

JarbasAl:
i have been having trouble with things not rendering in the faceplate, after a lot of frustration after some digging i found this

https://github.com/MycroftAI/enclosure-mark1/blob/master/protocols.txt#L141

according to the spec there is a bug, i have not tested this however (mark1 is not dev friendly!)

i have a game of life skill and some other [goodies](https://github.com/OpenJarbas/jarbas_utils/tree/dev/jarbas_utils/mark1) waiting on this :)

NOTE: pr made directly in browser, sorry if something is messed up

### DIFF
--- a/mycroft/client/enclosure/mark1/mouth.py
+++ b/mycroft/client/enclosure/mark1/mouth.py
@@ -108,11 +108,8 @@ class EnclosureMouth:
         message = "mouth.icon=" + x_offset + y_offset + clear_previous + code
         # Check if message exceeds Arduino's serial buffer input limit 64 bytes
         if len(message) > 60:
-            message1 = message[:31]
-            message2 = message[31:]
-            message1 += "$"
-            message2 += "$"
-            message2 = "mouth.icon=" + message2
+            message1 = message[:31] + "$"
+            message2 = "mouth.icon=$" + message[31:]
             self.writer.write(message1)
             time.sleep(0.25)  # writer bugs out if sending messages too rapidly
             self.writer.write(message2)


forslund:
I tried this on the device but it doesn't improve I'm afraid :(

Looking at the enclosure code [here](https://github.com/MycroftAI/enclosure-mark1/blob/master/lib/MycroftMouth/MycroftMouth.cpp#L187), it seems like the $ needs to be at the end of the code.

It could be the way the code is cut...looks like the second part gets very close to 64 character limit.

JarbasAl:
i have finally tested this on my mark1, it does not work with or without this change

this still makes things according to the spec i linked, so at least that needs to be updated

i think it makes more sense like in the spec than like it is coded in your link, so which one should be changed?

more importantly, how do we fix the actual bug?

forslund:
Like you I do think the approach with the $ sign in the beginning is better.

it also adds the possibility for some sanity checks

- if end of code is $ copy code and offsets to storage variables
- if starts with $
  - case 1 buffer with store exists, concatenate with stored buffer and reset buffer.
 -  case 2 the store is missing, drop current message
- if a message without $ first or last arrives reset the storage buffer



Then it does look like the $ aren't stripped from the code when storing the first part of the message ([here](https://github.com/MycroftAI/enclosure-mark1/blob/master/lib/MycroftMouth/MycroftMouth.cpp#L198)) which may be the problem.

forslund:
Did a quick test of the first part of how it handles the $-sign in the first part: https://gist.github.com/forslund/f4fb0698ed508b6db553ea2b153bd34a

PR #2431, opened 2019-12-17
Switch to the messagebus client module

forslund:
## Description
Use the external mycroft-messagebus-client pip package to keep the messagebus client synced across applications.

## How to test
Ensure that the testcases still work and that mycroft is still working as expected.

CLA [ Yes ]


ChanceNCounter:
Tested, working on Linux/Python3.7

PR #2433, opened 2019-12-19
Bugfix/settings sync

forslund:
## Description
During release testing two issues were identified with regard to settings meta upload
- A race condition between the pairing start of settingsmeta upload and skill loading upload
- Priority skills settingsmeta weren't uploaded

This resolves the first issue by adding a check that the started parameter has been set in the upload queue before attempting to upload settings after skill loading.

The second basically adds the skill loader to the queue when the settingsmeta when loading priority skills.

CLA [ Yes ]


forslund:
Verified on device. Self-merging

PR #2300, opened 2019-09-11
Properly call found_date check

forslund:
## Description
The check for no date or time in the sentence was never called, this calls it properly and adds a few simple tests.

## Contributor license agreement signed?
CLA [ Yes ]

PR #2438, opened 2019-12-23
Switch language formatting and parsing to use Lingua Franca

forslund:
## Description
This aims to replace the internal language parsing and formatting with the [lingua franca](https://github.com/MycroftAI/lingua-franca/) module. It still aims to be backwards compatible so some logic remains in core along with all python modules.

- It leaves all python intact but replaces every class and function defined under mycroft/util/lang with imports from lingua-franca.
- util/(format|parse).py still contains some logic but mainly refers to lingua-franca
- the data files in /mycroft/res has been completely removed.

## How to test
Run core make sure things behaves. Do some explorative testing.

CLA [ Yes ]


forslund:
The coverage will drop and is acceptable.

mycroft/util/format.py
```diff
@@ -20,79 +20,38 @@
...
-from mycroft.util.lang.format_da import nice_number_da
-from mycroft.util.lang.format_da import nice_time_da
-from mycroft.util.lang.format_da import pronounce_number_da
+from lingua_franca.format import (NUMBER_TUPLE, DateTimeFormat, join_list,
+                                  date_time_format, expand_options,
+                                  _translate_word)
+
+from lingua_franca.lang import get_full_lang_code, get_primary_lang_code
+
+from lingua_franca.lang.format_en import *
```
chrisveilleux: This pattern concerns me.  The wildcard imports are a Python anti pattern.  Also, this import list can get unwieldy as we add more language support.  Why not have nice_number, nice_time and pronounce_number functions that take a language parameter?

forslund:
Yes it should but...until a major release we need to keep the core-api as is.

mycroft/util/format.py
forslund: This is exactly what the file you're in does.

It is kept like this for backwards compatibility so it exports the expected symbols. In future releases this file will probably remain just importing the language-agnostic functions from lingua-franca.

chrisveilleux: Ok, please add a comment or TODO indicating this.

ChanceNCounter:
Seems to revert something about extract_datetime that breaks the weather skill. To repro: "Hey Mycroft, what's the weather like?" `NoneType is not subscriptable`

forslund:
Thanks for reporting @ChanceNCounter totally reproducible. will look into and fix.

forslund:
Hmm, seems like the lingua-franca behavior is the intended behavior: https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/util/parse.py#L248

forslund:
Likely something similar to this fix: https://github.com/MycroftAI/mycroft-core/pull/2300

PR #2449, opened 2020-01-08
make stop-mycroft.sh also stop if skills are started from other module.

andlo:
## Description
Make stop-mycroft.sh also stop if skills are started from other module.
If mycroft.skills are started from other module like ptvsd this make the stop function still posible to stop the skills service.

This is needed if running ptvsd remote debug to make sure skills are stoped as usal even though the proces is
```
python3 -m ptvsd --host 0.0.0.0 --port 5678 -m mycroft.skills
```
and not as normal
```
python3 -m mycroft.skills
```

## How to test
Starting and stopping and restarting skills should still work as usal

### DIFF
--- a/start-mycroft.sh
+++ b/start-mycroft.sh
@@ -106,7 +106,7 @@ function launch-process() {
 function require-process() {
     # Launch process if not found
     name-to-script-path ${1}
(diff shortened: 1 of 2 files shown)


forslund:
Thanks @andlo, Looks and works as expected. Merging.

PR #2453, opened 2020-01-13
Add readthedocs docs for settings_change_callback

forslund:
## Description
This was not clearly stated any where and the question popped up on the chat during the weekend.

CLA [ Yes ]


krisgesling:
Hey, what do you think about:
> Set to register a method that will be called every time settings for the Skill are changed.
> The referenced method should include any logic needed to handle the updated settings.

If you prefer the original, can I be pedantic and ask for a full stop at the end of the first line.

forslund:
Wording updated according to your suggestion. (still kept the "callback" in there)

PR #2461, opened 2020-01-22
Feat/message targetting

JarbasAl:
I could swear i made this PR before, but could not find it so here it goes again.

This is needed for the hive mind

# Description
The speech handling now checks the message context if it's the intended target for the message and will only speak in the following conditions:

- Explicitly targeted i.e. the destination is "audio"
- destination is set to None
- destination is missing completely

The idea is that for example when the android app is used to access Mycroft the device at home shouldn't start to speak.

# Targeting Theory

The context target parameter in the original message can be set to list with any number of intended targets:

```python
bus.emit(Message('recognizer_loop:utterance', data, 
				 context={'destination': ['audio', 'kde'],
						  'source': "remote_service"))
```

A missing destination or if the destination is set to None is interpreted as a multicast and should trigger all output capable processes (be it the mycroft-audio process, a web-interface, the KDE plasmoid or maybe the android app)

## Messages

- new message.forward method, keeps previous context.
	- message continues going to same destination
- message.reply method now swaps "source" with "destination"
	- message goes back to source

### Inside mycroft-core
- stt / cli will use "audio" or "debug_cli" as source
- intent service will .reply to utterance message
- all skill/intent service messages are .forward (from previous intent service .reply)
- cli is for debug, will print all utterances
- audio will check if "audio" or "debug_cli" is the destination

## How to test

- STT utterances trigger TTS
- Cli utterances trigger TTS
- [Remote Clients](https://github.com/JarbasAl/hive_mind) do not trigger TTS

### DIFF
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -44,6 +44,13 @@ def handle_speak(event):
     Configuration.set_config_update_handlers(bus)
     global _last_stop_signal
 
+    # if the message is targeted and audio is not the target don't
+    # don't synthezise speech
+    if (event.context and 'destination' in event.context and
+            event.context['destination'] and
+            'audio' not in event.context['destination']):
+        return
+
     # Get conversation ID
     if event.context and 'ident' in event.context:
         ident = event.context['ident']
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -46,9 +46,10 @@ def handle_speak(event):
 
     # if the message is targeted and audio is not the target don't
     # don't synthezise speech
-    if (event.context and 'destination' in event.context and
-            event.context['destination'] and
-            'audio' not in event.context['destination']):
+    event.context = event.context or {}
+    if event.context.get('destination') and not \
+            ('debug_cli' in event.context['destination'] or
+             'audio' in event.context['destination']):
         return
 
     # Get conversation ID
--- a/mycroft/client/speech/__main__.py
+++ b/mycroft/client/speech/__main__.py
@@ -35,24 +35,32 @@ config = None
 def handle_record_begin():
     """Forward internal bus message to external bus."""
     LOG.info("Begin Recording...")
-    bus.emit(Message('recognizer_loop:record_begin'))
+    context = {'client_name': 'mycroft_listener',
+               'source': 'audio'}
+    bus.emit(Message('recognizer_loop:record_begin', context=context))
 
 
 def handle_record_end():
     """Forward internal bus message to external bus."""
     LOG.info("End Recording...")
-    bus.emit(Message('recognizer_loop:record_end'))
+    context = {'client_name': 'mycroft_listener',
+               'source': 'audio'}
+    bus.emit(Message('recognizer_loop:record_end', context=context))
 
 
 def handle_no_internet():
     LOG.debug("Notifying enclosure of no internet connection")
-    bus.emit(Message('enclosure.notify.no_internet'))
+    context = {'client_name': 'mycroft_listener',
+               'source': 'audio'}
+    bus.emit(Message('enclosure.notify.no_internet', context=context))
 
 
 def handle_awoken():
     """Forward mycroft.awoken to the messagebus."""
     LOG.info("Listener is now Awake: ")
-    bus.emit(Message('mycroft.awoken'))
+    context = {'client_name': 'mycroft_listener',
+               'source': 'audio'}
+    bus.emit(Message('mycroft.awoken', context=context))
 
 
 def handle_wakeword(event):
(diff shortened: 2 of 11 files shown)


JarbasAl:
deprecates #2118

JarbasAl:
> Hello @JarbasAl! Thanks for updating this PR. We checked the lines you've touched for [PEP 8](https://www.python.org/dev/peps/pep-0008) issues, and found:
> 
> There are currently no PEP 8 issues detected in this Pull Request. Cheers! beers
> ##### Comment last updated at 2020-01-22 19:02:49 UTC

you are wrong pep8 bot

travis says

mycroft/skills/fallback_skill.py:65:36: E128 continuation line under-indented for visual indent
mycroft/skills/fallback_skill.py:89:44: E128 continuation line under-indented for visual indent

forslund:
Yay this I can review and include :)

mycroft/skills/fallback_skill.py
```diff
@@ -61,7 +61,7 @@ def make_intent_failure_handler(cls, bus):
 
         def handler(message):
             # indicate fallback handling start
-            bus.emit(message.reply("mycroft.skill.handler.start",
+            bus.emit(message.forward("mycroft.skill.handler.start",
                                    data={'handler': "fallback"}))
```
forslund: Here's the incorrect identation, data should move in 2

mycroft/skills/fallback_skill.py
```diff
@@ -73,19 +73,19 @@ def handler(message):
...
                 else:  # No fallback could handle the utterance
-                    bus.emit(message.reply('complete_intent_failure'))
+                    bus.emit(message.forward('complete_intent_failure'))
                     warning = "No fallback could handle intent."
                     LOG.warning(warning)
                     #  indicate completion with exception
-                    bus.emit(message.reply('mycroft.skill.handler.complete',
+                    bus.emit(message.forward('mycroft.skill.handler.complete',
                                            data={'handler': "fallback",
                                                  'exception': warning}))
```
forslund: And these two as well

JarbasAl:
>  I do think the CLI by default should show all interactions since it's used for debugging. Maybe an option to disable that? CLI should also produce audio by default.
> 


since this only makes sense (in my opinion) when debugging, why not simply check for the "debug" flag already present in the .conf ?

forslund:
Tested and seems to be working as intended. The code looks excellent from what I've seen (apart from the pep issues)

I want the cli to view the utterances from the speech client and utterance entered on the cli to be outputted as audio. The CLI is first and foremost intended as a debug tool and removing this makes it a lot less useful.

It looks to me like the changes to the Message.reply() method should be backwards compatible so we don't need to hold off waiting for the 20.02 release. (Although this thursdays release may be the last of 19.08.)

mycroft/client/speech/__main__.py
```diff
@@ -35,24 +35,32 @@
...
+    context = {'client_name': 'mycroft_listener',
+               'source': 'audio'}
+    bus.emit(Message('recognizer_loop:record_end', context=context))
 
 
 def handle_no_internet():
     LOG.debug("Notifying enclosure of no internet connection")
-    bus.emit(Message('enclosure.notify.no_internet'))
+    context = {'client_name': 'mycroft_listener',
+               'source': 'audio'}
```
JarbasAl: very minor nitpick, and not used in practice, but should i set destination here as "enclosure"?

JarbasAl:
when using self.speak inside a scheduled event the message.context was lost, added the option to pass the message context

this allows things like common query to forward messages correctly https://github.com/JarbasAl/skill-query/commit/618c97e8b36a6334323f25ad5f251a3427937e1b

forslund:
Ok I have some weirdness on my Desktop machine, seem to work on my laptop but I get no speech on my desktop. Will dig and see if I have some weird config set.

mycroft/audio/speech.py
```diff
@@ -44,6 +44,14 @@ def handle_speak(event):
     Configuration.set_config_update_handlers(bus)
     global _last_stop_signal
 
+    # if the message is targeted and audio is not the target don't
+    # don't synthezise speech
+    event.context = event.context or {}
+    if event.context.get('destination') and \
+            ('debug_cli' in event.context['destination'] or
+             'audio' in event.context['destination']):
```
forslund: should this be
```python
    if event.context.get('destination') and \
            ('debug_cli' not in event.context['destination'] or
             'audio' not in event.context['destination']):
```

That change seems to make the PR work for me.

NeonJarbas: actually it should be 

```python
if event.context.get('destination') and not \
            ('debug_cli' in event.context['destination'] or
             'audio' in event.context['destination']):
```

We want TTS to happen when "debug_cli" is destination and also when "audio" is destination (and on multicasts when destination is missing)

forslund:
This looks good now I think. Do you think you could rebase it and squash the pep-8 fixes and the logic fixes into the relevant main commits to clean up the commit history a bit?

forslund:
Rebased, and ready to merge! Thanks @JarbasAI, awesome work!

pcwii:
@JarbasAI @JarbasAl, Does this PR add the ability to know what skill originated the message? I think this would be handy also for filtering / directing intents to / from other devices. I am looking for this functionality in my Mesh-Skill. Thanks for all the hard work!

PR #2464, opened 2020-01-29
Add tzinfo to extracted datetime

forslund:
## Description
Lingua-franca returns the extracted datetime without the datetime from
the device info. This needs to be added by core through an "anchorDate" specifying the current time (including timezone info).

## How to test
Set your timezone to Kansas City run the skill tester for the latest alarm skill version and assure the tests passes.

CLA [ Yes ]

### DIFF
--- a/mycroft/util/parse.py
+++ b/mycroft/util/parse.py
@@ -209,8 +209,8 @@ def extract_datetime(text, anchorDate=None, lang=None, default_time=None):
         ... )
         None
     """
-    ret = lingua_franca.parse.extract_datetime(text, anchorDate, lang,
-                                               default_time)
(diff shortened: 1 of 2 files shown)


krisgesling:
If I set my Mycroft device and OS time to the KC timezone then the Alarm Skill tests pass on the current dev branch. 

If I only set my device time but not my OS time to the KC timezone then the tests still fail on this bugfix branch. 

So I'm assuming that some datetimes are being extracted from the device settings, and some from the system clock?

forslund:
I've updated with a correct(-er) approach, sending a now reference to lingua-franca. Tested this with the alarm skill using various device times (Hawaii, Kansas City, CET, Darwin (before and after midnight over there)

forslund:
Thank goshness

PR #2475, opened 2020-02-19
Addition of Systemd startup files & preparation for Watchdog

j1nx:
## Description
This Pull Request is to open a dialog about implementing Systemd service files as replacement for the different Bash files and systems.

Starting off originally as "User Services", but could be easily converted/installed to "System Services". 

There is one main (dummy) mycroft.service unit that handles the different sub unit files for;
- message-bus
- audio
- voice
- skills
- enclosure

image

Starting the mycroft.service unit will start all different sub units. All sub units can be restarted without interfering with the others. This system could replace the "./start-mycroft.sh all" script.

## How to test
Unit files for user services can be installed in a couple of [different places](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#User%20Unit%20Search%20Path). I believe there is no right or wrong here, but going to pick one; ~/.config/systemd/user/

So for now, to manually install these unit files for testing, copy them to that place;
`cp scripts/services/* ~/.config/systemd/user/`

We can then enable them to be auto started when the user logs in;
`systemctl --user enable mycroft.service mycroft-messagebus.service mycroft-audio.service mycroft-voice.service mycroft-skills.service mycroft-enclosure.service`

To start them manually now without a reboot, just run;
`systemctl --user start mycroft.service`
All different sub unit files will automatically be started as well. If for whatever reason one of the sub units need to be sarted/stopped/restarted that can be done with similar command (example);
`systemctl --user restart mycroft-voice.service`

If you restart your system now then Mycroft will be started automatically once you log in (the default for Picroft). After your last session is closed, your user's systemd instances (and with it, our Mycroft services) will shutdown. 

You can make your user's systemd instances independent from your user's sessions (so that our Mycroft services starts at boot time even if you don't log in and also keeps running until a shutdown/reboot) with the following command;
`sudo loginctl enable-linger $USER`

## Thing to discuss / Way forward
Above installation steps can be added/changed within the setup bash script. However a couple of things I want to discuss about the best way forward.
- Do we want to install these systemd units only for the user?
- Or do we want to offer them to be installed as systemwide services as well? (requires root)
- Do we want to include the enable-linger for the user unit files? (requires root)

Above unit files do automaticaly restart themselfs if the terminiate, however with Systemd we could implement Notify/Watchdog support. However this requires some changes to mycroft-core.

## Notifying systemd when the Service is Ready.
Our mycroft.service dummy unit is used to start mycroft-messagebus.service. All other units are only being started if the messagebus is started. However currently there is no way for systemd to know the mycroft-message bus is actually started properly. 

You can notify systemd once it has completed its initialization of a program. This allows systemd to delay starting these sub units until mycroft-messagebus service is really ready.

These notification is done using so called "[sd_notify](https://www.freedesktop.org/software/systemd/man/sd_notify.html)" system calls. We can use that with a simple additional requirements from "python-systemd". When mycroft-messagebus is fully started and ready we can let systemd know by;
`import systemd.daemon
systemd.daemon.notify('READY=1')`

We can then convert our simple unit files by changing them to "Type=notify" and all our unit files will only be started after the message-bus is ready to rock and role.

This is only a small thing as mycroft already restarts connection to the message-bus if it fails or not ready, however changing to Type=notify also opens the path to using proper Watchdog polling.

I have linked to watchdog info before, so if there is interest for it we can start that process as well. I think using systemd with notify and watchdog supports make the whole system very robust as;
- Different services are automatically be restarted if they stopped petting the dog
- If for some reason failed services keep restarting/failing we can auto reboot a device to get it back online automatically
- With hardware watchdog support, we can auto reboot if the whole system hangs.

Open for feedback. I am no expert to it also, so just want to open the disccusion about what path should be the best.

### DIFF
--- a/scripts/services/mycroft.service
+++ b/scripts/services/mycroft.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Mycroft A.I. Software stack.
+After=network-online.target
+
+[Service]
+Type=oneshot
+ExecStartPre=/bin/bash -c 'while ! ping -q -c 1 -W 1 1.1.1.1 >/dev/null 2>&1; do sleep 1; done'
+ExecStart=/bin/true
+RemainAfterExit=yes
+
+[Install]
+WantedBy=default.target
--- a/scripts/services/mycroft-audio.service
+++ b/scripts/services/mycroft-audio.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=Mycroft Audio
+PartOf=mycroft.service
+After=mycroft.service
+After=mycroft-messagebus.service
+
+[Service]
+#User=mycroft
+WorkingDirectory=%h/mycroft-core
+Environment=PYTHONPATH=%h/mycroft-core/.venv/bin
+ExecStart=%h/mycroft-core/.venv/bin/python -m mycroft.audio
+StandardOutput=file:/var/log/mycroft/audio.log
+StandardError=file:/var/log/mycroft/audio.error.log
+Restart=on-failure
+
+[Install]
+WantedBy=mycroft.service
--- a/scripts/services/mycroft-enclosure.service
+++ b/scripts/services/mycroft-enclosure.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=Mycroft Enclosure
+PartOf=mycroft.service
+After=mycroft.service
+After=mycroft-messagebus.service
+
+[Service]
+#User=mycroft
+WorkingDirectory=%h/mycroft-core
+Environment=PYTHONPATH=%h/mycroft-core/.venv/bin
+ExecStart=%h/mycroft-core/.venv/bin/python -m mycroft.client.enclosure
+StandardOutput=file:/var/log/mycroft/enclosure.log
+StandardError=file:/var/log/mycroft/enclosure.error.log
+Restart=on-failure
+
+[Install]
+WantedBy=mycroft.service
--- a/scripts/services/mycroft-messagebus.service
+++ b/scripts/services/mycroft-messagebus.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=Mycroft Messagebus
+PartOf=mycroft.service
+After=mycroft.service
+
+[Service]
+#User=mycroft
+WorkingDirectory=%h/mycroft-core
+Environment=PYTHONPATH=%h/mycroft-core/.venv/bin
+ExecStart=%h/mycroft-core/.venv/bin/python -m mycroft.messagebus.service
+StandardOutput=file:/var/log/mycroft/bus.log
+StandardError=file:/var/log/mycroft/bus.error.log
+Restart=on-failure
+
+[Install]
+WantedBy=mycroft.service
--- a/scripts/services/mycroft-skills.service
+++ b/scripts/services/mycroft-skills.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=Mycroft Skills
+PartOf=mycroft.service
+After=mycroft.service
+After=mycroft-messagebus.service
+
+[Service]
+#User=mycroft
+WorkingDirectory=%h/mycroft-core
+Environment=PYTHONPATH=%h/mycroft-core/.venv/bin
+ExecStart=%h/mycroft-core/.venv/bin/python -m mycroft.skills
+StandardOutput=file:/var/log/mycroft/skills.log
+StandardError=file:/var/log/mycroft/skills.error.log
+Restart=on-failure
+
+[Install]
+WantedBy=mycroft.service
--- a/scripts/services/mycroft-voice.service
+++ b/scripts/services/mycroft-voice.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=Mycroft Voice
+PartOf=mycroft.service
+After=mycroft.service
+After=mycroft-messagebus.service
+
+[Service]
+#User=mycroft
+WorkingDirectory=%h/mycroft-core
+Environment=PYTHONPATH=%h/mycroft-core/.venv/bin
+ExecStart=%h/mycroft-core/.venv/bin/python -m mycroft.client.speech
+StandardOutput=file:/var/log/mycroft/voice.log
+StandardError=file:/var/log/mycroft/voice.error.log
+Restart=on-failure
+
+[Install]
+WantedBy=mycroft.service


devs-mycroft:
Hello, @j1nx, thank you for helping with the Mycroft project! We welcome everyone
for Everyone.

you!

forslund:
Hi @j1nx,

thanks for starting the discussion and providing your systemd implementation.

First a bit of my view on this repo. The scripts here are aimed at a setting up a development environment for mycroft-core and for creating the python package They're not really meant for system installations. We've historically kept system integrations externally (debian package, docker image, and the currently non-working snap)

I think the systemd files here would be great to have as example code and to use instead of the semi-install thing we do offer in the dev_setup (the bin folder with startup shortcuts are added to the path).

To me the user approach sounds good (didn't know about it before, I'm a systemd rookie), it should avoid issues with pulseaudio instance being taken by the incorrect user. I know from experience that running at a system level can lock the user out from the audio system so it's better if mycroft runs as the logged in user.

In a dedicated system that is different but then we're really out of the scope for the mycroft-core repository.

sd_notify: I would also like to avoid having a direct dependency on systemd in core...we should wrap it in some way to make it optional, like using an environment variable perhaps or a systemd wrapper around the processes.

For example the Mark-2-pi system uses the wifi-setup service. we want this to be started after both the audio process is ready to speak and the messagebus is started so more of the services could benefit from having it. My original thought was to leave it outside of core and add the possibility to wrap these things with a callback when ready.

example audio_systemd.py
```python import systemd.daemon

def on_done():
    systemd.daemon.notify('READY=1')

from mycroft.messagebus.service.__main__ import main
main(done_callback=on_done)
```

Support for the on_done callback (or status_callback if we can make it more generic) needs to be added but should be quite trivial.

JarbasAl:
leaving my 2 cents

since mycroft is meant to run anywhere i think it should have no opinion regarding setup/boot, that would be the responsibility of different enclosures

i also view mycroft-core has a dev repo, i would not like it if dev-setup made any changes regarding launching it, i don't event use the launch scripts but instead launch services manually depending on my use case

maybe have a couple repos with launchers?

ChanceNCounter:
On the other hand, when running on a systemd-based distro, if this represents performance gains, it might make a good standalone. That said, Mycroft packages make me nervous in general, because someone has to maintain them, and because they give the impression they're from Forslund when really they're from yours truly, an unpaid idiot.

JarbasAl:
this should be a separate script that can be ran manually, dev_setup.sh could also ask if we want to run that

regarding the example above about the messagebus, i would maybe not do that, i prefer to give the services an headstart to speed up boot time, they will reconnect to the bus eventually. this is probably not a good default, but illustrates why i think core should have no opinion regarding the launch script

i do like the approach in general and would love to see that made painless somehow

JarbasAl:
some examples here on how i run mycroft:
- raspberry pi - systemd, running as root, pulseaudio also as root, start after pulse, usually with my own enclosure/admin_service 
- server, systemd, running as root - only skills + bus service are launched
- dev box, run each module has needed, nothing happens outside my IDE
- dev raspberry pi- start_script in .bashrc, launches when i SSH
- raspberry pi - start_script in .rclocal
- i randomly use my own script or the official start script
- sometimes my own script is actually interaction with a systemd service and not doing anything by itself
- i also do weird stuff in docker, but that's out of scope

scripts/services/mycroft.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Mycroft A.I. Software stack.
+After=network-online.target
```
JarbasAl: i dislike this one, if there is no internet i still want mycroft to start and give appropriate logs, and possibly handling it. i have some setups with cached dialog for this kind of situation

j1nx:
Great feedback and opening of discussion. Exactly what I had in mind and why I initiated it.

Am a bit occupied today/tomorrow but will get back into this discussion asap.

j1nx:
Right! Apologies, life catched up on me once again...

Slipped on quite some good feedback, comments and suggestions. Let's see how to progress from here on. Let's start on the topic "The right place for this".

I created the PR here at this repo only because it could be seen as replacement for the start_mycroft.sh scripts which are also provided from this repo. Secondly because the choice of installing them (and where, user or systemwide), could be added to the dev_setup.sh script. To be honest, I have no oppinion on where these scripts should be placed, so open to all / any suggestions on where to move them and continue the communication and development.

The reason I shared these systemd files is because I saw an increase of the demand for them. More and more people are installing Mycroft on their semi-production devices. Especially now since Snips bailed out for them.

So without going into the other shared thoughts and feedback as of yet, what would be the right thing to do in regards of moving these files?

- Seperate repo with only these scripts and a install.sh script?
- Seperate repo as enclosure repo? (I think an enclosure is more about hardware then software, although both are involved anyhow)

Open for your suggestions and decision @forslund and @JarbasAl (JarbasAl, I know you have or have not the official dev title, but consider you as one of them anyhow. Same for @ChanceNCounter)

scripts/services/mycroft.service
j1nx: Quick reply: network-online.target is for having the network up, but has nothing todo with the internet. This makes sure the service is started after lo, eth,wlan, etc are up.

j1nx:
OK, think I will go for option 1 as the whole enclosure code / discussion is also still "on going".

@forslund I like your idea about the notify. Need to read into it a bit to see if that is possible. Having systemd starting one python service, however receiving the notify from another.

Will close this PR after I have moved stuff and added the reference here.

JarbasAl:
i think this should stay out of the enclosure

individual enclosures could pull from the other repo as part of setup

ChanceNCounter:
If you're moving these to another repo, don't forget the license!

StuartIanNaylor:
Not sure why you are having --user scripts that run at logon.

Just run on boot as system wide with a User=pi

But apart from that just a service to start and stop a working mycroft from boot.

j1nx:
Ok moved the files into an own repository located here;

https://github.com/j1nx/mycroft-systemd

Splitted them in advanced into user and system folders. Next would be some sort of install script, but let's move any discussion to that repository as either issue or pull-request.

StuartIanNaylor:
I think your system service is likely to render no pulseaudio sound.

I had to create a dbus_pid file for the user that pulseaudio would still complain about not having a X11 session but pulseaudio would work.

/home/pi/mycroft-core/service-mycroft.sh
```
#!/usr/bin/env bash

DBUS_SESSION_BUS_ADDRESS=unix:path=/run/dbus/system_bus_socket
DBUS_SESSION_BUS_PID=`cat /run/dbus/pid`
pulseaudio --start
/home/pi/mycroft-core/start-mycroft.sh all
```

If that should be part mycroft-audio.service ... ?

Probably needs as stop-service

ExecStart = /home/pi/mycroft-core/service-mycroft.sh
ExecStop = /home/pi/mycroft-core/stop-mycroft.sh

Just test to see if you have a working mic with pulseaudio

PR #2476, opened 2020-02-20
Bugfix for padatious and converse responses

forslund:
## Description
- Forwards messages messages from the padatious service to keep the destination intact. 
- copies context in reply to make sure original message context stays the same.
- Fix error when the ident is missing during converse handling

## How to test
Make sure the hello world skill's "how are you" intent, qa skill and "what's the weather like" intent returns audio.

CLA [ Yes ]

### DIFF
--- a/mycroft/messagebus/message.py
+++ b/mycroft/messagebus/message.py
@@ -16,6 +16,7 @@ import json
 import re
 import inspect
 from mycroft.util.parse import normalize
+from copy import deepcopy
 
 
 class Message:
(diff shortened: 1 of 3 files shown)


JarbasAl:
this looks good! will do some testing later today, but it seems ready to go

nice work, and great catch with the deepcopy stuff, this could easily trip up someone

forslund:
Thanks @JarbasAI for all the good feedback. I'm goingto merge this as it is now, if you find anything let me know and I'll do a follow up PR.

PR #2477, opened 2020-02-24
Added a "save_path" configuration for saved wake words and utterances.

### LINKED ISSUE #2454: add config entry for path to save output files
# How to submit an Issue to a Mycroft repository
## Be clear about the software, hardware and version you are running
* I'm running a picroft
* With latest unstable
* With custom Wake Word

## Try to provide steps that we can use to replicate the Issue

Turn on saving of wake words and audio clips (set both of these to "true"):
    "record_wake_words": false,
    "save_utterances": false,

only over here: https://github.com/MycroftAI/mycroft-core/blob/master/mycroft/configuration/mycroft.conf 
does it mention where it saves these to.  
This should be a configurable path, much like the upload URL is, for either/both of these.  Perhaps "default_save_path" or something less awkward.

ludwhe:
## Description
Fixes #2454 

## How to test
- Set 'save_utterances' and 'record_wake_words' to true
- Set the 'save_path' to something else than "/tmp"

### DIFF
--- a/mycroft/client/speech/mic.py
+++ b/mycroft/client/speech/mic.py
@@ -227,7 +227,8 @@ class ResponsiveRecognizer(speech_recognition.Recognizer):
         self.save_utterances = listener_config.get('save_utterances', False)
 
         self.save_wake_words = listener_config.get('record_wake_words')
-        self.saved_wake_words_dir = join(gettempdir(), 'mycroft_wake_words')
+        self.save_path = listener_config.get('save_path', gettempdir())
+        self.saved_wake_words_dir = join(self.save_path, 'mycroft_wake_words')
 
         self.upload_lock = Lock()
         self.filenames_to_upload = []
@@ -647,7 +648,10 @@ class ResponsiveRecognizer(speech_recognition.Recognizer):
         if self.save_utterances:
             LOG.info("Recording utterance")
             stamp = str(datetime.datetime.now())
-            filename = "/tmp/mycroft_utterance%s.wav" % stamp
+            filename = "/{}/mycroft_utterances/{}.wav".format(
+                self.save_path,
+                stamp
+            )
             with open(filename, 'wb') as filea:
                 filea.write(audio_data.get_wav_data())
             LOG.debug("Thinking...")
--- a/mycroft/client/speech/mic.py
+++ b/mycroft/client/speech/mic.py
@@ -222,13 +222,18 @@ class ResponsiveRecognizer(speech_recognition.Recognizer):
         self.audio = pyaudio.PyAudio()
         self.multiplier = listener_config.get('multiplier')
         self.energy_ratio = listener_config.get('energy_ratio')
-        # check the config for the flag to save wake words.
 
+        # Check the config for the flag to save wake words, utterances
+        # and for a path under which to save them
         self.save_utterances = listener_config.get('save_utterances', False)
-
-        self.save_wake_words = listener_config.get('record_wake_words')
+        self.save_wake_words = listener_config.get('record_wake_words', False)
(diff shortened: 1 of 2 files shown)


mycroft/configuration/mycroft.conf
```diff
@@ -149,13 +149,18 @@
   // Override: REMOTE
   "listener": {
     "sample_rate": 16000,
+    // Set 'save_path' to configure the location of files stored if
+    // 'record_wake_words' and/or 'save_utterances' are set to 'true'.
+    // WARNING: Make sure that user 'mycroft' has write-access on the
+    // directory!
+    "save_path": "/tmp",
```
forslund: I wonder if this should be commented out as /tmp isn't always the temp directory. Commenting it out, it would use `gettempdir()` to determine the location which is safer.

ludwhe: No problem, just pushed an updated version!

mycroft/client/speech/mic.py
```diff
@@ -647,7 +648,10 @@ def listen(self, source, emitter, stream=None):
         if self.save_utterances:
             LOG.info("Recording utterance")
             stamp = str(datetime.datetime.now())
-            filename = "/tmp/mycroft_utterance%s.wav" % stamp
+            filename = "/{}/mycroft_utterance/{}.wav".format(
```
ludwhe: Missing an 's'

mycroft/client/speech/mic.py
```diff
@@ -647,7 +648,10 @@ def listen(self, source, emitter, stream=None):
         if self.save_utterances:
             LOG.info("Recording utterance")
             stamp = str(datetime.datetime.now())
-            filename = "/tmp/mycroft_utterance%s.wav" % stamp
+            filename = "/{}/mycroft_utterances/{}.wav".format(
```
forslund: This adds a new directory (`'save_dir'/mycroft_utterances`) and doesn't ensure that the directory exists. So currently the voice client will error out here unless the folder is manually.

Perhaps in the init there should be something like
```python
        self.saved_utterances_path = join(self.save_path, 'mycroft-utterances')
        if not isdir(self.saved_wake_words_dir):                
            os.mkdir(self.saved_wake_words_dir)
```

ludwhe: Good point, added that and did the same for the `wake_words_dir`, also made the `save_wake_words` flag explicitly `False` if unspecified. If the `save_*` flags are unspecified or `False`, the directories are not created.

mycroft/client/speech/mic.py
```diff
@@ -559,8 +559,6 @@ def _wait_until_wake_word(self, source, sec_per_buffer, emitter):
                         # Save wake word locally
                         audio = self._create_audio_data(byte_data, source)
                         mtd = self._compile_metadata()
-                        if not isdir(self.saved_wake_words_dir):
-                            os.mkdir(self.saved_wake_words_dir)
```
ludwhe: Now in `__init__`

forslund:
Looks great, gonna clean up the commit history and then merge it.

Thank you for contributing this!

StuartIanNaylor:
There is a worry with constant writes to flash that you could increase flash failure.
Prob why /tmp was chosen.

/tmp is quite cool as it uses ram but also when idle it will push out to swap, so it can use less memory than you think.

You can also use zram to create in effect a compressed /tmp drive.
I also created a demo repo https://github.com/StuartIanNaylor/zram-config
To show multiple usuages of zram even use it as an upper in a OverlayFs filemount that uses zram for all writes and merges down to SD flash on each each reboot or stop.

So its no bother that /tmp is being moved, but people might want to mitigate and be aware of increases in flash wear.

PR #2478, opened 2020-02-24
Make message.response keep source/dest info

forslund:
## Description
Message.response used reply, witch led to common play skills getting the desitnation wrong.

## How to test
Make sure the common play start message gets the context correctly.

CLA [ Yes ]


forslund:
Tiny amendment to #2476, self-merging

PR #2480, opened 2020-02-24
Don't forward the context instead use a bare message

forslund:
## Description
Without this the stop handling is really REALLY slow. The exact cause
will need to be researched but this works around the issue for now.

## How to test
Start playing the news and ensure that it doesn't stop directly.

CLA [ Yes ]


forslund:
Found the cause. The report_metric call at the end of the skill handler will starve the threadpool for the multicast mycroft.stop since it forwards the ident.

I will update the metrics reporting to use a queue and send in a batch mode as suggested by @JarbasAI.

Closing this for now, better stronger, faster PR tomorrow.

PR #2482, opened 2020-02-26
remove deprecated Settings class

chrisveilleux:
## Description
The Settings class was deprecated in 19.08 and is now being removed.

## How to test
Ensure skill settings are loaded and synchronized as expected. No functionality should change.

CLA [yes]


forslund:
Did a quick test run and it seems functional will do some more random poking before merging! :+1:

forslund:
Looks very good, I found a thing in the skill tester but apart from that it looks and behaves good.

There reference is in `test/integrationtests/skills/skill_tester.py`, an import, the TestSettings class [L197](https://github.com/MycroftAI/mycroft-core/blob/dev/test/integrationtests/skills/skill_tester.py#L197) and the use of it at [L345-347](https://github.com/MycroftAI/mycroft-core/blob/dev/test/integrationtests/skills/skill_tester.py#L345). Import and class can be removed and the usage could probably be replaced with 

```python
s.settings = test_case['settings']
```

If you don't have time to fix it I can add a commit to this tomorrow morning before merging.

PR #2496, opened 2020-03-09
Add tests for a variety of util functions

forslund:
## Description
- Add test cases
- Logic change in resolve_resource_file to make it work as the docstring indicated.
- split the `__init__.py` into subfiles based on functionality.

This now changes the read_stripped_lines to a generator. This changes the return type and may be considered breaking but it'll rarely be used indexed and there's currently nothing that's using the function as far as I've been able to determine.

## How to test
Ensure that mycroft is still functioning and that tests passes.

CLA [ Yes ]


mycroft/util/file_utils.py
```diff
@@ -0,0 +1,235 @@
...
+    # Finally look for it in the source package
+    filename = os.path.join(os.path.dirname(__file__), '..', 'res', res_name)
+    filename = os.path.abspath(os.path.normpath(filename))
+    if os.path.isfile(filename):
+        return filename
+
+    return None  # Resource cannot be resolved
+
+
+def read_stripped_lines(filename):
```
clusterfudge: This seems like an opportunity to optimize this to use a generator.

mycroft/util/network_utils.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def connected():
+    """Check connection by connecting to 8.8.8.8 and if google.com is
+    reachable if this fails, Check Microsoft NCSI is used as a backup.
+
+    Returns:
+        True if internet connection can be detected
+    """
+    if _connected_dns():
```
clusterfudge: It seems to me like the dns check is sufficient, and all the checks are potentially DoS and information disclosure concerns.

See [here](https://community.ntppool.org/t/recent-ntp-pool-traffic-increase/18/6) for an example of this type of problem in the wild.

mycroft/util/string_utils.py
```diff
@@ -0,0 +1,49 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""Common string utilities used by various parts of core."""
+
+import re
+
+
+def get_http(uri):
```
clusterfudge: Are these really heavily used utilities? They seem like unnecessary bloat.

test/unittests/util/test_time.py
```diff
@@ -0,0 +1,76 @@
...
+            'name': 'Central Standard Time',
+            'dstOffset': 3600000,
+            'offset': -21600000
+        }
+    }
+}
+
+
+@mock.patch('mycroft.configuration.Configuration')
+class TestTimeFuncs(TestCase):
```
clusterfudge: future improvement: A more common strategy to testing time is to introduce the concept of a Clock as an interface, and provide deterministic clocks at test time. This prevents the necessity of mocking global statics, which can cause cascading failures through tests (and introduce bugs!) when misused.

mycroft/util/file_utils.py
forslund: Would that be considered a non-breaking change? You can't index a generator...but it's quite a small break so maybe it's ok...

At the very least I can add a todo comment

mycroft/util/network_utils.py
forslund: The checks have grown a bit over time. The dns lookup doesn't work on all networks due to filtering, hence ncis check...

The google check was added to handle some weird edge cases filtering out Mark-1's where the AP wasn't shut off completely. It should be removed at some point

mycroft/util/string_utils.py
forslund: I'll mark the http and remove trailing slash functions for removal in the next major release

mycroft/util/network_utils.py
clusterfudge: yeah, I guess I'm just not fully understanding why we'd use both domains here. Outbound port filtering makes sense for DNS vs HTTP protocols, I guess we're trying another tactic with http/https ports?

A "good internet citizen" would probably try to connect to a mycroft-operated or public for-this-purpose endpoint. NCSI seems to hit that (if intended for msft users), but dns and google.com don't strike me as good choices here.

forslund:
Updated marking the redundant string functions as to be removed and changed the read_stripped_lines to a generator as suggested.

test/unittests/util/test_util.py
```diff
@@ -0,0 +1,226 @@
...
+                                                      env=Anything())
+
+
+class TestStringFunctions(TestCase):
+    def test_camel_case_split(self):
+        """Check that camel case string is split properly."""
+        self.assertEqual(camel_case_split('MyCoolSkill'), 'My Cool Skill')
+        self.assertEqual(camel_case_split('MyCOOLSkill'), 'My COOL Skill')
+
+    def test_get_http(self):
```
krisgesling: If you wanted to go comprehensive, we could add one to check that only the protocol is modified, eg:
```
self.assertEqual(get_http('http://https.com/'), 'http://https.com/')
```

forslund: The `get_http('https://github.com/https')` one sort of does that test, but I'll add the one with server url as well.

mycroft/util/process_utils.py
```diff
@@ -0,0 +1,105 @@
...
+        while True:
+            sleep(100)
+    except KeyboardInterrupt:
+        pass
+
+
+_log_all_bus_messages = False
+
+
+def create_echo_function(name, whitelist=None):
```
chrisveilleux: there is a lot going on here, maybe a TODO to clean this up later?

test/unittests/util/test_time.py
```diff
@@ -0,0 +1,76 @@
...
+from mycroft.util.time import (default_timezone, now_local, now_utc, to_utc,
+                               to_local, to_system)
+
+test_config = {
+    'location': {
+        'timezone': {
+            'code': 'America/Chicago',
+            'name': 'Central Standard Time',
+            'dstOffset': 3600000,
+            'offset': -21600000
```
chrisveilleux: replace magic numbers with named constants

test/unittests/util/test_util.py
```diff
@@ -0,0 +1,227 @@
+from os.path import (abspath, dirname, expanduser, join, normpath, isdir,
+                     exists)
+import shutil
+import tempfile
+from unittest import TestCase, mock
+
+from mycroft import MYCROFT_ROOT_PATH
+from mycroft.util import (resolve_resource_file, curate_cache,
```
chrisveilleux: convert to new directory structure (e.g. `from mycroft.util.audio_util`, etc.).  should these tests be broken up into one file per module being tested?

mycroft/util/audio_utils.py
```diff
@@ -0,0 +1,199 @@
...
+_ENVIRONMENT['PULSE_PROP'] = 'media.role=music'
+
+
+def _get_pulse_environment(config):
+    """Return environment for pulse audio depeding on ducking config."""
+    tts_config = config.get('tts', {})
+    if tts_config and tts_config.get('pulse_duck'):
+        return _ENVIRONMENT
+    else:
+        return os.environ
```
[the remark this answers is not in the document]
forslund: We don't want to change the global environ but the commands should be launched with a modified environment. We use a static copy to avoid copying every single time.

test/unittests/util/test_util.py
forslund: `mycroft.util` should still export all these methods so I think it's better for the test to pull from the module and not the submodules

test/unittests/util/test_time.py
forslund: Good idea, I never think about adding todos for these things because my editor always warns me that methods/functions are too complex :)

mycroft/util/audio_utils.py
```diff
@@ -0,0 +1,199 @@
...
+    play_cmd = config.get("play_wav_cmdline")
+    play_wav_cmd = str(play_cmd).split(" ")
+    for index, cmd in enumerate(play_wav_cmd):
+        if cmd == "%1":
+            play_wav_cmd[index] = (get_http(uri))
+    try:
+        return subprocess.Popen(play_wav_cmd, env=environment)
+    except Exception as e:
+        LOG.error("Failed to launch WAV: {}".format(play_wav_cmd))
+        LOG.debug("Error: {}".format(repr(e)), exc_info=True)
```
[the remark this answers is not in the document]
forslund: Most of the time in this case the exception information confuses more than it helped so a simpler error message with the most relevant information was added in it's place we didn't want to remove the possibility to get the entire stacktrace logged completely so it was set as debug only.

forslund:
I didn't originally intend to do too much changes to the functions themselves, just organize them a bit better but I'll continue the refactoring and clean up the code as much as I get time to this weekend.

chrisveilleux:
> I didn't originally intend to do too much changes to the functions themselves, just organize them a bit better but I'll continue the refactoring and clean up the code as much as I get time to this weekend.

Understood, but this is how we make the code better, by chipping away at it like this.  If we don't make these changes as we touch the code, when do we?

forslund:
The next time? I have this as an ongoing weekend morning project, chipping away pushing PR's at logical intervals.

However now I've almost completed the second pass of this so I can just as well add ot to this PR even if it makes the reviewing much more involved

test/unittests/util/test_time.py
forslund: I'm sceptical, this mirrors how it looks in the default config. Feels like it's better to keep it as close to the source as possible.

mycroft/util/file_utils.py
```diff
@@ -0,0 +1,238 @@
...
+def create_file(filename):
+    """ Create the file filename and create any directories needed
+
+        Args:
+            filename: Path to the file to be created
+    """
+    try:
+        os.makedirs(os.path.dirname(filename))
+    except OSError:
+        pass
```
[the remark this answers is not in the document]
forslund: ensure_directory_exists would need to be changed a bit since by default it sets the rights to 777 and that's not something that is always what you want.

mycroft/util/audio_utils.py
```diff
@@ -0,0 +1,199 @@
...
+    LOG.debug('Devices: ')
+    pa = pyaudio.PyAudio()
+    pattern = re.compile(device_name)
+    for device_index in range(pa.get_device_count()):
+        dev = pa.get_device_info_by_index(device_index)
+        LOG.debug('   {}'.format(dev['name']))
+        if dev['maxInputChannels'] > 0 and pattern.match(dev['name']):
+            LOG.debug('    ^-- matched')
+            return device_index
+    return None
```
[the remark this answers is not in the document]
forslund: It's better to be explicit in cases where it can either return a value or `None`

mycroft/util/file_utils.py
```diff
@@ -0,0 +1,238 @@
...
+        bytes_needed = int(bytes_needed + 1.0)
+
+        # get all entries in the directory w/ stats
+        entries = (os.path.join(directory, fn) for fn in os.listdir(directory))
+        entries = ((os.stat(path), path) for path in entries)
+
+        # leave only regular files, insert modification date
+        entries = ((stat[ST_MTIME], stat[ST_SIZE], path)
+                   for stat, path in entries if S_ISREG(stat[ST_MODE]))
+
```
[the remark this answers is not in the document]
forslund: I did some refactoring, still loops a couple of times, collecting them all, sorting them and then removing the oldest

devs-mycroft:
Testing a comment

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2496))

mycroft/util/audio_utils.py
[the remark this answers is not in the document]
chrisveilleux: Why don't we want to change the global environment?  Is there some breakage that occurs if we do?  If this code remains, I would at least like to see some comments around why this approach was taken.

Also, can we move the _ENVIRONMENT attribute into the _get_pulse_environment function?

mycroft/util/audio_utils.py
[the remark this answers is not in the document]
chrisveilleux: What was confusing about the exception information?  The only difference between this code and a log.exception() call is that the traceback appears in a different log message.  Making a user turn on debug mode and recreate the issue just to get the stack trace seems counter-intuitive to me.

mycroft/util/audio_utils.py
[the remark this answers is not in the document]
chrisveilleux: I've always been a proponent of a single return statement in each function.  A better solution to this, IMO, is to define a variable to be the return value and initialize it to None.  Then assign a new value if the logic dictates it.  this also makes debugging easier as you can watch the variable inside the function.

Resolving this for now but want to revisit.

mycroft/util/audio_utils.py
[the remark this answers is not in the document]
forslund: It was originally a plain exception but was a bit verbose and added very little useful information so it was reduced to this a couple of years ago making the default info less verbose while keeping the option to have the backtrace for the rare occations where it's of use.

test/unittests/util/test_time.py
chrisveilleux: I disagree.  At least put a comment as to what the numbers represent so someone doesn't have to come along later to do the math in their heads.

mycroft/util/audio_utils.py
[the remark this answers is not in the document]
forslund: I can definitely add a comment.

Breaking is a quite hard term but the audio behavior will be changed for the entire Mycroft processes and all their future subprocess. We just want the commands we're launching through these methods to have the specified PULSE_PROP making pulse duck them when things like TTS is executed.

I do not want to have it inside the function since the copy action is quite heavy and I do not want it done on every call. Since it will be the same each time it's used it's kept at module level.

mycroft/util/file_utils.py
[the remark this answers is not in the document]
forslund: This was updated to use the `ensure_directory_exist()` with an extra parameter.

test/unittests/util/test_time.py
forslund: Comments I agree is a good thing, added those.

mycroft/util/audio_utils.py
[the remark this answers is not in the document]
forslund: I added the comment explaining the use of the modified environment a bit more.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2496))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2496))

mycroft/util/audio_utils.py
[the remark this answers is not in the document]
chrisveilleux: I don't agree with verbosity as a rationale for omitting a traceback from an error message.  Consistency in how this is done is more important.  One-off differences like this make very little sense to me.  Resolving because I am done fighting it.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2496))

forslund:
I did a final update to the Logging so the known error (FileNotFoundError) will log a short informative  message while an unknown error (Exception) will show the stack trace.

PR #2500, opened 2020-03-11
Clear is speaking signal on exceptions

forslund:
## Description
This fixes lingering isSpeaking signals if an exception occurs during playback. A dummy package is added to the tts playback queue

## How to test
Check that the unittest passes and that tts is working as expected.

CLA [ Yes ]


forslund:
This seems to have been merged as part of voight kamff. Closing

PR #2506, opened 2020-03-20
Voight Kampff test suite

forslund:
Add the voight_kampff test suite.

This allows creating behavioral tests for skills using the behave framework. It specifies a couple of pre-defined Given, When and Then covering the most common test cases but can be expanded with custom "step files".

In addition of being a test suite for skill creators, it will be used as an integration test validating pull requests to mycroft-core, making sure the essential skills work.

There are very little changes to core, but a couple of modifications are included in all these commits:
- To allow the test to easier identify the dialog files correctly a meta field is added to the speak messages with dialog and origin information.
- A fix to handling converse errors.
- Adds a "dummy" tts which doesn't do any audio output

CLA [ Yes ]

### DIFF
--- a/mycroft/tts/tts.py
+++ b/mycroft/tts/tts.py
@@ -471,6 +471,7 @@ class TTSFactory:
     from mycroft.tts.responsive_voice_tts import ResponsiveVoice
     from mycroft.tts.mimic2_tts import Mimic2
     from mycroft.tts.yandex_tts import YandexTTS
+    from mycroft.tts.dummy_tts import DummyTTS
 
     CLASSES = {
         "mimic": Mimic,
@@ -483,7 +484,8 @@ class TTSFactory:
         "watson": WatsonTTS,
         "bing": BingTTS,
         "responsive_voice": ResponsiveVoice,
-        "yandex": YandexTTS
+        "yandex": YandexTTS,
+        "dummy": DummyTTS
     }
 
     @staticmethod
--- a/mycroft/tts/tts.py
+++ b/mycroft/tts/tts.py
@@ -51,6 +51,7 @@ class PlaybackThread(Thread):
         self._terminated = False
         self._processing_queue = False
         self.enclosure = None
+        self.p = None
         # Check if the tts shall have a ducking role set
         if Configuration.get().get('tts', {}).get('pulse_duck'):
             self.pulse_env = _TTS_ENV
@@ -102,8 +103,9 @@ class PlaybackThread(Thread):
                         self.p = play_mp3(data, environment=self.pulse_env)
                     if visemes:
                         self.show_visemes(visemes)
-                    self.p.communicate()
-                    self.p.wait()
+                    if self.p:
+                        self.p.communicate()
+                        self.p.wait()
                 report_timing(ident, 'speech_playback', stopwatch)
 
                 if self.queue.empty():
@@ -312,6 +314,13 @@ class TTS(metaclass=ABCMeta):
         sentence = self.validate_ssml(sentence)
 
         create_signal("isSpeaking")
+        try:
+            self._execute(sentence, ident, listen)
+        except Exception:
+            # If an error occurs end the audio sequence
+            self.queue.put((None, None, None, None, None))
+
+    def _execute(self, sentence, ident, listen):
         if self.phonetic_spelling:
             for word in re.findall(r"[\w']+", sentence):
                 if word.lower() in self.spellings:
--- a/test/integrationtests/voight_kampff/features/steps/utterance_responses.py
+++ b/test/integrationtests/voight_kampff/features/steps/utterance_responses.py
@@ -106,29 +106,13 @@ def _match_dialog_patterns(dialogs, sentence):
         return False, debug
 
 
-def match_dialog_patterns(dialogs, sentence):
-    """Match sentence against a list of dialog patterns.
-
-    Returns simple True/False and not index
-    """
-    index, debug = _match_dialog_patterns(dialogs, sentence)
-    return index is not False, debug
-
-
-def expected_dialog_check(utterance, skill_path, dialog):
-    """Check that expected dialog file is used. """
-    dialogs, load_debug = load_dialog_list(skill_path, dialog)
-    match, match_debug = match_dialog_patterns(dialogs, utterance)
-    return match, load_debug + match_debug
-
-
 @given('an english speaking user')
-def given_impl(context):
+def given_english(context):
     context.lang = 'en-us'
 
 
 @when('the user says "{text}"')
-def when_impl(context, text):
+def when_user_says(context, text):
     context.bus.emit(Message('recognizer_loop:utterance',
                              data={'utterances': [text],
                                    'lang': 'en-us',
@@ -137,11 +121,11 @@ def when_impl(context, text):
                              context={'client_name': 'mycroft_listener'}))
 
 
-def then_timeout(then_func, context, timeout=TIMEOUT):
+def then_wait(msg_type, then_func, context, timeout=TIMEOUT):
     count = 0
     debug = ''
     while count < TIMEOUT:
-        for message in context.speak_messages:
+        for message in context.bus.get_messages(msg_type):
             status, test_dbg = then_func(message)
             debug += test_dbg
             if status:
@@ -155,13 +139,11 @@ def then_timeout(then_func, context, timeout=TIMEOUT):
 
 @then('"{skill}" should reply with dialog from "{dialog}"')
 def then_dialog(context, skill, dialog):
-    skill_path = context.msm.find_skill(skill).path
-
     def check_dialog(message):
-        utt = message.data['utterance'].lower()
-        return expected_dialog_check(utt, skill_path, dialog)
+        utt_dialog = message.data.get('meta', {}).get('dialog')
+        return (utt_dialog == dialog.replace('.dialog', ''), '')
 
-    passed, debug = then_timeout(check_dialog, context)
+    passed, debug = then_wait('speak', check_dialog, context)
     if not passed:
         print(debug)
     assert passed
@@ -183,7 +165,7 @@ def then_anything(context, skill):
         result = message is not None
         return (result, debug)
 
-    passed = then_timeout(check_any_messages, context)
+    passed = then_wait('speak', check_any_messages, context)
     if not passed:
         print('No speech recieved at all.')
     assert passed
@@ -197,7 +179,7 @@ def then_exactly(context, skill, text):
         result = utt == text.lower()
         return (result, debug)
 
-    passed, debug = then_timeout(check_exact_match, context)
+    passed, debug = then_wait('speak', check_exact_match, context)
     if not passed:
         print(debug)
     assert passed
@@ -211,7 +193,7 @@ def then_contains(context, text):
         result = text.lower() in utt
         return (result, debug)
 
-    passed, debug = then_timeout(check_contains, context)
+    passed, debug = then_wait('speak', check_contains, context)
 
     if not passed:
         print('No speech contained the expected content')
@@ -219,7 +201,7 @@ def then_contains(context, text):
 
 
 @then('the user says "{text}"')
-def when_impl(context, text):
+def then_user_follow_up(context, text):
     time.sleep(2)
     wait_while_speaking()
     context.bus.emit(Message('recognizer_loop:utterance',
--- a/test/integrationtests/voight_kampff/features/steps/utterance_responses.py
+++ b/test/integrationtests/voight_kampff/features/steps/utterance_responses.py
@@ -163,7 +163,7 @@ def then_dialog(context, skill, dialog):
         assert_msg = debug
         assert_msg += mycroft_responses(context)
 
-    assert passed, assert_msg
+    assert passed, assert_msg or 'Mycroft didn\'t respond'
 
 
 @then('"{skill}" should reply with "{example}"')
--- a/test/integrationtests/voight_kampff/features/steps/utterance_responses.py
+++ b/test/integrationtests/voight_kampff/features/steps/utterance_responses.py
@@ -243,4 +243,7 @@ def then_messagebus_message(context, message_type):
         if cnt > 20:
             assert False, "Message not found"
             break
+        else:
+            cnt += 1
+
         time.sleep(0.5)
--- a/test/integrationtests/voight_kampff/features/steps/utterance_responses.py
+++ b/test/integrationtests/voight_kampff/features/steps/utterance_responses.py
@@ -33,15 +33,15 @@ TIMEOUT = 10
 SLEEP_LENGTH = 0.25
 
 
-def find_dialog(skill_path, dialog):
+def find_dialog(skill_path, dialog, lang):
     """Check the usual location for dialogs.
 
     TODO: subfolders
     """
     if exists(join(skill_path, 'dialog')):
-        return join(skill_path, 'dialog', 'en-us', dialog)
+        return join(skill_path, 'dialog', lang, dialog)
     else:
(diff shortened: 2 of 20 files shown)


JarbasAl:
shouldn't the "meta" field simply be a part of message.context ?

forslund:
It could definitely be moved there, I did consider it but felt like the context should be more "standardized" and not depend to much on the message type. Like the skill shouldn't interfere with the context and let the "surrounding machinery" handle that field. Not sure if that makes sense?

I don't really have a strong opinion one way or another, but that was my thoughts at the time.

mycroft/tts/tts.py
```diff
@@ -51,6 +51,7 @@ def __init__(self, queue):
         self._terminated = False
         self._processing_queue = False
         self.enclosure = None
+        self.p = None
```
chrisveilleux: Use more descriptive name for this variable.

test/integrationtests/voight_kampff/features/steps/utterance_responses.py
```diff
@@ -0,0 +1,209 @@
...
+        return None
+
+
+def _match_dialog_patterns(dialogs, sentence):
+    """Match sentence against a list of dialog patterns.
+
+    Returns index of found match.
+    """
+    # Allow custom fields to be anything
+    d = [re.sub(r'{.*?\}', r'.*', t) for t in dialogs]
```
chrisveilleux: more descriptive name

test/integrationtests/voight_kampff/features/steps/utterance_responses.py
```diff
@@ -0,0 +1,209 @@
...
+    # Remove left over '}'
+    d = [re.sub(r'\}', r'', t) for t in d]
+    d = [re.sub(r' .* ', r' .*', t) for t in d]
+    # Merge consequtive .*'s into a single .*
+    d = [re.sub(r'\.\*( \.\*)+', r'.*', t) for t in d]
+    # Remove double whitespaces
+    d = ['^' + ' '.join(t.split()) for t in d]
+    debug = 'MATCHING: {}\n'.format(sentence)
+    for i, r in enumerate(d):
```
chrisveilleux: more descriptive names

test/integrationtests/voight_kampff/features/steps/utterance_responses.py
```diff
@@ -0,0 +1,209 @@
...
+SLEEP_LENGTH = 0.25
+
+
+def find_dialog(skill_path, dialog):
+    """Check the usual location for dialogs.
+
+    TODO: subfolders
+    """
+    if exists(join(skill_path, 'dialog')):
+        return join(skill_path, 'dialog', 'en-us', dialog)
```
chrisveilleux: should this use context.lang instead of hardcoding 'en-us'?

test/integrationtests/voight_kampff/features/steps/utterance_responses.py
```diff
@@ -0,0 +1,209 @@
...
+    """
+    dialog_path = find_dialog(skill_path, dialog)
+
+    debug = 'Opening {}\n'.format(dialog_path)
+    return load_dialog_file(dialog_path), debug
+
+
+def dialog_from_sentence(sentence, skill_path):
+    """Find dialog file from example sentence."""
+    dialog_paths = join(skill_path, 'dialog', 'en-us', '*.dialog')
```
chrisveilleux: use context.lang instead of hardcoding 'en-us'?

test/integrationtests/voight_kampff/features/steps/utterance_responses.py
```diff
@@ -0,0 +1,209 @@
...
+    dialog = dialog_from_sentence(example, skill_path)
+    print('Matching with the dialog file: {}'.format(dialog))
+    assert dialog is not None, 'No matching dialog...'
+    then_dialog(context, skill, dialog)
+
+
+@then('"{skill}" should reply with anything')
+def then_anything(context, skill):
+    def check_any_messages(message):
+        return (True, '') if message else (False, '')
```
chrisveilleux: what do these tuples represent?  another statement that assigns the tuple to a variable might help with readability.

test/integrationtests/voight_kampff/generate_feature.py
```diff
@@ -0,0 +1,61 @@
...
+"""
+
+
+def generate_feature(skill, skill_path):
+    """Generate a feature file provided a skill name and a path to the skill.
+    """
+    test_path = Path(join(skill_path, 'test', 'intent'))
+    case = []
+    if test_path.exists() and test_path.is_dir():
+        for f in sorted(glob(join(str(test_path), '*.json'))):
```
chrisveilleux: more descriptive name

test/integrationtests/voight_kampff/test_setup.py
```diff
@@ -0,0 +1,163 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+import argparse
+from argparse import RawTextHelpFormatter
+from glob import glob
+from os.path import join, dirname, expanduser, basename
+from pathlib import Path
```
chrisveilleux: can we use pathlib or os.path instead of both?

test/integrationtests/voight_kampff/test_setup.py
```diff
@@ -0,0 +1,163 @@
...
+that anyi specified extra skills also gets installed into the environment.
+"""
+
+FEATURE_DIR = join(dirname(__file__), 'features') + '/'
+
+
+def copy_feature_files(source, destination):
+    """Copy all feature files from source to destination."""
+    # Copy feature files to the feature directory
+    for f in glob(join(source, '*.feature')):
```
chrisveilleux: more meaningful name

test/integrationtests/voight_kampff/test_setup.py
```diff
@@ -0,0 +1,163 @@
...
+        shutil.copyfile(f, join(destination, basename(f)))
+
+
+def copy_step_files(source, destination):
+    """Copy all python files from source to destination."""
+    # Copy feature files to the feature directory
+    for f in glob(join(source, '*.py')):
```
chrisveilleux: more meaningful name

test/integrationtests/voight_kampff/test_setup.py
```diff
@@ -0,0 +1,163 @@
...
+    with open(expanduser(config)) as f:
+        conf_dict = yaml.safe_load(f)
+
+    if not args.test_skills and 'test_skills' in conf_dict:
+        args.test_skills = conf_dict['test_skills']
+    if not args.extra_skills and 'extra_skills' in conf_dict:
+        args.extra_skills = conf_dict['extra_skills']
+    if not args.platform and 'platform' in conf_dict:
+        args.platform = conf_dict['platform']
+    return
```
chrisveilleux: a bare return statement should not be necessary at the end of a function

test/integrationtests/voight_kampff/test_setup.py
```diff
@@ -0,0 +1,163 @@
...
+                              'Ex: "cocktails, laugh"'))
+    parser.add_argument('-r', '--random-skills', default=0,
+                        help='Number of random skills to install.')
+    parser.add_argument('-d', '--skills-dir')
+    parser.add_argument('-c', '--config',
+                        help='Path to test configuration file.')
+
+    args = parser.parse_args(cmdline_args)
+    if args.test_skills:
+        args.test_skills = args.test_skills.replace(',', ' ').split()
```
chrisveilleux: don't need to call replace() and split() here.  you can just do `split(',')`

test/integrationtests/voight_kampff/test_setup.py
```diff
@@ -0,0 +1,163 @@
...
+                        help=('Comma-separated list of extra skills to '
+                              'install.\n'
+                              'Ex: "cocktails, laugh"'))
+    parser.add_argument('-r', '--random-skills', default=0,
+                        help='Number of random skills to install.')
+    parser.add_argument('-d', '--skills-dir')
+    parser.add_argument('-c', '--config',
+                        help='Path to test configuration file.')
+
+    args = parser.parse_args(cmdline_args)
```
chrisveilleux: split out into an argument parsing function, perhaps?

test/integrationtests/voight_kampff/test_setup.py
```diff
@@ -0,0 +1,163 @@
...
+        args.extra_skills = args.extra_skills.replace(',', ' ').split()
+
+    if args.config:
+        load_config(args.config, args)
+
+    if args.platform is None:
+        args.platform = "mycroft_mark_1"
+
+    msm = MycroftSkillsManager(args.platform, args.skills_dir)
+    run_setup(msm, args.test_skills, args.extra_skills, args.random_skills)
```
chrisveilleux: why not just pass `args` to `run_setup`?  then MSM could be instantiated within run_setup and only one argument needs to be passed

test/integrationtests/voight_kampff/test_setup.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def run_setup(msm, test_skills, extra_skills, num_random_skills):
```
chrisveilleux: this is a really long function.  can it be broken up?

test/integrationtests/voight_kampff/test_setup.py
```diff
@@ -0,0 +1,163 @@
...
+    # Install random skills from uninstalled skill list
+    random_skills = [s for s in msm.all_skills if s not in msm.local_skills]
+    shuffle(random_skills)  # Make them random
+    random_skills = random_skills[:num_random_skills]
+    for s in random_skills:
+        msm.install(s)
+
+    print_install_report(msm.platform, test_skills,
+                         extra_skills + [s.name for s in random_skills])
+    return
```
chrisveilleux: bare returns at end of function is unnecessary

test/integrationtests/voight_kampff/test_setup.py
```diff
@@ -0,0 +1,163 @@
...
+            # Generate feature file if no data exists yet
+            print('No feature files exists for {}, '
+                  'generating...'.format(skill_name))
+            # No feature files setup, generate automatically
+            feature = generate_feature(skill_name, skill.path)
+            with open(join(FEATURE_DIR, skill_name + '.feature'), 'w') as f:
+                f.write(feature)
+
+    # Install random skills from uninstalled skill list
+    random_skills = [s for s in msm.all_skills if s not in msm.local_skills]
```
chrisveilleux: Shouldn't this be optional?  We don't want every test run to have random skills, just those that run outside of pull requests.

test/integrationtests/voight_kampff/test_setup.py
```diff
@@ -0,0 +1,163 @@
...
+        'platform': platform,
+        'test_skills': test_skills,
+        'extra_skills': extra_skills
+        })
+    print(yml)
+    print('----------------------------')
+
+
+if __name__ == '__main__':
+    main(sys.argv[1:])
```
chrisveilleux: this does not have to be a single line.  IMO, this is a good place to call functions in an order than tells a story about when is being done.  for example:
`parse_args()`
`apply_config()`
`install_skills()`
`collect_feature_files()`
`install_random_skillls()`
`print_install_report()`
in six lines of code, you just told the person reading the code the what this script is doing at the high level.

test/integrationtests/voight_kampff/tools.py
```diff
@@ -0,0 +1,94 @@
...
+
+def wait_for_dialog(bus, dialogs, timeout=TIMEOUT):
+    """Wait for one of the dialogs given as argument.
+
+    Arguments:
+        bus (InterceptAllBusClient): Bus instance to listen on
+        dialogs (list): list of acceptable dialogs
+        timeout (int): how long to wait for the messagem, defaults to 10 sec.
+    """
+
```
chrisveilleux: remove blank line

test/integrationtests/voight_kampff/test_setup.py
```diff
@@ -0,0 +1,163 @@
...
+from argparse import RawTextHelpFormatter
+from glob import glob
+from os.path import join, dirname, expanduser, basename
+from pathlib import Path
+from random import shuffle
+import shutil
+import sys
+
+import yaml
+
```
chrisveilleux: remove blank line.  all third party software goes in the same section, even if said software was written by us

mycroft/tts/tts.py
forslund: p is something that existed already, it was just not set up so it was guaranteed to exist. It's not in the scope of this PR to change the part of the code

test/integrationtests/voight_kampff/test_setup.py
forslund: Good suggestion, I'll rearrange things to be more similar to this layout. I'll keep the main function though and make sure the main control flow is representated.

test/integrationtests/voight_kampff/test_setup.py
forslund: This only get's the skill-list, lines below selects the actual random skills from the list of non-local skills.

test/integrationtests/voight_kampff/test_setup.py
forslund: In this case I prefer to keep it a single letter since
- it's an iterator variable
- it's inside a two line function with descriptive name
- it makes the code easier to read

test/integrationtests/voight_kampff/test_setup.py
forslund: In this case I prefer to keep it a single letter since
- it's an iterator variable
- it's inside a two line function with descriptive name
- it makes the code easier to read

test/integrationtests/voight_kampff/test_setup.py
forslund: I want to allow simple space separated list as well to a bit more lenient of semi-bad syntax.

forslund:
I think I've addressed all concerns here, either with a comment or changed code.

test/integrationtests/voight_kampff/test_setup.py
```diff
@@ -0,0 +1,226 @@
...
+    install_or_upgrade_skills(msm, all_skills)
+    collect_test_cases(msm, args.test_skills)
+
+    print_install_report(msm.platform, args.test_skills,
+                         args.extra_skills + random_skills)
+
+
+if __name__ == '__main__':
+    main(sys.argv[1:])
```
chrisveilleux: you don't need to pass the command line arguments to the arg parser.  it looks at sys.argv by default.

forslund: I personally like the traceability. main get's an argument to operate on so all actions can be traced back to it without any questions of where the settings originate.

test/integrationtests/voight_kampff/run_test_suite.sh
```diff
@@ -0,0 +1,25 @@
...
+fi
+# Start all mycroft core services.
+${SCRIPT_DIR}/../../../start-mycroft.sh all
+# Run the integration test suite.  Results will be formatted for input into
+# the Allure reporting tool.
+echo "Running behave with the arguments \"$@\""
+behave $@
+RESULT=$?
+# Stop all mycroft core services.
+/opt/mycroft/mycroft-core/stop-mycroft.sh all
```
krisgesling: Hey I just noticed that the start script got switched to a relative path, but the stop script got missed.

forslund: Good eyes, will fix!

forslund: Updated.

PR #2512, opened 2020-03-25
upgrade Lingua Franca to 0.2.1

ChanceNCounter:
Apparently the VSCode PR extension doesn't want to edit the template before creating a PR.

Upgrades Lingua Franca. Release notes at that repo. I have mixed feelings about this PR.

On the one hand, this release fixes critical flaws in extract_number(). On the other hand, there is another PR pending at Lingua Franca which would solve other, less critical, but equally frustrating issues with the same parser. At Ake's discretion, that PR could turn directly into a 0.2.2 release, putting a bow on the problem.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -21,7 +21,7 @@ google-api-python-client==1.6.4
 fasteners==0.14.1
 PyYAML==5.1.2
 
-lingua-franca==0.2.0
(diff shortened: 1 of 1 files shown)


forslund:
Let's merge this now and we'll see if we bump lingua franca another point this weekend.

PR #2514, opened 2020-03-26
Add configurable API url to Watson TTS

krisgesling:
## Description
The new IBM Watson TTS service now provides a unique url for your TTS
API resource. This adds a configuration option to override the default
url. As the url provided by IBM contains the api_path, we check for it 
and if found remove it, as the RemoteTTS.__request method appends it 
when making the call.

## How to test
Create an IBM cloud account and add Watson TTS as described in the [docs](https://mycroft-ai.gitbook.io/docs/using-mycroft-ai/customizations/tts-engine#ibm-watson).
Add a `url` attribute to your config using the url provided with your API key. This can be added with or without the API path "/v1/synthesize", and with or without trailing slashes.

### DIFF
--- a/mycroft/tts/ibm_tts.py
+++ b/mycroft/tts/ibm_tts.py
@@ -16,7 +16,6 @@
 from .tts import TTSValidator
 from .remote_tts import RemoteTTS
 from mycroft.configuration import Configuration
-from mycroft.util import remove_last_slash
 from requests.auth import HTTPBasicAuth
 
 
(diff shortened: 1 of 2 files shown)


forslund:
The `remove_trailing_slash` is suggested for removal in #2496, so maybe replace that with a simpler `.rstrip('/')

I also wonder if we want to do some generalization, allowing all remoteTTS to specify a custom url using the same config parameter, by changing the `self.url = url` line in remote_tts.py to `self.url = config.get("url", self.url)`.

I'm also wondering if the remoteTTS could use some extra customizability... but I'll add a separate issue for that.

krisgesling:
Yeah good suggestion re generalization. This seems like a good candidate.

Wondering if it's worth checking for the `url.endswith(api_path)` in `remote_tts.py` too?
The only case it would break for is if an API was in the form `api.com/endpoint/endpoint`
hence the `url == 'api.com/endpoint'`,
and `api_path == '/endpoint'`
and the url would then get stripped to become just `api.com`

This seems like a very remote case.

krisgesling:
Maybe I'll leave it where it is for now, and as you said we can consider further changes to remote_tts in a later PR.

forslund:
Looks great!

Issue #2509, opened 2020-03-23
Add aditional lang folder for skill loader to look into


andlo:
Some skills are missing translations or the translation on poodle isnt yet merged. At the same time when adding translation to poodle it is nearly imposible to test it right away.
What I am proposing is adding a aditional language folder outside of skills where aditional language files can be plased.
That would make it posible to translate localy or fetch translation from poodle and use it right away.

My proposal is adding a setting to mycroft.comf for aditional_language=/opt/mycroft/aditional_language/

Then when skill loader looks for lang files it looks first in skillfolder and if none exists it looks at aditional_lanuage for skill folder and lang files there.

Then a skil can be made that fetches langfiles from poodle and place them in aditional_language folder. Or a skill that auto translate and place files in there as wel or a combo of the two. 

In mycroft-skills.py the functions that reads lang files could be changed from 
```
   def load_vocab_files(self, root_directory):
        """ Load vocab files found under root_directory.

        Arguments:
            root_directory (str): root folder to use when loading files
        """
        keywords = []
        vocab_dir = join(root_directory, 'vocab', self.lang)
        locale_dir = join(root_directory, 'locale', self.lang)
        if exists(vocab_dir):
            keywords = load_vocabulary(vocab_dir, self.skill_id)
        elif exists(locale_dir):
            keywords = load_vocabulary(locale_dir, self.skill_id)
        else:
            LOG.debug('No vocab loaded')

        # For each found intent register the default along with any aliases
        for vocab_type in keywords:
            for line in keywords[vocab_type]:
                entity = line[0]
                aliases = line[1:]
                self.intent_service.register_adapt_keyword(vocab_type,
                                                           entity,
                                                           aliases)

    def load_regex_files(self, root_directory):
        """ Load regex files found under the skill directory.

        Arguments:
            root_directory (str): root folder to use when loading files
        """
        regexes = []
        regex_dir = join(root_directory, 'regex', self.lang)
        locale_dir = join(root_directory, 'locale', self.lang)
        if exists(regex_dir):
            regexes = load_regex(regex_dir, self.skill_id)
        elif exists(locale_dir):
            regexes = load_regex(locale_dir, self.skill_id)

        for regex in regexes:
            self.intent_service.register_adapt_regex(regex)
```
to something like this
```
   def load_vocab_files(self, root_directory):
        """ Load vocab files found under root_directory.

        Arguments:
            root_directory (str): root folder to use when loading files
        """
        keywords = []
[...]

PR #2519, opened 2020-03-28
additional lang folders

andlo:
aditional language

## Description
This tryes to add additional language folder as described in #2509

Additional lanuage folders are placed for each skill under ~/.mycroft/translations

Added settings for this in mycroft.cong under general skillsettings
```
  // General skill values
  "skills": {
    "translations_dir": "~/.mycroft/translations",
    // prefear translations over skills own translation
    "prefere_translations": false
    }
```
Added so the additional language folder is read when adding rescource files

It also add the additional language folder to the watching for changes in skill_loader so canges and addings to files at that place will reload the skill.

## How to test
* Use other language than en-us
* move files from hellow-world-skill (all files in other lang) from cocab and dialog older to ~/.mycroft/translations/mycroft-hello-world.mycroftai/<lang> - (so all resource files in same folder!
* See that mycroft picksup resource files from new place.
* make change in one file and se that he reloads skil and picksup changes

CLA [ X] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

### DIFF
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -118,7 +118,11 @@
     // priority skills to be loaded first
     "priority_skills": ["mycroft-pairing", "mycroft-volume"],
     // Time between updating skills in hours
-    "update_interval": 1.0
+    "update_interval": 1.0,
+    // Directory for aditional translations
+    "translations_dir": "~/.mycroft/translations",
+    // prefear translations over skills own translation
+    "prefere_translations": false
   },
 
   // Address of the REMOTE server
--- a/mycroft/skills/mycroft_skill/mycroft_skill.py
+++ b/mycroft/skills/mycroft_skill/mycroft_skill.py
@@ -19,8 +19,7 @@ import sys
 import re
 import traceback
 from itertools import chain
-from os import walk
-from os.path import join, abspath, dirname, basename, exists
+from os.path import join, abspath, dirname, basename, exists, expanduser, isdir
 from threading import Event, Timer
 
 from adapt.intent import Intent, IntentBuilder
@@ -87,7 +86,6 @@ def get_non_properties(obj):
     Returns:
         Set of attributes that are not a property.
     """
-
     def check_class(cls):
         """Find all non-properties in a class."""
         # Current class
@@ -142,7 +140,6 @@ class MycroftSkill:
         #: Mycroft global configuration. (dict)
         self.config_core = Configuration.get()
         self.dialog_renderer = None
-
         #: Filesystem access to skill specific folder.
         #: See mycroft.filesystem for details.
         self.file_system = FileSystemAccess(join('skills', self.name))
@@ -156,6 +153,12 @@ class MycroftSkill:
         # Delegator classes
         self.event_scheduler = EventSchedulerInterface(self.name)
         self.intent_service = IntentServiceInterface()
+
+        # Aditional translations
+        self.translations_dir = expanduser(self.config_core.get(
+                                          'skills').get('translations_dir'))
+        self.prefere_translations = self.config_core.get(
+                                    'skills').get('prefere_translations')
 
     @property
     def enclosure(self):
@@ -537,7 +540,7 @@ class MycroftSkill:
         cache_key = lang + voc_filename
         if cache_key not in self.voc_match_cache:
             # Check for both skill resources and mycroft-core resources
-            voc = self.find_resource(voc_filename + '.voc', 'vocab')
+            voc = self.find_resource(voc_filename + '.voc')
             if not voc:  # Check for vocab in mycroft core resources
                 voc = resolve_resource_file(join('text', lang,
                                                  voc_filename + '.voc'))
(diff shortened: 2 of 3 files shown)


mycroft/configuration/mycroft.conf
```diff
@@ -118,7 +118,11 @@
     // priority skills to be loaded first
     "priority_skills": ["mycroft-pairing", "mycroft-volume"],
     // Time between updating skills in hours
-    "update_interval": 1.0
+    "update_interval": 1.0,
+    // Directory for aditional translations
+    "translations_dir": "~/.mycroft/translations",
```
JarbasAl: could this maybe be made relative to "data_dir" param?

andlo: yes it could, but I would like to have translations in one logical place. 

having translations in one place makes it easier to acces, reviec and see which skills have additional translations. 

usecase is a auto-translation skill that autotranslate skills that are missing current language and a poodle sync skill that pulls down latest translations from translate.mycroft.ai (and maybe some day also push back translations)

andlo: also it is easier to monitor one folder for changes and need of reload skills

JarbasAl: i'll leave this up to you and @forslund, no strong opinion either way

i don't see any issue re what you just said, it is still a single folder.... data_dir is a param in mycroft.conf, it is already used internally in a couple places, it would just use /opt/mycroft/translations

what i meant was
```python
translations_dir = join(data_dir, translations_path)
```

not data_dir from the skills or anything like that

andlo: Oh I see - that were actual my first idea, but using ~/.mycroft makes sure there is read and write access across platforms and implementations.

andlo:
I cant figure out why test fails. It seems to be that I use od.path.expanduser() but that is called with the string from the mycroft setting. So it shouldnt be NoneType.

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -655,57 +658,70 @@ def translate(self, text, data=None):
         """
         return self.dialog_renderer.render(text, data or {})
 
-    def find_resource(self, res_name, res_dirname=None):
+    def find_resource(self, res_name, res_type=None):
```
forslund, line 661: keep the res_dirname name for backwards compatibility

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -1116,70 +1132,62 @@ def acknowledge(self):
...
-        # If "<skill>/dialog/<lang>" exists, load from there.  Otherwise
-        # load dialog from "<skill>/locale/<lang>"
-        dialog_dir = join(root_directory, 'dialog', self.lang)
-        if exists(dialog_dir):
-            self.dialog_renderer = load_dialogs(dialog_dir)
-        elif exists(join(root_directory, 'locale', self.lang)):
-            locale_path = join(root_directory, 'locale', self.lang)
-            self.dialog_renderer = load_dialogs(locale_path)
-        else:
-            LOG.debug('No dialog loaded')
```
forslund, line 1129: `init_dialog()` is part of the public skill api and needs to remain. At least until 20.08.

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -1116,70 +1132,62 @@ def acknowledge(self):
...
+
         Arguments:
             root_directory (str): root folder to use when loading files.
         """
         root_directory = root_directory or self.root_dir
-        self.init_dialog(root_directory)
-        self.load_vocab_files(root_directory)
-        self.load_regex_files(root_directory)
-
-    def load_vocab_files(self, root_directory):
```
forslund, line 1142: This is also part of the public API and shouldn't be removed

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -1116,70 +1132,62 @@ def acknowledge(self):
...
-        # For each found intent register the default along with any aliases
-        for vocab_type in keywords:
-            for line in keywords[vocab_type]:
-                entity = line[0]
-                aliases = line[1:]
-                self.intent_service.register_adapt_keyword(vocab_type,
-                                                           entity,
-                                                           aliases)
-
-    def load_regex_files(self, root_directory):
```
forslund, line 1167: Same goes for this.

andlo: Great and thanks. I will redo my work and keep the funcions and see if I can figure out what jenkins says....

andlo:
Great and thanks. I will redo my work and keep the funcions and see if I can figure out what jenkins says....

HKalbasi:
What is the status of this PR?

ChanceNCounter:
@HKalbasi see the Roadmap link in the status change above. I haven't gotten used to checking that, but it's cool.

Anyway, this was moved to "Longer term," which is described as

> Work that is needed but can wait until “Next” [category to the right] items are finished

denics:
Hi all, please reconsider this to be not that much longer term. This would be extremely useful especially for non en-en users.

krisgesling:
Hey, I can't give you a date when the internal team will start work on it, but better multi-lingual support is one of our top 5 priorities. So it is going to get a lot of attention this year. Personally I think something like this PR is the way to go too.

denics:
Hey @krisgesling now that the new Mark II is out, what is the status of this?

PR #2526, opened 2020-04-01
feat/ cli GUI client

JarbasAl:
## Description
Adds a very simple CLI GUI, useful for debugging and skill developers. 

Original a quick and dirty script, but @AIIX suggested i made a PR

simply_gui

## How to test
```./start-mycroft.sh gui```

### DIFF
--- a/mycroft/client/debug_gui/__main__.py
+++ b/mycroft/client/debug_gui/__main__.py
@@ -1,0 +1,159 @@
+from os import getpid
+from os.path import basename
+import json
+from time import sleep
+from pprint import pformat
+
+from mycroft.util.log import LOG
+from mycroft.messagebus import Message, get_websocket
+from mycroft.messagebus.client import MessageBusClient
+
+
+class bcolors:
+    # helper to print in color
+    HEADER = '\033[95m'
+    OKBLUE = '\033[94m'
+    OKGREEN = '\033[92m'
+    WARNING = '\033[93m'
+    FAIL = '\033[91m'
+    ENDC = '\033[0m'
+    BOLD = '\033[1m'
+    UNDERLINE = '\033[4m'
+
+
+class DebugGUI:
+    def __init__(self, bus, name=None, debug=False):
+        self.bus = bus
+        self.loaded = []
+        self.skill = None
+        self.page = None
+        self.vars = {}
+        self.gui_ws = None
+        self.name = name or self.__class__.__name__.lower()
+        self.debug = debug
+        self.connected = False
+        self.buffer = []
+
+    def run(self):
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by AIIX** [comment]
The dev branch seems to have an issue with establishing multiple gui client connections and gui debug client doesn't work when gui is already running a similar issue to: https://github.com/MycroftAI/mycroft-core/issues/2525

It works nicely on the bugfix workaround-multi-displays branch, so this branch should be merged in dev for this to work correctly: https://github.com/MycroftAI/mycroft-core/tree/bugfix/workaround-multi-displays

**Comment by JarbasAl** [comment]
im making a cli GUI, this is now included as a debugger

https://github.com/JarbasAl/Mycroft_cli_GUI

**Comment by krisgesling** [comment]
Hey sorry I haven't looked at this properly before. I really like it as a developer focused GUI debugging tool.

I've been leaning toward core being as lightweight and agnostic as possible, so that it can be as "composable" as possible through build recipes. I'm wondering if it makes the most sense to have it as a separate package, rather than included in core? 

Same with tools like MSK. As we move to more end-user installations, I don't see a need to have it on every end-user device. I think it makes sense on Picroft as it's a hackers/makers release. But on the Mark II - not necessarily.

I'm really just thinking out loud here, so interested in others thoughts on this.

**Comment by JarbasAl** [comment]
this is part of my TUI  https://github.com/JarbasAl/Mycroft-TUI  it can be used as a standalone script, its still WIP and im still figuring out how i want to package it, but its available for anyone to use

Makes no difference to me if its part of core or not, if you feel its just bloat feel free to close this PR

**Comment by krisgesling** [comment]
Would it be possible to load it as a plugin using the [methods Ake's proposed](https://github.com/MycroftAI/mycroft-core/pull/2594/commits/74cb9c48f0c9f5c31faf1ab729d3fa9a59504194)?

**Comment by JarbasAl** [comment]
> Would it be possible to load it as a plugin using the [methods Ake's proposed](https://github.com/MycroftAI/mycroft-core/pull/2594/commits/74cb9c48f0c9f5c31faf1ab729d3fa9a59504194)?

i dont think that makes sense, this is not a plugin, it just connects to the bus

i can just put it up on pip, then people can do something like
```
pip install mycroft-gui-debugger
(.venv) mycroft-gui-debugger
```

krisgesling:
Yeah fair call. If it is on PyPI then we can easily add something to provide a simple installation for it. 
> "If you're a dev, do you want these tools installed - [x, y, z]"

Although maybe that should really be in MSK. 

Either way, I'll close this for now. If anyone disagrees, they can re-open.

PR #2530, opened 2020-04-03
Feature/email jenkins results

chrisveilleux:
## Description
Add steps to Jenkinsfile to email the results of integration tests on success or failure

## How to test
- Make a branch off of this one and create local version of it
- add the branch name to the "when" clause in the integration test stage
- push change to origin

CLA [yes]

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -73,8 +73,8 @@ pipeline {
                         label: 'Publish Report to Web Server',
                         script: '''scp allure-report.zip root@157.245.127.234:~;
                             ssh root@157.245.127.234 "unzip -o ~/allure-report.zip";
-                            ssh root@157.245.127.234 "rm -rf /var/www/voight-kampff/${BRANCH_ALIAS}";
-                            ssh root@157.245.127.234 "mv allure-report /var/www/voight-kampff/${BRANCH_ALIAS}"
+                            ssh root@157.245.127.234 "rm -rf /var/www/voight-kampff/core/${BRANCH_ALIAS}";
+                            ssh root@157.245.127.234 "mv allure-report /var/www/voight-kampff/core/${BRANCH_ALIAS}"
                         '''
                     )
                     echo 'Report Published'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -80,6 +80,13 @@ pipeline {
                     echo 'Report Published'
                 }
                 failure {
+                    script {
+                        // Create comment for Pull Requests
+                        if (env.CHANGE_ID) {
+                            echo 'Sending PR comment'
+                            pullRequest.comment('Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/' + env.BRANCH_ALIAS + '))')
+                        }
+                    }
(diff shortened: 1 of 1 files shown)


forslund:
Should we send success-emails? I'd say it's better to only send things when there is something worth noting.

forslund:
Well Github shows the jobs that have finished and their status so I think that information is available and only really of interest to the one reviewing a PR. since a contributor won't have the possibility to access the Jenkins job and restart it afaik.

krisgesling:
I assumed the allure report would be accessible to anyone with the link so they can't restart the job but they could see why their build failed.

forslund:
Yeah exactly, for a failure that's usful but for a success it's redundant imho.

forslund:
I realize now that the results can't be viewed at all except from the link in the e-mail. So maybe it's good to have for now. I do think we should try to get report returned to the the PR instead if the build is successful.

Jenkinsfile
```diff
@@ -73,12 +73,85 @@ pipeline {
...
+                                &nbsp(Requires account on Mycroft's Jenkins instance)
+                            </p>
+                            <br>
+                            <p>
+                                <a href='https://reports.mycroft.ai/core/${BRANCH_ALIAS}'>
+                                    Report of Test Results
+                                </a>
+                            </p>""",
+                        replyTo: 'devops@mycroft.ai',
+                        to: 'devs@mycroft.ai',
```
forslund, line 147: This also sends to devs instead of dev.

Jenkinsfile
```diff
@@ -73,12 +73,85 @@ pipeline {
...
+                        body: """
+                            <p>
+                                All integration tests passed.  No further action required.
+                            </p>
+                            <br>
+                            <p>
+                                <a href='${BUILD_URL}'>
+                                    Jenkins Build Details
+                                </a>
+                                &nbsp(Requires account on Mycroft's Jenkins instance)
```
forslund, line 138: The non-breaking space doesn't render correctly in the e-mail

Jenkinsfile
```diff
@@ -73,12 +73,85 @@ pipeline {
...
+                    // Send success email containing a link to the Jenkins build
+                    // and the results report to Mycroft developers, the developers
+                    // of the pull request and the developers that caused the
+                    // last failed build.
+                    echo 'Sending Success Email'
+                    emailext (
+                        subject: "SUCCESS - Core Integration Tests - Build ${BRANCH_NAME} #${BUILD_NUMBER}",
+                        body: """
+                            <p>
+                                All integration tests passed.  No further action required.
```
forslund, line 131: There's an extra space after the `.`

chrisveilleux: I am an old, two space after the sentence guy.  I sometimes forget that HTML ignores the second space...

Jenkinsfile
chrisveilleux: Could you be more specific?  I noticed a space in the link after the text but not sure how to get rid of it.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2530))

PR #2543, opened 2020-04-20
Fix uploading settings meta

forslund:
## Description
This changes the order so the started flag is set before the settingsmeta
the send() method is called.

The send() method now exits if the started flag isn't set causing the
skill settings not to be uploaded.

The issue was introduced in PR #2529 when adding the possibility to stopping the upload loop.

## How to test
Startup mycroft-core and check the skills log so that settingsmeta are uploaded.

CLA [ Yes ]

### DIFF
--- a/mycroft/skills/skill_manager.py
+++ b/mycroft/skills/skill_manager.py
@@ -48,8 +48,8 @@ class UploadQueue:
 
     def start(self):
         """Start processing of the queue."""
-        self.send()
         self.started = True
+        self.send()
 
     def stop(self):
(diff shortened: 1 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2543))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2543))

PR #2559, opened 2020-05-02
Check the skill directory and XDG_CONFIG_DIR for settings

PureTryOut:
## Description
This makes the skills read their settings from either XDG_CONFIG_DIR or `/etc/mycroft` (if `XDG_CONFIG_DIR` is unset), and write to XDG_CONFIG_DIR. This however happens only if `/opt/mycroft/skills/<skillname>/settings.json` (the current default location) doesn't already exist, to make sure we don't break existing setups.

## How to test
1. Run `mycroft-skills` like before and verify if all skills still load correctly
2. Quit the running process
3. **Move** a `settings.json` from any skill to either `~/.config/mycroft/skills/<skillname>/settings.json` or `/etc/mycroft/skills/<skillname>/settings.json`. For `skill-pairing` the skillname would be "PairingSkill" for example
4. Run `mycroft-skills` again and verify if the affected skill still loads correctly and if it saves it settings to `~/.config/mycroft/skills/<skillname>/settings.json`.

CLA [x]

### DIFF
--- a/mycroft/skills/mycroft_skill/mycroft_skill.py
+++ b/mycroft/skills/mycroft_skill/mycroft_skill.py
@@ -21,7 +21,10 @@ import traceback
 from itertools import chain
 from os import walk
 from os.path import join, abspath, dirname, basename, exists
+from pathlib import Path
 from threading import Event, Timer
+
+from xdg import BaseDirectory
 
 from adapt.intent import Intent, IntentBuilder
 
@@ -123,16 +126,6 @@ class MycroftSkill:
         #: Member variable containing the absolute path of the skill's root
         #: directory. E.g. /opt/mycroft/skills/my-skill.me/
         self.root_dir = dirname(abspath(sys.modules[self.__module__].__file__))
-        if use_settings:
-            self.settings = get_local_settings(self.root_dir, self.name)
-            self._initial_settings = deepcopy(self.settings)
-        else:
-            self.settings = None
-
-        #: Set to register a callback method that will be called every time
-        #: the skills settings are updated. The referenced method should
-        #: include any logic needed to handle the updated settings.
-        self.settings_change_callback = None
 
         self.gui = SkillGUI(self)
 
@@ -141,6 +134,18 @@ class MycroftSkill:
         self.bind(bus)
         #: Mycroft global configuration. (dict)
         self.config_core = Configuration.get()
+
+        self.settings = None
+        self.settings_write_path = None
+
+        if use_settings:
+            self._init_settings()
+
+        #: Set to register a callback method that will be called every time
+        #: the skills settings are updated. The referenced method should
+        #: include any logic needed to handle the updated settings.
+        self.settings_change_callback = None
+
         self.dialog_renderer = None
 
         #: Filesystem access to skill specific folder.
@@ -156,6 +161,37 @@ class MycroftSkill:
         # Delegator classes
         self.event_scheduler = EventSchedulerInterface(self.name)
         self.intent_service = IntentServiceInterface()
+
+    def _init_settings(self):
+        """Setup skill settings."""
+
+        # To not break existing setups,
+        # save to skill directory if the file exists already
+        self.settings_write_path = Path(self.root_dir)
+
+        # Otherwise save to XDG_CONFIG_DIR
+        if not self.settings_write_path.joinpath('settings.json').exists():
+            self.settings_write_path = Path(BaseDirectory.save_config_path(
+                'mycroft', 'skills', basename(self.root_dir)))
+
+        # To not break existing setups,
+        # read from skill directory if the settings file exists there
+        settings_read_path = Path(self.root_dir)
+
+        # Then, check XDG_CONFIG_DIR
+        if not settings_read_path.joinpath('settings.json').exists():
+            for dir in BaseDirectory.load_config_paths('mycroft',
+                                                       'skills',
+                                                       basename(
+                                                           self.root_dir)):
+                path = Path(dir)
+                # If there is a settings file here, use it
+                if path.joinpath('settings.json').exists():
+                    settings_read_path = path
+                    break
+
+        self.settings = get_local_settings(settings_read_path, self.name)
+        self._initial_settings = deepcopy(self.settings)
 
     @property
     def enclosure(self):
@@ -271,7 +307,7 @@ class MycroftSkill:
             if remote_settings is not None:
                 LOG.info('Updating settings for skill ' + self.name)
                 self.settings.update(**remote_settings)
-                save_settings(self.root_dir, self.settings)
+                save_settings(self.settings_write_path, self.settings)
                 if self.settings_change_callback is not None:
                     self.settings_change_callback()
 
@@ -544,7 +580,7 @@ class MycroftSkill:
 
             if not voc or not exists(voc):
                 raise FileNotFoundError(
-                        'Could not find {}.voc file'.format(voc_filename))
+                    'Could not find {}.voc file'.format(voc_filename))
             # load vocab and flatten into a simple list
             vocab = read_vocab_file(voc)
             self.voc_match_cache[cache_key] = list(chain(*vocab))
@@ -811,7 +847,7 @@ class MycroftSkill:
             """Store settings and indicate that the skill handler has completed
             """
             if self.settings != self._initial_settings:
-                save_settings(self.root_dir, self.settings)
+                save_settings(self.settings_write_path, self.settings)
                 self._initial_settings = self.settings
             if handler_info:
                 msg_type = handler_info + '.complete'
@@ -1232,8 +1268,9 @@ class MycroftSkill:
         self.settings_change_callback = None
 
         # Store settings
-        if self.settings != self._initial_settings:
-            save_settings(self.root_dir, self.settings)
+        if self.settings != self._initial_settings and Path(
+                self.root_dir).exists():
+            save_settings(self.settings_write_path, self.settings)
 
         if self.settings_meta:
             self.settings_meta.stop()
--- a/mycroft/skills/settings.py
+++ b/mycroft/skills/settings.py
@@ -93,18 +93,22 @@ def get_local_settings(skill_dir, skill_name) -> dict:
 def save_settings(skill_dir, skill_settings):
     """Save skill settings to file."""
     settings_path = Path(skill_dir).joinpath('settings.json')
-    if Path(skill_dir).exists():
-        with open(str(settings_path), 'w') as settings_file:
-            try:
-                json.dump(skill_settings, settings_file)
-            except Exception:
-                LOG.exception('error saving skill settings to '
-                              '{}'.format(settings_path))
-            else:
-                LOG.info('Skill settings successfully saved to '
-                         '{}' .format(settings_path))
-    else:
-        LOG.info('Skill folder no longer exists, can\'t save settings.')
+
+    # Either the file already exists in /opt, or we are writing
+    # to XDG_CONFIG_DIR and always have the permission to make
+    # sure the file always exists
+    if not Path(settings_path).exists():
+        settings_path.touch(mode=0o644)
+
+    with open(str(settings_path), 'w') as settings_file:
+        try:
+            json.dump(skill_settings, settings_file)
+        except Exception:
+            LOG.exception('error saving skill settings to '
+                          '{}'.format(settings_path))
+        else:
+            LOG.info('Skill settings successfully saved to '
+                     '{}' .format(settings_path))
 
 
 def get_display_name(skill_name: str):
--- a/mycroft/skills/skill_loader.py
+++ b/mycroft/skills/skill_loader.py
@@ -259,7 +259,8 @@ class SkillLoader:
         if first_run:
             LOG.info("First run of " + self.skill_id)
             self.instance.settings["__mycroft_skill_firstrun"] = False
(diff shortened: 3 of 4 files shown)


mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -157,6 +162,29 @@ def __init__(self, name=None, bus=None, use_settings=True):
         self.event_scheduler = EventSchedulerInterface(self.name)
         self.intent_service = IntentServiceInterface()
 
+    def _init_settings(self, prefix):
```
forslund, line 165: prefix isn't actually used since you changed it to use XDG_CONFIG_DIR

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -157,6 +162,29 @@ def __init__(self, name=None, bus=None, use_settings=True):
...
+        """Setup skill settings."""
+
+        #: First, check the skill directory to not break existing setups
+        self.settings_dir = self.root_dir
+
+        #: Then, check XDG_CONFIG_DIR
+        if not isdir(self.settings_dir):
```
forslund, line 172: Won't this always be true? since this is the path to the skill folder?

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -157,6 +162,29 @@ def __init__(self, name=None, bus=None, use_settings=True):
...
+            for dir in BaseDirectory.load_config_paths('mycroft'):
+                path = dir + 'skills' + basename(self.root_dir)
+                #: If there is a settings file here, use it
+                if isdir(path):
```
forslund, line 176: On first start will this path actually exist? shouldn't it be created if the XDG_CONFIG_DIR is to be used?

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -157,6 +162,29 @@ def __init__(self, name=None, bus=None, use_settings=True):
...
+    def _init_settings(self, prefix):
+        """Setup skill settings."""
+
+        #: First, check the skill directory to not break existing setups
+        self.settings_dir = self.root_dir
+
+        #: Then, check XDG_CONFIG_DIR
+        if not isdir(self.settings_dir):
+            for dir in BaseDirectory.load_config_paths('mycroft'):
+                path = dir + 'skills' + basename(self.root_dir)
```
forslund, line 174: use the join() to join the paths `join(dir, 'skills', basename(self.root_dir))`

mycroft/skills/settings.py
```diff
@@ -90,21 +111,26 @@ def get_local_settings(skill_dir, skill_name) -> dict:
     return skill_settings
 
 
-def save_settings(skill_dir, skill_settings):
+def save_settings(skill_dir, skill_name, skill_settings):
```
forslund, line 114: Let's not change the logic and interface here, it should do the save in the specified directory. The correct directory should be sent in through the skill_directory argument.

mycroft/skills/mycroft_skill/mycroft_skill.py
PureTryOut: It might not exist, but that shouldn't be a problem as afaik the skills currently start without settings as well on first start. At least on my system this works fine on first start without any settings anywhere.

When writing the settings we do make sure XDG_CONFIG_DIR exists though.

mycroft/skills/settings.py
```diff
@@ -74,8 +76,27 @@
 def get_local_settings(skill_dir, skill_name) -> dict:
     """Build a dictionary using the JSON string stored in settings.json."""
     skill_settings = {}
+    settings_path = None
```
forslund, line 79: Let's not change the logic here, instead determine the correct path in MycroftSkill and use that as an argument

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2559))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2559))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2559))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2559))

requirements.txt
```diff
@@ -30,3 +30,4 @@ fann2==1.0.7
 padaos==0.1.9
 precise-runner==0.2.1
 petact==0.1.2
+pyxdg>=0.26
```
forslund, line 33: Let's keep the requirement locked pyxdg==0.26 (using the loose requirements thing)

PureTryOut: Good call, fixed.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2559))

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -141,6 +135,48 @@ def __init__(self, name=None, bus=None, use_settings=True):
...
+                '/etc/mycroft/skills/').joinpath(self.name)
+
+        if use_settings:
+            self._init_settings()
+
+        #: Set to register a callback method that will be called every time
+        #: the skills settings are updated. The referenced method should
+        #: include any logic needed to handle the updated settings.
+        self.settings_change_callback = None
+
```
forslund, line 179: This whole block is quite big perhaps move it into the _init_settings() and pass the use_settings as a parameter

mycroft/skills/settings.py
```diff
@@ -93,18 +93,22 @@ def get_local_settings(skill_dir, skill_name) -> dict:
...
-                LOG.info('Skill settings successfully saved to '
-                         '{}' .format(settings_path))
-    else:
-        LOG.info('Skill folder no longer exists, can\'t save settings.')
+
+    #: Either the file already exists in /opt, or we are writing
+    #: to XDG_CONFIG_DIR and always have the permission to make
+    #: sure the file always exists
+    if not Path(settings_path).exists():
+        settings_path.touch(mode=0o644)
```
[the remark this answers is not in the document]
forslund:
```python
def save_settings(skill_dir, skill_settings):
    """Save skill settings to file."""
    try:
        settings_path = Path(skill_dir).joinpath('settings.json')
        #: Either the file already exists in /opt, or we are writing
        #: to XDG_CONFIG_DIR and always have the permission to make
        #: sure the file always exists
        if not Path(settings_path).exists():
            settings_path.touch(mode=0o644)
        with open(str(settings_path), 'w') as settings_file:
            json.dump(skill_settings, settings_file)
    except Exception:
        LOG.exception('error saving skill settings to '
                      '{}'.format(settings_path))
    else:
        LOG.info('Skill settings successfully saved to '
                 '{}' .format(settings_path))

File: mycroft/skills/mycroft_skill/mycroft_skill.py [hunk 10]
```diff
@@ -141,6 +135,48 @@ def __init__(self, name=None, bus=None, use_settings=True):
...
+        self.settings = None
+
+        #: To not break existing setups,
+        #: save to skill directory if the file exists already
+        self.settings_write_path = Path(self.root_dir)
+
+        #: Otherwise save to XDG_CONFIG_DIR
+        if not self.settings_write_path.joinpath('settings.json').exists():
+            self.settings_write_path = Path(BaseDirectory.save_config_path(
+                'mycroft', 'skills', self.name))
```
forslund, line 148: possibly this should be the dirname and not self.name since self.name is settable by the skill and may be changed over time while the skill directory name will always be the same unless a user manually moves the skill (or the package moves the skill)

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -141,6 +135,48 @@ def __init__(self, name=None, bus=None, use_settings=True):
...
+                path = Path(dir)
+                #: If there is a settings file here, use it
+                if path.joinpath('settings.json').exists():
+                    self.settings_read_path = path
+                    break
+
+        #: Lastly, check /etc/mycroft
+        if not self.settings_read_path.joinpath('settings.json').exists():
+            self.settings_read_path = Path(
+                '/etc/mycroft/skills/').joinpath(self.name)
```
forslund, line 170: Is the /etc/mycroft addition needed?

mycroft/skills/settings.py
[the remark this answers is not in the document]
forslund: It also strikes me that removed skills will save their settings here on their final shutdown. Which may not be a bad thing...

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -141,6 +135,48 @@ def __init__(self, name=None, bus=None, use_settings=True):
...
+        self.settings_write_path = Path(self.root_dir)
+
+        #: Otherwise save to XDG_CONFIG_DIR
+        if not self.settings_write_path.joinpath('settings.json').exists():
+            self.settings_write_path = Path(BaseDirectory.save_config_path(
+                'mycroft', 'skills', self.name))
+
+        self.settings_read_path = None
+
+        #: To note break existing setups,
```
forslund, line 152: note -> not

forslund: Also the `#:` comments are things that will be picked up by the [sphinxdocs](https://mycroft-core.readthedocs.io/en/latest/) and these should be more on how to use the member variables, this is probably a note to core developers and not skill developers so the `:` can be removed. (goes for a bunch of other places as well)

mycroft/skills/settings.py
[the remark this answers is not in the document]
PureTryOut: The directory will at that point in the function always exists. If we're writing to a XDG directory, the folder will be created by `BaseDirectory.save_config_path()` if it does not exist yet, so you never have to check if the folder exists. In your local test you're probably not using `BaseDirectory.save_config_path()`, in which case the folder might not exist and will then indeed error out.

As for your last edit, in case of not saving to the skill directory, it should still work fine no? With your proposed change we'll not save the settings on shutdown ever, even if we would've saved it to the XDG directory which would've been fine.

mycroft/skills/mycroft_skill/mycroft_skill.py
PureTryOut: Not necessarily, but that is what most programs on Linux adhere too. Just check your average `/etc` folder on any Linux distributions, most programs have their system-wide settings there.

PureTryOut:
> Do we need to differentiate between settings read path and settings write path?

Yes. While you can read from `/etc/xdg` or `/etc/mycroft`, you should never write to them, that's where `~/.config` is for.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2559))

mycroft/skills/settings.py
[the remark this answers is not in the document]
forslund: With the last edit this should be fine

mycroft/skills/mycroft_skill/mycroft_skill.py
forslund: The settings are something the skills handle them selves, read only defaults should really not be introduced imho. Basically the only places I'd expect to see the skill settings would be `$XDG_CONFIG_HOME/...` and the settings dir, none of the read only places is really of interest.

forslund: Does the settings_read_path need to be a member though? it's read at startup and then the only thing of interest to the skill is the write path

PureTryOut: Ok, I removed the check for `/etc/mycroft`. Also made `settings_read_path` a local variable only.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2559))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2559))

forslund:
Think you might have rebased on top of master instead of dev, PR is now 10000 commits

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2559))

PureTryOut:
Ok why is the master branch still a thing if it's never used and updated? :see_no_evil: 

Reset the branch and should be fixed now.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2559))

forslund:
It is used, it's the "stable" branch pointing to the latest release basically, but due to some initial script snafu's it has a bunch of extra commits (like 500k), but since it's a branch that's used no-one has wanted to drop it and re-create it.

PureTryOut:
How strange.

Anyhow, anything more I need to fix or change?

forslund:
I think all looks good, gonna do a final test run on the Mark-1 hardware and then it should be merged.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2559))

Issue #2517, opened 2020-03-27
Relax dependency version requirements


PureTryOut:
I'm packaging Mycroft for Alpine Linux, and am currently working my way through all the Python dependencies. I noticed that mycroft-core currently expects really specific versions of requirements (e.g. `padaos==0.1.9`), instead of just the minimal version it needs for it's functionality. This makes it hard to package, as rolling distributions tend to package the newest version of every library.

This means that even newer Mycroft libraries don't get used. For example [`precise-runner`](https://github.com/MycroftAI/mycroft-precise) 0.3.1 exists, but mycroft-core still expects 0.2.1, so no one actually gets to use the improvements made there.

Please start relaxing the version requirements, do something like `precise-runner>=0.2.1`. It'll make it a lot easier for rolling distributions to package mycroft and makes sure people get the benefits of all improvements made to all libraries.

forslund:
Hi thanks for starting this discussion.

We've been sticking with fixed versions to make sure upstream bugs don't affect us and try to upgrade in a controlled fashion. We've been hit a couple of times where some users get problems due to a specific version of a package containing a new bug. it usually goes away after a while.

precise-runner hasn't been updated because of issues with the precise-engine 0.3.1 so we can't actually move to that version. (We upgraded and then reverted the upgrade due to issues with it)

One alternative is to use the ~= to make sure the api is the same but including bugfixes / non-breaking issues but again that's no guarantee and a person installing one day can have an environment that works flawlessly while an user installing the next day has terrible issues.

I would love to see your packaging and help in any way we can. The common issue is the install of dependencies by skills. On debian based systems we've used dh-virtualenv to package a sandboxed virtual environment that skills freely can install their dependencies into. I'd very much like to see your solution.

PureTryOut:
> I would love to see your packaging and help in any way we can.

Awesome, thanks! That is a great response! Do note that it's very WIP, I literally just started this today. You can find the packaging here https://gitlab.alpinelinux.org/PureTryOut/aports/commits/testing_mycroft.

> We've been hit a couple of times where some users get problems due to a specific version of a package containing a new bug

[...]

PR #2562, opened 2020-05-04
Allow building Mycroft with loose requirements

forslund:
## Description
Allow pip install to use "loose" requirements (>= instead of ==) by setting the `MYCROFT_LOOSE_REQUIREMENTS` environment variable.

The intention of this is to allow installs with desktop packages instead of using pip to be smoother while allowing the git installation to still guarantee the expected module versions. (Discussed in #2517).

This will not be guaranteed to work, but allows packagers for distributions to install their own versions of the Mycroft dependencies through their packaging manager.

Example
`MYCROFT_LOOSE_REQUIREMENTS=true pip install .`

## How to test
Create a test virtual environment, activate it and run `MYCROFT_LOOSE_REQUIREMENTS=true pip install .` Make sure the installed versions are of the packages are the latest (for example requests should be the latest version), (depending on the arch and system this MAY FAIL to setup due to fann2 1.1.2 which has a broken install on some systems. If this occurs install the 1.0.7 version separately.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 #
 from setuptools import setup, find_packages
+import os
 import os.path
 
 BASEDIR = os.path.abspath(os.path.dirname(__file__))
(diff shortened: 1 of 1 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2562))

PureTryOut:
Thanks for this! Note that the same should happen for `lingua-franca`.

PR #2568, opened 2020-05-06
CI: Reduce Jenkins reporting verbosity

forslund:
# Description
- E-mails are only sent on failing branch builds
- PR comment is only sent on failure

The rationale here is that the build status is available from github so sending success emails / PR comments provides no extra useful information. A failure however is more likely to need immediate response from the PR author. Branch build failures will be reported by e-mail to the dev-team so these things can be caught quickly. This mirrors the behavior of Travis and PEP8-speaks.

See [this test run](https://hudson.mycroft.team/blue/organizations/jenkins/mycroft-core/detail/feature%2FJenkinsfile-verbosity/3/pipeline) for an example of e-mail notifications still being sent on failure.

CLA [ Yes ]

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -81,89 +81,50 @@ pipeline {
                 }
                 failure {
                     script {
-                        // Create comment for Pull Requests
                         if (env.CHANGE_ID) {
+                            // Create comment for Pull Requests
                             echo 'Sending PR comment'
                             pullRequest.comment('Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/' + env.BRANCH_ALIAS + '))')
+                        } else {
+                            // if it's the build of a branch send failure email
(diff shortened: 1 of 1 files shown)


krisgesling:
This would be my preferred reporting level personally. Though currently I just filter all the success emails, so doesn't bother me a great deal.

For PR's, if it fails, anyone subscribed to that thread will be notified, hence prompts the author to take another look. If it succeeds, no new message required as the expectation is already for pull requests to get reviewed.

For branch builds, if it fails, get emailed and go investigate what's gone wrong.

If there's disagreement still, we could get some feedback from the community on how they'd like it.

forslund:
I think what we agreed on on last time we discussed this was to always send an email to dev@mycroft.ai and I neede to update this PR to reflect that before it's good to merge.

forslund:

PR #2578, opened 2020-05-08
Use XDG Base directories for settings, cache and runtime data

PureTryOut:
## Description
This makes Mycroft use the [XDG Base Directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).
Data is now saved to:
* settings to `$XDG_CONFIG_DIR/mycroft`, which most of the time equals to `~/.config/mycroft`
* data to `$XDG_DATA_DIRS/mycroft`, which most of the time equals to `~/.local/share/mycroft`
* cache to `$XDG_CACHE_HOME/mycroft`, which most of the time equals to `~/.cache`

Since this changes the default location of pretty much everything, _please test it well_!

Goal is to support and read from the old `~/.mycroft` location if it already exists, but save to the XDG Base Directories and read from it on new installations.

Together with #2559 this should be full XDG Base Directory compliance.

## How to test
First of all, run this PR with an existing installation, and check if everything is loaded from the usual locations. Then shut it down and restart to make sure it has saved to the new locations.
Then, completely remove any trace of Mycroft settings from your system. Easiest would be to fully deinstall (remove `/opt/mycroft`), remove `~/.mycroft`, and the XDG Base directories mentioned above.
Now reinstall Mycroft (with this PR applied) and start and connect it. Check if `~/.mycroft` is _not_ created anymore and `~/.config/mycroft`, `~/.local/share/mycroft` and `~/.cache/mycroft` _are_ with the correct files and directories in them.

CLA [x]


devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2578))

test/util.py
```diff
@@ -1,7 +1,10 @@
-from mycroft.configuration.config import LocalConf, DEFAULT_CONFIG
+from os.path import join, dirname
+from mycroft.configuration.config import LocalConf
 from copy import deepcopy
+from xdg import BaseDirectory
 
-__config = LocalConf(DEFAULT_CONFIG)
+__config = LocalConf(join(dirname(dirname(__file__)),
```
PureTryOut, line 6: Not a fan of this, but I'm not sure how else to resolve this without reintroducing `DEFAULT_CONFIG`...

PureTryOut:
> I still haven't finished writing the specs for this feature you've already implemented 

Woops, sorry. I got too excited for this :see_no_evil:

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2578))

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2578))

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2578))

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2578))

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2578))

PureTryOut:
> I still haven't finished writing the specs for this feature you've already implemented so I'll start by sending some of the notes I made on this PR this weekend.

@forslund I realize you're less active nowadays but is there a chance you have finished the specs you were writing? I'd like to finish this PR at some point :wink:

forslund:
Sure I'll try to have my thoughts jotted down within the week.

forslund:
Hi, sorry for the long delay in coming back with this. You seem to have implemented much of what I describe already.

- FilesystemAccess
 - the mycroft skill filesystem "scratchpad": This is a very internal system and can just be moved by the startup of the skills service

 - identity location: This is sort of an internal object and can be moved but it's sort of tricky to do a clean move since it's accessed by all Mycroft processess. Easiest is probably to allow both the old and an XDG based location but when writing new identity information use the new location.

- mycroft configuration locations: Easiest and most compatible would be to just insert the XDG config locations into the list making it something like 
[DEFAULT_CONFIG (mycroft internal), XDG_SYSTEM_CONFIG, SYSTEM_CONFIG (/etc/mycroft), USER_CONFIG (~/.mycroft), XDG_USER_CONFIG, patch]
Possibly set the USER_CONFIG path to XDG_USER_CONFIG if the USER_CONFIG doesn't exist

- precise model download location: Should check all the XDG_locations /share/mycroft/models and the current one. Would be good if we could make a package out of this one as well and include it in /usr/share somewhere.

- precise binary download: This could do a move to an XDG based data directory since it's really only used by a single process. (this download should in my opinion be removed and instead a package should be used or downloaded through the setup script.

- padatious intent cache: This could be moved the users XDG_DATA directory completely. May not need a move even. Padatious will regenerate the cache if it doesn't find it in the new location.

May not be much news but these were the thoughts I had (had to rebuild my document)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2578))

forslund:
My main reasoning was to avoid needing to have config file merging on disk, just in memory (which would use the existing system).

If the XDG version exists and a user creates the .mycroft/mycroft.conf it will need to merge the changes from that file into the the file in the XDG path. it's less things that can go wrong. If you feel like the merge is better I'm fine with that as well.

PureTryOut:
> If the XDG version exists and a user creates the .mycroft/mycroft.conf

I think we should in that case just ignore the new file in the old location, it's just the wrong location to put things. We'll move over existing files from the old location, but if someone puts a new file there and then complains it isn't being used we just point them to the new location. People will get used to it eventually.
If you want we could even leave a `$HOME/.mycroft/this-location-is-not-used-anymore.txt` file behind with the contents telling where it is now, but I don't personally see a need for it.

I rather not merge config files from an old and a new location, it just complicates things.

krisgesling:
We've been recommending the use of the config manager rather than directly editing files for a little while now, but a big challenge here is that `~/.mycroft/mycroft.conf` has been around for a very long time. So there are lots of references in our docs, blog posts and out in the community to it. Regardless of whether it's a correct or incorrect location, if people are following a tutorial / guide / documentation and it just quietly fails, that is a pretty frustrating experience. 

So my preference would be to support the existing location in some way but flagging it as deprecated so we can remove it entirely in a future release.

PureTryOut:
Those are valid points. I now made sure to check both old and new locations and merge them together.

I'm trying to notify that the old location is deprecated and will be removed in a future version, but I can't find the prints in any output anywhere. Where are `print` statements supposed to show up if at all? Is there some other way I should notify the user?

forslund:
They should, when using the start-mycroft.sh, be output to the relevant log file in /var/log/mycroft/*.log

However I think it maybe  better to use `LOG.warning` to print these kinds of messages.

mycroft/client/speech/hotword_factory.py
```diff
@@ -191,7 +192,17 @@ def __init__(self, key_phrase="hey mycroft", config=None, lang="en-us"):
...
+        old_path = join(expanduser('~'), './mycroft/mycroft.conf')
+        if isfile(old_path):
+            local_conf = old_path
+        else:
+            for dir in BaseDirectory.load_config_paths('mycroft'):
+                local_conf = LocalConf(join(dir, 'mycroft.conf'))
+                # If the current config contains the precise key use it,
+                # otherwise continue to the next file
+                if local_conf.get('precise', None) is not None:
+                    break
```
forslund, line 205: This may not work as intended if the precise conf is made in the xdg location but the old .mycroft location still exists. Should be quite a rare case though...

Also does the `BaseDirectory.load_config_paths()` function also include locations not writable by the user? in that case the `local_conf.store()` below may fail and throw an exception.

mycroft/client/enclosure/mark1/__init__.py
```diff
@@ -193,7 +198,8 @@ def process(self, data):
 
             LOG.info("Setting opt_in to: " + word)
             new_config = {'opt_in': enable}
-            user_config = LocalConf(USER_CONFIG)
+            user_config = LocalConf(os.path.join(
+                BaseDirectory.save_config_path('mycroft'), 'mycroft.conf'))
```
forslund, line 202: could the USER_CONFIG vaariable be used again now that it's defined to the as the save_config_path?

forslund (commented):
To me this mainly looks good now!

There's a couple of (possible) bugs I noted and a slight incompatibility. Not sure we should bother about the incompatibility though. It's small and should not be a critical issue. (like causing exceptions or so)

bin/mycroft-config
```diff
@@ -95,13 +95,13 @@ function validate_config_file() {
...
-_conf_file="~/.mycroft/mycroft.conf"
+_conf_file="~/.config/mycroft/mycroft.conf"
 function name_to_path() {
     case ${1} in
         "system")     _conf_file="/etc/mycroft/mycroft.conf" ;;
-        "user")       _conf_file=$(readlink -f ~/.mycroft/mycroft.conf) ;;
+        "user")       _conf_file=$(readlink -f ~/.config/mycroft/mycroft.conf) ;;
         "default")    _conf_file="$DIR/../mycroft/configuration/mycroft.conf" ;;
-        "remote")     _conf_file="/var/tmp/mycroft_web_cache.json" ;;
+        "remote")     _conf_file="~/.cache/mycroft/web_cache.json" ;;
```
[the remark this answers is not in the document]
PureTryOut: You're right, replaced for `$HOME`.

mycroft/client/speech/hotword_factory.py
PureTryOut: I guess we should give preference to the new XDG location, and only use the old location if no `precise` key is found yet?

`load_config_paths()` can indeed return unwriteable locations, I now made sure to always save to a writeable location but read from unwriteable ones too.

mycroft/skills/skill_updater.py
```diff
@@ -96,9 +97,9 @@ def installed_skills_file_path(self):
                     '.mycroft-skills'
                 )
             else:
-                self._installed_skills_file_path = os.path.expanduser(
-                    '~/.mycroft/.mycroft-skills'
-                )
+                self._installed_skills_file_path = os.path.join(
+                    BaseDirectory.save_data_path('mycroft',
+                                                 'skills'))
```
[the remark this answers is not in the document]
PureTryOut: https://github.com/MycroftAI/mycroft-skills-manager/pull/88

mycroft/client/speech/hotword_factory.py
forslund: Yeah make preference to the new XDG based location sounds like the best option

PureTryOut: Cool, that's what I did now.

That's all conversations resolved then :tada:

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2578))

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2578))

PureTryOut:
Darn pycodestyle messing me up :stuck_out_tongue: 
Should be good to go now!

forslund:
My `identity/identity2.json` doesn't seem to be moved as expected. I did some debugging and I think the declaration of `old_path` variable needs to be moved to after the `path` is overridden to point to the new location in the .config/mycroft folder.

Maybe the new path (xdg location) shouldn't shadow the input argument in that method?

PureTryOut:
You're right! :scream: 

I turned the order around, it's migrated properly now. Did you see anything else?

forslund:
Not during the short time I had yesterday to play with it. (I did have some wierdness but that's probably my own doing)

Gonna take some more time tonight to test some more.

mycroft/configuration/config.py
```diff
@@ -182,30 +183,71 @@ def get(configs=None, cache=True):
...
+                configs.append(RemoteConf())
+
+            # Then use the system config (/etc/mycroft/mycroft.conf)
+            configs.append(LocalConf(SYSTEM_CONFIG))
+
+            # Then use the config that comes with the package
+            configs.append(LocalConf(DEFAULT_CONFIG))
+
+            # Lastly use patched config
+            configs.append(LocalConf(Configuration.__patch))
```
forslund, line 246: Looked over the failing VK tests and [this](https://hudson.mycroft.team/blue/organizations/jenkins/mycroft-core/detail/PR-2578/29/pipeline#log-391) seems to be related to the code above.

`Configuration.__patch` is a dict and not a file so I think the line should be changed to ` configs.append(Configuration.__patch)`

It probably works when the `__patch` is not set since the path is checked before attempting to load (path evalutating to `False` will spawn an empty `LocalConf`)

PureTryOut: Made the change, let's see.

forslund: The Jenkins server seems to be down again :(

forslund:
Hum hum hum, the VK test seems to hang on the QA skills tests :/ will see if I can run them manually and see if the logs say anything extra

forslund:
I got really stuck on why the config patching didn't work in the VK tests, in the end it turned out to be quite simple. The `__patch` dict should have top priority and not the least priority.

I ran into a couple of other minor problems are still in the migration of files. Gonna add separate comments on each issue I saw.

When figuring this out I made some commits to show the needed changes https://github.com/forslund/mycroft-core/commits/PureTryOut-xdg-config

mycroft/configuration/config.py
```diff
@@ -182,30 +183,71 @@ def get(configs=None, cache=True):
...
+                configs.append(RemoteConf())
+
+            # Then use the system config (/etc/mycroft/mycroft.conf)
+            configs.append(LocalConf(SYSTEM_CONFIG))
+
+            # Then use the config that comes with the package
+            configs.append(LocalConf(DEFAULT_CONFIG))
+
+            # Lastly use patched config
+            configs.append(Configuration.__patch)
```
forslund, line 246: This needs to be moved to the top to be applied last.

mycroft/filesystem/__init__.py
```diff
@@ -30,7 +31,13 @@ def __init__(self, path):
...
         if not isinstance(path, str) or len(path) == 0:
             raise ValueError("path must be initialized as a non empty string")
-        path = join(expanduser('~'), '.mycroft', path)
+
+        # Migrate from the old location if it still exists
+        old_path = join(expanduser('~'), '.mycroft', path)
+        if isdir(old_path):
+            os.rename(old_path, path)
+
+        path = join(BaseDirectory.save_config_path('mycroft'), path)
```
forslund, line 40: This path needs to be moved to right after old_path is created...

mycroft/filesystem/__init__.py
```diff
@@ -30,7 +31,13 @@ def __init__(self, path):
     def __init_path(path):
         if not isinstance(path, str) or len(path) == 0:
             raise ValueError("path must be initialized as a non empty string")
-        path = join(expanduser('~'), '.mycroft', path)
+
+        # Migrate from the old location if it still exists
+        old_path = join(expanduser('~'), '.mycroft', path)
+        if isdir(old_path):
+            os.rename(old_path, path)
```
forslund, line 38: I changed this to `shutils.move`, but I think the real issue was that `path` was not pointing to the correct place (since it hasn't been concatenated with the XDG config dir). Not technically necessary, but shutils.move is more safe than os.rename since it works across mount points.

PureTryOut:
Made the changes, let's see! Honestly, VK is one big mystery to me.

forslund:
Just realized this will bork (and did bork) the VK test runner since it moves the identity file (and doesn't restore it)... Could you edit the [Jenkinsfile@L60](https://github.com/MycroftAI/mycroft-core/blob/dev/Jenkinsfile#L60) to be
`                        -v "$HOME/voight-kampff/identity:/root/.config/mycroft/identity" \`

to make sure the identity file isn't moved from the shared volume.

PureTryOut:
Hmm, still failing. Seems testing the skills times out or something?

forslund:
Right now the identity file is broken (I think) so many of the tests using it (weather, Q and A) are failing and it gets 10 seconds timeout of all of those.

We can retry when the identity file for VK is restored

krisgesling:
New identity file in place, triggered a new run :crossed_fingers:

PureTryOut:
Uh, seems it ignores the updated Jenkinsbuild file and keeps mounting the file to `/root/.mycroft/identity`...

https://hudson.mycroft.team/blue/organizations/jenkins/mycroft-core/detail/PR-2578/35/pipeline#step-39-log-1

forslund:
```
Connecting to https://api.github.com to check permissions of obtain list of PureTryOut for MycroftAI/mycroft-core
Loading trusted files from base branch dev at 81eae60b61071746d3f5837c1fe5b8b64ecd0e64 rather than db5a3eba75d86c000240fed66bf26bc30e98cb77
Obtained Jenkinsfile from 81eae60b61071746d3f5837c1fe5b8b64ecd0e64
‘Jenkinsfile’ has been modified in an untrusted revision
```

Seems like Jenkins doesn't trust you :(

I'll see if I can figure out how to  make it trust you / this change of the Jenkinsfile

mycroft/skills/event_scheduler.py
```diff
@@ -54,14 +55,19 @@ class EventScheduler(Thread):
...
         self.events = {}
         self.event_lock = Lock()
 
         self.bus = bus
         self.is_running = True
-        self.schedule_file = join(data_dir, schedule_file)
+        old_path = join(expanduser(Configuration.get()['data_dir']),
+                        schedule_file)
+        if isfile(old_path):
+            self.schedule_file = old_path
```
JarbasAl, line 67: can we `shutil.move` the file to the new directory automatically instead?

mycroft/client/text/text_client.py
```diff
@@ -171,6 +172,25 @@ def load_settings():
     global max_log_lines
     global show_meter
 
+    config_file = None
+
+    # Old location
+    path = os.path.join(os.path.expanduser("~"), ".mycroft_cli.conf")
+    if os.path.isfile(path):
+        config_file = path
```
JarbasAl, line 180: can we `shutil.move` the file to the xdg directory automatically instead?

mycroft/client/speech/hotword_factory.py
```diff
@@ -359,8 +384,23 @@ def found_wake_word(self, frame_data):
 class PorcupineHotWord(HotWordEngine):
     def __init__(self, key_phrase="hey mycroft", config=None, lang="en-us"):
         super(PorcupineHotWord, self).__init__(key_phrase, config, lang)
+
+        xdg_data_path = None
+
+        # Use the old path if available
+        old_path = join(expanduser('~'), '.mycroft', 'Porcupine')
+        if os.path.isdir(old_path):
+            xdg_data_path = old_path
```
JarbasAl, line 393: can we `shutil.move` the file to the xdg directory automatically instead?

mycroft/skills/event_scheduler.py
PureTryOut: Please read https://github.com/MycroftAI/mycroft-core/pull/2578#issuecomment-655919861 and following comments, we already discussed if we should just move everything over or keep old locations.

JarbasAl: just use the combo_lock and it should be safe to move?

JarbasAl: other possibility is to support both locations for a while, but copy the file to new location, and on load if new location exists, delete from old location, this way you do a 2 step move in an async way that is safe across processes

PureTryOut: @forslund

forslund: I think this should be safe to move since it's not really exposed to the user. This should only be used from the skills process anyways so combo-lock shouldn't be necessary

mycroft/client/text/text_client.py
forslund: I think as this is an user-exposed config file (similar to the config files) it shouldn't be moved automatically until the location has been deprecated for a while.

forslund:
If you make changes don't push to this branch since it'll destroy the jenkins identity file :/  ... Update in a branch and let me know. I'll pull them into the branch for the PR I created

mycroft/client/speech/hotword_factory.py
forslund: Also user installed so I don't think it should be moved. But will not be an issue when #2722 or the porcupine plugin is used.

mycroft/client/text/text_client.py
JarbasAl: add a deprecation warning like in config then?

PureTryOut:
Done, my branch `xdg-config-jarbas` contains the changes.

Too bad another review with more requested changes comes in so late :cry:

j1nx:
> Done, my branch `xdg-config-jarbas` contains the changes.
> 
> Too bad another review with more requested changes comes in so late 😢

Tried out that branch on OVOS. The following error pops up;

```
  File "/usr/bin/mycroft-skills", line 11, in <module>
    load_entry_point('mycroft-core==20.8.0', 'console_scripts', 'mycroft-skills')()
  File "/usr/lib/python3.8/site-packages/mycroft/skills/__main__.py", line 197, in main
    event_scheduler = EventScheduler(bus)
  File "/usr/lib/python3.8/site-packages/mycroft/skills/event_scheduler.py", line 70, in __init__
    self.schedule_file = new_path
NameError: name 'new_path' is not defined
```

PureTryOut:
Yup known problem @j1nx, fixed in https://github.com/MycroftAI/mycroft-core/pull/2794

j1nx:
Right! To manu different branches / tries because of this VK stuff. Will check/test the other one for you.

PureTryOut:
Closing this as it's being replaced by #2794 and it seems it causes confusion by being still open. It's the same changes, but with stuff to make VK work (which I don't have permissions for in this PR).

PR #2584, opened 2020-05-12
dev_setup.sh: add support for Alpine Linux

PureTryOut:
## Description
This adds Alpine Linux support to the `dev_setup.sh` script.

Also make the Arch check use `is_os`, as it already thought my Alpine installation was Arch Linux just because I had pacman installed. Distros often have package managers for other distros available in their repositories for bootstrapping, so you can't guess an OS just by their package managers installed.

## How to test
Run the script on Alpine Linux and see if everything builds and runs correctly

CLA [x]

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -294,7 +294,7 @@ function os_is() {
 }
 
 function os_is_like() {
-    grep "^ID_LIKE=" /etc/os-release | awk -F'=' '/^ID_LIKE/ {print $2}' | sed 's/\"//g' | grep -P -q '(^|\s)'"$1"'(\s|$)'
+    grep "^ID_LIKE=" /etc/os-release | awk -F'=' '/^ID_LIKE/ {print $2}' | sed 's/\"//g' | grep -q "\\b$1\\b"
 }
 
 function redhat_common_install() {
@@ -368,6 +368,10 @@ function redhat_install() {
 
 }
 
+function alpine_install() {
+    $SUDO apk add alpine-sdk git python3 py3-pip py3-setuptools py3-virtualenv mpg123 vorbis-tools pulseaudio-utils fann-dev automake autoconf libtool pcre2-dev pulseaudio-dev alsa-lib-dev swig python3-dev portaudio-dev libjpeg-turbo-dev
+}
+
 function install_deps() {
     echo 'Installing packages...'
     if found_exe zypper ; then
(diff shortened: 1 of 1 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2584))

dev_setup.sh
```diff
@@ -390,10 +394,14 @@ function install_deps() {
         # Fedora
         echo "$GREEN Installing packages for Fedora...$RESET"
         fedora_install
-    elif found_exe pacman; then
+    elif found_exe pacmani && os_is arch ; then
```
forslund, line 397: macmani?

forslund:
I'm getting a bunch of errors when running this in my alpine VM, logs here: http://termbin.com/ygs2

wonder if the `grep -P` can be replaced with a normal grep regexp, seems to only be used in `os_is_like()`

The error collecting packages I'm not sure about

forslund:
Am I missing something in my alpine setup to get those packages?

dev_setup.sh
PureTryOut: Ha, yes I am. Woops

PureTryOut:
Ah my system has the GNU coreutils installed, which I didn't think about. Seems to be a dep of the script. That makes it a tad annoying to fix, as we'd need to make the whole script POSIX compliant.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2584))

PureTryOut:
As for the packages it couldn't find, make sure you have the `community` repository enabled as they're all in that repository.

forslund:
I'll see if I can figure out what actually needs the -P flag...

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2584))

forslund:
Seems like the following is a working replacement for the os_is_like method:

```
function os_is_like() {
    grep "^ID_LIKE=" /etc/os-release | awk -F'=' '/^ID_LIKE/ {print $2}' | sed 's/\"//g' | grep -q "\\b$1\\b"
}
```

I needed to add `swig`, `python3-dev`, `portaudio-dev` and `libjpeg-dev` to the list of deps to get it to finish the dev_setup script (libjpeg-dev is technically only a Mark-1 dependency and can probably be removed in the future)

PureTryOut:
Thanks for that find! I updated the PR with your grep and package changes.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2584))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2584))

forslund:
Looks good, merging.

PR #2588, opened 2020-05-18
Bugfix/remove fallback handler

### LINKED ISSUE #2586: self.remove_fallback(handler) it does not work
i need the remove.fallback function after recording in my wakeword skill. I acted according to instructions https://mycroft-ai.gitbook.io/docs/skill-development/skill-types/fallback-skill. however, I always get the message. "mycroft.skills.fallback_skill:remove_fallback:139 | Could not remove fallback!". the fallback function is retained and is not deactivated. During the skill shutdown, the function is executed, the error message remains and fallback is deactivated.

forslund:
## Description
Allow removal of a fallback handler without knowing the wrapper. Fixes #2586 

- Additionally adds basic setup/teardown tests for the fallback skills.

## How to test
Check that the new tests pass and make sure fallback query can be reloaded correctly.

CLA [ Yes ]

### DIFF
--- a/mycroft/skills/fallback_skill.py
+++ b/mycroft/skills/fallback_skill.py
@@ -48,6 +48,7 @@ class FallbackSkill(MycroftSkill):
     utterance will not be see by any other Fallback handlers.
     """
     fallback_handlers = {}
+    wrapper_map = []  # Map containing (handler, wrapper) tuples
 
     def __init__(self, name=None, bus=None, use_settings=True):
         super().__init__(name, bus, use_settings)
@@ -98,18 +99,25 @@ class FallbackSkill(MycroftSkill):
         return handler
 
     @classmethod
-    def _register_fallback(cls, handler, priority):
+    def _register_fallback(cls, handler, wrapper, priority):
         """Register a function to be called as a general info fallback
         Fallback should receive message and return
         a boolean (True if succeeded or False if failed)
 
         Lower priority gets run first
         0 for high priority 100 for low priority
+
+        Arguments:
+            handler (callable): original handler, used as a reference when
+                                removing
+            wrapper (callable): wrapped version of handler
+            priority (int): fallback priority
         """
         while priority in cls.fallback_handlers:
             priority += 1
 
-        cls.fallback_handlers[priority] = handler
+        cls.fallback_handlers[priority] = wrapper
+        cls.wrapper_map.append((handler, wrapper))
 
     def register_fallback(self, handler, priority):
         """Register a fallback with the list of fallback handlers and with the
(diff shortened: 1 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2588))

test/unittests/skills/test_fallback_skill.py
```diff
@@ -0,0 +1,49 @@
...
+        self.assertEqual(len(FallbackSkill.fallback_handlers), 0)
+
+        # Create skill adding a single handler
+        fb_skill = setup_fallback(SimpleFallback)
+        self.assertEqual(len(FallbackSkill.fallback_handlers), 1)
+
+        fb_skill.remove_fallback(fb_skill.fallback_handler)
+        # Both internal trackers of handlers should be cleared now
+        self.assertEqual(len(FallbackSkill.fallback_handlers), 0)
+        self.assertEqual(len(FallbackSkill.wrapper_map), 0)
```
krisgesling, line 40: Nit pick as it's already tested in the lifecycle test, however I'm still wondering if we should check the length of the wrapper_map after adding the Fallback Skill in this test as well. 

I can't really see any way that it would happen, but if the length of the wrapper map never changed from 0 this would still pass.

test/unittests/skills/test_fallback_skill.py
```diff
@@ -0,0 +1,49 @@
...
+class TestFallbackSkill(TestCase):
+    def test_life_cycle(self):
+        """Test startup and shutdown of a fallback skill.
+
+        Ensure that an added handler is removed as part of default shutdown.
+        """
+        self.assertEqual(len(FallbackSkill.fallback_handlers), 0)
+        fb_skill = setup_fallback(SimpleFallback)
+        self.assertEqual(len(FallbackSkill.fallback_handlers), 1)
+        self.assertEqual(FallbackSkill.wrapper_map[0][0],
```
krisgesling, line 22: Keeping my nitpick hat on, do we also want to check the length of the object here? If somehow multiple tuples were assigned it could still pass.

mycroft/skills/fallback_skill.py
```diff
@@ -122,8 +130,16 @@ def wrapper(*args, **kwargs):
...
+        self.instance_fallback_handlers.append(handler)
+        self._register_fallback(handler, wrapper, priority)
+
+    @classmethod
+    def _remove_registered_handler(cls, wrapper_to_del):
+        for priority, handler in list(cls.fallback_handlers.items()):
+            if handler == wrapper_to_del:
+                del cls.fallback_handlers[priority]
+        else:
+            LOG.warning('No fallback matching {}'.format(wrapper_to_del))
```
krisgesling, line 142: I'm hitting this warning when I reload fallback-query  by touching `__init__.py`. 
Should the `handle_question` fallback be getting removed?

forslund: That doesn't sound correct, let me check...

test/unittests/skills/test_fallback_skill.py
forslund: It's probably a good idea to check, but perhaps it's enough to add that check in the lifecycle test so it's tested at exactly one place.

mycroft/skills/fallback_skill.py
forslund: Ok, found the mistake. Since the code above doesn't break to exit the for loop (to be able to handle multiple registered wrappers) the warning will always occur. I'll fix that.

test/unittests/skills/test_fallback_skill.py
forslund: I added this check in the lifecycle test

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2588))

PR #2598, opened 2020-05-27
Fix audio service startup without Internet

forslund:
## Description
If started up without Network connection the audio service would crash
importing the google_tts module as it tried to fetch language list from
google.

This restructures it to be a just-in-time check. It also adds a cached
list of supported languages to use if an error occurs fetching language
codes.

A small fix allowing "en-US" to be matched to "en-us" instead of the
generic "en" is also included.

## How to test
Insert a `raise Exception` at [L74](https://github.com/MycroftAI/mycroft-core/compare/dev...forslund:bugfix/audio-startup?expand=1#diff-d536e19534c7002f55e0187c8e33b9feR74) and make sure that the audio service process can start.

CLA [ Yes ]

### DIFF
--- a/mycroft/tts/google_tts.py
+++ b/mycroft/tts/google_tts.py
@@ -17,17 +17,83 @@ from gtts.lang import tts_langs
 
 from .tts import TTS, TTSValidator
 
-supported_langs = tts_langs()
+from mycroft.util.log import LOG
+
+# Live list of languages
+# Cached list of supported languages (2020-05-27)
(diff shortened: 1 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2598))

krisgesling:
Looks good and tested on Linux and Picroft. Merging.

PR #2616, opened 2020-06-14
self.gui.show_animated_image

JarbasAl:
Adds some missing stuff from SkillGui docstrings

adds a new method and qml template to show animated images (gif files)

PR #2617, opened 2020-06-15
fix send_event and revert change from pull/2616 

AIIX:
## Description
Reverts a change in https://github.com/MycroftAI/mycroft-core/pull/2616 that causes the GUI to crash as None type is unexpected, when events send params as objects

## How to test
self.send_event("someskill.event", {"clicked": True}


forslund:
The `params` are changed from None to empty dict `{}` at [L122](https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/enclosure/gui.py#L122)

If the default argument is {} modifications params may not be cleared

PR #2618, opened 2020-06-16
Add VK test - specific skill should not reply

krisgesling:
## Description
Adds a test tool opposite to then_wait. Instead of the test succeeding when the criteria is met, the test fails if the given criteria is detected.

This is then used to test if a specific Skill responds to an utterance.

## Improvements needed
It feels like there is a better way to do this, but it's working for me right now, so thought I'd put it out in the world for critique. Suggestions welcome. 

Would also prefer to utilize MSM to search for the name of the skill in the test which would add more flexibility to what test authors enter. Currently it requires that you enter the class name of the Skill.

## How to test
With the following feature file for the News Skill:
```
Feature: mycroft-news

  Scenario Outline: play music with names similar to news channels
    Given an english speaking user
     When the user says "<play some music>"
     Then "NewsSkill" should not reply

    Examples:
      | play some music |
      | what time is it |
      | what's the weather |
      | play the bbc news |
      | what's news |
```

First two should pass, last two should fail.

### DIFF
--- a/test/integrationtests/voight_kampff/tools.py
+++ b/test/integrationtests/voight_kampff/tools.py
@@ -51,6 +51,23 @@ def then_wait(msg_type, criteria_func, context, timeout=TIMEOUT):
     return False, debug
 
 
+def then_wait_fail(msg_type, criteria_func, context, timeout=TIMEOUT):
+    """Wait for a specified time, failing if criteria is fulfilled.
+
+    Arguments:
+        msg_type: message type to watch
+        criteria_func: Function to determine if a message fulfilling the
+                       test case has been found.
+        context: behave context
+        timeout: Time allowance for a message fulfilling the criteria
+
+    Returns:
+        tuple (bool, str) test status and debug output
+    """
+    status, debug = then_wait(msg_type, criteria_func, context, timeout)
+    return (not status, debug)
+
+
 def mycroft_responses(context):
     """Collect and format mycroft responses from context.
 
--- a/test/integrationtests/voight_kampff/__init__.py
+++ b/test/integrationtests/voight_kampff/__init__.py
@@ -14,4 +14,5 @@
 #
 
 from .tools import (emit_utterance, wait_for_dialog, then_wait,
-                    mycroft_responses, print_mycroft_responses)
+                    then_wait_fail, mycroft_responses,
+                    print_mycroft_responses)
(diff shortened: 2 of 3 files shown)


test/integrationtests/voight_kampff/tools.py
```diff
@@ -51,6 +51,35 @@ def then_wait(msg_type, criteria_func, context, timeout=TIMEOUT):
...
+        for message in context.bus.get_messages(msg_type):
+            status, test_dbg = criteria_func(message)
+            if status:
+                debug += test_dbg
+                context.matched_message = message
+                context.bus.remove_message(message)
+                return False, debug
+        context.bus.new_message_available.wait(0.5)
+    # Timed out return debug from test
+    return True, debug
```
forslund, line 80: The `test_failed` variable doesn't sem to be used.

The answer to this is probably "No" but I'll ask anyway,

could this function be

```python
    def then_wait_fail(msg_type, criteria_func, context, timeout=TIMEOUT):
    """Wait for a specified time failing if criteria is fulfilled.
    [...]
    """
    res = then_wait(msg_type, criteria_func, context, timeout)
    return (not res[0], res[1])
```

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2618))

forslund:
Regarding using msm to lookup the skill as you suggest

The context contains msm so a check could be done if `self.skill_id` is added to `meta`[here](https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/skills/mycroft_skill/mycroft_skill.py#L1111)

```python
    skill = context.msm.local_skills.get(skill_name) or context.msm.find(skill_name)
    if basename(skill.path) == skill_id_from_speak_message:
        do_stuff()
```

Or something similar perhaps?

krisgesling:
Thanks Ake, negating the existing function does work. I went through a lot of variations trying to use the existing functions rather than adding another. By the time I resigned myself to adding a new function to tools.py I'd lost sight of the easy answer!

Agree that we should add this, and I'll revisit the Skill searching and adding skill_id to the meta dict later.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2618))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2618))

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2618))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2618))

PR #2621, opened 2020-06-19
prevent audio playing between mic record and utt handling

krisgesling:
## Description
This creates a separate `restore_volume` handler for the end of mic recording. Previously the same handler was used for speech and recording.

When the recording finishes, there should either be an utterance to be handled, or a `speech.recognition.unknown` message meaning the transcription failed because the user didn't say anything (or it was a false activation).

If the latter and no utterance was detected, then we immediately restore the volume. Otherwise we do not restore the volume, and the utterance is passed onto the intent handlers. Once Mycroft responds to that utterance, the `_restore_volume` handler attached to the `audio_output_end` message is triggered.

This means that if audio is playing and:
- a user says "Hey Mycroft, stop" 
  - then no more audio plays
- a user asks a question such as "Hey Mycroft, what's 7 times 3" 
  - then audio is paused until the response is complete. 
    Once the response has finished, audio resumes.

## Need to confirm
Are there other paths that I'm not considering?
Is there a chance that neither a failed transcription, nor Mycroft replying would occur and therefore leave the audio in an undesired state?

Ensure that no snippets of news audio are played during interactions. Should still play between questions.

### DIFF
--- a/mycroft/audio/audioservice.py
+++ b/mycroft/audio/audioservice.py
@@ -298,22 +298,41 @@ class AudioService:
     def _restore_volume_after_record(self, message=None):
         """
             Restores the volume when Mycroft is done recording.
-            If no utterance detected, restores immediately.
-            Else keeps volume low until Mycroft responds to the utterance.
+            If no utterance detected, restore immediately.
+            If no response is made in reasonable time, then also restore.
 
             Args:
                 message: message bus message, not used but required
         """
         def __restore_volume():
             LOG.debug('restoring volume')
-            self.volume_is_low = False
-            if not self.volume_is_low:
-                self.current.restore_volume()
-            self.bus.remove('recognizer_loop:speech.recognition.unknown',
-                            __restore_volume)
+            self.current.restore_volume()
 
-        if self.current:
-            self.bus.on('recognizer_loop:speech.recognition.unknown',
+        def wait_for_speak(bus, timeout=8):
+            """Wait for a speak Message on the bus.
+
+            Arguments:
+                bus (Mycroft MessageBus): Bus instance to listen on
+                timeout (int): how long to wait, defaults to 8 sec
+            """
+            self.speak_msg_detected = False
+
+            def detected_speak(message=None):
+                self.speak_msg_detected = True
+            self.bus.on('speak', detected_speak)
+            time.sleep(timeout)
+            self.bus.remove('speak', detected_speak)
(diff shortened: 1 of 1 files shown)


devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2621))

forslund:
Would there be an issue with skill handlers that don't produce a spoken sentence, for example just using the skill.acknowledge()?

krisgesling:
I've been trying to think of the best way to know that "a Skill has finished". It's also possible that Skills do something and don't respond at all, even an acknowledge tone. Certainly not recommended but still very possible.

So perhaps we can re-use the VK then_wait method with a timeout of ~4 seconds (?), and if no speak command was detected then we restore the volume anyway. This should give most Skills plenty of time to do something. That way even if an expected action isn't taken, it falls back to restoring audio playback.

forslund:
Sounds like a good solution. An alternative is to add something to the [on_end()](https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/skills/mycroft_skill/mycroft_skill.py#L846) skill handler callback, but the timeout should probably be there first step since it'll be pretty safe.

krisgesling:
Have now modified so that the volume will restore after 8 seconds regardless of what happens. So if a Skill consumes an intent and does absolutely nothing or just plays an acknowledgement beep, the playing audio will be restored.

Had to push it to a higher 8 seconds as unknown intents take longer to respond. This should be the minority of cases so I don't see it as a problem that the audio stops playing for slightly longer.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2621))

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2621))

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2621))

krisgesling:
Great feedback, thanks Ake :)

Have changed each of them.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2621))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2621))

Issue #2639, opened 2020-07-21
Ensure services are fully ready before reporting readiness


krisgesling:
This is partially a rubber duck issue to document my own thoughts, but also interested in broader input on the best way to resolve.

Currently on boot the SkillsManager loads all Skills and emits a `mycroft.skills.initialized` message
On detecting this message PadatiousService starts training and if it's the initial train emits `mycroft.ready` once it's complete.

`mycroft.ready` is intended to signify that the device is ready to use, however there are some circumstances in which that is not entirely the case.

### 1. Initial training was for priority_skills only
In this instance only the Volume and Pairing Skill are loaded when Padatious runs it's training. Once the initial training is completed, it declares Mycroft ready and the "Mycroft is now ready for use" dialog is spoken by the device. 
At this point, asking Mycroft anything not handled by those Skills responds with "Please wait a moment as I finish booting up" because the Unknown Fallback doesn't yet exist. 

**Expectation:**
All default Skills for that platform should be loaded and trained prior to the `mycroft.ready` message being emitted.

### 2. Significant Skill updates required
If a device is reasonably outdated or there has been significant updates to a number of Mycroft Skills, it can take time for new dependencies to install and updates to be applied. 

In this case, the old versions of the Skills are loaded and trained. Mycroft then reports itself ready. Following this, all the Skills start updating. Utterances received during this time may hit the old Skill, or the new Skill prior to intents being trained.

**Question:**
Do we want to run an update of all Skills immediately after boot, before declaring Mycroft is ready?
This slows the boot time, but improves the stability and predictability of responses.

### 3. Precise not downloaded
The Precise executable is a fairly big download and can take some time. This is now shipped as part of the Picroft image but is not part of Mycroft-core and we can't assume it is there. 

**Question:**
Should we be waiting for all assets and services to be ready before declaring Mycroft as a whole ready?
Eg each service [bus, skills, audio, voice, enclosure] must itself report being ready, before the global `mycroft.ready` message is sent.
Should we be checking that all services report being ready, before declaring that Mycroft as a whole is ready? Eg through the service hooks proposed in #2601

forslund:
Good write-up, I've below posted some thoughts and some history on these issues, but the main conclusion is that the proposals above are good ideas according to me.

1, is this when no skills are installed or is it always the case? If skills are installed but not loaded before the initialized message is sent there's some bug in the load_new_skills() method.

[...]

PR #2642, opened 2020-07-24
Feature - check if services ready on boot

krisgesling:
**DO NOT MERGE**

## Description
Some example code following on from issue #2639 

This shifts the `mycroft.ready` message from the Padatious Service to the Enclosure.

It sends a request to each service over the message bus to respond if it's alive. As such it assumes that both the bus and the enclosure services are alive as they're required for successfully transmission of the messages.

In it's current state it will always timeout as there's no handler added to the Skills service. I'm sure there's also a better way to handle checking the speech service, and general refactoring is needed.

## How to test
Boot Mycroft - should startup, train all intents, then check each service for 2 mins and timeout.

Remove skills service from list to check and it should emit `mycroft.ready` as normal.

## Questions
* Is this a useful approach? 
* Does it matter that it relies on the enclosure running?
* Presuming we have watchdogs running on each service, what do we do if the boot completely times out? Restart the device?

### DIFF
--- a/mycroft/client/speech/__main__.py
+++ b/mycroft/client/speech/__main__.py
@@ -171,18 +171,19 @@ def handle_open():
     EnclosureAPI(bus).reset()
 
 
-def main():
-    global bus
-    global loop
-    global config
-    reset_sigint_handler()
-    PIDLock("voice")
-    bus = MessageBusClient()  # Mycroft messagebus, see mycroft.messagebus
-    Configuration.set_config_update_handlers(bus)
-    config = Configuration.get()
+def on_ready():
+    LOG.info('Speech client is ready.')
 
-    # Register handlers on internal RecognizerLoop bus
-    loop = RecognizerLoop()
+
+def on_stopping():
+    LOG.info('Speech service is shutting down...')
+
+
+def on_error(e='Unknown'):
+    LOG.error('Audio service failed to launch ({}).'.format(repr(e)))
+
+
+def connect_loop_events(loop):
     loop.on('recognizer_loop:utterance', handle_utterance)
     loop.on('recognizer_loop:speech.recognition.unknown', handle_unknown)
     loop.on('speak', handle_speak)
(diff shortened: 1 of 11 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2642))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2642))

mycroft/client/speech/__main__.py
```diff
@@ -171,18 +171,29 @@ def handle_open():
...
+def is_alive(message=None):
+    """Respond to is_alive status request."""
+    status = alive_status.is_set()
+    if message:
+        bus.emit(message.response({'status': status}))
+    return status
+
+
+def on_ready():
+    alive_status.set()
```
forslund, line 183: Since these can be overridden the setting in the alive_status here may not be the best way to do this.

krisgesling: ah good point, maybe I'll shift them to where the hooks are called from.

Does the overall approach make sense though? Should we be moving straight to an external monitoring service or something else?

forslund:
I wonder if a separate class for tracking the status and handling the reporting and callbacks would be a good idea. Something along the lines of: https://github.com/forslund/mycroft-core/commit/da941672b3d5dbdd64ee5de8dd4b9a8adc61aa77

Each process creates an object based on the class and it handles callbacks and reporting readiness over messagebus

mycroft/client/speech/__main__.py
forslund: It sort of makes sense with the other things the enclosure does. I don't have a better suggestion...

forslund:
It should improve things but there's still the case when default skills aren't installed that needs to be handled. But that's probably a separate issue

j1nx:
Ok, then I think it is best that I create an seperate issue at that repo referencing to this one.

krisgesling:
Closing in favour of #2645 

Created in a new PR to maintain discussion and it's context here.

PR #2644, opened 2020-07-29
Always install default skills before initial load

forslund:
## Description
Following the discussion in #2639, this adds a check for the default skills before considered being initialized, if a default skill is missing the a skill update will occur before the mycroft.skills.initialized message is sent to trigger Padatious training. The check seem to be quite fast due to msm caching and will not cause the skills startup time to be delayed unless a skill is really missing.

This checks if all default skills are installed before continuing with
initial load.

For devices with no skills installed on first start this should ensure
that skills are installed before sending the mycroft.skills.initialized
to start precise training and generating the mycroft.ready message

## How to test
Remove a default skill and check that the skill is installed during startup.

CLA [ Yes ]

### DIFF
--- a/mycroft/skills/skill_updater.py
+++ b/mycroft/skills/skill_updater.py
@@ -219,6 +219,15 @@ class SkillUpdater:
                 raise
         self.installed_skills.add(skill.name)
 
+    def defaults_installed(self):
+        """Check if all default skills are installed.
+
+        Returns:
+            True if all default skills are installed, else False.
+        """
+        defaults = self.msm.list_all_defaults()[self.msm.platform]
+        return all([skill.is_local for skill in defaults])
+
     def _get_device_skill_state(self, skill_name):
         """Get skill data structure from name."""
         device_skill_state = {}
--- a/mycroft/skills/skill_updater.py
+++ b/mycroft/skills/skill_updater.py
@@ -31,6 +31,11 @@ ONE_HOUR = 3600
 FIVE_MINUTES = 300  # number of seconds in a minute
 
 
+def skill_is_blacklisted(skill):
+    blacklist = Configuration.get()['skills']['blacklisted_skills']
+    return os.path.basename(skill.path) in blacklist or skill.name in blacklist
+
+
 class SkillUpdater:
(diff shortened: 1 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2644))

j1nx:
Currently in the middle of rebuilds, but hopefully tomorrow I can and will test this in combination with the pairing skill and a fresh firstboot.

mycroft/skills/skill_updater.py
```diff
@@ -219,6 +219,15 @@ def install_or_update(self, skill):
...
         self.installed_skills.add(skill.name)
 
+    def defaults_installed(self):
+        """Check if all default skills are installed.
+
+        Returns:
+            True if all default skills are installed, else False.
+        """
+        defaults = self.msm.list_all_defaults()[self.msm.platform]
+        return all([skill.is_local for skill in defaults])
```
krisgesling, line 229: Do we need to account for people blacklisting default Skills?

forslund: Hmm, blacklisting skills still installs them, it just doesn't load them afaik. It's an update we can do but we should probably change the update method as well.

forslund: I've pushed an update to exclude blacklisted skills from the default check. I've also looked at excluding blacklisted skills from the install/update procedure which is fairly easy to add, if we think that should be added as well?

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2644))

mycroft/skills/skill_updater.py
krisgesling: Working great. 

Yeah I think that makes sense too. If someone has blacklisted a Skill we don't need to install it. Though if there is some security update, perhaps we do want to update? 

I see that update and install are handled separately so we could do updates but not installs? Really though, this is a nice to have so I'm going to merge this PR as is and we can do the above in the future.

PR #77, opened 2016-06-01
Issues 76 - Fix changes to wikipedia skill so that tests pass

ethanaward:
I changed the wikipedia skill so that it triggers on 'what does wikipedia say about'. I also removed 'who is' and 'who was' in the vocab, as those will be handled by Wolfram.

PR #2650, opened 2020-08-07
set active lang to configured if missing from STT and normalize

forslund:
## Description
I ran across this when checking some questions I had surrounding [mycroft-timer PR #77](https://github.com/MycroftAI/mycroft-timer/pull/77)

If the STT doesn't send a lang code fallback to configured language instead of hardcoded en-us. This should at least be more correct than the previous behaviour.

## How to test
Check that Mycroft operates as usual and that the new unittest passes

CLA [ Yes ]

### DIFF
--- a/mycroft/skills/intent_service.py
+++ b/mycroft/skills/intent_service.py
@@ -149,6 +149,19 @@ class ContextManager:
         return result
 
 
+def _get_message_lang(message):
+    """Get the language from the message or the default language.
+
+    Arguments:
+        message: message to check for language code.
+
+    Returns:
+        The languge code from the message or the default language.
+    """
+    default_lang = Configuration.get().get('lang', 'en-us')
+    return message.data.get('lang', default_lang)
+
+
 class IntentService:
     def __init__(self, bus):
         self.config = Configuration.get().get('context', {})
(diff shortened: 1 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2650))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2650))

krisgesling:
Nice!

Is there going to be a benefit in re-using this elsewhere? ie should this be an imported util?


In terms of testing, is it worth us adding a test for a globally configured lang code?
Something like:
```
    def test_config_lang(self):
        msg = Message('test msg', data={})
        config = {'lang': 'it-it'}
        Configuration.patch(Message('patch', data={'config': config}))
        self.assertEqual(_get_message_lang(msg), 'it-it')
        Configuration.patch_clear(msg)
```

Sidenote - It seems like `patch_clear` doesn't really need any arguments, the Message arg looks like it's just been copy-pasted from the `patch` method. Will do a PR to propose removing that.

krisgesling:
disregard the `patch_clear` sidenote, just saw that it's got a messagebus handler so the Message arg makes total sense.

forslund:
I've added an explicit testcase for the case when the config doesn't contain a language code and changed it so the configured language code in the test is 'it-it' which should indeed be better than using the 'en-us' one which would be invisible.

I don't _think_ it's of much use outside of the intentservice, the function could technically be placed in `mycroft.util.lang`, but since this is marked as "private" with the underscore, placing it here for now should be fine. If another use is found it can be added there and this can be removed.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2650))

PR #2657, opened 2020-08-12
Allow override for animations in platform skills

AIIX:
## Description
Allows setting an override for animations like talking, thinking, listening in platform skills like skill-mark-2 for interactive GUI pages, This helps skills developers decide weather they want the general animations disabled for a page, for example when expecting response in a speech dialog without moving focus away from the content displayed on the page. 

The option by default is set to always show all animations.

## How to test
self.gui.show_page("example.qml", override_animations=True)


JarbasAl:
:+1: i like this

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2657))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2657))

forslund:
Sounds like a good idea.

I think the original idea was that the get_response() start of listening shouldn't generate a listening -> thinking -> speaking cycle if a skill page is shown so this should maybe be default behaviour.

I'm away from my gui setup at the moment but I will test this together with your Mark-2 PR on Monday and see it all in action.

krisgesling (approved):

PR #2662, opened 2020-08-14
Do not call max if files list is empty

krisgesling:
## Description
Method that checks modified date of files attempts to call `max()` even if the list of files is empty. Results in:
```
ValueError: max() arg is an empty sequence
```

This change checks if the list is populated first.

## How to test
Run dev branch on a Mark 1, after booting uninstall a Skill.

### DIFF
--- a/mycroft/skills/skill_loader.py
+++ b/mycroft/skills/skill_loader.py
@@ -99,7 +99,10 @@ def _get_last_modified_time(path):
                 all_files.append(os.path.join(root_dir, f))
 
     # check files of interest in the skill root directory
-    return max(os.path.getmtime(f) for f in all_files)
+    if all_files:
+        return max(os.path.getmtime(f) for f in all_files)
+    else:
+        return 0
 
 
 class SkillLoader:


mycroft/skills/skill_loader.py
```diff
@@ -99,7 +99,8 @@ def _get_last_modified_time(path):
                 all_files.append(os.path.join(root_dir, f))
 
     # check files of interest in the skill root directory
-    return max(os.path.getmtime(f) for f in all_files)
+    if all_files:
+        return max(os.path.getmtime(f) for f in all_files)
```
forslund, line 103: Should there be an 
```python
else
    return 0
```

I'd think a None would result in error when comparing with last_loaded

krisgesling: Good point - changed

... and this is why I'm not doing the core release right now!

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2662))

mycroft/skills/skill_loader.py
forslund: Also it being 21:42 Friday evening I hope. Looks good now.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2662))

PR #2663, opened 2020-08-16
Voight Kampff given timeout

forslund:
## Description
Following some discussions with @krisgesling.

This adds a couple of given statements for setting the timeout for each of the steps in the scenario. After scenario completion the step timeout is returned to to original 10 seconds. Hopefully we won't be needed to go above 10 seconds in too many cases, but the option can be good to have for computation heavy tasks or tasks relying on an external data source.

## How to test
Create a simple test for a skill and add a 25 second delay (to not trigger within the retry) before the speak command. ensure that this test fails then add a `And a 30 second timeout` after the given step.

CLA [ Yes ]

### DIFF
--- a/test/integrationtests/voight_kampff/features/environment.py
+++ b/test/integrationtests/voight_kampff/features/environment.py
@@ -97,6 +97,7 @@ def before_all(context):
     sleep(10)
 
     context.bus = bus
+    context.step_timeout = 10  # Reset the step_timeout to 10 seconds
     context.matched_message = None
     context.log = log
     context.original_config = {}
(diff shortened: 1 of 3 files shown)


devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2663))

chrisveilleux (approved):
Looks good.  Was surprised to not see any scenarios using the new givens.  I guess that comes later...

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2663))

PR #2668, opened 2020-08-18
remove hardcoded timeouts - now in mycroft.conf

krisgesling:
## Description
Removes hardcoded recording timeout values that were moved to the default `mycroft.conf`.

They were left for backwards compatibility. Removing in preparation for 20.08

## How to test
- [ ] with default config say something short (like "time") and listener should timeout 3 seconds after your speech.
- [ ] with default config talk non-stop and listener should timeout after 10 seconds.
- [ ] set `listener.recording_timeout` parameter to 2.0 and ensure no utterance can exceed 2 seconds.
- [ ] set `listener.recording_timeout_with_silence` parameter to 10.0 and listener will record a full 10 seconds of silence.

### DIFF
--- a/mycroft/client/speech/mic.py
+++ b/mycroft/client/speech/mic.py
@@ -323,16 +323,6 @@ class ResponsiveRecognizer(speech_recognition.Recognizer):
     # before a phrase will be considered complete
     MIN_SILENCE_AT_END = 0.25
 
-    # TODO: Remove in 20.08
-    # The maximum seconds a phrase can be recorded,
-    # provided there is noise the entire time
-    RECORDING_TIMEOUT = 10.0
-
-    # TODO: Remove in 20.08
-    # The maximum time it will continue to record silence
-    # when not enough noise has been detected
-    RECORDING_TIMEOUT_WITH_SILENCE = 3.0
-
     # Time between pocketsphinx checks for the wake word
     SEC_BETWEEN_WW_CHECKS = 0.2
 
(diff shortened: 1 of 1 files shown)


mycroft/client/speech/mic.py
```diff
@@ -374,14 +364,12 @@ def __init__(self, wake_word_recognizer, watchdog=None):
 
         # The maximum seconds a phrase can be recorded,
         # provided there is noise the entire time
-        self.recording_timeout = listener_config.get('recording_timeout',
-                                                     self.RECORDING_TIMEOUT)
+        self.recording_timeout = listener_config.get('recording_timeout')
```
forslund, line 367: Maybe use brackets here to generate a lookup error if it's missing from config. This will default to `None` and cause issues when used.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2668))

mycroft/client/speech/mic.py
krisgesling: should i just include the previous hardcoded values as a sane backup default?

forslund: This is a case where I'm not quite sure what I think, if these are completely missing the default config is either missing or messed up. Sane defaults are generally a good thing though and the user would get a warning that the config failed to load if that were the case. So you're probably right.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2668))

forslund (approved):
Looks all good.

PR #2679, opened 2020-08-21
Refactor/remove imp

forslund:
## Description
Remove the remaining uses of the old `imp module`

- Audio service loading
- Loading test environment setup code in the old skills integration tests

## How to test
Check that audio services load and that the test runner can run for example the spotify skill (which uses a test environment)

CLA [ Yes ]

### DIFF
--- a/mycroft/audio/audioservice.py
+++ b/mycroft/audio/audioservice.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import imp
+import importlib
 import sys
 import time
 from os import listdir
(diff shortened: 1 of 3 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2679))

krisgesling:
Looks perfect, and working as expected though when I ran it against spotify as you suggested, it failed to import `mock`

It looks like we only use `unittest.mock` in core so it's not available by default and not a requirement in the Skill.

krisgesling:
mock included in docker image and will be updated in Spotify Skill in next update
Merging.

PR #2682, opened 2020-08-27
Feat/self.gui.connected

JarbasAl:
Allows skill to check if some gui is connected

goes well together with #2660 

see https://github.com/JarbasSkills/skill-dagon/blob/cps_media/__init__.py#L96 or https://github.com/JarbasSkills/skill-euronews/blob/feat/cps_media/__init__.py#L119 for usage

### DIFF
--- a/mycroft/enclosure/gui.py
+++ b/mycroft/enclosure/gui.py
@@ -41,6 +41,7 @@ class SkillGUI:
 
     @property
     def connected(self):
+        """Returns True if at least 1 gui is connected, else False"""
         if self.skill.bus:
             reply = self.skill.bus.wait_for_response(
                 Message("gui.status.request"), "gui.status.request.response")
--- a/mycroft/enclosure/gui.py
+++ b/mycroft/enclosure/gui.py
@@ -42,10 +42,10 @@ class SkillGUI:
     @property
     def connected(self):
         if self.skill.bus:
-            data = self.skill.bus.wait_for_response(
+            reply = self.skill.bus.wait_for_response(
                 Message("gui.status.request"), "gui.status.request.response")
-            if data:
-                return data["connected"]
+            if reply:
+                return reply.data["connected"]
         return False
 
     @property
--- a/mycroft/enclosure/gui.py
+++ b/mycroft/enclosure/gui.py
@@ -43,7 +43,7 @@ class SkillGUI:
     def connected(self):
         if self.skill.bus:
             data = self.skill.bus.wait_for_response(
-                Message("gui.status.request"), "gui.status.reply")
+                Message("gui.status.request"), "gui.status.request.response")
             if data:
                 return data["connected"]
         return False
--- a/mycroft/enclosure/gui.py
+++ b/mycroft/enclosure/gui.py
@@ -40,6 +40,15 @@ class SkillGUI:
         self.config = Configuration.get()
 
     @property
+    def connected(self):
+        if self.skill.bus:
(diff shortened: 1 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2682))

forslund:
Looks nice and simple! Will test it properly tonight or tomorrow

Two things:
- Docstrings are missing
- is there a reason you're using a custom response message type and not `message.response(...)` to generate the default response message with the default type?

penrods:
Nice!  Are you guys still trying to maintain anything like this list of messages that traverse the messagebus?
https://docs.google.com/spreadsheets/d/1RsBpMOGx4o8A_zF_UELYQSLK8DlIScHGnfzJ3iSN0AE/edit#gid=0

I never loved using a Google Sheet for this (outside of the repo), but having something like this seems useful.

@krisgesling @chrisveilleux

krisgesling:
Hey @penrods I have moved that list of message types to the public documentation and trying to keep it updated there instead:
https://mycroft-ai.gitbook.io/docs/mycroft-technologies/mycroft-core/message-types

It's a good point that we need to add in the `gui.*` messages though

JarbasAl:
>     * is there a reason you're using a custom response message type and not `message.response(...)` to generate the default response message with the default type? (assuming it has something to do with using reply...)

because of message context, needs to use .reply for proper targeting/hivemind routing, but i would argue that message.response should be refactored to use this internally ```self.reply(xxxx)``` instead of ```response_message = Message(xxxx)``.

forslund:
It can't always do that, since it is sometimes used within a service (I did change it briefly when merging your updates for the reply but it broke things severely when chained messages were emitted). I do think we should have an optional reply flag to the response however.

For now can you do change the response message type to match the default response? `gui.status.request.response`?

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2682))

mycroft/enclosure/gui.py
```diff
@@ -39,6 +39,15 @@ def __init__(self, skill):
         self.on_gui_changed_callback = None
         self.config = Configuration.get()
 
+    @property
+    def connected(self):
+        if self.skill.bus:
+            data = self.skill.bus.wait_for_response(
```
krisgesling, line 45: I may have been awake for too long, but would this return a Message that has a data object rather than the data object directly?

krisgesling:
Should also have said, beyond the little question above, I think this is great. Definitely going to be needed very soon.

mycroft/enclosure/gui.py
JarbasAl: good catch, you are correct

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2682))

JarbasAl:
> Should also have said, beyond the little question above, I think this is great. Definitely going to be needed very soon.

i already need it xD

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2682))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2682))

krisgesling (approved):
Working great :+1:

Issue #2658, opened 2020-08-12
extend common play - audiobooks / podcasts / video / games


JarbasAl:
We should add support for media types under common play, it currently assumes "play X" is always about music, but it could be video or audiobooks, at very least the trigger phrase needs to be expanded to handle other use cases, which can also be used for media type detection (but wont be enough)

Apologies for the messy issue, instead of rewriting all the info bellow i thought it would be easier to screenshot the chat

![Image Pasted at 2020-8-12 21-38](https://user-images.githubusercontent.com/33701864/90065934-e9b79c00-dce4-11ea-984e-d9e32362e4c7.png)
![Image Pasted at 2020-8-12 21-328](https://user-images.githubusercontent.com/33701864/90065948-ede3b980-dce4-11ea-8b83-9b552a0ffe59.png)

![2020-08-12_21-56](https://user-images.githubusercontent.com/33701864/90067247-ba099380-dce6-11ea-85d0-f886c27d14a4.png)

![2020-08-12_21-54](https://user-images.githubusercontent.com/33701864/90067089-83337d80-dce6-11ea-861d-fd5c7e9f8fdb.png)
![2020-08-12_21-55](https://user-images.githubusercontent.com/33701864/90067095-862e6e00-dce6-11ea-8ae0-20b95eed3fb7.png)

![image](https://user-images.githubusercontent.com/33701864/90067330-dc9bac80-dce6-11ea-9877-f6152d85bb45.png)

krisgesling:
Hey Jarbas, good discussion already, thanks for getting it going.

The other big area that I see coming up (particularly with the GUI) are "playing" games. They can use Padatious intents to closely match their game name but is it worth allowing them to respond to the `CPS_match_query_phrase` with a media type of `game`?

AIIX:
Hey so I got down to working on adding a video type to playback skill that can support "Video" media type GUI side, It isn't exactly clear as to how the video type is supposed to be handled as currently at least on the "Audio"' type Mycroft Internally manages playback but for GUI video side, video playback will be managed by an component in the user interface so I believe it would be nice to discuss the structure of the data CPS will provide GUI and GUI needs to send back to playback skill

PR #2619, opened 2020-06-17
Add method for updating playback information from skills

forslund:
## Description
This adds the method used by the [npr-news-skill](https://github.com/mycroftai/skill-npr-news), [spotify-skill](https://github.com/forslund/spotify-skill) to communicate their info to the playback control showing the default playback UI.

The method has the common parameters specified but is setup to be able to add new functionallity without core changes.

Points of discussion:
- Is this the way we want to go or should we re-think it somewhat?
- There are a couple of extra fields that might be added such as *track length* and *current position* to be able to show a progress bar or similar.
- The status should be implemented to be able to say that playback has stopped clearing the information. Should this be a string, enum or just a bool?
Status that could be of interest is `playing`, `paused` and `stopped` and maybe `exited` so a bool
probably not fine grained enough I'd personally would probably prefer an enum for this.
## How to test
Remove the matching methods from the npr-news skill and check that messages are still sent when playback starts.

## Contributor license agreement signed?
CLA [ Yes ]

PR #2683, opened 2020-08-27
feat/gui video common play

JarbasAl:
Starts integrating gui video playback into the common play framework

- adds self.gui.play_video 
- makes gui video playback respond to playback control skill
	- pause
	- resume
	- stop

see usage here https://github.com/JarbasSkills/skill-dagon/blob/feat/gui_common_play/__init__.py#L124


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2683))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2683))

mycroft/enclosure/gui.py
```diff
@@ -339,6 +340,37 @@ def show_url(self, url, override_idle=None,
...
+        """ Play video stream """
+        self["playStatus"] = "play"
+        self["video"] = url
+        self["title"] = title
+        self.video_info = {"title": title, "url": url}
+        self.show_page("SYSTEM_VideoPlayer.qml",
+                       override_idle=override_idle)
+
+    @property
+    def is_playing_video(self):
```
forslund, line 353: Missing docstring

mycroft/enclosure/gui.py
```diff
@@ -339,6 +340,37 @@ def show_url(self, url, override_idle=None,
...
+        return self.video_info is not None
```
forslund, line 354: Can the UI set the "playStatus" member to indicate that it has stopped playback? in that case should that also be considered?

forslund (changes requested):
I think this looks good I think. The thing I notice is that unlike the show_* this doesn't clear the current namespace when starting to show video.

We might want to think about how the show_* methods clears the namespace...It may cause skill creators some headache if they mix a show_* with normal pages. But that's outside of the scope of this PR.

I made a couple of minor comments, the missing docstring should be added before merge. And I'll leave the qml to be reviewed by @AIIX (since I'm terrible with it)

JarbasAl:
seems like we are moving to a centralize playback module per https://docs.google.com/document/d/1ebS60QLna4_AOQxQYpoaZrpkBzPXjL-7VFPJ7-xVvIQ/edit   maybe this no longer makes sense if its going to be replaced soon?

forslund:
My thinking is that this could still be of use since the playback module described there would need to talk to the gui in some fashion and this is the best api for that.

mycroft/enclosure/gui.py
JarbasAl: maybe add bus messages to clicks in the ui?

mycroft/enclosure/gui.py
```diff
@@ -339,6 +340,37 @@ def show_url(self, url, override_idle=None,
         self.show_page("SYSTEM_UrlFrame.qml", override_idle,
                        override_animations)
 
+    def play_video(self, url, title="", override_idle=True):
```
AIIX, line 343: should add override_animations=None

mycroft/enclosure/gui.py
```diff
@@ -339,6 +340,37 @@ def show_url(self, url, override_idle=None,
...
+        """ Play video stream """
+        self["playStatus"] = "play"
+        self["video"] = url
+        self["title"] = title
+        self.video_info = {"title": title, "url": url}
+        self.show_page("SYSTEM_VideoPlayer.qml",
```
AIIX, line 349: should add override_animations

krisgesling:
IMHO, I think it will be safer to get the architecture nailed down first then we can be sure about the changes.

JarbasAl:
> IMHO, I think it will be safer to get the architecture nailed down first then we can be sure about the changes.

i still think this should be merged, it's a simple and straightforward change, i don't the SPEC going in any direction where this change will be invalid

furthermore the SPEC is in very early stages and any implementation is not even being considered much less assigned, we are talking about delaying this change for months without any alternative available

if this is the official stance i will probably be closing this PR and adding this to my jarbas_utils package and start importing it in my skills, otherwise i will be duplicating a lot of code and have to maintain it in sync across many repos

when the SPEC gets off the idea stage into something concrete i will gladly update this

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2683))

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -255,9 +255,42 @@ def bind(self, bus):
             self.event_scheduler.set_id(self.skill_id)
             self._enclosure = EnclosureAPI(bus, self.name)
             self._register_system_event_handlers()
+
             # Initialize the SkillGui
             self.gui.setup_default_handlers()
 
+            # TODO can we rename this namespace to mycroft.playback.XXX ?
```
JarbasAl, line 262: small note, this renaming seems to make even more sense now according to SPEC https://docs.google.com/document/d/1ebS60QLna4_AOQxQYpoaZrpkBzPXjL-7VFPJ7-xVvIQ/edit#heading=h.pr2syrcbyvq6

but out of scope for this PR

krisgesling: Yeah agreed. 

I do however wonder if these handlers should also live in mycroft.enclosure.gui?

krisgesling (commented):
Yeah, I think I was bit rash last week. This will be needed whether we take the new direction or not. So I'm in favour of it being merged when ready

mycroft/res/ui/SeekControl.qml
```diff
@@ -0,0 +1,272 @@
...
+                        border.width: 1
+                        border.color: "white"
+                    }
+                    
+                    contentItem: Item {
+                        Image {
+                            width: parent.width > 600 ? Kirigami.Units.iconSizes.medium : Kirigami.Units.iconSizes.small
+                            height: width
+                            anchors.centerIn: parent
+                            source: "images/back.png"
```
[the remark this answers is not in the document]
JarbasAl: this was taken from @AIIX youtube skill

mycroft/enclosure/gui.py
JarbasAl: this is essentially an internal method for use in pause and stop methods, and you are correct that is_video_displayed is more descriptive

that said i think it makes sense to expose this to users, and i think splitting into those two makes sense, will add those changes :+1:

mycroft/skills/mycroft_skill/mycroft_skill.py
JarbasAl: self.gui is initialized before the bus is ready, the events can only be registered when/after the skill bind method is called

since the SkillGUI class takes the skill object as a param (which means it is not used outside skills) i don't see anything wrong in having those here, moving them around would require some refactoring to ensure event handlers were properly registered and increase complexity

JarbasAl: disregard previous comment, just noticed that SkillGUI has a setup_default_handlers method called in the skill bind method

it makes more sense to move these to the gui class since they dont use the skill class for anything, good catch!

mycroft/enclosure/gui.py
```diff
@@ -339,6 +381,54 @@ def show_url(self, url, override_idle=None,
...
+        return GUIPlaybackStatus.UNDEFINED
+
+    def pause_video(self):
+        """Pause video playback."""
+        if self.is_video_displayed:
+            self["playStatus"] = "pause"
+
+    def stop_video(self):
+        """Stop video playback."""
+        # TODO detect end of media playback from gui and call this
```
JarbasAl, line 421: @AIIX do you have any suggestions for this TODO ?

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2683))

mycroft/enclosure/gui.py
krisgesling: Yeah nice. 

With the addition of the `playback_status` would it be worth basing `is_video_displayed` off this status rather than the existence of `video_info`? 

`video_info` works, I'm just wondering if there's a higher chance of them getting out of sync with future changes. Such as adding a queuing method where `track_info` gets set to buffer a video prior to playback.

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -255,6 +255,7 @@ def bind(self, bus):
             self.event_scheduler.set_id(self.skill_id)
             self._enclosure = EnclosureAPI(bus, self.name)
             self._register_system_event_handlers()
+
```
krisgesling, line 258: This is nitpicky, but can we remove this so the file doesn't get touched by this PR?

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2683))

mycroft/res/ui/SeekControl.qml
```diff
@@ -0,0 +1,272 @@
...
+
+            }
+        }
+    }
+    
+
+    function formatTime(timeInMs) {
+        if (!timeInMs || timeInMs <= 0) return "0:00"
+        var seconds = timeInMs / 1000;
+        var minutes = Math.floor(seconds / 60)
```
krisgesling, line 267: Should we add in an `hours` that is only shown if time > 1hr. 

Would also need to pad the minutes `if (hours > 0 && minutes < 10)`

mycroft/res/ui/SeekControl.qml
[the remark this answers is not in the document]
AIIX: it can use an svg but are there any svg's available for this, not sure if we can just pick something of the internet cause of potential license issues or is there any creatives common license compatibility that too should be good

mycroft/enclosure/gui.py
AIIX: yes there is, we can detect when the media has ended or stopped playing, or status has changed and call the above https://doc.qt.io/qt-5/qml-qtmultimedia-video.html#status-prop

MetaGamer:
CC-licensed icons can be found at thenounproject.com. You can choose CC-BY 
for free or commercial with no attribution resulted for a couple USD.

AIIX:
Created an media player Icon Set that can be freely used with a CC0 1.0. license, free for commercial use and no attributions required. https://github.com/AIIX/SimpleMedia-Icons/

mycroft/res/ui/SeekControl.qml
AIIX: still need to implement this, If i understand correctly an hour would be represented like "1:05:04" "H:M:S"  format, and on duration change once it changes below an hour on displayed video would still be represented like "0:50:04" "H:M:S" format or "50.04" "M:S" format ?

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2683))

devops-mycroft:
Awesome, thanks. I think the previous and next icons are reversed?

mycroft/res/ui/SeekControl.qml
krisgesling: Yeah, good question, I hadn't thought of that but I like the idea of keeping the "H:M:S" format if the total duration is > 1hr but the time remaining drops below 1:00:00.
Where as if the total duration was < 1hr it would only ever show the "M:S" format. 

I think it's less cognitive load for the user, it indicates that we're under an hour to go, and less chance of strange things happening when rendering the time remaining to screen. 

@dschweppe do you have an opinion?

krisgesling:
Nice, I like the subtle rounded nature of the icons. Makes them feel like they are actually rounded buttons. Unless I'm just seeing things? 

I do wonder if the thinner aspects of some (eg repeat) might be harder to see on small screens, but we can test it out and tweak things down the line if needed.

AIIX:
> Nice, I like the subtle rounded nature of the icons. Makes them feel like they are actually rounded buttons. Unless I'm just seeing things?
> 
> I do wonder if the thinner aspects of some (eg repeat) might be harder to see on small screens, but we can test it out and tweak things down the line if needed.

yes they are rounded buttons based of the current video player seek controls design it gives them a subtle lift, making flat ones should be pretty easy if ever needed, also width/thickness and stuff is tweakable for any button, have left the inkscape edit layer intact so anyone can make changes as needed.

AIIX:
> Awesome, thanks. I think the previous and next icons are reversed?
> […](#)
> On September 21, 2020 3:11:44 PM "'Aditya Mehra' via DevOps" ***@***.***> wrote: Created a media player Icon Set that can be freely used with a CC0 1.0. license, free for commercial use and no attributions required. https://github.com/AIIX/SimpleMedia-Icons/ — You are receiving this because you commented. Reply to this email directly, view it on GitHub, or unsubscribe.

it was fixed in the previous commit as media-next.svg and media-previous.svg

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2683))

mycroft/enclosure/gui.py
```diff
@@ -53,6 +61,42 @@ def setup_default_handlers(self):
...
+                             self.__handle_gui_pause)
+        self.skill.add_event('mycroft.audio.service.resume',
+                             self.__handle_gui_resume)
+        self.skill.add_event('mycroft.audio.service.stop',
+                             self.__handle_gui_stop)
+        self.skill.add_event('mycroft.audio.service.track_info',
+                             self.__handle_gui_track_info)
+        self.skill.add_event('mycroft.audio.queue_end',
+                             self.__handle_gui_stop)
+        self.skill.gui.register_handler('video.media.playback.ended',
```
krisgesling, line 75: Was just giving this a run on a Mark II and this message doesn't seem to get triggered. Given it overrides idle this leaves the screen black until you trigger some other GUI event. I'm guessing this isn't happening on other platforms?

Also wondering if we should make this consistent with the other messages, perhaps `mycroft.video.playback.ended`

AIIX: is setup_default_handlers() even called anywhere, I don't see the handle even registering alone with this PR, maybe a missing piece in some other PR ?

krisgesling: Nevermind, it was my bad, `setup_default_handlers()` is called. I hadn't pulled in the latest changes on the device.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2683))

JarbasAl:
will re-open upon clarification of privacy ramifications of the latest mycroft partnership

code is free, you can copy and open a new PR if you do not want to wait for me re-opening it

krisgesling:
I've spent some more time trying to get this going. However I'm still just getting a black screen on both Ubuntu 20.04 and a Mark II whether I try playing a local mp4 or a url fetch from youtube.

I made a quick test Skill that does nothing but play a video:
https://github.com/krisgesling/play-a-video-skill/

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2683))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2683))

mycroft/enclosure/gui.py
```diff
@@ -478,7 +478,8 @@ def stop_video(self):
         # TODO detect end of media playback from gui and call this
         if self.is_video_displayed:
             self["playStatus"] = "stop"
-            self.clear()
+            self.skill.bus.emit(Message("mycroft.gui.screen.close",
```
krisgesling, line 481: Should this be `self.gui.release()` now?

krisgesling: Other than that, it's all working so good to merge :tada

AIIX: self.gui.clear() seems to cause an issue where it clears the data but also causes an issue of sending repeated events to mycroft-gui so its better to send the event directly for now. In any case self.gui.release() was never fully implemented. This PR (https://github.com/MycroftAI/skill-mark-2/pull/57) implements proper handling of the "mycroft.gui.screen.close" event for both this PR and self.gui.release() to work correctly and is needed for this PR as well.

krisgesling:
Hey @AIIX I've just done some minor clean up and added a series of tests, if you could double check it, I think we're good to go.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2683))

krisgesling:
@ken-mycroft - from our brief discussion today - here's the spec I mentioned:
https://docs.google.com/document/d/1ebS60QLna4_AOQxQYpoaZrpkBzPXjL-7VFPJ7-xVvIQ/edit?usp=sharing

As suggested by the spec, I think we should be moving toward an integrated "media" playback service and that Skills should interact with that so the system can more easily know and express it's own state. Either way the GUI needs a way to play that video, so I think this is moving in the right direction however in this PR's current state the interface for playing a video would be directly in the GUI module.

ken-mycroft:
Please forgive me for my ignorance, but why could I not simply add a fourth and fifth type (.mov and .mp4) to the existing simple audio service and let it figure out which player to use like it does now with .mp3 vs .wav (and I believe ogg). In this way nothing above it would change and it would conform to existing commands like play, stop, etc. We would simply need to add a new config value (mp4_cmd=qvlc) and add the qvlc plugin (for QT environments) to the requirements.txt file. Skills would continue to play media the same way they do now, or am I missing something very basic here?

AIIX:
> Please forgive me for my ignorance, but why could I not simply add a fourth and fifth type (.mov and .mp4) to the existing simple audio service and let it figure out which player to use like it does now with .mp3 vs .wav (and I believe ogg). In this way nothing above it would change and it would conform to existing commands like play, stop, etc. We would simply need to add a new config value (mp4_cmd=qvlc) and add the qvlc plugin (for QT environments) to the requirements.txt file. Skills would continue to play media the same way they do now, or am I missing something very basic here?

Mycroft-Core handles audio playback externally via VLC or whatever different from the Qt implementation, The same is not applicable to videos and not even possible to do as any video playback directly requires a video surface creation and this is directly handled with Qt's official multimedia backend and video qml type https://doc.qt.io/qt-5/qml-qtmultimedia-video.html which assumes all control and responsibility for video playback of the streaming or local video using Linux's multimedia gstreamer provider.

So the major difference between audio and video via CPS is: 
- Audio: Playback Control flows from Mycroft Core to GUI with all control assumed by common play framework  
- Video: Playback Control flows from GUI to Mycroft Core & CPS via events
 
Qt does not officially support qvlc and qml vlc extensions are not mature and don't seem even maintained anymore. Additionally there is no support for opening external application windows and controlling external Linux application window behavior.

AIIX:
Additionally most of the spec sheet being mentioned here has already been implemented by @JarbasAl in another project and can be found here https://github.com/OpenVoiceOS/OVOS-workshop/blob/master/ovos_workshop/frameworks/cps/__init__.py

ken-mycroft:
Quick question, is this https://github.com/OpenVoiceOS/OVOS-workshop/blob/master/ovos_workshop/frameworks/cps/__init__.py included in this PR? My concern is backward compatibility in non qt environments and this code seems to handle that, albeit in an interesting way.

AIIX:
> Quick question, is this https://github.com/OpenVoiceOS/OVOS-workshop/blob/master/ovos_workshop/frameworks/cps/__init__.py included in this PR? My concern is backward compatibility in non qt environments and this code seems to handle that, albeit in an interesting way.

No I believe that implementation by @JarbasAl https://github.com/OpenVoiceOS/OVOS-workshop/blob/master/ovos_workshop/frameworks/cps/__init__.py is a total refactor of the CPS framework based on the spec sheet with few more extensions,

This PR is only meant to handle show_video in the Qt context currently based on the current CPS implementation in mycroft-core, any kinds of backward compatibility etc should probably be handled separately under some other PR. As @JarbasAl was the original author of this PR he might know better. 

There were also some additional PR's by Jarbas for CPS spec implementation but were closed. I do not know if the above refactor done under OVOS relates but some of them are:

- https://github.com/MycroftAI/mycroft-core/pull/2660
- https://github.com/MycroftAI/skill-playback-control/pull/32

JarbasAl:
Just want to note that o tried to do this upstream across several prs, due to lack of interest I moved dev to ovos and am not following that spec

Not sure on status of this pr, but it should be quite basic and one of the first pieces, no conflicts with any of the other work I did

JarbasAl:
@ken-mycroft Here is what i could find, these are the PRs related to the better common play framework that i made before deciding to make this under https://github.com/OpenVoiceOS/OVOS-workshop . There are some more PRs in some of the official skills also (news skill ?)

- Track Status - [skill-playback-control/pull/35](https://github.com/MycroftAI/skill-playback-control/pull/35)
    - better handling of playback status, see [MycroftAI/mycroft-core#2619](https://github.com/MycroftAI/mycroft-core/pull/2619) and [MycroftAI/mycroft-core#2674](https://github.com/MycroftAI/mycroft-core/pull/2674)
- MatchType - [skill-playback-control/pull/32](https://github.com/MycroftAI/skill-playback-control/pull/32) and [mycroft-core/pull/2660](https://github.com/MycroftAI/mycroft-core/pull/2660)
    - playback now has a "media type" see [mycroft-core/issues/2658](https://github.com/MycroftAI/mycroft-core/issues/2658)
    - depending on intent additional info is used to help playback selection
    - integrates with video / media skills, no longer focused on music only
- Resume - [skill-playback-control/pull/38](https://github.com/MycroftAI/skill-playback-control/pull/38)
    - When told to "play" if music is paused now it restarts see [skill-playback-control/issues/18](https://github.com/MycroftAI/skill-playback-control/issues/18)
- Timeout - [skill-playback-control/pull/33](https://github.com/MycroftAI/skill-playback-control/pull/33)
    - improves the common play framework handling, timeout is now configurable,
      default is 5 seconds, previoulsy many skills failed to play because 
      the timeout was too small
      
      
These PRs partially informed the SPEC linked above

JarbasAl:
closing this PR in favor of the new audio backend plugin

this PR is very old and outdated

PR #2686, opened 2020-09-06
{start,stop}-mycroft.sh: make POSIX sh compatible and fix shellcheck issues

PureTryOut:
## Description
Makes the start and stop scripts POSIX sh compatible, and fixes any `shellcheck` issues found.

This doesn't really depend on #2802, but without it `bin/mycroft-start` and `bin/mycroft-stop` will break. Thus, #2802 should be merged first.

## How to test
Run all the usual variants of this test like normal (e.g. `./start-mycroft.sh`) and see if anything behaves differently than before

### DIFF
--- a/start-mycroft.sh
+++ b/start-mycroft.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 # Copyright 2017 Mycroft AI Inc.
 #
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-SOURCE="${BASH_SOURCE[0]}"
+SOURCE="$0"
 
 script=${0}
 script=${script##*/}
@@ -22,7 +22,7 @@ cd -P "$( dirname "$SOURCE" )" || exit 1 # Enter scripts folder or fail!
 DIR="$( pwd )"
 VIRTUALENV_ROOT=${VIRTUALENV_ROOT:-"${DIR}/.venv"}
 
-function help() {
+help() {
     echo "${script}:  Mycroft command/service launcher"
     echo "usage: ${script} [COMMAND] [restart] [params]"
     echo
@@ -60,7 +60,7 @@ function help() {
 }
 
 _module=""
-function name-to-script-path() {
+name_to_script_path() {
     case ${1} in
         "bus")               _module="mycroft.messagebus.service" ;;
         "skills")            _module="mycroft.skills" ;;
@@ -77,47 +77,47 @@ function name-to-script-path() {
     esac
 }
 
-function source-venv() {
+source_venv() {
     # Enter Python virtual environment, unless under Docker
     if [ ! -f "/.dockerenv" ] ; then
-        source "${VIRTUALENV_ROOT}/bin/activate"
+        . "${VIRTUALENV_ROOT}/bin/activate"
     fi
 }
 
 first_time=true
-function init-once() {
+init_once() {
     if ($first_time) ; then
         echo "Initializing..."
         "${DIR}/scripts/prepare-msm.sh"
-        source-venv
+        source_venv
         first_time=false
     fi
 }
 
-function launch-process() {
-    init-once
+launch_process() {
+    init_once
 
-    name-to-script-path "${1}"
+    name_to_script_path "${1}"
 
     # Launch process in foreground
     echo "Starting $1"
     python3 -m ${_module} "$@"
 }
 
-function require-process() {
+require_process() {
     # Launch process if not found
-    name-to-script-path "${1}"
+    name_to_script_path "${1}"
     if ! pgrep -f "python3 (.*)-m ${_module}" > /dev/null ; then
         # Start required process
-        launch-background "${1}"
+        launch_background "${1}"
     fi
 }
 
-function launch-background() {
-    init-once
+launch_background() {
+    init_once
 
     # Check if given module is running and start (or restart if running)
-    name-to-script-path "${1}"
+    name_to_script_path "${1}"
     if pgrep -f "python3 (.*)-m ${_module}" > /dev/null ; then
         if ($_force_restart) ; then
             echo "Restarting: ${1}"
@@ -131,7 +131,7 @@ function launch-background() {
     fi
 
     # Security warning/reminder for the user
-    if [[ "${1}" == "bus" ]] ; then
+    if [ "${1}" = "bus" ] ; then
         echo "CAUTION: The Mycroft bus is an open websocket with no built-in security"
         echo "         measures.  You are responsible for protecting the local port"
         echo "         8181 with a firewall as appropriate."
@@ -141,29 +141,29 @@ function launch-background() {
     python3 -m ${_module} "$@" >> "/var/log/mycroft/${1}.log" 2>&1 &
 }
 
-function launch-all() {
+launch_all() {
     echo "Starting all mycroft-core services"
-    launch-background bus
-    launch-background skills
-    launch-background audio
-    launch-background voice
-    launch-background enclosure
+    launch_background bus
+    launch_background skills
+    launch_background audio
+    launch_background voice
+    launch_background enclosure
 }
 
-function check-dependencies() {
+check_dependencies() {
     if [ -f .dev_opts.json ] ; then
         auto_update=$( jq -r ".auto_update" < .dev_opts.json 2> /dev/null)
     else
         auto_update="false"
     fi
-    if [ "$auto_update" == "true" ] ; then
+    if [ "$auto_update" = "true" ] ; then
         # Check github repo for updates (e.g. a new release)
         git pull
     fi
 
-    if [ ! -f .installed ] || ! md5sum -c &> /dev/null < .installed ; then
+    if [ ! -f .installed ] || ! md5sum -c > /dev/null 2>&1 < .installed ; then
         # Critical files have changed, dev_setup.sh should be run again
-        if [ "$auto_update" == "true" ] ; then
+        if [ "$auto_update" = "true" ] ; then
             echo "Updating dependencies..."
             bash dev_setup.sh
         else
@@ -179,82 +179,91 @@ function check-dependencies() {
 
 _opt=$1
 _force_restart=false
+
+if [ $# -eq 0 ]; then
+	help
+	return
+fi
+
 shift
-if [[ "${1}" == "restart" ]] || [[ "${_opt}" == "restart" ]] ; then
+if [ "${1}" = "restart" ] || [ "${_opt}" = "restart" ] ; then
     _force_restart=true
-    if [[ "${_opt}" == "restart" ]] ; then
+    if [ "${_opt}" = "restart" ] ; then
         # Support "start-mycroft.sh restart all" as well as "start-mycroft.sh all restart"
         _opt=$1
     fi
-    shift
+
+    if [ $# -gt 0 ]; then
+	    shift
+    fi
 fi
 
-if [[ ! "${_opt}" == "cli" ]] ; then
-    check-dependencies
+if [ ! "${_opt}" = "cli" ] ; then
+    check_dependencies
 fi
 
 case ${_opt} in
     "all")
-        launch-all
+        launch_all
         ;;
 
     "bus")
-        launch-background "${_opt}"
+        launch_background "${_opt}"
         ;;
     "audio")
-        launch-background "${_opt}"
+        launch_background "${_opt}"
         ;;
     "skills")
-        launch-background "${_opt}"
+        launch_background "${_opt}"
         ;;
     "voice")
-        launch-background "${_opt}"
+        launch_background "${_opt}"
         ;;
 
     "debug")
-        launch-all
-        launch-process cli
+        launch_all
+        launch_process cli
         ;;
 
     "cli")
-        require-process bus
-        require-process skills
-        launch-process "${_opt}"
+        require_process bus
+        require_process skills
+        launch_process "${_opt}"
         ;;
 
     # TODO: Restore support for Wifi Setup on a Picroft, etc.
     # "wifi")
-    #    launch-background ${_opt}
+    #    launch_background ${_opt}
     #    ;;
     "unittest")
-        source-venv
+        source_venv
         pytest test/unittests/ --cov=mycroft "$@"
         ;;
     "singleunittest")
-        source-venv
+        source_venv
         pytest "$@"
         ;;
     "skillstest")
-        source-venv
+        source_venv
         pytest test/integrationtests/skills/discover_tests.py "$@"
         ;;
     "vktest")
-        source "$DIR/bin/mycroft-skill-testrunner" vktest "$@"
+        "$DIR/bin/mycroft-skill-testrunner" vktest "$@"
         ;;
     "audiotest")
-        launch-process "${_opt}"
+        launch_process "${_opt}"
         ;;
     "wakewordtest")
-        launch-process "${_opt}"
+        launch_process "${_opt}"
         ;;
     "sdkdoc")
-        source-venv
+        source_venv
         cd doc || exit 1  # Exit if doc directory doesn't exist
         make "$@"
         cd ..
         ;;
     "enclosure")
-        launch-background "${_opt}"
+        launch_background "${_opt}"
         ;;
 
     *)
--- a/start-mycroft.sh
+++ b/start-mycroft.sh
@@ -189,7 +189,6 @@ if [[ "${1}" == "restart" ]] || [[ "${_opt}" == "restart" ]] ; then
     shift
 fi
 
-
 if [[ ! "${_opt}" == "cli" ]] ; then
     check-dependencies
 fi
--- a/stop-mycroft.sh
+++ b/stop-mycroft.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 # Copyright 2017 Mycroft AI Inc.
 #
@@ -14,13 +14,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-SOURCE="${BASH_SOURCE[0]}"
+# This script is never sourced but always directly executed, so this is safe to do
+SOURCE="$0"
 
 script=${0}
 script=${script##*/}
 cd -P "$( dirname "$SOURCE" )" || exit 1 # quit if change of folder fails
 
-function help() {
+help() {
     echo "${script}:  Mycroft service stopper"
     echo "usage: ${script} [service]"
     echo
@@ -40,36 +41,36 @@ function help() {
     exit 0
 }
 
-function process-running() {
-    if [[ $( pgrep -f "python3 (.*)-m mycroft.*${1}" ) ]] ; then
+process_running() {
+    if [ "$( pgrep -f "python3 (.*)-m mycroft.*${1}" )" ] ; then
         return 0
     else
         return 1
     fi
 }
 
-function end-process() {
-    if process-running "$1" ; then
+end_process() {
+    if process_running "$1" ; then
         # Find the process by name, only returning the oldest if it has children
         pid=$( pgrep -o -f "python3 (.*)-m mycroft.*${1}" )
-        echo -n "Stopping $1 (${pid})..."
-        kill -SIGINT "${pid}"
+        printf "Stopping %s (%s)..." "$1" "${pid}"
+        kill -s INT "${pid}"
 
         # Wait up to 5 seconds (50 * 0.1) for process to stop
         c=1
         while [ $c -le 50 ] ; do
-            if process-running "$1" ; then
+            if process_running "$1" ; then
                 sleep 0.1
-                (( c++ ))
+                c=$((c + 1))
             else
                 c=999   # end loop
             fi
         done
 
-        if process-running "$1" ; then
+        if process_running "$1" ; then
             echo "failed to stop."
             pid=$( pgrep -o -f "python3 (.*)-m mycroft.*${1}" )            
-            echo -n "  Killing $1 (${pid})..."
+            printf "  Killing %s (%s)...\n" "$1" "${pid}"
             kill -9 "${pid}"
             echo "killed."
             result=120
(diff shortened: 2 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2686))

PureTryOut:
Hmm, not sure why Travis fails.

```
$ ./start-mycroft.sh unittest
./start-mycroft.sh: 85: .: Can't open /home/travis/build/MycroftAI/mycroft-core/.venv/bin/activate
```

This works fine for me locally.

forslund:
The python env in Travis exists already so the .venv/bin/activate is not actually used here. Earlier the command failed and carried on with the already active venv. (See here for example https://travis-ci.org/github/forslund/mycroft-core/jobs/635713266#L1410)...so either add `|| true` or perhaps the better way is to do a check if a venv already is active...

The following oneliner should be best according to stack overflow
`python -c 'import sys;sys.exit(not (hasattr(sys, "real_prefix") or (hasattr(sys, "base_prefix") and sys.base_prefix != sys.prefix)))'`

Rather long but still a single line...

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2686))

PureTryOut:
I do not care if it's one line or more to be honest, but then again I'm not sure how to split that up into multiple :joy: 

Thanks for the explanation, I'll add that check.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2686))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2686))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2686))

start-mycroft.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
```
fermulator, line 1: we should still be using `env`
```
#!/usr/bin/env sh
```

start-mycroft.sh
```diff
@@ -14,15 +14,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-SOURCE="${BASH_SOURCE[0]}"
+# This script is never sourced but always directly executed, so this is safe to do
```
fermulator, line 17: I mean ... technically we should embed all "work" in a `main()` function to avoid sourcing collisions ... but I agree not a huge issue.

start-mycroft.sh
```diff
@@ -14,15 +14,16 @@
...
+SOURCE="$0"
 
 script=${0}
 script=${script##*/}
-cd -P "$( dirname "$SOURCE" )"
+cd -P "$( dirname "$SOURCE" )" || exit
```
fermulator, line 22: silent failure? ideally log the error here

start-mycroft.sh
```diff
@@ -77,51 +78,54 @@ function name-to-script-path() {
     esac
 }
 
-function source-venv() {
-    # Enter Python virtual environment, unless under Docker
```
fermulator, line 81: what is the requirement/reasoning for not using python venv within docker? (is it pre-loaded?)

start-mycroft.sh
```diff
@@ -77,51 +78,54 @@ function name-to-script-path() {
...
-    if [ ! -f "/.dockerenv" ] ; then
-        source ${VIRTUALENV_ROOT}/bin/activate
+source_venv() {
+    # Enter Python virtual environment, unless under Docker or when a virtualenv is already active
+    virtualenv=$(python3 -c 'import sys; sys.stdout.write("1") if (hasattr(sys, "real_prefix") or (hasattr(sys, "base_prefix") and sys.base_prefix != sys.prefix)) else sys.stdout.write("0")')
```
fermulator, line 83: i tried this on my `hass` venv environment 

OUTSIDE venv:
```
fermulator@hass:~$ python3 -c 'import sys; sys.stdout.write("1") if (hasattr(sys, "real_prefix") or (hasattr(sys, "base_prefix") and sys.base_prefix != sys.prefix)) else sys.stdout.write("0")'
0 
```

INSIDE venv
```
python3 -c 'import sys; sys.stdout.write("1") if (hasattr(sys, "real_prefix") or (hasattr(sys, "base_prefix") and sys.base_prefix != sys.prefix)) else sys.stdout.write("0")'
1
```

:heavy_check_mark:

start-mycroft.sh
```diff
@@ -131,39 +135,39 @@ function launch-background() {
...
-    if [[ "${1}" == "bus" ]] ; then
+    if [ "${1}" = "bus" ] ; then
         echo "CAUTION: The Mycroft bus is an open websocket with no built-in security"
         echo "         measures.  You are responsible for protecting the local port"
         echo "         8181 with a firewall as appropriate."
     fi
 
     # Launch process in background, sending logs to standard location
-    python3 -m ${_module} $_params >> /var/log/mycroft/${1}.log 2>&1 &
+    python3 -m ${_module} "$_params" >> /var/log/mycroft/"${1}".log 2>&1 &
```
fermulator, line 145: i would enclose the ENTER blob log file path in quotes (cleaner)

```
... >> "/var/log/mycroft/${1}.log" ...
```

start-mycroft.sh
```diff
@@ -131,39 +135,39 @@ function launch-background() {
...
         auto_update="false"
     fi
-    if [ "$auto_update" == "true" ] ; then
+    if [ "$auto_update" = "true" ] ; then
         # Check github repo for updates (e.g. a new release)
         git pull
     fi
 
-    if [ ! -f .installed ] || ! md5sum -c &> /dev/null < .installed ; then
+    if [ ! -f .installed ] || ! md5sum -c 2>&1 > /dev/null < .installed ; then
```
fermulator, line 168: what's the reason to also shove `stderr` to `/dev/null`?

start-mycroft.sh
```diff
@@ -180,80 +184,80 @@ function check-dependencies() {
...
         pytest "$@"
         ;;
     "skillstest")
-        source-venv
+        source_venv
         pytest test/integrationtests/skills/discover_tests.py "$@"
         ;;
     "vktest")
-        source "$DIR/bin/mycroft-skill-testrunner" vktest "$@"
+        bash "$DIR/bin/mycroft-skill-testrunner" vktest "$@"
```
fermulator, line 245: shouldn't this script define its own shebang? (ideally we wouldn't force `bash` from here)

stop-mycroft.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
```
fermulator, line 1: similar comments on this file with `start-mycroft.sh` (fix throughout) as applicable

stop-mycroft.sh
```diff
@@ -40,37 +41,37 @@ function help() {
...
             fi
         done
 
-        if process-running $1 ; then
+        if process_running "$1" ; then
             echo "failed to stop."
             pid=$( pgrep -o -f "python3 (.*)-m mycroft.*${1}" )            
-            echo -n "  Killing $1 (${pid})..."
-            kill -9 ${pid}
+            printf "  Killing %s (%s)..." "$1" "${pid}"
```
fermulator, line 73:
```
printf "  Killing %s (%s)..." "${1}" "${pid}"
```

also, when I tried this I don't know why it trails a `$`

```
$ FOO=blah
$ BAR=23

$ printf "  Killing %s (%s)..." "$FOO" "$BAR"
  Killing blah (23)...$ ```

**Review by fermulator — CHANGES_REQUESTED** [review]

File: start-mycroft.sh (hunk 4 above)
Comment (forslund) [reply start-mycroft.sh:81]: If I recall correctly they wanted to skip this step in the docker since it isn't strictly needed (the docker environment is enough of a sandbox). However I believe the current docker setup (in [docker-mycroft](https://github.com/mycroftai/docker-mycroft) is currently using the venv

File: start-mycroft.sh (hunk 8 above)
Comment (forslund) [reply start-mycroft.sh:245]: It does have it's own shebang so bash could be omitted... @krisgesling do you remember if there was any thought behind sourcing it? (Making it the same level as the rest of the function calls?)

Comment (krisgesling) [reply start-mycroft.sh:245]: May have been related to something else I was trying and just got left in there. I can't think of any reason that it would be needed. Remove it I think.

File: stop-mycroft.sh (hunk 10 above)
Comment (forslund) [reply stop-mycroft.sh:73]: Is that your prompt maybe?

There should be a `\n` at the end of the string

File: start-mycroft.sh (hunk 1 above)
Comment (PureTryOut) [reply start-mycroft.sh:1]: I don't see a reason why. For one, env is a GNU Coreutils thing, and I'd like this to work on systems without those installed. Secondly, we can assume that on POSIX systems `/bin/sh` is always available and it's a POSIX compatible shell. If it's not, that's broken on the distro in question and not something that we should care about, as that's just a system not being POSIX.

File: start-mycroft.sh (hunk 2 above)
Comment (PureTryOut) [reply start-mycroft.sh:17]: Agreed, I might do that in the future but that's out of the scope of this PR for now.

File: start-mycroft.sh (hunk 6 above)
Comment (PureTryOut) [reply start-mycroft.sh:145]: That's a personal preference but one I can get behind, resolved.

File: stop-mycroft.sh (hunk 10 above)
Comment (PureTryOut) [reply stop-mycroft.sh:73]: Strange, Bash doesn't seem to append the `\n` but zsh does. I'll add a `\n` anyway to force it.

File: start-mycroft.sh (hunk 7 above)
Comment (PureTryOut) [reply start-mycroft.sh:168]: No reason really, that's just a habit of mine I guess. Good call, I'll change it.

File: start-mycroft.sh (hunk 3 above)
Comment (PureTryOut) [reply start-mycroft.sh:22]: That's... A bit hard to do. It'll exit with whatever the return code was of `cd` which should be enough indication in itself on what went wrong, but to actually log something is a bit tricky.

https://github.com/koalaman/shellcheck/wiki/SC2242

I'm not entirely sure how to capture the result of `cd` and check it later to then log it in some conditional statement.

**Comment by devops-mycroft** [comment]
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2686))

**Comment by devops-mycroft** [comment]
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2686))

File: start-mycroft.sh (hunk 3 above)
Comment (forslund) [reply start-mycroft.sh:22]: I do think cd will log the failure (atleast in my tests it does). And as you said exit-status is preserved so this should be fine

File: stop-mycroft.sh [hunk 11]
```diff
@@ -90,30 +91,28 @@ OPT=$1
 shift
```
krisgesling, line 91: It seems that `sh` doesn't handle shifting when there are no arguments.
```
./mycroft-core/stop-mycroft.sh: 91: shift: can't shift that many
```

So perhaps we need a conditional on here?

krisgesling (commented):
Hey, this looks great, thanks!

Only one small issue with `shift` that I think we need to resolve.

Thanks @fermulator and @forslund for reviewing and testing this already too!

stop-mycroft.sh
PureTryOut: That is strange, that shouldn't be the case. Afaik if none argument is specified, it should just do `N+1`.
Shellcheck doesn't complain either. How did you get this error? I can't reproduce it personally.

I could do `shift 1` but that seems redundant. What conditional would you suggest?

krisgesling: Apparently the expected behaviour in [ash](https://linux.die.net/man/1/ash):
> Shift the positional parameters n times. A shift sets the value of $1 to the value of $2, the value of $2 to the value of $3, and so on, decreasing the value of $# by one. If n is greater than the number of positional parameters, shift will issue an error message, and exit with return status 2.

What about:
```
OPT=$1
if [ $# -gt 0 ] ; then
    shift
fi
```

PureTryOut: Yup thanks for that pointer. My system doesn't run Ash (`/bin/sh` points to Busybox) for me so I didn't catch this.

Modified it, could you try again?

stop-mycroft.sh
```diff
@@ -87,33 +88,33 @@ result=0  # default, no change
 
 
 OPT=$1
-shift
+if [ $# -gt 0]; then
```
krisgesling, line 91: little typo, missing the space between `0 ]`

PureTryOut: :see_no_evil:

start-mycroft.sh
```diff
@@ -180,80 +184,80 @@ function check-dependencies() {
...
-if [[ "${1}" == "restart" ]] || [[ "${_opt}" == "restart" ]] ; then
+if [ "${1}" = "restart" ] || [ "${_opt}" = "restart" ] ; then
     _force_restart=true
-    if [[ "${_opt}" == "restart" ]] ; then
+    if [ "${_opt}" = "restart" ] ; then
         # Support "start-mycroft.sh restart all" as well as "start-mycroft.sh all restart"
         _opt=$1
     fi
     shift
 fi
```
krisgesling, line 194: Sorry I just realised we also run into the same problem here. It's not intended behaviour to use this script without args, but I'd rather we show the help info if someone does try to use it incorrectly.

I guess the easiest method would be to throw a conditional on both? But it's getting late here so I may just be getting lazy...

PureTryOut: The key bit being "It's not intended behaviour to use this script without args".

I checked if `$#` equals 0 now, and if it does show the help and exit. If it's at least 1 or more, both shifts should be successful now I think?

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2686))

start-mycroft.sh
krisgesling: The other edge case would be `mycroft-start restart`
which now fails with a shift error when called using start-mycroft.sh
and directs to update dependencies for bin/mycroft-start

PureTryOut: I'm not sure how that still happens but ok, I added a check for shift within the restart block too.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2686))

start-mycroft.sh
```diff
@@ -179,81 +183,90 @@ function check-dependencies() {
...
+if [ "${1}" = "restart" ] || [ "${_opt}" = "restart" ] ; then
     _force_restart=true
-    if [[ "${_opt}" == "restart" ]] ; then
+    if [ "${_opt}" = "restart" ] ; then
         # Support "start-mycroft.sh restart all" as well as "start-mycroft.sh all restart"
         _opt=$1
     fi
-    shift
+
+    if [ $# -gt 1 ]; then
```
krisgesling, line 200: I think we want this as `-gt 0` as the first param has already been shifted off.

test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+echo "$#"
```
krisgesling, line 3: Looks like this got committed by accident

start-mycroft.sh
```diff
@@ -14,15 +14,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-SOURCE="${BASH_SOURCE[0]}"
+# This script is never sourced but always directly executed, so this is safe to do
+SOURCE="$0"
```
krisgesling, line 18: Don't know how I glossed over this previously but the script is sourced in the `bin/mycroft-start` helper command.

This means the $PWD will get set to `mycroft-core/bin` which causes issues in `check_dependencies` when it is reading from files stored in the parent `mycroft-core` directory.

krisgesling (commented):
Sorry, looks like we aren't quite done...

start-mycroft.sh
PureTryOut: Hmm yes you're right. POSIX doesn't have an easy way to do this, this will get ugly...

PureTryOut: It seems it's just not possible to determine the path of a source script from within itself using just POSIX features...
https://stackoverflow.com/questions/29832037/how-to-get-script-directory-in-posix-sh/29835459#29835459

So, not sure how to resolve this. `pwd` seems to always return the directory in which the script is located (at least for me with Busybox, GNU Coreutils seem to do the same), no matter if it's sourced or not, but I can not find anything about that behaviour online so I'm not sure if we should rely on it. Otherwise it would be as easy as `SOURCE=$(pwd)`.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2686))

start-mycroft.sh
krisgesling: Thinking out loud here... in the helper commands could we set `$BASH_SOURCE=$(pwd)` or an equivalent variable. Then test for it's existence in the start/stop scripts rather than assume it is there or not?

PureTryOut: What helper commands do you mean? `bin/mycroft-start`?

stop-mycroft.sh
```diff
@@ -40,37 +41,37 @@ function help() {
...
-function end-process() {
-    if process-running $1 ; then
+end_process() {
+    if process_running "$1" ; then
         # Find the process by name, only returning the oldest if it has children
         pid=$( pgrep -o -f "python3 (.*)-m mycroft.*${1}" )
-        echo -n "Stopping $1 (${pid})..."
-        kill -SIGINT ${pid}
+        printf "Stopping %s (%s)..." "$1" "${pid}"
+        kill -SIGINT "${pid}"
```
forslund, line 57: Was looking at this to see if it could be merged got an issue with this:

```
Stopping skills (6297)..../stop-mycroft.sh: 57: kill: Illegal option -S
failed to stop.
  Killing skills (6297)...
killed.
Stopping audio (6300)..../stop-mycroft.sh: 57: kill: Illegal option -S
failed to stop.
  Killing audio (6300)...
killed.
Stopping speech (6303)..../stop-mycroft.sh: 57: kill: Illegal option -S
failed to stop.
  Killing speech (6303)...
killed.
Stopping enclosure (6306)..../stop-mycroft.sh: 57: kill: Illegal option -S
failed to stop.
  Killing enclosure (6306)...
killed.
Stopping messagebus.service (17057)..../stop-mycroft.sh: 57: kill: Illegal option -S
failed to stop.
  Killing messagebus.service (17057)...
killed.
```

seems like kill in sh doesn't work with signal names?

PureTryOut: Is `-SIGINT` equivelant to `-s SIGINT`? I believe sh just requires the `-s` option before specifying what signal to send. I just pushed a change, could you try again?

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2686))

stop-mycroft.sh
forslund: Now it's `Stopping skills (5419)..../stop-mycroft.sh: 57: kill: invalid signal number or name: SIGINT` `kill -l` in sh suggests it should be just `INT`

PureTryOut: How strange, fixed.

forslund: Yes, sh seems to be Dash on my machine.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2686))

stop-mycroft.sh
forslund: Working now. :+1:

start-mycroft.sh
forslund: Looked a bit on this and the helper commands in bin that uses start-mycroft.sh and stop-mycroft.sh seem to fail (when called from outside of the mycroft-core root folder). I wonder if this was what Kris meant. but I'm not sure.

I think this is due to the fact that the files are sourced and not executed as a script. for me it seems to work if instead of sourcing the start-mycroft.sh the script is executed.

For example in mycroft-start the change would be
```diff
diff --git a/bin/mycroft-start b/bin/mycroft-start
index 65723c2c37..bda0967048 100755
--- a/bin/mycroft-start
+++ b/bin/mycroft-start
@@ -18,4 +18,4 @@ SOURCE="${BASH_SOURCE[0]}"
 cd -P "$( dirname "$SOURCE" )"/..
 DIR="$( pwd )"
 
-. "$DIR/start-mycroft.sh" $@
+"$DIR/start-mycroft.sh" $@
```

PureTryOut: I've just gone ahead and made this depend on #2802, as that'll completely replace `bin/mycroft-start` for a new script anyway. Maybe at some point we can get rid of these location dependent scripts entirely with just `bin/mycroft-{start,stop}`, who knows.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2686)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2686/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2686/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2686/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2686/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2686/logs/enclosure.log)

PureTryOut:
Just rebased this. @krisgesling anything holding this back?

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2686))

start-mycroft.sh
fermulator: sure thing! sounds good (if a dev system doesn't have coreutils installed that's odd; but agreed that `/bin/sh` is most portable)

fermulator (approved):

start-mycroft.sh
```diff
@@ -179,83 +179,91 @@ function check-dependencies() {
...
-fi
 
+    if [ $# -gt 0 ]; then
+	    shift
+    fi
+fi
 
-if [[ ! "${_opt}" == "cli" ]] ; then
+if [ ! "${_opt}" = "cli" ] ; then
     check-dependencies
```
forslund: I think this should be replaced with the `check_dependencies`

forslund (dismissed):
Looks really good and seems to work well. I did find a place where the `check-dependencies`call hadn't been updated (L202)

start-mycroft.sh
```diff
@@ -141,29 +141,29 @@ function launch-background() {
...
         auto_update="false"
     fi
-    if [ "$auto_update" == "true" ] ; then
+    if [ "$auto_update" = "true" ] ; then
         # Check github repo for updates (e.g. a new release)
         git pull
     fi
 
-    if [ ! -f .installed ] || ! md5sum -c &> /dev/null < .installed ; then
+    if [ ! -f .installed ] || ! md5sum -c 2>&1 > /dev/null < .installed ; then
```
forslund: For me to silence this completely when md5sum detects a difference is detected I had to change the order of the redirects:
```sh
md5sum -c > /dev/null 2>&1 < .installed
```

krisgesling (approved):
With Ake's final catches addressed, this seems good to go!

:tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada:

PR #2687, opened 2020-09-06
{start,stop}-mycroft.sh: allow system-wide setups

### LINKED ISSUE #2191: Make start-mycroft.sh independent from dev_setup.sh
The dev_setup.sh script included in the source release would not function as it requires the directory to be a git repo.

It has to be fixed in order for the source release to be packaged.

PureTryOut:
## Description
This allows `start-mycroft.sh` and `stop-mycroft.sh` to start and stop MyCroft when it's installed system-wide, assuming the files are installed as `start-mycroft` and `stop-mycroft` (so without the `.sh` extension). This is especially useful for distribution packaging which won't have a git checkout with the source and a virtualenv available and can probably be used for Pip as well.

Note that this PR depends on #2686 and shouldn't be merged without it. I'll rebase this PR on the latest `dev` when #2686 is merged.

## How to test
Install MyCroft system wide (e.g. `python3 setup.py install --prefix=/usr --root=/`, make sure you have all the deps installed system-wide!), copy/rename `start-mycroft.sh` to `start-mycroft` and optionally put it in `/usr/bin`. Then just run `start-mycroft all`. Do the same for `stop-mycroft.sh`.

Fixes #2191

### DIFF
--- a/start-mycroft.sh
+++ b/start-mycroft.sh
@@ -18,7 +18,7 @@ SOURCE="${BASH_SOURCE[0]}"
 
 script=${0}
 script=${script##*/}
-cd -P "$( dirname "$SOURCE" )"
+cd -P "$( dirname "$SOURCE" )" || exit
 DIR="$( pwd )"
 VIRTUALENV_ROOT=${VIRTUALENV_ROOT:-"${DIR}/.venv"}
 
@@ -80,7 +80,8 @@ function name-to-script-path() {
 function source-venv() {
     # Enter Python virtual environment, unless under Docker
     if [ ! -f "/.dockerenv" ] ; then
-        source ${VIRTUALENV_ROOT}/bin/activate
+        # shellcheck source=/dev/null
+        source "${VIRTUALENV_ROOT}"/bin/activate
     fi
 }
 
@@ -97,19 +98,19 @@ function init-once() {
 function launch-process() {
     init-once
 
-    name-to-script-path ${1}
+    name-to-script-path "${1}"
 
     # Launch process in foreground
     echo "Starting $1"
-    python3 -m ${_module} $_params
+    python3 -m ${_module} "$_params"
 }
 
 function require-process() {
     # Launch process if not found
-    name-to-script-path ${1}
+    name-to-script-path "${1}"
     if ! pgrep -f "python3 (.*)-m ${_module}" > /dev/null ; then
         # Start required process
-        launch-background ${1}
+        launch-background "${1}"
     fi
 }
 
@@ -117,11 +118,11 @@ function launch-background() {
     init-once
 
     # Check if given module is running and start (or restart if running)
-    name-to-script-path ${1}
+    name-to-script-path "${1}"
     if pgrep -f "python3 (.*)-m ${_module}" > /dev/null ; then
         if ($_force_restart) ; then
             echo "Restarting: ${1}"
-            "${DIR}/stop-mycroft.sh" ${1}
+            "${DIR}/stop-mycroft.sh" "${1}"
         else
             # Already running, no need to restart
             return
@@ -138,7 +139,7 @@ function launch-background() {
     fi
 
     # Launch process in background, sending logs to standard location
-    python3 -m ${_module} $_params >> /var/log/mycroft/${1}.log 2>&1 &
+    python3 -m ${_module} "$_params" >> /var/log/mycroft/"${1}".log 2>&1 &
 }
 
 function launch-all() {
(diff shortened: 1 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2687))

start-mycroft.sh
```diff
@@ -131,39 +158,49 @@ function launch-background() {
...
     fi
 
-    # Launch process in background, sending logs to standard location
-    python3 -m ${_module} $_params >> /var/log/mycroft/${1}.log 2>&1 &
+    # Launch process in background
+    # Send logs to old standard location if it exists
+    # Otherwise send to XDG Base Directories cache location
+    logdir="/var/log/mycroft/${1}.log"
+    if [ ! -f "$logdir" ]; then
+	    if [ ! -z ${XDG_CACHE_HOME+x} ]; then
```
PureTryOut, line 172: I'm not entirely sure if the cache dir is the right directory. Maybe it should go to `XDG_DATA_HOME` (so `$HOME/.local/share`) instead?

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2687))

start-mycroft.sh
krisgesling: Based purely on the [directory descriptions](https://wiki.archlinux.org/index.php/XDG_Base_Directory) on the Arch wiki I'd lean toward `XDG_DATA_HOME`. But a very quick search doesn't seem to give any definitive answers. Might post in Chat and see if anyone has strong opinions

ChanceNCounter: The XDG spec designates `XDG_DATA_HOME` for "user-specific data," whereas `XDG_CACHE_HOME` is meant for "user-specific non-essential (cached) data."

By extension, data in `XDG_DATA_HOME` *is* essential.

It's worth mentioning that this question has been asked a million times and still has no clear answer. Some people would like to flush the cache dir from time to time, because it says *cache* on it. You don't want logfiles going *poof* on power loss. On the other hand, some people want to track the data dir in a VCS, because dotfiles. You probably don't want to commit your logs to a VCS.

I vote `XDG_CACHE_HOME`. They're non-essential data, and the likelihood of somebody *actually* flushing `XDG_CACHE_HOME` is much slimmer than somebody tracking `XDG_DATA_HOME` (that'll happen a lot.)

---

At one point, Debian proposed yet another envvar, for a directory to store things like logs and histfiles. That was a long time ago. The XDG mailing list isn't very active. Indeed, the most active thread right now is about who's gonna take over an abandoned spec...

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2687))

start-mycroft.sh
j1nx: Considering /var/log/ in many distributions and / or setups is mounted on a tmpfs and so be gone at a reboot I am with @PureTryOut and @ChanceNCounter classifying it as non-essential data and therefor XDG_CACHE_HOME fits the description better.

AIIX: There are two cache directories defined in the linux file hireachy spec man pages:

SYSTEM Level: /var/cache https://man7.org/linux/man-pages/man7/file-hierarchy.7.html#PERSISTENT_VARIABLE_SYSTEM_DATA

User Level: https://man7.org/linux/man-pages/man7/file-hierarchy.7.html#HOME_DIRECTORY

XDG_CACHE_HOME env var is *generally* defined inside the home directory as per the hierarchy above and is accessible at user level, tho this location might not always be guaranteed since some distros do choose their own standards, but then it's upto that unique distro not following the standard implemention to patch their own paths at packaging level.

PureTryOut: We're going with XDG_CACHE_HOME, thanks for your opinions guys!

start-mycroft.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
```
fermulator, line 1: BLOCKED by https://github.com/MycroftAI/mycroft-core/pull/2686/files

(pending that, then a rebase or merge to update - then we can review this diff a little more cleanly)

fermulator (changes requested):
blocked by https://github.com/MycroftAI/mycroft-core/pull/2686/files

PureTryOut:
I have different plans to do system-wide (and per-user) setups actually. Closing this in favour of a new PR soon to come.

PR #2689, opened 2020-09-09
Fix Audioservice startup

forslund:
## Description
Handle ntp sync in audioservice ready wait. Events (and [python threading in general](https://bugs.python.org/issue41710)) don't use monotonic time and can be affected by an ntp sync.

The log from @krisgesling shows the following

```
2020-04-01 17:24:12.190 | INFO     |   467 | __main__:main:50 | Starting Audio Services
2020-04-01 17:24:12.198 | INFO     |   467 | mycroft.messagebus.client.client:on_open:114 | Connected
2020-04-01 17:24:12.204 | INFO     |   467 | mycroft.audio.audioservice:get_services:61 | Loading services from /home/mycroft/mycroft-core/mycroft/audio/services/
2020-04-01 17:24:12.215 | INFO     |   467 | mycroft.audio.audioservice:load_services:105 | Loading chromecast
2020-09-02 02:02:49.463 | ERROR    |   467 | __main__:on_error:34 | Audio service failed to launch ('No audio services loaded').
```

Notice the timestamp difference on the two last entries. Between these the time jumped ~5 months which is longer than the 3 minute timeout on the wait. **Likely** (possibly/maybe) network connection was established during this time adjusting the time and thus ending the Event wait.

This adds logic to repeat the wait if the monotonic time don't agree with the event timeout.

## How to test
It's sort of tricky to test in a good way. Firstly make sure the audioservice starts in normal circumstances (run the service, add a delay in loading services and make sure the result of the wait is as expected).

Secondly try to repeat the actual issue using the `date` command:
- inject a long sleep (`time.sleep(300)` on L204
- launch the audio service
- run `sudo date +%T -s "HH:MM:SS"` with a suitable clock time in the future

Make sure the service doesn't exit directly, but waits the expected 3 minutes before shutting down.

### DIFF
--- a/mycroft/audio/audioservice.py
+++ b/mycroft/audio/audioservice.py
@@ -17,11 +17,12 @@ import sys
 import time
 from os import listdir
 from os.path import abspath, dirname, basename, isdir, join
-from threading import Lock, Event
+from threading import Lock
 
 from mycroft.configuration import Configuration
 from mycroft.messagebus.message import Message
 from mycroft.util.log import LOG
+from mycroft.util.monotonic_event import MonotonicEvent
 
 from .services import RemoteAudioBackend
 
@@ -152,7 +153,7 @@ class AudioService:
         self.play_start_time = 0
         self.volume_is_low = False
 
-        self._loaded = Event()
+        self._loaded = MonotonicEvent()
         bus.once('open', self.load_services_callback)
 
     def load_services_callback(self):
@@ -161,7 +162,6 @@ class AudioService:
             service and default and registers the event handlers for the
             subsystem.
         """
-
         services = load_services(self.config, self.bus)
         # Sort services so local services are checked first
         local = [s for s in services if not isinstance(s, RemoteAudioBackend)]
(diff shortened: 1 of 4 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2689))

krisgesling:
Great work on figuring this out!

Have confirmed that switching the clock forward during loading of the audioservice causes the timeout to trigger and the service loading fails. Also confirmed the fix works as expected using the methods you suggested. 

As discussed in chat, it looks like this also exists in the hotword factory so will wait for that before merging.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2689))

forslund:
Updated PR with a refactor and handling of the wait in the hotword_factory.py as well as the one in audioservice

krisgesling:
Working great, thanks again!!

PR #2705, opened 2020-09-26
Remove DSS English-only error

el-tocino:
## Description
As deepspeech now supports other languages, the error raised for English-only is not correct.

## How to test
Try using a non-english DSS instance to handle STT services.

CLA [x]

### DIFF
--- a/mycroft/stt/__init__.py
+++ b/mycroft/stt/__init__.py
@@ -316,8 +316,6 @@ class DeepSpeechServerSTT(STT):
 
     def execute(self, audio, language=None):
         language = language or self.lang
-        if not language.startswith("en"):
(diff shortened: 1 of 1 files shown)


el-tocino:
Works with german now, as well:  https://chat.mycroft.ai/community/pl/kz1x68ewrjgsx8q5punr8g9hhr

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2705))

krisgesling (approved):
I was going to suggest a list of supported languages but the DS service needs to be self hosted anyway, so presumably anyone using it will know better than we do if they have a model for their language loaded. 

This also means we don't have to keep the list updated.

PR #2707, opened 2020-09-29
Fix remote settings overwrite at startup

forslund:
## Description
This is a suggestion to improve the usability of the local settings, an implementation of an idea mentioned in #2698

Currently Mycroft always mangels any local settings changes at startup. This PR adds a cache of the last settings from home on disk and only updates skills if there is an actual change.

This means that as if a member is changed locally (manually edited or changed by the skill itself) it will be kept until another  change is made to the skill on Home.

In my opinion this is a more intuitive way of handling the changes, but it's definitely up for discussion / rejection.

## How to test
Ensure that skill updates from home works as normally. Edit a settings.json while Mycroft is turned off and make sure that the change doesn't get overwritten when mycroft starts again.

CLA [ Yes ]

### DIFF
--- a/mycroft/skills/settings.py
+++ b/mycroft/skills/settings.py
@@ -57,9 +57,11 @@ SkillSettings Usage Example:
 """
 import json
 import os
+from os.path import dirname
 import re
 from pathlib import Path
 from threading import Timer
+from xdg.BaseDirectory import xdg_cache_home
 
 import yaml
 
@@ -68,6 +70,7 @@ from mycroft.configuration import Configuration
 from mycroft.messagebus.message import Message
 from mycroft.util import camel_case_split
 from mycroft.util.log import LOG
+from mycroft.util.file_utils import ensure_directory_exists
 from .msm_wrapper import build_msm_config, create_msm
 
 ONE_MINUTE = 60
@@ -309,6 +312,42 @@ class SettingsMetaUploader:
         return success
 
 
+# Path to remote cache
+REMOTE_CACHE = Path(xdg_cache_home, 'mycroft', 'remote_skill_settings.json')
+
+
+def load_remote_settings_cache():
+    """Load cached remote skill settings.
+
+    Returns:
+        (dict) Loaded remote settings cache or None of none exists.
+    """
+    remote_settings = {}
+    if REMOTE_CACHE.exists():
+        try:
+            with open(str(REMOTE_CACHE)) as cache:
+                remote_settings = json.load(cache)
+        except Exception as error:
+            LOG.warning('Failed to read remote_cache ({})'.format(error))
+    return remote_settings
+
+
+def save_remote_settings_cache(remote_settings):
+    """Save updated remote settings to cache file.
+
+    Arguments:
+        remote_settings (dict): downloaded remote settings.
+    """
+    try:
+        ensure_directory_exists(dirname(str(REMOTE_CACHE)))
+        with open(str(REMOTE_CACHE), 'w') as cache:
+            json.dump(remote_settings, cache)
+    except Exception as error:
+        LOG.warning('Failed to write remote_cache. ({})'.format(error))
+    else:
+        LOG.debug('Updated local cache of remote skill settings.')
+
+
 class SkillSettingsDownloader:
     """Manages download of skill settings.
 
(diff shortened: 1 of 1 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2707))

JarbasAl:
this looks good, its a partial solution but ensures the settings updated event means exactly that

2 way sync is still needed however

issue clarification:
- update setting X locally
- update setting Y in web interface
- settings update replaces X

Out of scope for this PR either way, we could separate local settings from remote settings somehow, but any approach like that is just hacky, the proper fix is selene side

forslund:
I definitely agree that the two-way sync is what's needed. But in addition to that I think this is needed as well as core shouldn't send settings updated messages to all skills at startup even if they've been synced.

ChanceNCounter:
I've confirmed that it works as described:

1. Dismiss Mycroft, get feedback
2. Disable Dismissal feedback in Home
3. Ensure that the change synced, dismiss Mycroft, no feedback
4. Shut Mycroft down
5. Enable Dismissal feedback in Home
6. Start Mycroft, still no feedback from Dismissal ✅ 
7. Go back to Home, don't even touch the setting, just push "save" again
8. Ensure that the change synced, dismiss Mycroft, get feedback

The code looks straightforward, I can't see what could possibly be wrong with it, and it runs as described. Only thing I can't say is whether it'll break what's underneath that allows it to work. Where's the main change itself? I don't see anything to do with the overwriting loop.

forslund:
The main change is [L353](https://github.com/MycroftAI/mycroft-core/pull/2707/commits/8c8a210b8dc98bb7282b013dd993b05e9f0896b0#diff-3e08b7eb36a753ba228c9661639c672c4012acd93c18492f2ec48ff461fe2927R353), the settings already only request skills to reload settings if there is a difference. At above mentioned line the copy of the settings used for comparison when identifying changes is restored so the initial download after startup also has a reference.

The behavior you describe doesn't sound quite correct, gonna see if I can replicate that. A change on home while the device is turned off should trigger an update, the idea is that if there is no change on home since last fetch but a local change has been made, the local change should not be overwritten as soon as Mycroft is restarted.

forslund:
I haven't been able to replicate your behavior using my test skill, going to try the dismissal skill as well.

I did find and remove an unused member variable assignment.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2707))

ChanceNCounter:
To be honest, I thought the behavior I saw was what you intended. It made sense to me.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2707)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2707/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2707/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2707/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2707/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2707/logs/enclosure.log)

krisgesling:
Hey Chance, thanks for checking it out.

I wasn't clear from your first message of whether you tried changing the setting locally on the device?

I was thinking along the lines of

1. Dismiss Mycroft, get feedback
2. Disable Dismissal feedback in Home
3. Ensure that the change synced, dismiss Mycroft, no feedback
4. Modify Dismissal Skill settings locally to enable feedback
5. Shut Mycroft down
6. Start Mycroft - should get feedback from Dismissal showing remote settings haven't synced
7. Go back to Home, enable feedback, and save
8. On Home, disable feedback, and save
8. Ensure that the change synced, dismiss Mycroft, get no feedback

Might be even easier to see with something like the News Skill where you can switch between many default stations. Or something with a text input.

ChanceNCounter:
Thanks, Gez! I was able to confirm the behavior as described, this works and I'm not sure what happened the first time around. It may simply be that I didn't wait long enough before reporting back, and didn't notice it overwriting on sync.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2707))

krisgesling:
Excellent, thanks Chance - if you are happy to submit an approving code review I'll merge it in :)

ChanceNCounter (approved):
Working as described on Debian/dev branch.

krisgesling (approved):
Reviewed by Chance

PR #2713, opened 2020-10-01
Feature/mozilla tts

domcross:
## Description
Implement TTS module for Mozilla-TTS server

## How to test
Fire up Mozilla-TTS server, set mycroft.conf: tts.module "mozilla" and configure tts.mozilla.url, mycroft-speak "My name is Mycroft and I am funky"

## Contributor license agreement signed?
CLA [X] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

PR #2719, opened 2020-10-10
mozilla TTS validation

JarbasAl:
adds validation to mozilla TTS, if a wrong url is entered this will now be handled and fallback to default TTS engine, instead of the user simply having no TTS output 

removed unused lang param set to "de"

i messed up in posting my review of #2713 , changes have been tested at [HelloChatterbox/text2speech](https://github.com/HelloChatterbox/text2speech/blob/master/text2speech/modules/mozilla_tts.py)

### DIFF
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -308,7 +308,6 @@
       "voice": "m1"
     },
     "mozilla": {
-      "lang": "de",
       "url": "http://0.0.0.0:5002/api/tts"
     }
   },
--- a/mycroft/tts/mozilla_tts.py
+++ b/mycroft/tts/mozilla_tts.py
@@ -49,7 +49,7 @@ class MozillaTTSValidator(TTSValidator):
         pass
 
     def validate_connection(self):
-        url = self.tts.config['url']
+        url = self.tts.config['url'].replace("/api/tts", "")
         response = requests.get(url)
         if not response.status_code == 200:
             raise ConnectionRefusedError
--- a/mycroft/tts/mozilla_tts.py
+++ b/mycroft/tts/mozilla_tts.py
@@ -49,8 +49,10 @@ class MozillaTTSValidator(TTSValidator):
         pass
 
     def validate_connection(self):
-        # TODO
-        pass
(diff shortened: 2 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2719))

krisgesling:
Hey I just tried this against a local Mozilla TTS server and it seems that a non-empty text parameter is required.

Passing no params resulted in:
```
File "/home/user/TTS/.venv/lib/python3.6/site-packages/TTS/server/synthesizer.py", line 143, in split_into_sentences
    text = " " + text + "  <stop>"
TypeError: must be str, not NoneType
[INFO] 127.0.0.1 - - [12/Oct/2020 15:56:41] "GET /api/tts HTTP/1.1" 500 -
```

and passing an empty string:
```
ValueError: zero-size array to reduction operation maximum which has no identity
[INFO] 127.0.0.1 - - [12/Oct/2020 16:00:11] "GET /api/tts?text= HTTP/1.1" 500 -
```

krisgesling:
Just had a look at their [server app routes](https://github.com/mozilla/TTS/blob/master/TTS/server/server.py) and there's only two by default:
- / 
- /api/tts

So we could send a non-empty text param to the synthesizer or ping the root instead?

JarbasAl:
ups, i intended to ping the root, not the tts endpoint.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2719))

forslund:
Works exactly as advertised :+1:

Since the mozilla tts hasn't been merged to master yet, do you think the config should be changed to just the root? So in the config it would be
```
"url": "http://0.0.0.0:5002"
```

and the execute call would join the `/api/tts` to the url from the config instead of the validation removing it?

JarbasAl:
totally agree, i thought this was on master already, but if its only on dev i will make that change!

JarbasAl:
will re-open upon clarification of privacy ramifications of the latest mycroft partnership

code is free, you can copy and open a new PR if you do not want to wait for me re-opening it

PR #2725, opened 2020-10-15
allow variants of arch (archarm, ...) to fetch dependencies (dev_setup.sh)

emphasize:
Problem: Arch Linux Arm doesn't fetch dependencies in the installation process

Just added `&& ( ... || os_is_like arch)` to let variants like archarm pass the filter

since os-release (archarm)

```
ID=archarm
ID_LIKE=arch
```

Alternatively this could be handled with `os_is archarm` , yet i've seen many (core) arch variants which could run into the same problem. This should also pass Manjaro, i guess.

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -398,14 +398,14 @@ function install_deps() {
         # Fedora
         echo "$GREEN Installing packages for Fedora...$RESET"
         fedora_install
-    elif found_exe pacman && os_is arch ; then
+    elif found_exe pacman && (os_is arch || os_is_like arch); then
         # Arch Linux
         echo "$GREEN Installing packages for Arch...$RESET"
         arch_install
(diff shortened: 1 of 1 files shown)


devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2725))

forslund:
Looks like good changes, thanks. Gonna ask briefly on the chat if an arch / manjaro user can test it.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2725))

forslund:
@ChanceNCounter tested it with Manjaro and is working as expected.

forslund (approved):
And I tested that other systems are unaffected. seems to work as intended. Approved!

PR #2737, opened 2020-10-26
msm appstore warning

JarbasAl:
this adds a log warning users in case they configure a different appstore, currently this can only be done by replacing the mycroft marketplace, discussion about supporting multiple appstores can be found here https://github.com/MycroftAI/mycroft-skills-manager/pull/75

this warning should be tweaked and probably moved around once msm supports multiple appstores

This makes it clear there is no assurance of skill safety, i wonder if skill installer skill should be vocal about it instead of just logging it

### DIFF
--- a/mycroft/skills/msm_wrapper.py
+++ b/mycroft/skills/msm_wrapper.py
@@ -86,6 +86,11 @@ def create_msm(msm_config: MsmConfig) -> MycroftSkillsManager:
     especially during the boot sequence when this function is called multiple
     times.
     """
+    if msm_config.repo_url != "https://github.com/MycroftAI/mycroft-skills":
+        LOG.warning("You have enabled a third-party skill store.\n"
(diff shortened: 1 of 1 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2737))

krisgesling:
Will mark this as for discussion while that happens over in the linked PR in MSM

JarbasAl:
regardless of discussion in other thread i think this is a good addition, we do want this warning no? 

it is not depending on the outcome of the other discussion, if anything it might need a refactor later (and that later does not seem to be anytime soon, too much stuff to be implemented before we get to that point)

this adds the warning to the current mechanism to change appstores, the discussion is about improving/expanding that mechanism

krisgesling (approved):
Ah my bad. Yeah I think this is good. Thanks!

PR #2738, opened 2020-10-26
initial release of sj201 rev B hardware support. note this change als…

ken-mycroft:
…o requires the mark2 skill init update as well

## Description
This PR adds Mark2 hardware support for the sj201 revision B which entirely uses the usb port to control all devices

## How to test
LEDs, Switches and volume all work

Yes


devops-mycroft:
Hello, @ken-mycroft, thank you for helping with the Mycroft project! We welcome everyone
for Everyone.

you!

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2738))

forslund:
The mycroft.enclosure module was designed as an interface towards the enclosure process. Should this be implemented in the enclosure and just the interface be offered in the /mycroft/enclosure? Like with the mark-1 and gui parts?

I'm also curious about the sudo, generally use of sudo from an application should be avoided. (I can't see the software called through sudo so I can only speculate about the reason) I imagine some gpio or usb or i2c device is used? If this is the case the pi user should be granted rights to the device instead of using sudo. This can in many cases be done using a fairly simple udev rule.

Is `sj201_revB/vfctrl_usb` available somewhere?

j1nx:
> Nice seeing the SJ201 added 👍
> 
> The `mycroft.enclosure` module was designed as an interface towards the enclosure process. Should this be implemented in the enclosure and just the interface be offered in the /mycroft/enclosure? Like with the mark-1 and gui parts?
> 
> I'm also curious about the sudo, generally use of sudo from an application should be avoided. (I can't see the software called through sudo so I can only speculate about the reason) I imagine some gpio or usb or i2c device is used? If this is the case the pi user should be granted rights to the device instead of using sudo. This can in many cases be done using a fairly simple udev rule.
> 
> Is `sj201_revB/vfctrl_usb` available somewhere?

I said the same the other day. It looks like they choose the neopixel ring for the leds connected to GPIO 18 on the SJ201 which indeed can't be controlled without root privileges.

If they would re-route that to GPIO 10 (pin 19) they could use the same SPI interface and by adding the user to the right groups via udev. But apparently that route isn't really stable and act funny.
https://raspberrypi.stackexchange.com/questions/85109/run-rpi-ws281x-without-sudo

Not sure but believe @ken-mycroft already searched the internet extensivily so assume he is aware.

krisgesling:
Closing in favor of newer PR

PR #2749, opened 2020-11-06
Fix readthedocs build

forslund:
## Description
After readthedocs was fixed enough to update again a couple of issues were observed, the most obvious one that the docs for MycroftSkill derived classes are missing. See [here](https://mycroft-core.readthedocs.io/en/latest/source/mycroft.html#mycroftskill-class-base-class-for-all-mycroft-skills)

This was due to `xdg` not being mocked when generating the docs, and thus erroring out when sphinx imported those modules. This adds `xdg` as a `autodoc_mocked_import`.

Finding this out I realized the same isse ẃas causing the previous issue with `yaml` so that has been cleaned up as well.

I also noticed that the `.readthedocs` file was not named correctly and wasn't actually used. This renames it and fixes an incorrect path making it usable.

## How to test
I generated readthedocs from this branch [here](https://mycroft-core.readthedocs.io/en/bugfix-readthedocs/) Make sure the MycroftSkills module and related modules are generated properly.

CLA [ Yes ]

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -16,4 +16,4 @@ formats: all
 python:
   version: 3.7
   install:
-    - requirements: docs/requirements.txt
+    - requirements: doc/requirements.txt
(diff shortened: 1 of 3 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2749))

krisgesling (approved):
You're awesome, thanks!

PR #2766, opened 2020-11-21
self.gui.release()

JarbasAl:
Signal that this skill is no longer using the GUI, allow different platforms to properly handle this event.
     
Also calls self.clear() to reset the state variables

Currently self.gui.clear() has been wrongly used for this purpose, for example in my [evil-laugh-skill](https://github.com/JarbasSkills/skill-laugh/blob/master/__init__.py#L86) this worked ok if mark2 skill was installed (due to a different bug related to idle screen) which makes it seem like its working, but in bigscreen it causes a black screen to linger around

Platforms can close the window or go back to previous page, clearing the page data is not enough

more recently in https://github.com/MycroftAI/mycroft-timer/pull/82 self.gui.clear() is also being wrongly used for this purpose

image

### DIFF
--- a/mycroft/enclosure/gui.py
+++ b/mycroft/enclosure/gui.py
@@ -349,6 +349,15 @@ class SkillGUI:
         self.show_page("SYSTEM_UrlFrame.qml", override_idle,
                        override_animations)
 
+    def release(self):
+        """Signal that this skill is no longer using the GUI,
+        allow different platforms to properly handle this event.
+        Also calls self.clear() to reset the state variables
+        Platforms can close the window or go back to previous page"""
+        self.clear()
+        self.skill.bus.emit(Message("mycroft.gui.screen.close",
+                                    {"skill_id": self.skill.skill_id}))
+
     def shutdown(self):
         """Shutdown gui interface.
 


JarbasAl:
alternative ways to accomplish this would be to self.gui.remove_page manually for all pages, but it seemed weird and not intuitive, pages are currently not tracked by SkillGui class and this approach would require a bigger refactor and wouldn't give different platforms as much freedom. 

A bus message also provides an event that any other process can react to so i consider it a better approach

@AIIX can provide more context on what the emitted bus message does, since it is already handled by Bigscreen and Android i believe

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2766))

AIIX:
I like the idea of standardizing a method to tell the GUI to stop showing a screen and let the platform skill control the events that take place when it receives the "mycroft.gui.screen.close" event, for the Mark-2 skill I basically see it replacing the event on the following line https://github.com/MycroftAI/skill-mark-2/blob/master/__init__.py#L256 which restores the idle screen on receiving this standardized message and for other platforms this standardized "mycroft.gui.screen.close" message can be reacted to differently as per their own platform skills. 

self.gui.clear method should only be used for clearing session data for a particular skill apparently it should also ignore __override flags when clearing the session data, as this allows for skill displaying the current page to be able to wipe all the session data slate clean and push in new session data if required, without removing the displayed page, maybe this should be renamed to self.gui.clear.data to bring in more clarity (a larger change for another PR if its currently used at to many places)

forslund:
Really good suggestion and nice and simple implementation!

krisgesling:
Yeah nice one.

I'm wondering if we should call it `self.gui.close()` to keep it consistent with the bus message?

Any other arguments for / against either - release vs close?

JarbasAl:
close sounds like you are closing the actual GUI, but the GUI will still be available

i think release matches better what it actually does. Since most likely you want the previous "active skill" to take over or if there isnt another skill go to desktop / default screen

In the case of an individual voice app, eg, `mycroft-gui-app --hideTextInput --skill=peertube-skill.aiix.home` this is the same as closing the window and showing the desktop. This would be the usage in [bigscreen](https://github.com/AIIX/skill-bigscreen-platform)

But if you are just running the GUI in a [mark2](https://github.com/MycroftAI/skill-mark-2) this is not equivalent to closing. You want the [idle screen to take over](https://github.com/MycroftAI/skill-mark-2/blob/master/__init__.py#L129) instead

In android you would want to return to the [homescreen](https://github.com/AIIX/AndroidHomescreen) which shows the icons of all apps (but that homescreen is also running in the GUI)


Maybe we could change the message if that makes sense, but @AIIX would need to handle that in all the platforms linked above, and also in the [gui protocol spec](https://github.com/MycroftAI/mycroft-gui/blob/master/transportProtocol.md)

AIIX:
> Maybe we could change the message if that makes sense, but @AIIX would need to handle that in all the platforms linked above, and also in the [gui protocol spec](https://github.com/MycroftAI/mycroft-gui/blob/master/transportProtocol.md)

I don't think the emitted message should be changed as its already used internally within mycroft-gui for global back feature and several other places, "mycroft.gui.screen.close" is pretty easy to make sense of that your targeting exactly the mycroft.gui namespace with screen.close action requested from a specific skill_id.

JarbasAl:
@AIIX is right, the "skill_id" in the message is pretty unambiguous "close this skill screen"

i do prefer self.gui.release for reasons described above, when used in skills it's more ambiguous than in the message

krisgesling:
If no one else feels strongly, I think `release` is good. 

Consistency can reduce complexity, but not if it makes less sense in that context. So I think we go with whatever makes the most sense in the context of Skill developers.

I have also proposed some extra wording for `clear` in the PR linked above to redirect dev's to this new method.

krisgesling (approved):
Tested and working great, thanks again Jarbas!

PR #2772, opened 2020-12-08
Upgrade to Lingua Franca 0.3+

ChanceNCounter:
**WIP**

Upgrades Lingua Franca to v0.3.1 (current pypi release.) This includes a major refactor, in which @JarbasAl has reorganized and standardized naming conventions. After that, I overhauled the way Lingua Franca handles localized code. The lib now crawls itself whenever a module is imported, caches localized functions in loaded languages, and imports them dynamically.

Furthermore (as implied above) Lingua Franca no longer loads any of its important functions by default. Instead, a given language's functions must be loaded by calling `lingua_franca.load_language(<lang code>)`, or by passing a list of language codes to `lingua_franca.load_languages()`. Lastly, one can change the "default" language - the one which Lingua Franca will use if a language is not specified when you call a function - using `lingua_franca.set_default_lang(<lang code>)`. This will load the specified language, if it isn't already loaded.

This PR adds or alters Mycroft-core's wrappers around certain LF functions, to reflect the fact that passing `lang=None` to parsers and formatters is deprecated, and will be unsupported in a future version. Skill authors are encouraged to omit the lang parameter entirely, so that their skills will always use the currently-set default language.

At the moment, up to two languages are loaded: the user's configured language from mycroft.conf, and English. Both are loaded along with the Skill Manager. The Skill Manager was chosen simply because it's a component that *will* be running before Mycroft has the opportunity to call other LF functions.

### DIFF
--- a/mycroft/skills/__main__.py
+++ b/mycroft/skills/__main__.py
@@ -38,7 +38,7 @@ from mycroft.util import (
     reset_sigint_handler,
     wait_for_exit_signal
 )
-from mycroft.util.lang import set_active_lang
+from mycroft.util.lang import set_default_lang
 from mycroft.util.log import LOG
 from .core import FallbackSkill
 from .event_scheduler import EventScheduler
@@ -191,7 +191,7 @@ def main(ready_hook=on_ready, error_hook=on_error, stopping_hook=on_stopping,
     mycroft.lock.Lock('skills')
     config = Configuration.get()
     # Set the active lang to match the configured one
-    set_active_lang(config.get('lang', 'en-us'))
+    set_default_lang(config.get('lang', 'en-us'))
 
     # Connect this process to the Mycroft message bus
     bus = _start_message_bus_client()
--- a/mycroft/skills/intent_service.py
+++ b/mycroft/skills/intent_service.py
@@ -17,7 +17,7 @@ from copy import copy
 import time
 
 from mycroft.configuration import Configuration
-from mycroft.util.lang import set_active_lang
+from mycroft.util.lang import set_default_lang
 from mycroft.util.log import LOG
 from mycroft.util.parse import normalize
 from mycroft.metrics import report_timing, Stopwatch
@@ -149,7 +149,7 @@ class IntentService:
     def reset_converse(self, message):
         """Let skills know there was a problem with speech recognition"""
         lang = _get_message_lang(message)
-        set_active_lang(lang)
+        set_default_lang(lang)
         for skill in copy(self.active_skills):
             self.do_converse(None, skill[0], lang, message)
 
@@ -269,7 +269,7 @@ class IntentService:
         """
         try:
             lang = _get_message_lang(message)
-            set_active_lang(lang)
+            set_default_lang(lang)
 
             utterances = message.data.get('utterances', [])
             combined = _normalize_all_utterances(utterances)
--- a/mycroft/skills/skill_manager.py
+++ b/mycroft/skills/skill_manager.py
@@ -28,6 +28,8 @@ from .settings import SkillSettingsDownloader
 from .skill_loader import SkillLoader
 from .skill_updater import SkillUpdater
 
+from lingua_franca import load_languages, set_default_lang
+
 SKILL_MAIN_MODULE = '__init__.py'
 
 
@@ -41,6 +43,7 @@ class UploadQueue:
     After all queued settingsmeta has been processed and the queue is empty
     the queue will set the self.started flag.
     """
+
     def __init__(self):
         self._queue = []
         self.started = False
@@ -139,6 +142,9 @@ class SkillManager(Thread):
         self.skill_updater = SkillUpdater()
         self._define_message_bus_events()
         self.daemon = True
+
+        self.lang_code = self.config.get("lang", 'en-us')
+        load_languages([self.lang_code, 'en-us'])
 
     def _define_message_bus_events(self):
         """Define message bus events with handlers defined in this class."""
--- a/mycroft/util/format.py
+++ b/mycroft/util/format.py
@@ -29,122 +29,15 @@ import warnings
 from calendar import leapdays
 from enum import Enum
 
-import lingua_franca.format
+from lingua_franca import get_default_lang
 # These are the main functions we are using lingua franca to provide
 from lingua_franca.format import (NUMBER_TUPLE, DateTimeFormat, join_list,
                                   date_time_format, expand_options,
-                                  _translate_word)
+                                  _translate_word,
+                                  nice_number, nice_time, pronounce_number,
+                                  nice_date, nice_date_time, nice_year)
 
 from padatious.util import expand_parentheses
-
-
-def nice_number(number, lang=None, speech=True, denominators=None):
-    """Format a float to human readable functions
-    This function formats a float to human understandable functions. Like
-    4.5 becomes 4 and a half for speech and 4 1/2 for text
-    Args:
-        number (int or float): the float to format
-        lang (str): code for the language to use
-        speech (bool): format for speech (True) or display (False)
-        denominators (iter of ints): denominators to use, default [1 .. 20]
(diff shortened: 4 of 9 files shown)


ChanceNCounter:
yeap. i'm about ready to kill my git extension.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2772))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2772))

ChanceNCounter:
Per MycroftAI/lingua-franca#174, before moving to LF0.4 (unreleased but ready) a setting should be added to `mycroft.conf` toggling `lingua_franca.config.load_langs_on_demand`.

In the long run, when language switching is more robust, I think it would be better if the Skill API provided for "fallback languages." The issue was raised by a user who upgraded to LF0.3.1 against *current* mycroft-core/dev, and encountered a hardcoded fallback to English in the weather skill.

This PR currently works around the problem by simply loading English alongside the user's primary language. If Mycroft wants to support falling back to languages *other* than English, the Skill API will need to adapt; in the meantime, it's only possible with `load_langs_on_demand`.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2772)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2772/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2772/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2772/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2772/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2772/logs/enclosure.log)

r0d0dendr0n:
I confirm that applying this commit gives Mycroft a proper fallback to English, without the need to override `load_langs_on_demand` in LF. This fixes the issues with skills having hardcoded English values (like the weather skill). I believe that loading English as a hardcoded default is generaly a good idea, because all the skills I've seen are written in English. It might be better though to give a configuration option to load a list of languages.

I am currently developing a skill that I use to play with my son. It speaks the sound of an animal we name (https://github.com/r0d0dendr0n/skill-animal-say). It's a work in progress, but Polish version is quite functional. Because the functions list is not closed, I don't want do too much translating at the moment. This decision however makes it harder to share this skill with anyone not native to English or Polish, because their Mycroft would be set to their native language and English. Even if they know Polish, they won't be able to use it without changing Mycroft's primary language.

ChanceNCounter:
Yep. Stuff like that comes up a lot. These changes to Lingua Franca were partially intended to facilitate that, but they're only one part of the puzzle.

The `load_language()` and `set_default_language()` workflow slots nicely into a hypothetical algorithm that loads or unloads whole STT and TTS models, wake words, and stuff like that, which gets you as far as changing your default language on the fly.

Trying to fall back on languages other than English *first*, and sort of go down a list, is an important notion for situations where skills are supported in what might be an "intermediate" language for the user (this most recently came up with respect to Catalan speakers, most of whom live in Spain, and are therefore more likely to understand Castillian Spanish than English.

But you begin with an interesting problem and add more problems from there: how do you make it easy for users to set up STT and TTS modules on a per-language basis, config-wise? Do you limit this functionality to STT and TTS engines that can support multiple languages, or use different engines where one is more suitable? How jarring would it be for Mycroft to speak in a completely different voice when you ask it a question in Polish versus English?

And how do you handle that fallback stuff? Should Mycroft start by trying to parse STT in the default language, then try the next language if it detects junk? Should it try them all every time, returning confidence values? That'll either be slow or computationally expensive.

These are solvable problems, but there are a lot of moving parts, only some of which are actually made by MycroftAI or this community. I'd certainly encourage you to join the neverending brainstorm over there on Mattermost.

r0d0dendr0n:
I think that the user should provide a configuration with priorities to TTS and STT engines, along with their supported languages. Something like this:
`"tts": {
    "mimic: {
        "priority": 1,
        "lang": ["en-us", "pl-pl", "de-de"],
        "charset": "UTF-8",
        "path": "..."
    },
    "my_http_service": {
        "priority": 2,
        "lang": ["pl-pl"],
        "charset": "UTF-8",
        "url": "...."
    }
}`
This way Mycroft could try to utilize multiple TTS engines, sorted by their priority weight. If my http service would give an error for Polish text, Mimic would take over. For an English text Mimic would be used first. This is only an example. Another one is that we could provide a priority list of engines language wise, independently from engines. This way we could say for pl-pl use A, B, C, for en-en use B, C, D and for de-de use A, C. Of course it would not be consistent to provide different voices, but I guess that it's a low price and would not be mandatory.
It would be slow, but now if we get TTS errors, neither we have any speech.

STT is more tricky than TTS. I don't have a "the way to go" solution but I can propose some fantasies.
# Give a skill to switch to a different STT ("Hey Mycroft, speak Spanish from now on." "Si Senior" :))
# Use some neural network to distinguish users by their voice sample and respond in their preconfigured language. (This does not rule out the previous one.)
# YouTube recognizes my speech when I use Polish or English... so maybe leave it up to the engine? But the engine could provide text and the recognized language. Of course there is a problem with multilingual sentences, like "Hey Mycroft, what is a Krankenwagen".

I guess I should check out Mycroft's Mattermost.

krisgesling:
Hey Chance, I get one other error on this for extracting timezone data. What do you think of comparing datetime objects created from the tzinfo instead of the tz itself?

```Python
    def test_extract_timezone(self):
        """Check that extract_datetime returns the expected timezone."""
        default_tz = default_timezone()
        dt, _ = extract_datetime("today")
        # As default_timezone() returns tzlocal() and extract_datetime() may
        # return a tzfile we cannot directly compare dt.tzinfo to default_tz.
        self.assertEqual(dt, dt.astimezone(default_tz))
```

I'm also going to close and re-open this to trigger the new tests.

krisgesling:
Just remembered the Python3.5 issues. Wondering if we abandon the attempt to get it in 20.08. 

We can cut a 20.08.1 release and then we can start merging in breaking changes.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2772)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2772/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2772/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2772/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2772/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2772/logs/enclosure.log)

ChanceNCounter:
Yeah, I think straight to 0.4.0 is a good idea. I don't mind holding the release, but there are languages waiting on 0.4, so the sooner the better at this point.

I can't speak to the error because I have absolutely no idea what I'm looking at:

```
Failing scenarios:

  features/date.feature:64  when is a date in the future -- @1.6 when is a date

JarbasAl:
the catalan community did a sprint hoping to get support for their language, they translated everything in less than 1 month, i feel we shouldn't tell them to hold out 8 more months until next release...

krisgesling:
Hey to clarify, I mean do the v20.8.1 release, then add this for 21.2.0 where we have officially dropped support for Python 3.5

Definitely not waiting a whole extra release cycle.

JarbasAl:
> Hey to clarify, I mean do the v20.8.1 release, then add this for 21.2.0 where we have officially dropped support for Python 3.5
> 
> Definitely not waiting a whole extra release cycle.

right, i sometimes get confused with the versioning scheme..... if it makes it into the next major release that would be great :+1:

ChanceNCounter:
Close in favor of #2890

PR #2776, opened 2020-12-09
move speech listener visualizer from mark-2 skill to speech client

AIIX:
## Description
Moves the speech visualizer code from mark-2 to the speech client for improving performance of the animations

## How to test
Disable all the listner volume visualizer code in mark-2 skill
Requires: https://github.com/MycroftAI/mycroft-gui-mark-2/tree/feature/mart/volumeinshell
Requires: https://github.com/MycroftAI/mycroft-gui/tree/feature/mart/volumeinshell


forslund:
If the level signal should be generated here I suggest extending the mic object to send the rms data instead of adding a separate thread + mic stream.

forslund:
You should maybe even be able to use the sound data from [here](https://github.com/MycroftAI/mycroft-core/blob/6f8bae6ba136c9dd66ca47aaadd75e214d006190/mycroft/client/speech/mic.py#L446) or even the calculated "energy" a couple of lines below.

forslund:

PR #2778, opened 2020-12-13
fa-ir localization initialized

HKalbasi:
## Description
I have translated strings in `res/text`.
## How to test
Currently it does not work. I tested this with this config:
```
{
  "lang": "fa-ir",
  "tts": {
    "espeak": {
      "lang": "persian",
      "voice": "m1"
    }
  },
  "max_allowed_core_version": 20.8
}
```
Mycroft react to `Hey Mycroft` and print my Farsi words in terminal (means stt part works) but no response, neither in terminal nor in voice (I expect some response like "I don't understand"). Maybe I should translate some basic skills?

Waiting for email.


krisgesling:
Hey there, thanks for contributing to Mycroft!

If you haven't seen it already, this page in our docs outlines all the pieces that need to be implemented to use Mycroft in another language: https://mycroft-ai.gitbook.io/docs/using-mycroft-ai/customizations/languages

In short, to respond in Persian the Skills would also need to translated. I've just added it to our [translate.mycroft.ai](https://translate.mycroft.ai/fa/) platform though it will take a few hours to populate. 

I'm interested to see how our intent parsers go, though I don't foresee any big issues. You can test it out by adding the relevant language directory into any Skill under `/opt/mycroft/skills/skill-name`. In each Skill will be one or more directories named: `dialog`, `locale`, and `vocab`. Within these directories are the language specific folders.

In terms of the CLA, I haven't seen anything come through. Can you email me directly at kris.gesling @ our domain? To kick that off, I just need your name, email address and whether you're contributing as an individual or through an employer.

krisgesling:
Hey there - I've just submitted a PR to include the Farsi support you added to Lingua Franca - are you able to re-test this and let us know if it's ready to go?

Thanks again for all the work you've put into adding a new language!

krisgesling:
Given this only adds translations, doesn't change any code, and we have Farsi support in Lingua Franca. I think it's worth merging. If something is broken, it's not because of these strings and the configurations.json looks right too.

I'm going to close and re-open the PR to trigger the new status checks. If anyone doesn't think this should be merged, please say so.

mycroft/res/text/fa-ir/configurations.json
```diff
@@ -0,0 +1,33 @@
+{
+  "unit system": {
+    "metric": {"system_unit": "metric"},
+    "imperial": {"system_unit": "imperial"}
+  },
+  "location": {
+    "stockholm": {
```
JarbasAl, line 7: this whole file doesn't look right

HKalbasi: What it should be? I just copy pasted it from `en-us` and other languages seems doesn't have it.

ChanceNCounter: That file offers defaults related to the Voight-Kampff test system. Farsi doesn't need a copy.

krisgesling: Yeah, it doesn't hurt to have it, but it also won't be used. It's been removed now.

Merging.

krisgesling:
:disappointed:

I stupidly did not see that this was against the master branch

I don't think this will cause any issues with our release process. 
@chrisveilleux @forslund what do you think?
Is it worth reverting or should we just leave it alone and add this to the dev branch too?

I've also updated our branch protections to only allow merges to master by the account that runs our CI processes. So theoretically it should prevent me from doing this ever again.

ChanceNCounter:
It looks like the incoming branch was also forked off of master, so if it's PRed to dev, it'll bring the 500,000 circular backmerges along with it.

I think the best way forward is to rebase @HKalbasi's 4 commits onto dev. As for master, I have no skin in the game, but I'd set fire to it and force-push something functional. Whatever consequences that might have for tooling can't possibly be as time-consuming to fix as what happens when somebody forks master, which is a natural thing for many devs to do.

I'm not saying master should be kept in sync, but if master weren't a mess, the rebase after this happens would be less painful, and might not always be necessary.

krisgesling:
Yeah the master history thankfully wasn't my fault. That was before I joined the team. We don't want to mess with the history of master as it can have very strange consequences for a whole range of people.

I'd be cherry-picking the squashed commit onto dev so that it remained consistent with the history of master, and should merge cleanly as it's only adding files.

forslund:
Cherry-picking it to dev like that should work but you probably want to do a test run of the release-script locally to make sure it's still clean. if it's not clean maybe a revert of the PR to master would work.

PR #2781, opened 2020-12-18
Fix missing remove page implementation, show idle if all pages removed

AIIX:
## Description
- Added missing check for "remote" and "system" page urls in remove pages GUI API
- Added check and emit show idle screen if current page position is 0 and no other pages are left in the namespace, ensures the idle screen is activated for the mark-2 enclosure when all pages are removed from the skill

## How to test
 ```
# Add a page in some function
self.gui.show_page("somepage.qml")
self.gui.show_page("somepage2.qml")
self.gui,remove_page("somepage2.qml")
self.gui.remove_page("somepage.qml")

# if all pages are removed it should go back to idle screen for platform mycroft_mark_2
```

### DIFF
--- a/mycroft/client/enclosure/base.py
+++ b/mycroft/client/enclosure/base.py
@@ -240,10 +240,9 @@ class Enclosure:
                    })
         # Remove the page from the local reprensentation as well.
         self.loaded[0].pages.pop(pos)
-        # Add a check to return to idle from position 0
-        if (pos == 0 and self.config.get("platform") == "mycroft_mark_2"
-                and len(self.loaded[0].pages) == 0):
-            self.bus.emit(Message("mycroft.mark2.reset_idle"))
+        # Add a check to return any display to idle from position 0
+        if (pos == 0 and len(self.loaded[0].pages) == 0):
+            self.bus.emit(Message("mycroft.display.reset_idle"))
 
     def __insert_new_namespace(self, namespace, pages):
         """ Insert new namespace and pages.
--- a/mycroft/client/enclosure/base.py
+++ b/mycroft/client/enclosure/base.py
@@ -240,6 +240,10 @@ class Enclosure:
                    })
         # Remove the page from the local reprensentation as well.
         self.loaded[0].pages.pop(pos)
+        # Add a check to return to idle from position 0
+        if pos == 0 and self.config.get("platform") == "mycroft_mark_2" and
+        len(self.loaded[0].pages) == 0:
+            self.bus.emit(Message("mycroft.mark2.reset_idle"))
 
     def __insert_new_namespace(self, namespace, pages):
         """ Insert new namespace and pages.
--- a/mycroft/client/enclosure/base.py
+++ b/mycroft/client/enclosure/base.py
@@ -242,7 +242,7 @@ class Enclosure:
         self.loaded[0].pages.pop(pos)
         # Add a check to return any display to idle from position 0
         if (pos == 0 and len(self.loaded[0].pages) == 0):
-            self.bus.emit(Message("mycroft.display.reset_idle"))
+            self.bus.emit(Message("mycroft.device.show.idle"))
 
     def __insert_new_namespace(self, namespace, pages):
         """ Insert new namespace and pages.
--- a/mycroft/client/enclosure/base.py
+++ b/mycroft/client/enclosure/base.py
@@ -241,8 +241,8 @@ class Enclosure:
         # Remove the page from the local reprensentation as well.
         self.loaded[0].pages.pop(pos)
         # Add a check to return to idle from position 0
-        if pos == 0 and self.config.get("platform") == "mycroft_mark_2" and
-        len(self.loaded[0].pages) == 0:
+        if (pos == 0 and self.config.get("platform") == "mycroft_mark_2"
+                and len(self.loaded[0].pages) == 0):
             self.bus.emit(Message("mycroft.mark2.reset_idle"))
 
     def __insert_new_namespace(self, namespace, pages):
--- a/mycroft/enclosure/gui.py
+++ b/mycroft/enclosure/gui.py
@@ -228,9 +228,15 @@ class SkillGUI:
         # Convert pages to full reference
         page_urls = []
         for name in page_names:
-            page = self.skill.find_resource(name, 'ui')
+            if name.startswith("SYSTEM"):
+                page = resolve_resource_file(join('ui', name))
+            else:
+                page = self.skill.find_resource(name, 'ui')
             if page:
-                page_urls.append("file://" + page)
+                if self.config.get('remote'):
+                    page_urls.append(self.remote_url + "/" + page)
+                else:
+                    page_urls.append("file://" + page)
             else:
                 raise FileNotFoundError("Unable to find page: {}".format(name))
 

### CODE REVIEW

**Comment by devops-mycroft** [comment]
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2781))

**Comment by devops-mycroft** [comment]
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2781))

File: mycroft/client/enclosure/base.py [hunk 1]
```diff
@@ -240,6 +240,10 @@ def __remove_page(self, namespace, pos):
                    })
         # Remove the page from the local reprensentation as well.
         self.loaded[0].pages.pop(pos)
+        # Add a check to return to idle from position 0
+        if (pos == 0 and self.config.get("platform") == "mycroft_mark_2"
+                and len(self.loaded[0].pages) == 0):
+            self.bus.emit(Message("mycroft.mark2.reset_idle"))
```
krisgesling, line 246: Should this be more generic for anything with a GUI, rather than just the Mark II?

mycroft/enclosure/gui.py
```diff
@@ -228,9 +228,15 @@ def remove_pages(self, page_names):
         # Convert pages to full reference
         page_urls = []
         for name in page_names:
-            page = self.skill.find_resource(name, 'ui')
+            if name.startswith("SYSTEM"):
```
krisgesling, line 231: The changes to this file seem unrelated, were they meant to be included in this PR? or is it a separate thing?

krisgesling (commented):
Hey, thanks for this AIIX. I've been noticing this behaviour too.

mycroft/client/enclosure/base.py
AIIX: since there is no other platform that uses the idle, resting screen except mark II would keep it this way until there is another platform that uses this, but if we want to make it more generic then we can have something more on the lines of "mycroft.devices.reset_idle", will need an added event here (https://github.com/MycroftAI/skill-mark-2/blob/d81aeadd1efcec9309e8d2b5e4519f611e992be9/__init__.py#L221) for it the same event listener as "mycroft.mark2.reset_idle" just with the change message type. Also the issue I see currently with the generic method is that we don't know which platforms / enclosure / devices support an idle screen and which don't.

mycroft/enclosure/gui.py
AIIX: Remove pages had an incomplete implementation and this fits fine with the PR, remove pages was not looking for the correct paths for SYSTEM_html.qml and any other qml file that is handled by resources in core, this fixes that issue

mycroft/client/enclosure/base.py
forslund: Feels a bit weird to have a platform flag in the base class.

Could we instead make a method that the Mark2 enclosure overrides so the code is

```python
        if pos == 0 and and len(self.loaded[0].pages) == 0:
            self.on_empty_namespace()
```

then in the base enclosure just implement that as an empty method

```python
    def on_empty_namespace(self):
        """Override to add special action when a namespace is empty."""
```

And in the mark2 enclosure have the bus emit
```python
    def on_empty_namespace(self):
        """Go to idle screen."""
        self.bus.emit(Message("mycroft.mark2.reset_idle"))
```


Alternatively add a special message on empty namespace that the Mark-2 skill can capture and perform the same action as on the "mycroft.mark2.reset_idle" message...

krisgesling: Could we emit a `mycroft.gui.reset_idle` message and then we'll add that into the Mark 2 enclosure (or the Skill for now)?

Any systems that don't have a handler for this message will just disregard it.

AIIX: mycroft.gui is reserved namespace for transport protocol so I wouldn't use it for targeting skills, I have added it under "mycroft.display.reset_idle" since we are targeting a display and also removed any platform checks for mk2, any platform displaying the idle screen can now listen to this event when all pages have been removed to reset to idle.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2781))

mycroft/client/enclosure/base.py
krisgesling: Familiarising myself with these bus messages types - I think we want to "show" idle rather than reset it. Based off the current usage, resetting / restoring suggests removing any idle override back to the default idle screen. Whilst I assume we just want to show whatever is registered as the current idle screen.

I think `mycroft.display.show_idle` would probably be best. I could also see an argument for `mycroft.display.idle.show`.

Certainly outside of this PR, but why isn't the idle screen handled by the GUI framework itself? I'm presuming this is because it's not wanted in other implementations like Big Screen - hence added where it's needed?

krisgesling: Actually there's also an existing bus message being used `mycroft.device.show.idle`. We should probably use this instead. If we want to change the message structure in the future we would need to update this anyway.

krisgesling: but everything is working great with the fixes - including removal of the `SYSTEM` pages :+1:

AIIX: > Familiarising myself with these bus messages types - I think we want to "show" idle rather than reset it. Based off the current usage, resetting / restoring suggests removing any idle override back to the default idle screen. Whilst I assume we just want to show whatever is registered as the current idle screen.
> 
> I think `mycroft.display.show_idle` would probably be best. I could also see an argument for `mycroft.display.idle.show`.
> 
> Certainly outside of this PR, but why isn't the idle screen handled by the GUI framework itself? I'm presuming this is because it's not wanted in other implementations like Big Screen - hence added where it's needed?

There is no concept of idle screen in gui protocol, the protocol only provides communication mechanism and methods to display qml pages, idle screen is a concept on mycroft core gui framework side that is handled by the mark-2 skill and the resting screen class, this resting screen class could have ideally lived in the enclosure code being configurable in some way by a platform if it should use it or not (in case of mobile / bigscreen where it can be configured to be turned off) because its not useful in environments that aren't a smart speaker or only a dedicated mycroft device, but @forslund might know better about its implementation.

AIIX: > mycroft.device.show.idle

merge request updated, latest commit now uses mycroft.device.show.idle

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2781)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2781/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2781/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2781/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2781/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2781/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2781))

krisgesling (approved):

mycroft/client/enclosure/base.py
krisgesling: Thanks, that does make sense. 

We will shift the resting screen class across to the enclosure, along with most of the Mark 2 Skill I think.

PR #2783, opened 2020-12-19
Upgrade requests to 2.25.1

PureTryOut:
## Description
Upgrades requests to 2.25.1

## How to test
Run the unit tests.
Verify with a clean installation that all things that get downloaded still work, and network connectivity is still correct.

CLA [x]


JarbasAl:
cant this be relaxed? some other package/skill will eventually require a lower requests version

i would say the version should be lowered not increased, and versions should be pinned with >= not ==

this is a cleanup that should be done across all requirements imho

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2783))

krisgesling:
I think we need a broader discussion on how these should be handled. In the past we've had things randomly break from newer package incompatibility so I'm not sure that I'm keen on open ended >= definitions. 

It does seem like we could allow flexibility within known working minor versions, eg: `requests>=2.20.0,<2.26.0`

Given we have the testing in a much better place now, we should also be able to do this in a more automated fashion, testing new dependency versions as they're released against not just core, but all Skills (and in future plugins) in the Marketplace.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2783))

forslund (approved):

PR #2788, opened 2020-12-30
Do not download Precise if executable already exists

krisgesling:
This is partially a conversation starter unless it really turns out to be this simple.

Currently we repeatedly download Precise on first boot unless both the Precise engine tarball and extracted contents are present. IMO Precise should be preloaded with any distribution of Mycroft. If Precise is not present or fails, it can still fall back PocketSphinx but then clearly something has gone very wrong.

If the Precise engine is going to be updated, then this should be provided in an update of the Mycroft distribution.

I am however assuming that there are edge cases or side effects that I'm not considering. So very interested in why this PR is a bad idea?

## How to test
- Delete the Precise tarball at `~/.mycroft/precise/precise-engine_*.tar.gz`
- With the Precise engine executable in place, mycroft-core should not attempt to download Precise.
- Remove `~/.mycroft/precise/precise-engine/precise-engine`

### DIFF
--- a/mycroft/client/speech/hotword_factory.py
+++ b/mycroft/client/speech/hotword_factory.py
@@ -232,6 +232,8 @@ class PreciseHotword(HotWordEngine):
     def update_precise(self, precise_config):
         """Continously try to download precise until successful"""
         precise_exe = None
+        if exists(self.install_destination):
+            precise_exe = self.install_destination
         while not precise_exe:
             try:
                 precise_exe = self.install_exe(precise_config['dist_url'])
@@ -240,11 +242,8 @@ class PreciseHotword(HotWordEngine):
             except Exception as e:
                 LOG.error(
                     'Precise could not be downloaded({})'.format(repr(e)))
-                if exists(self.install_destination):
-                    precise_exe = self.install_destination
-                else:
-                    # Wait one minute before retrying
-                    sleep(60)
+                # Wait one minute before retrying
+                sleep(60)
         return precise_exe
 
     @property


forslund:
Not a bad idea, as long as we don't expect it to upgrade precise.

Personally I think we should rip out all the update code and just do it in dev_setup.sh and in the mycroft-core packaging.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2788)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2788/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2788/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2788/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2788/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2788/logs/enclosure.log)

forslund:
[Here](https://github.com/forslund/mycroft-core/commit/00a2eed81ea5a2c4a1c58b5981a59164ff969c81)'s a quick first attempt on ripping out the download all together, with the added feature of using a precise installed by the distribution if found.

JarbasAl:
also see changes in https://github.com/JarbasLingua/jarbas-wake-word-plugin-precise

PureTryOut:
As a distro packager I'm in favor of assuming it has been pre-installed as well. On Alpine Linux the pre-compiled thing that gets downloaded doesn't work anyway due to Musl, so it being downloaded automatically makes no sense.

krisgesling:
I think we move straight to the plugin. I'm just pushing this into a Mark 2 feature branch as a stop gap.

PR #2794, opened 2021-01-08
Pure-try-out's XDG config PR with changed Jenkinsfile

forslund:
## Description
This is PR #2578 with added Jenkinsfile change for the identity location.


PureTryOut:
Oh, can you not push to my branch directly?

forslund:
I can try that as well

forslund:
The Jenkinsfile changes in your PR still seems to be rejected :/

forslund:
This one should at least technically work, though I think the identity file was broken again yesterday when the other PR was run without updating the Jenkinsfile.

I did try open a PR from your branch but since there's an existing PR I could not. @PureTryOut can you try to create a new branch based on your xdg-config branch with some tiny change (so they are not the same exact commit hashes) and I could try to open a PR from that one...

PureTryOut:
Done, the branch `xdg-config-new`.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

krisgesling:
With VK in it's current state - that is a pass ^^

PureTryOut:
Can't wait for it to be finally merged :raised_hands:

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

forslund:
Rebased against dev to resolve conflicts.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

forslund:
As far as I can determine the get_response reply from VK is sent before the skill has set up the handler. Not quite sure why this is the case for this PR in particular...Will be digging into it during this week but if anyone has some insights I'd be interested to hear

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

forslund:
Down to a single error...which might just be VK being VK...

PureTryOut:
So close :hugs:

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -427,16 +430,19 @@ def validator_default(utterance):
         on_fail_fn = on_fail if callable(on_fail) else on_fail_default
         validator = validator or validator_default
 
+        # Setup the converse handler
```
JarbasAl, line 433: is this included accidentally? all changes in this whole file doesn't seem XDG related

PureTryOut: I believe it's to fix some issues with VK that this PR uncovered? @forslund made the change so knows more about it.

forslund: That should probably be reverted, it's from one of the many attempts at figuring out the issue with the failing timer skill. It didn't fix it completely so it needs to be looked at again. I do believe the timer skill does something really funky, looking at the logs it throws a couple of random errors (even for successful test runs)

mycroft/util/log.py
```diff
@@ -84,7 +84,10 @@ def init(cls):
         # Check configs manually, the Mycroft configuration system can't be
         # used since it uses the LOG system and would cause horrible cyclic
         # dependencies.
-        confs = [SYSTEM_CONFIG, USER_CONFIG]
+        confs = []
+        for dir in BaseDirectory.load_config_paths('mycroft'):
+            confs.append(join(dir, 'mycroft.conf'))
+        confs.append('/etc/mycroft/mycroft.conf')
```
JarbasAl, line 90: Shouldn't this one be the very first config loaded? this way it takes precedence over the user config, everywhere else in core it doesn't

PureTryOut: But the user config should always takes precedence over the system config. It doesn't make sense the other way around.
This way you can configure something system-wide but have different users have separate options.

forslund: It gets reversed later, I assume it had to do with how the XDG is ordered.

forslund: Jarbas seems to be right, the merge below will apply the /etc/... last making it take precedence

PureTryOut: Oh he is saying it currently takes precedence over the user config? That would be wrong indeed then. @forslund since it's now your PR, could you make this change? :sweat_smile:

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

forslund:
Rebased on top of latest dev. Was able to reproduce some errors in the alarm skill but need to check it some more

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2794))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2794))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2794))

mycroft/util/log.py
krisgesling: _Note: Github won't let me create a new thread about the same line of code. I can see that this now gets reversed on the following line._

Separate to that, should this be appending `SYSTEM_CONFIG` rather than a hard coded string?

mycroft/configuration/config.py
```diff
@@ -183,30 +184,72 @@ def get(configs=None, cache=True):
...
+            remote (boolean): False if the Mycroft Home settings shouldn't
+                              be loaded
         Returns:
             (dict) merged dict of all configuration files
         """
         if not configs:
-            configs = [LocalConf(DEFAULT_CONFIG), RemoteConf(),
-                       LocalConf(SYSTEM_CONFIG), LocalConf(USER_CONFIG),
-                       Configuration.__patch]
+            configs = configs or []
```
krisgesling, line 212: Nitpick - configs must be Falsey in this code path - so presumably this can just be:
```
configs = []
```

mycroft/configuration/config.py
```diff
@@ -174,7 +175,7 @@ class Configuration:
     __patch = {}  # Patch config that skills can update to override config
 
     @staticmethod
-    def get(configs=None, cache=True):
+    def get(configs=None, cache=True, remote=True):
```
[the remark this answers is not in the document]
krisgesling: So after reviewing the rest of the code, is it because we now have an unknown number of configuration paths courtesy of `xdg.BaseDirectory.load_config_paths()`?

mycroft/filesystem/__init__.py
```diff
@@ -31,7 +33,13 @@ def __init__(self, path):
...
         if not isinstance(path, str) or len(path) == 0:
             raise ValueError("path must be initialized as a non empty string")
-        path = join(expanduser('~'), '.mycroft', path)
+
+        old_path = join(expanduser('~'), '.mycroft', path)
+        path = join(xdg.BaseDirectory.save_config_path('mycroft'), path)
+
+        # Migrate from the old location if it still exists
+        if isdir(old_path):
+            shutil.move(old_path, path)
```
krisgesling, line 42: Is there any danger in overwriting data if both paths exist?

Same with the other use of `shutil.move()`

bin/mycroft-config
```diff
@@ -95,13 +95,13 @@ function validate_config_file() {
     return $result
 }
 
-_conf_file="~/.mycroft/mycroft.conf"
+_conf_file="~/.config/mycroft/mycroft.conf"
 function name_to_path() {
     case ${1} in
         "system")     _conf_file="/etc/mycroft/mycroft.conf" ;;
-        "user")       _conf_file=$(readlink -f ~/.mycroft/mycroft.conf) ;;
+        "user")       _conf_file=$(readlink -f ~/.config/mycroft/mycroft.conf) ;;
```
krisgesling, line 102: should this use `$XDG_CONFIG_HOME` instead of the hard coded path?

mycroft/client/enclosure/mark1/__init__.py
```diff
@@ -164,6 +165,9 @@ def process(self, data):
         if "unit.factory-reset" in data:
             self.bus.emit(Message("speak", {
                 'utterance': mycroft.dialog.get("reset to factory defaults")}))
+            subprocess.call(
+                'rm ~/.config/mycroft/identity/identity2.json',
```
krisgesling, line 169: Should we be using `xdg.BaseDirectory.save_config_path('mycroft')` here?

mycroft/configuration/config.py
```diff
@@ -183,30 +184,72 @@ def get(configs=None, cache=True):
         Args:
             configs (list): List of configuration dicts
             cache (boolean): True if the result should be cached
+            remote (boolean): False if the Mycroft Home settings shouldn't
```
krisgesling, line 187: nitpick but I'd prefer we keep this generic - it may be another backend that isn't "Mycroft Home". I also think we're likely to rename "Home" given it can refer to sooooo many things.

can we just use "remote"?

mycroft/configuration/config.py
```diff
@@ -183,30 +184,72 @@ def get(configs=None, cache=True):
...
+
+            # Then use the system config (/etc/mycroft/mycroft.conf)
+            configs.append(LocalConf(SYSTEM_CONFIG))
+
+            # Then use the config that comes with the package
+            configs.append(LocalConf(DEFAULT_CONFIG))
+
+            # Make sure we reverse the array, as merge_dict will put every new
+            # file on top of the previous one
+            configs = reversed(configs)
```
krisgesling, line 252: I originally wondered why each config gets appended and then reversed (as opposed to appending in the correct order to begin with), but presumably this is to handle the order of `load_config_paths()`?

krisgesling (commented):
Hey, this has been a long journey, and very excited to see the tests now passing!

I'm sorry to throw more questions into the mix but hoping we can settle them quickly.

krisgesling (commented):
I just saw the HolmesV reference in the conversation history and noticed some more things:
- for caching do we want to use `xdg.BaseDirectory.save_cache_path`? This could be in a separate PR. Doesn't seem to have any cross over.
- do we need to update `dev_setup.sh` to symlink the new Skills directory?

mycroft/configuration/config.py
```diff
@@ -183,30 +184,72 @@ def get(configs=None, cache=True):
...
+            configs.append(Configuration.__patch)
+
+            # Then use XDG config
+            # This includes both the user config and
+            # /etc/xdg/mycroft/mycroft.conf
+            for dir in BaseDirectory.load_config_paths('mycroft'):
+                configs.append(LocalConf(join(dir, 'mycroft.conf')))
+
+            # Then check the old user config
+            if isfile(OLD_USER_CONFIG):
```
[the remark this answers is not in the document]
forslund: OLD_USER_CONFIG should only be the old hard coded path `~/.mycroft/mycroft.conf` so shouldn't be able to conflict as far as I can tell...

mycroft/configuration/config.py
forslund: It is due to the order returned by the `load_config_paths()` generator as far as I've understood it.

mycroft/filesystem/__init__.py
forslund: it is but I think it's the preferable case...if this move failed someway leaving the two files the move should be performed again since we can't know the state of the target file.

forslund:
Many good comments, I will fix them up during the weekend. Would like to hear @PureTryOut's comments as well and suggestions for the updates

mycroft/configuration/config.py
```diff
@@ -183,30 +184,72 @@ def get(configs=None, cache=True):
...
+                            OLD_USER_CONFIG)
+                LOG.warning(" Note that this location is deprecated and will" +
+                            " not be used in the future")
+                LOG.warning(" Please move it to " +
+                            BaseDirectory.save_config_path('mycroft'))
+                configs.append(LocalConf(OLD_USER_CONFIG))
+
+            # Then check the XDG user config
+            if isfile(USER_CONFIG):
+                configs.append(LocalConf(USER_CONFIG))
```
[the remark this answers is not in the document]
PureTryOut: No, line 221 checks for `OLD_USER_CONFIG`, which is a different location (`~/.mycroft/mycroft.conf`). This line checks for the XDG config (by default `~/.config/mycroft/mycroft.conf`).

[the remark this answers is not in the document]
PureTryOut: Oh sorry I saw that wrong. Yes it is duplicating it, you're right.

mycroft/configuration/config.py
```diff
@@ -183,30 +184,72 @@ def get(configs=None, cache=True):
...
-                       LocalConf(SYSTEM_CONFIG), LocalConf(USER_CONFIG),
-                       Configuration.__patch]
+            configs = configs or []
+
+            # First use the patched config
+            configs.append(Configuration.__patch)
+
+            # Then use XDG config
+            # This includes both the user config and
+            # /etc/xdg/mycroft/mycroft.conf
```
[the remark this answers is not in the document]
PureTryOut: The loop at line 220 doesn't include `/etc/mycroft` if that is what you mean.

I'd say any file outside of `$HOME` should be given a system level priority yes.

I guess `/etc/xdg/mycroft/mycroft.conf` should get the same priority as `/etc/mycroft/mycroft.conf`, which means it should be taken out of the loop and next to line 245.

mycroft/configuration/config.py
PureTryOut: `load_config_paths()`:

> Returns an iterator which gives each directory named ‘resource’ in the configuration search path. Information provided by earlier directories should take precedence over later ones, and the user-specific config dir comes first.

So yes, because of that. The most important directories come first, the less important ones later.

mycroft/util/log.py
PureTryOut: It should yes

mycroft/configuration/config.py
PureTryOut: Sure we can, that seems fine

forslund:
Might have missed this but I think the new skills location is in a separate PR?

mycroft/configuration/config.py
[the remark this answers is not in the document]
PureTryOut: I'm not entirely sure anymore to be honest, it has been ages since I've touched this code (hint hint, please don't wait so long before reviewing/accepting community PR's :wink:). Checking the code I don't fully see why this was made an option. We call this function with `remote=False` most of the time but at least 1 place uses `remote=True` (or well, it doesn't pass a value for it which means it defaults to True). I'm guessing I had a reason for that but yes, this was ages ago.

mycroft/client/enclosure/mark1/__init__.py
PureTryOut: Yes, yes we should.

mycroft/client/enclosure/__main__.py
```diff
@@ -70,8 +70,8 @@ def main(ready_hook=on_ready, error_hook=on_error, stopping_hook=on_stopping):
     only the GUI bus will be started.
     """
     # Read the system configuration
-    system_config = LocalConf(SYSTEM_CONFIG)
-    platform = system_config.get("enclosure", {}).get("platform")
+    config = Configuration.get(remote=False)
+    platform = config.get("enclosure", {}).get("platform")
```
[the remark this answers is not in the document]
PureTryOut: It's to deal with the many possible configs scenario yes. XDG comes with it's `/etc/xdg` system level configuration but Mycroft also supports plain `/etc`. Both of these have to be checked. We can just check those 2 directories manually here as well, which might make the `Configuration.get()` function cleaner.

bin/mycroft-config
PureTryOut: Yes it should. `${XDG_CONFIG_HOME:-$HOME/.config}`  to be exact.

PureTryOut:
@forslund yes that is in #2803.

I've responded everywhere (if I just left a thumbs up then I agreed to the comment). Normally I'd fix this up myself but this PR not being mine anymore makes things quite difficult and messy... Darn VK.

PureTryOut:
@krisgesling

> for caching do we want to use xdg.BaseDirectory.save_cache_path? This could be in a separate PR. Doesn't seem to have any cross over.

Yes. I'm not sure about any caching done by Mycroft so I left it out here. If you point me to the places where it caches things then I can do that in a separate PR once this and #2803 are in (so not too many PR's stay open that have to be rebased and changed constantly).

> do we need to update dev_setup.sh to symlink the new Skills directory?

Symlink from `/opt/mycroft` you mean? My personal goal of these PR's are to get rid of `/opt` usage completely, so I personally wouldn't be happy with a symlink there. But I guess it can be done if the script detects there is already an installation in `/opt`, for new installations it would be ignored then.

mycroft/configuration/locations.py
```diff
@@ -13,11 +13,16 @@
...
+import xdg
 
 DEFAULT_CONFIG = join(dirname(__file__), 'mycroft.conf')
 SYSTEM_CONFIG = os.environ.get('MYCROFT_SYSTEM_CONFIG',
                                '/etc/mycroft/mycroft.conf')
-USER_CONFIG = join(expanduser('~'), '.mycroft/mycroft.conf')
+# Make sure we support the old location still
+# Deprecated and will be removed eventually
+OLD_USER_CONFIG = join(expanduser('~'), '.mycroft/mycroft.conf')
+USER_CONFIG = join(xdg.BaseDirectory.save_config_path('mycroft'),
```
JarbasAl, line 24: a string definition should not create a folder, especially by simply importing this, other places would be ok-ish

eg, if every service launches at the exact same time there is a millisecond window where 2 services try to create the directory and one of them throws an exception and fails to load, this sounds hard to cause but i can replicate consistently in my setup

mycroft/client/text/text_client.py
```diff
@@ -171,6 +172,35 @@ def load_settings():
...
     global show_meter
 
+    config_file = None
+
+    # Old location
+    path = os.path.join(os.path.expanduser("~"), ".mycroft_cli.conf")
+    if os.path.isfile(path):
+        LOG.warning(" ===============================================")
+        LOG.warning(" ==             DEPRECATION WARNING           ==")
+        LOG.warning(" ===============================================")
```
JarbasAl, line 182: i suggest this whole Configuration section is refactored into smaller pieces, see this file https://github.com/HelloChatterbox/HolmesV/commit/2608fe0e236854bfb9d8ea9f141b51623bc47a45#diff-7a26324169efbe9ae270dc54e1e9476fff06521e22b5294f4e55624081b92bbdR186 

(ignore the fact that i have a flag to determine if xdg is enabled or not, that is for backwards compat with mycroft-core, will remove that once this PR is finally merged)

mycroft/configuration/locations.py
PureTryOut: In that case it should be replaced with `xdg.BaseDirectory.load_first_config('mycroft')`.

mycroft/configuration/config.py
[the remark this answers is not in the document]
JarbasAl: i think this was to avoid cyclic import errors due to RemoteConf using `from mycroft.api import is_paired`

mycroft/util/log.py
```diff
@@ -84,7 +84,11 @@ def init(cls):
         # Check configs manually, the Mycroft configuration system can't be
         # used since it uses the LOG system and would cause horrible cyclic
```
JarbasAl, line 85: just gonna leave this here https://github.com/HelloChatterbox/HolmesV/commit/0696559fd17f31ffb40f8099c839d98a23fd7096

mycroft/configuration/locations.py
JarbasAl: what if config does not exist? what does that return?

the main usage outside of core for this is to edit the user config, `LocalConf(USER_CONFIG).store()`, this is what creates the file if it does not exist, so shouldnt it be always the same place? if this operation creates different files it gets messy. 

`USER_CONFIG = join(xdg.BaseDirectory.xdg_config_home, BASE_FOLDER, CONFIG_FILE_NAME)`

this does not impact loading of other xdg dirs, but ensures the string points to a valid deterministic path, its kinda like the other places where you have a read location and a write location

mycroft/configuration/config.py
JarbasAl: comment below with suggested refactor into smaller functions addresses this

mycroft/configuration/locations.py
JarbasAl: https://github.com/takluyver/pyxdg/pull/15 also fixes this, but i still think its not clean to create a directory on a string definition, doesnt feel like the proper place. eg, if i want to check if that directory exists it gets created on import, so i can't do that check

mycroft/client/enclosure/__main__.py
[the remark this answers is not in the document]
JarbasAl: i think using all configs follows the [Principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)

If as a dev/user i change that string in the file i'm told belongs to user and it does nothing I will be scratching my head

Note that there are skills depending on this string, if you install some "enclosure skill" it will have write access to user config as part of its setup, but not to the system level.

This string is really a flag that changes behavior, its no longer a simple identifier about what hardware its running.  With the mark2 loading different QML and such based on this string i can also imagine skills wanting to override this to disable stuff like the UI scaling and providing a new "skin".

forslund:
When starting to fix this I see the minor clean up I did last weekend isn't here...Big thanks to @JarbasAI for the prepared implementation suggestions will try to cherry-pick some of them.

mycroft/configuration/locations.py
PureTryOut: Why would you ever want to check if that directory exists though? Is there _any_ use-case for _not_ having the config folder present? Who cares that it's created on string definition. You just ask for the location of the folder, and you get it. With pyxdg you can safely assume it exists, which is fine as long as there is no use-case for not having the folder.

JarbasAl: in OVOS we do some checks like these to determine in what platform we are running for example, so yes, there are use edge cases like that. For instance since this PR has been open for over a year i have used this to determine if xdg is supported at all, this is part of what allows ovos_utils to know its running in the alpine patched version of mycroft-core but not on HolmesV ;)

PureTryOut: > You are welcome to disagree with me

I definitely disagree here lol :wink: 

> in OVOS we do some checks like these to determine in what platform we are running for example

You check _what platform_ you are running on by checking if this folder exists? Or just if XDG is supported or not? If the latter, that would definitely be the only use-case ever for that check. I really see no reason to ever check if that directory exists in a pyxdg codebase as it's safe to assume the directory exists if you use the pyxdg save method to get the path (which you should).

JarbasAl: getting a path != creating a path

let's just agree to disagree, the race condition issue stands and your suggestion wont work, we can also block this PR until a new release of pyxdg with PR linked above

> You check what platform you are running on by checking if this folder exists? Or just if XDG is supported or not?

its a datapoint in a larger heuristic for platform detection, OVOS approach is more capability based than hardcoded list of platform strings

JarbasAl: duh, disregard what i said, i forgot about `__ensure_folder_exists`, you can use that to workaround the race condition mentioned above, and update it with same fix from pyxdg PR, it has the same issue

mycroft/client/enclosure/__main__.py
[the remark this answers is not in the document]
chrisveilleux: I think I agree with Jarbas on this, if I am reading his comment right.  I have always been confused why certain parts of the system only need certain config files.  Why can't we just always use all the configs?  What benefit is there to only loading a partial list of the config files?

mycroft/configuration/config.py
```diff
@@ -17,15 +17,15 @@
 import re
 import json
 import inflection
-from os.path import exists, isfile
+from os.path import exists, isfile, join, expanduser, dirname
```
chrisveilleux, line 20: see previous comments on import organization and using `pathlib` instead of `os`

mycroft/configuration/config.py
```diff
@@ -17,15 +17,15 @@
...
 from requests import RequestException
+from xdg import BaseDirectory
 
+from .locations import DEFAULT_CONFIG, OLD_USER_CONFIG, USER_CONFIG
```
chrisveilleux, line 24: why were these moved?  I think in general we put the relative imports after the `mycroft` package imports... or is that just me?

mycroft/messagebus/send_func.py
```diff
@@ -32,10 +30,7 @@ def send(message_to_send, data_to_send=None):
     data_to_send = data_to_send or {}
 
     # Calculate the standard Mycroft messagebus websocket address
-    config = Configuration.get([DEFAULT_CONFIG,
-                                SYSTEM_CONFIG,
-                                USER_CONFIG],
-                               cache=False)
+    config = Configuration.get(cache=False, remote=False)
```
chrisveilleux, line 33: confused as to why it matters if the remote config is accessed here.

mycroft/filesystem/__init__.py
chrisveilleux: should we add a comment regarding when this code will be removed?  is there a release... say 22.02... where we no longer support the old path?

chrisveilleux (changes requested):

mycroft/messagebus/send_func.py
forslund: This just replicates the existing behavior instead of the list of locations it uses the remote=False flag. So I don't think investigating this belongs in this PR

Originally it was set up like this to speed up the command line for sending messages to mycroft core.

mycroft/client/enclosure/__main__.py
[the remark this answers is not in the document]
forslund: Excluding remote is generally a good idea if we know that a setting will not originate from there. The speedup is significant.

mycroft/client/speech/hotword_factory.py
```diff
@@ -253,7 +277,10 @@ def update_precise(self, precise_config):
 
     @property
     def folder(self):
-        return join(expanduser('~'), '.mycroft', 'precise')
+        old_path = join(expanduser('~'), '.mycroft', 'precise')
+        if os.path.isdir(old_path):
+            return old_path
+        return join(xdg.BaseDirectory.save_data_path('mycroft', 'precise'))
```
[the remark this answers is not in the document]
forslund: This particular call to join doesn't even do anything and can be removed completely :)

mycroft/messagebus/send_func.py
forslund: Since this only replicate the earlier behavior using the flag instead of the location list, I think this can be kept as is.

Originally remote was dropped to get some speedup when using the command-line application to inject messages.

mycroft/client/enclosure/__main__.py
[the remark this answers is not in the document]
JarbasAl: agree in excluding remote, especially because it should never touch these values, just disagree with the user config being removed

one is an optimization and you fully control it, the other is user facing and just confusing if ignored

mycroft/configuration/config.py
JarbasAl: is this an official requirement now?

forslund: I think pathlib is often a good thing now that we don't need to support python 3.5 where pathlib was very wonky to work with. Pathlib sometimes makes code more convoluted though and is not the best in all circumstances IMO.

Anyway, regarding this PR: I don't think mixing in pathlib into code that is already using os.path functions extensively is desirable so a single call is now made using pathlib while the rest of the function uses os.path functions. Just makes it more of a jumble.

I think we can go over a couple of classes and functions and make them use pathlib instead of os.path in separate PRs to make the code more coherent.

mycroft/client/enclosure/__main__.py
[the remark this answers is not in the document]
forslund: Agreed, I think this is good including the User config in this. I guess the old version was a quick hack. Is the consensus that we leave it like it is now, since as far as I can tell it includes all local config files?

mycroft/client/text/text_client.py
```diff
@@ -171,6 +172,35 @@ def load_settings():
     global max_log_lines
     global show_meter
 
+    config_file = None
+
+    # Old location
+    path = os.path.join(os.path.expanduser("~"), ".mycroft_cli.conf")
+    if os.path.isfile(path):
+        LOG.warning(" ===============================================")
```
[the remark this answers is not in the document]
forslund: Made it a single log statement

mycroft/filesystem/__init__.py
forslund: Added comment about this

mycroft/client/enclosure/mark1/__init__.py
```diff
@@ -15,6 +15,7 @@
 import subprocess
 import time
 import sys
+import os
```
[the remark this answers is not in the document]
forslund: Removed the import

mycroft/client/speech/hotword_factory.py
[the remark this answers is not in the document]
forslund: Removed join

mycroft/configuration/config.py
forslund: The location of these has been restored

mycroft/util/log.py
forslund: Cherry-picked this to remove the duplication of code. Very nice indeed

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

mycroft/client/enclosure/mark1/__init__.py
forslund: Code updated according to this.

mycroft/configuration/config.py
forslund: Updated

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

mycroft/client/enclosure/__main__.py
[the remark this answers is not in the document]
PureTryOut: This includes all configs but the remote one yes. I agree with you guys here, exclude remote but otherwise use all available configs in the proper priority.

PureTryOut:
Please squash this PR a bit before merging, the commit list is starting to look a bit like a mess...

forslund:
Indeed I will, just wanting to keep the things I do for the review visible

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

mycroft/filesystem/__init__.py
forslund: There actually was an issue with the move if the target existed as @krisgesling suggested. This has now been handled

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

mycroft/configuration/config.py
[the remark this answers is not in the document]
krisgesling: ah yeah, cos the xdg path would always use a `$PATH/mycroft/stuff` subdirectory, so the "dot mycroft" can't conflict

mycroft/configuration/config.py
chrisveilleux: fair enough.  can we at least order the imports according to PEP8?

mycroft/messagebus/send_func.py
chrisveilleux: is it really that much faster?  Regardless, we don't have to address it in this PR.  just asking a question.

mycroft/client/enclosure/__main__.py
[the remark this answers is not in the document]
chrisveilleux: One thing that still bugs me about this... a config that is not in the remote config right now could be added in the future.   For example, when the Mark II gets the the "consumer ready" point, we should have a remote config that allows users to select their language.  How do we know if doing so breaks this code (or any other code that excludes remote)?

One other question, why is parsing and extra JSON file and merging it with others such an expensive operation?

mycroft/configuration/config.py
forslund: Reordered

PureTryOut:
I guess this is blocked on VK for whatever reason failing again now? :cry:

I'm getting a bit tired of waiting to be honest. I made the PR ages ago and only this month I finally got reviews from Mycroft team members.
And VK is being awfully annoying and hard to figure out why something goes wrong, often it doesn't even have anything to do with the PR in question.
I was excited to get Mycroft in a state so desktop users could better install and update their instance, thus getting more users interested, but my excitement is fading slowly...

krisgesling:
Hi Bart, I do get the frustration. VK is better than what we had previously but it still needs a lot of work. Are you seeing the current failures if you run the tests locally or is it only in the CI?

In terms of engagement with this PR, from my perspective Ake has been helping out with it, and he's either the most (or second most) knowledgeable person in the world when it comes to Mycroft and Voight Kampff. So in thinking about where we allocate resources, this in my opinion was already being well resourced - albeit not by a paid staff member. Once the PR was in a final draft state, we get a fresh set of eyes across it, and in this case identified some other fixes and improvements. This can be a frustrating turn of events, but with such a significant change it's pretty important that we get it right.

I think this is a great change, however it is not a strategic priority for the company. We are very focused on shipping the Mark II and if we're looking from that lens, Mycroft becoming XDG compliant does not help us move toward that goal in any way. In fact it delays shipping by adding complexity and potentially introducing bugs - so we want to be pretty careful with making this change.

I'm certainly not saying we don't want to support anything but the Mark II. We still absolutely want to merge it. I think it's an important change, and as you say makes it much easier for desktop packaging and desktop users in general. There's just a limit on how much time we can spend on features that don't align to our current goals. These are not easy decisions to make. Choosing **not to do something** can often be harder than choosing **to do it**, but trying to do everything or be all things for all people doesn't often end up with good results.

In very good news, we'll soon be posting a range of new positions to expand our dev team. The team will still be focused on the Mark II, but it does lighten the load and provide more capacity for us to support changes like this one.

forslund:
I haven't been able to reproduce the issues locally I'm afraid. I've pushed a bunch of PR's improving some edge cases where VK might miss messages to hopefully improve this.

The weird thing is that the VK tests work reliably on exact the commit / history that first passed but doing a rebase on latest dev makes it fail.

goldyfruit:
@PureTryOut keep up the good work :+1:

PureTryOut:
@krisgesling I understand this PR isn't a priority since it doesn't do anything for the Mark 2, but this PR has been open for 7 months now, and the PR it's based on was made even 7 months earlier. That means these changes were being proposed **14 months ago**. Yes @forslund has been doing a great job and I owe him my thanks, but he is not the person with the final say. It took **more than a year** to finally get core team members to look at this properly. "Not a priority" and "have it hanging for 14 months" are basically complete opposites of each other. Is it really too much to ask for a review of a team member within _a few months_?
Yes I understand changes like these are dangerous to just merge, especially when shipping hardware that shouldn't have the software break on it, but getting at least a review in within a reasonable time frame would go a long way to prevent such dangerous problems.

And please note that I have more PR's open that depend on this. With the current pace, after this is merged it'll take another year for the rest to get in. Hopefully those extra developers you mentioned will help, as if every community PR that is not directly focused on the Mark 2 has to wait this long, you'll lose _a lot_ of potentionally interested developers.

As for VK, I have no clue how to run those tests locally and even if I did, I wouldn't know how to determine what's wrong. The unit tests all succeed, both locally and on the CI here, that's all I can confirm.

krisgesling:
I don't know that I've found a solution, but I've dug into two problems which might move us closer.

1. The Alarm Skill has a pretty complicated approach to clearing out any previously set alarm times. This has a bunch of back and forth on the message bus to trigger the delete intent and confirm etc. I'm testing replacing this with a Skill API method so we can then just call a method that assigns the alarm list to an empty list. This should have the added benefit of speeding the tests up and _might_ make the tests more consistent.

2. I've found that I'm getting a lot of converse errors and it seems like `handle_converse_request` is getting called twice. The first time not all Skills are being returned, and so the handler returns an error that the Skill can't be found. The test then fails as it receives the error message, but the behaviour might still succeed because the second call of the handler correctly calls the Skills converse method.

The strange thing about this 2nd issue is that I can't work out why the handler is called twice. Maybe something in the message bus client?

forslund:
I think the duplication is due to the fact that there are multiple active skills. it will send one request for each active skill (or until it finds an active skill that wants to handle the query)

Using the skill api to simplify and speed up the alarm skill will be a great improvement. Hopefully the make timer skill great again will improve the timer tests stability...

forslund:
Also worth noting is that the alarm skill is not an issue since https://github.com/MycroftAI/skill-alarm/pull/115 was merged I wonder if there exist similar things in the timer skill still?

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2794)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2794/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2794/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2794/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2794/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2794/logs/enclosure.log)

forslund:
A status update here, the current failure is something that is common to all VK runs at the moment. The weather skill seem to have introduced some issue and gives two failures consistently when ever VK is run.

PureTryOut:
So... Nothing to do with this PR then and it's ready to be merged? Can someone press that button then pretty please? :hugs:

JarbasAl:
> So... Nothing to do with this PR then and it's ready to be merged? Can someone press that button then pretty please? hugs

sounds like VK stops more good code from making it into core than it stops bad code :sweat_smile:

krisgesling:
Hey @forslund awesome work!

It looks like this is just rebased with the latest changes from core is that right?

I've tagged the failing tests in Weather as xfail so should have a clean run in another hour or so.

forslund:
Yes it's just rebased on top of the latest updates in dev

forslund:
Aaaand...VK finally approves...

krisgesling (approved):
Start automatic ground launch sequencer
    Retract orbiter access arm
    Start auxiliary power units
    Arm solid rocket booster range safety safe and arm devices
    Start orbiter aerosurface profile test
    Start main engine gimbal profile test
    Retract gaseous oxygen vent arm
    Crew members close and lock visors
    Orbiter transfer to internal power
    Ground launch sequencer is go for auto sequence start
    Activate launch pad sound suppression system
    Activate main engine hydrogen burnoff system
    Main engine start
    ...
    Solid rocket booster ignition...
    ...
    We have liftoff!
    I repeat, we - have - liftoff!

PureTryOut:
YYYYYEEEESSSS! Thank you very much! Now up to the rest of the XDG PR's :sweat_smile:

PR #2800, opened 2021-01-14
Fix Mimic 2 cache directory error

chrisveilleux:
## Description
Fixed an error that occurs when the pre-loaded cache directory already exists but the cache_text.txt file does not.

## How to test
Remove the cache_text.txt file from the pre-loaded cache directory and reboot the device.  There should be no errors regarding creating the file.

CLA [Y]

### DIFF
--- a/mycroft/tts/cache_handler.py
+++ b/mycroft/tts/cache_handler.py
@@ -45,14 +45,23 @@ wifi_setup_path = '/usr/local/mycroft/mycroft-wifi-setup/dialog/en-us'
 cache_dialog_path = [res_path, wifi_setup_path]
 
 
-def generate_cache_text(cache_text_file):
+def generate_cache_text(cache_audio_dir, cache_text_file):
     """
     This prepares a text file with all the sentences
     from *.dialog files present in
     mycroft/res/text/en-us and mycroft-wifi setup skill
     Args:
+        cache_audio_dir (path): DEPRECATED path to store .wav files
         cache_text_file (file): file containing the sentences
     """
+    # TODO: remove in 21.08
+    if cache_audio_dir is not None:
+        LOG.warning(
+            "the cache_audio_dir argument is deprecated. ensure the directory "
+            "exists before executing this function. support for this argument "
+            "will be removed in version 21.08"
+        )
+        os.makedirs(cache_audio_dir)
     try:
         if not os.path.isfile(cache_text_file):
             text_file = open(cache_text_file, 'w')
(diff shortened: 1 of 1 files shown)


devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2800)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2800/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2800/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2800/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2800/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2800/logs/enclosure.log)

krisgesling (approved):
Looks great. 

I also like the shift so that `generate_cache_text` only does what the method name suggests.

mycroft/tts/cache_handler.py
```diff
@@ -45,28 +45,26 @@
 cache_dialog_path = [res_path, wifi_setup_path]
 
 
-def generate_cache_text(cache_audio_dir, cache_text_file):
+def generate_cache_text(cache_text_file):
```
JarbasAl, line 48: this is deprecating the old API, maybe throw a deprecation warning and support the argument for a while

krisgesling: Ah yeah good point. 

Should we also make them private methods and only expose a single `cache_handler`? Or are there uses for these independently?

Actually it would definitely need a little more refactoring to be properly generic.

JarbasAl: i'm not sure this even fits in mycroft-core tbh, feels like an helper script that should be there in the first place.

but now that its there, there might be users in the wild depending on it, eg, OVOS could be using that to generate its own cache (it isn't afaik)

making them private would also break the api compatibility, so still needs a deprecation warning, that way next major version tells everyone to stop depending on it, and the one after can make them private/remove them

krisgesling: Yeah, agree we shouldn't break the API. 

Given it's currently hardcoded to Mimic2, we could extract it to that plugin. At the moment this seems like the most likely course of action to me.

Alternatively, if there is interest and someone willing to do it, they could refactor it for use in any TTS in which case we could leave it in core as a common method.

chrisveilleux:
Fixed backwards compatibility issue.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2800))

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2800)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2800/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2800/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2800/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2800/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2800/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2800))

PR #2802, opened 2021-01-15
Allow system-wide and user installations rather than just virtual environments

PureTryOut:
This is a replacement for #2687. It does not fully depend on any other PR, but is recommended to use with #2794, #2803, https://github.com/MycroftAI/mycroft-skills-manager/pull/88 https://github.com/MycroftAI/mycroft-skills-manager/pull/90 to stop Mycroft doing anything in directories it might not have access to anymore.

## Description
This PR modifies `bin/mycroft-start` and `bin/mycroft-stop` to allow starting Mycroft services no matter where they installed and where the scripts itself are located. These are intended to be the main way to launch and run Mycroft for end-users, instead of them requiring to basically setup a full dev environment.

For end users this makes running `dev_setup.sh` unnecessary, unless they want to actually develop for Mycroft. In theory a PyPi package should also be possible to create now so users can just run `pip install mycroft-core` and have it work, but that might be something for later.

It also modifies the README to include new installations instructions using `setup.py`.

## How to test

1. Create a new empty virtual environment: `python3 -m venv .venv`.
2. Run `python3 setup.py install`
3. Run `mycroft-start debug` and observe everything be launched like normal
4. Run `mycroft-stop` and observe everything being stopped like normal
5. In a new terminal (outside of a virtual environment) run `python3 setup.py install`
6. Run `mycroft-start debug` and observe everything be launched like normal
7. Run `mycroft-stop` and observe everything being stopped like normal

Although testing the user installation (outside of virtual environment) should be enough, a system-wide installation can also be tested by running `sudo python3 setup.py install`.

CLA [x]

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,29 +32,49 @@ Mycroft is a hackable open source voice assistant.
 
 ## Getting Started
 
-First, get the code on your system!  The simplest method is via git ([git installation instructions](https://gist.github.com/derhuerst/1b15ff4652a867391f03)):
-- `cd ~/`
-- `git clone https://github.com/MycroftAI/mycroft-core.git`
-- `cd mycroft-core`
-- `bash dev_setup.sh`
+Mycroft might be packaged by your distribution already, in that case installing it might be as simple as `apk add mycroft-core` or equivalent for your distribution.
+Otherwise, the simplest method is downloading the latest release on the [Github releases page](https://github.com/MycroftAI/mycroft-core/releases) and unpacking it somewhere.
+Then run the following commands in the unpacked directory.
 
+```sh
+pip install . # This actually installs Mycroft and it's dependencies
+```
 
-This script sets up dependencies and a [virtualenv][about-virtualenv].  If running in an environment besides Ubuntu/Debian, Arch or Fedora you may need to manually install packages as instructed by dev_setup.sh.
+For development on Mycroft it's recommended to use Git instead ([git installation instructions](https://gist.github.com/derhuerst/1b15ff4652a867391f03)).
+
+```sh
+cd ~/
+git clone https://github.com/MycroftAI/mycroft-core.git
+cd mycroft-core
+./dev_setup.sh
+```
+
+This script sets up dependencies and a [virtualenv][about-virtualenv].  If running in an environment besides Ubuntu/Debian, Arch or Fedora you may need to manually install packages as instructed by `dev_setup.sh`.
 
 [about-virtualenv]:https://virtualenv.pypa.io/en/stable/
 
-NOTE: The default branch for this repository is 'dev', which should be considered a work-in-progress. If you want to clone a more stable version, switch over to the 'master' branch.
+**NOTE:** The default branch for this repository is 'dev', which should be considered a work-in-progress. If you want to clone a more stable version, switch over to the 'master' branch.
 
 ## Running Mycroft
 
-Mycroft provides `start-mycroft.sh` to perform common tasks. This script uses a virtualenv created by `dev_setup.sh`.  Assuming you installed mycroft-core in your home directory run:
-- `cd ~/mycroft-core`
-- `./start-mycroft.sh debug`
+Mycroft provides `mycroft-start` to perform common tasks. Assuming you installed mycroft-core via your systems package manager or via `setup.py`:
 
-The "debug" command will start the background services (microphone listener, skill, messagebus, and audio subsystems) as well as bringing up a text-based Command Line Interface (CLI) you can use to interact with Mycroft and see the contents of the various logs. Alternatively you can run `./start-mycroft.sh all` to begin the services without the command line interface.  Later you can bring up the CLI using `./start-mycroft.sh cli`.
+```sh
+mycroft-start debug
+```
+
+The "debug" command will start the background services (microphone listener, skill, messagebus, and audio subsystems) as well as bringing up a text-based Command Line Interface (CLI) you can use to interact with Mycroft and see the contents of the various logs.
+Alternatively you can run `mycroft-start all` to begin the services without the command line interface.
+Later you can bring up the CLI using `mycroft-start cli`.
 
 The background services can be stopped as a group with:
-- `./stop-mycroft.sh`
+
+```sh
+mycroft-stop
+```
+
+If you want to develop for Mycroft, please use `start-mycroft.sh` and `stop-mycroft.sh` instead of the aforementioned commands.
+These provide extra options for development purposes.
 
 ## Using Mycroft
 
--- a/bin/mycroft-start
+++ b/bin/mycroft-start
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 # Copyright 2019 Mycroft AI Inc.
 #
@@ -14,8 +14,155 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-SOURCE="${BASH_SOURCE[0]}"
-cd -P "$( dirname "$SOURCE" )"/.. || exit
-DIR="$( pwd )"
+script=${0}
+script=${script##*/}
 
-. "$DIR/start-mycroft.sh" "$@"
+help() {
+	echo "${script}:  Mycroft command/service launcher"
+	echo "usage: ${script} [COMMAND] [restart] [params]"
+	echo
+	echo "Services COMMANDs:"
+	echo "  all                      runs core services: bus, audio, skills, voice"
+	echo "  debug                    runs core services, then starts the CLI"
+	echo "  audio                    the audio playback service"
+	echo "  bus                      the messagebus service"
+	echo "  skills                   the skill service"
+	echo "  voice                    voice capture service"
+	echo "  enclosure                mark_1 enclosure service"
+	echo
+	echo "Tool COMMANDs:"
+	echo "  cli                      the Command Line Interface"
+	echo
+	echo "Options:"
+	echo "  restart                  (optional) Force the service to restart if running"
+	echo
+	echo "Examples:"
+	echo "  ${script} all"
+	echo "  ${script} all restart"
+	echo "  ${script} bus"
+	echo "  ${script} voice"
+
+	exit 1
+}
+
+name_to_script_path() {
+	case ${1} in
+		"bus")		_module="mycroft.messagebus.service" ;;
+		"skills")	_module="mycroft.skills" ;;
+		"audio")	_module="mycroft.audio" ;;
+		"voice")	_module="mycroft.client.speech" ;;
+		"cli")		_module="mycroft.client.text" ;;
+		"enclosure")	_module="mycroft.client.enclosure" ;;
+
+		*)
+			echo "Error: Unknown name '${1}'"
+			exit 1
+	esac
+}
+
+require_process() {
+	name_to_script_path "${1}"
+	if ! pgrep -f "python3 (.*)-m ${_module}" > /dev/null; then
+		launch_background "${1}"
+	fi
+}
+
+launch_process() {
+	name_to_script_path "${1}"
+
+	# Launch process in foreground
+	echo "Starting $1"
+	python3 -m ${_module} "$_params"
+}
+
+launch_background() {
+	name_to_script_path "${1}"
+
+	if pgrep -f "python3 (.*)-m ${_module}" > /dev/null; then
+		if ($_force_restart); then
+			echo "Restarting: ${1}"
+			mycroft-stop "${1}"
+		else
+			# Already running, no need to restart
+			return
+		fi
+	else
+		echo "Starting background service $1"
+	fi
+
+	# Security warning/reminder for the user
+	if [ "${1}" = "bus" ] ; then
+		echo "CAUTION: The Mycroft bus is an open websocket with no built-in security"
+		echo "         measures.  You are responsible for protecting the local port"
+		echo "         8181 with a firewall as appropriate."
+	fi
+
+	# Launch process in background
+	# Send logs to XDG Base Directories cache location
+	logdir="${XDG_STATE_HOME:-$HOME/.local/state}/mycroft"
+
+	if [ ! -d "$logdir" ]; then
+		mkdir -p "$logdir"
+	fi
+
+	python3 -m ${_module} "$_params" >> "$logdir/${1}.log" 2>&1 &
+}
+
+launch_all() {
+	echo "Starting all mycroft-core services"
+	launch_background bus
+	launch_background skills
+	launch_background audio
+	launch_background voice
+	launch_background enclosure
+}
+
+_opt=$1
+_force_restart=false
+
+if [ $# -gt 0 ]; then
+	shift
+fi
+
+if [ "${1}" = "restart" ] || [ "${_opt}" = "restart" ]; then
+	_force_restart=true
+	if [ "${_opt}" = "restart" ]; then
+		# Support "start-mycroft restart all" as well as "start-mycroft all restart"
+		_opt=$1
+	fi
+	shift
+fi
+_params=$*
+
+case ${_opt} in
+	"all")
+		launch_all
+		;;
+	"bus")
+		launch_background "${_opt}"
+		;;
+	"audio")
+		launch_background "${_opt}"
+		;;
+	"skills")
+		launch_background "${_opt}"
+		;;
+	"voice")
+		launch_background "${_opt}"
+		;;
+	"debug")
+		launch_all
+		launch_process cli
+		;;
+	"cli")
+		require_process bus
+		require_process skills
+		launch_process "${_opt}"
+		;;
+	"enclosure")
+		launch_background "${_opt}"
+		;;
+	*)
+		help
+		;;
+esac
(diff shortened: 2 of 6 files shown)


devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2802)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2802/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2802/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2802/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2802/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2802/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2802))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2802))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2802))

PureTryOut:
I've updated the launch script to log to the new `$XDG_STATE_HOME` as defined in the updated XDG Base Directory specification.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2802))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2802))

ChanceNCounter:
Are system installs desirable? I'm really asking, not being sarcastic. Makes me nervous.

PureTryOut:
I mean, that's literally how people have it installed now. `/opt` _is_ a system-wide install. Now it just allows Mycroft.ai to ship via Pip which allows user to easily update and stuff.

I think you're more worried about distribution packaging which is fair, but not the point of this PR (although this does help that).

ChanceNCounter:
`/opt` is a system-wide install, but other bits aren't always. I'm more wondering how this would affect, for example, user-level config.

PureTryOut:
I'm not sure I understand what you mean? No matter where mycroft is installed, (using the XDG PR's) the configuration will always be read from the same locations. It doesn't matter if you have a system or a local install.

ChanceNCounter:
If I log out of my system, log in as a different user, and restart Mycroft, what do I encounter?

Is the device paired to the same account? Does it use the same Skill preferences? Does that mean it's using the same API keys and etc.? Now we need users who put Mycroft on the family PC to understand what's private and what's shared, so they can make sure they put settings in the right place.

PureTryOut:
I'm not sure how this PR makes that any different from before.

With Mycroft installed to `/opt` like usual, what happens if you log out of your system and log into a different user? Whatever happens there, will happen with this PR and a system-wide setup. I just fail to see how this PR changes such behaviour.

ChanceNCounter:
Under the current model, most of a given Mycroft build lives where it lives. `/opt/mycroft` is largely an exception, and XDG compliance only changes that to a certain extent, given that some config can, should, and will remain in a user's home folder.

PureTryOut:
I'm afraid I still don't understand what you're getting at. Are you saying that even though data and configs are in `/opt/mycroft`, the code and the executables itself are wherever the user `git clone`'ed mycroft-core to? I'm not sure how moving mycroft-core to a system-wide directory would break or mess things up.

With a system-wide installation, Mycroft still runs as _your_ user reading the usual config locations (e.g. `/etc/mycroft` and `~/.config/mycroft`). It also stores it's (temporary) data in your user folder still. So if you start Mycroft, then log out and log in as a different user and start Mycroft, it'll just run as 2 separate processes each acting as if no other Mycroft process is running, both storing their changed settings and data to their own user's folder. That is, if this is merged _after_ the XDG PR's are in (#2794 and #2803) which I recommended in the original comment. If this is merged with `/opt` still in use then it will definitely mess things up when you run it as 2 different users at the same time yes.

This basically just removes the need of having duplicated files (the mycroft-core code) on your system if you want to use it with multiple users, and will allow easy installation via e.g. Pip.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2802))

PureTryOut:
A lot of messing around later, CI now succeeds :tada: 

This should be good to go now.

README.md
```diff
@@ -32,29 +32,49 @@ Mycroft is a hackable open source voice assistant.
 
 ## Getting Started
 
-First, get the code on your system!  The simplest method is via git ([git installation instructions](https://gist.github.com/derhuerst/1b15ff4652a867391f03)):
-- `cd ~/`
-- `git clone https://github.com/MycroftAI/mycroft-core.git`
-- `cd mycroft-core`
-- `bash dev_setup.sh`
```
krisgesling, line 39: I think we need to keep git installation as the primary recommended way for people to install Mycroft from this repository. To me, this readme is for developers.

README.md
```diff
@@ -32,29 +32,49 @@ Mycroft is a hackable open source voice assistant.
...
+Mycroft might be packaged by your distribution already, in that case installing it might be as simple as `apk add mycroft-core` or equivalent for your distribution.
```
krisgesling, line 35: I'm happy for us to add a community releases section, but it can't be the primary recommendation. We're also probably better off pointing to the [main docs](https://mycroft-ai.gitbook.io/docs/using-mycroft-ai/get-mycroft) so we don't need to maintain this in two locations.

README.md
```diff
@@ -32,29 +32,49 @@ Mycroft is a hackable open source voice assistant.
...
 
-NOTE: The default branch for this repository is 'dev', which should be considered a work-in-progress. If you want to clone a more stable version, switch over to the 'master' branch.
+**NOTE:** The default branch for this repository is 'dev', which should be considered a work-in-progress. If you want to clone a more stable version, switch over to the 'master' branch.
 
 ## Running Mycroft
 
-Mycroft provides `start-mycroft.sh` to perform common tasks. This script uses a virtualenv created by `dev_setup.sh`.  Assuming you installed mycroft-core in your home directory run:
-- `cd ~/mycroft-core`
-- `./start-mycroft.sh debug`
+Mycroft provides `mycroft-start` to perform common tasks. Assuming you installed mycroft-core via your systems package manager or via `setup.py`:
```
krisgesling, line 60: I'd say something like: 
> Assuming you added the `mycroft-core/bin` directory to your $PATH or installed it as a Python package.

dev_setup.sh
```diff
@@ -638,15 +594,5 @@ chmod +x bin/mycroft-say-to
...
 
-# create and set permissions for logging
-if [[ ! -w /var/log/mycroft/ ]] ; then
-    # Creating and setting permissions
-    echo 'Creating /var/log/mycroft/ directory'
-    if [[ ! -d /var/log/mycroft/ ]] ; then
-        $SUDO mkdir /var/log/mycroft/
-    fi
-    $SUDO chmod 777 /var/log/mycroft/
-fi
```
krisgesling: Do we need to create the new directories - looks like this gets done in `bin/mycroft-start` but wouldn't if people are following the `dev_setup.sh` > `start-mycroft.sh` path?

krisgesling (changes requested):
Hey I like the changes here, I think it's just the README that I think we need to work through. 

At a foundational level, I don't think we should consider these "non-development setups". The only non-development installations that exist (in my opinion) are the Mark 1 and the Mark II.

I'm extremely cautious about pointing people to distro packaging too. We haven't had a great experience of it in the past which I'm sure you've heard about. Clearly you're the big exception, however we still have no visibility over how this is managed, how updates are made, and what the final experience looks like for users. Eg if I'm a Ubuntu user, my first inclination is to try apt install it, Ubuntu instead points to the Snap. I install the Snap and I have a horrible experience. It was built by the community with the best intentions and we took a quick stab at bringing it up to speed but it needs time and focus to get it there. In the end all we get are pissed off people that don't take another look at the project.

I'm sure Alpine is much better, however I can't imagine that it's a polished desktop experience because that just doesn't exist yet. Desktop hasn't been a focus for us because it requires a complete UX design process, and as you know our attention is on the Mark II right now. You can use the basic functions of Mycroft on a PC (or a phone) but it's setting up expectations with users that it's intended for those platforms, and that's just not the case yet. 

This is still great stuff, and it should help us to work with distro maintainers and other people integrating Mycroft for us to release complete products in the future. I just wanted to be clear why I don't want us to present the software as a "ready-to-use product" in this readme.

dev_setup.sh
PureTryOut: Seems I missed that part! It should also log to `$XDG_STATE_HOME/mycroft`, changed it!

README.md
PureTryOut: I replaced the whole section to just point to the handbook instead :+1:

PureTryOut:
> At a foundational level, I don't think we should consider these "non-development setups". The only non-development installations that exist (in my opinion) are the Mark 1 and the Mark II.

That's a shame, I think Mycroft has a lot of potential running on phones and desktops as well.

> I'm extremely cautious about pointing people to distro packaging too. We haven't had a great experience of it in the past which I'm sure you've heard about. Clearly you're the big exception, however we still have no visibility over how this is managed, how updates are made, and what the final experience looks like for users. Eg if I'm a Ubuntu user, my first inclination is to try apt install it, Ubuntu instead points to the Snap. I install the Snap and I have a horrible experience. It was built by the community with the best intentions and we took a quick stab at bringing it up to speed but it needs time and focus to get it there. In the end all we get are pissed off people that don't take another look at the project.

I get that 100%. It's a shame these things happened in the past and it makes distributions look bad. I like to think not every distribution is like that but even some big names like openSUSE still have a very old version packaged (18.8.13 in their rolling distribution!) and don't even respond to emails asking about it.

I suppose as a project you don't have to tell the user about any distro packaging at all. You could provide a PyPi package which you maintain yourself (and thus have full control and insight over!), and if the user wants to get it from distro packaging instead they can look for it for themselves.

> I'm sure Alpine is much better, however I can't imagine that it's a polished desktop experience because that just doesn't exist yet. Desktop hasn't been a focus for us because it requires a complete UX design process, and as you know our attention is on the Mark II right now. You can use the basic functions of Mycroft on a PC (or a phone) but it's setting up expectations with users that it's intended for those platforms, and that's just not the case yet.

I get that your priorities are elsewhere but I already consider Mycroft quite usable on desktop, especially in combination with the widget provided by KDE Plasma. You don't have to promote it, but I wouldn't completely dismiss it either :wink: 

> This is still great stuff, and it should help us to work with distro maintainers and other people integrating Mycroft for us to release complete products in the future. I just wanted to be clear why I don't want us to present the software as a "ready-to-use product" in this readme.

That's fine! I'm happy your not dismissing it and I hope you're fine with me improving the situation for Alpine Linux/postmarketOS users wherever I can at least!

I applied the changes you suggested so this should be ready to go!

krisgesling:
Hey, I certainly don't think people shouldn't run Mycroft in other places. I agree there is huge potential there. It's just a question of what we consider a shipable product that we promote to non-techies.

Back to this PR though, I gave it a test run and hit three issues.

1. Looks like we need to update the default log location in at least one more spot:
```
mycroft/client/text/__main__.py
```
This is what gets called when you run the CLI client.

2. When stopping services I get:
```
mycroft-core/.venv/bin/mycroft-stop: 53: kill: Illegal option -S
```

3. If using a bare `mycroft-stop` you get a shift error. Now is requiring `mycroft-stop all`

Are you seeing the same?

PureTryOut:
> Looks like we need to update the default log location in at least one more spot:

Good call, done.

> ```
> mycroft-core/.venv/bin/mycroft-stop: 53: kill: Illegal option -S
> ```

Strange, that option isn't used anymore (it was in the past but replaced for `-SIGINT` due to afaik Debian's default kill not supporting it). It was however present still in `stop-mycroft.sh`, where I replaced it now.

> If using a bare mycroft-stop you get a shift error. Now is requiring mycroft-stop all

I'm not seeing the same no, for me this works just fine. Can you post the error you're getting?

PureTryOut:
@krisgesling it's been a few months again... Could you either approve/merge this or tell me what to do to get it merged?

krisgesling:
Hi Bart, this one falls in the same category as #2803

What this really needs is someone to comprehensively test it out and stake their reputation on it being solid. Then we can merge it in. 

If any core contributor community members have time to review and test this it would be much appreciated.

forslund:
I can try to make some time this weekend and during next week to test.

PureTryOut:
Well I'm willing to stake my reputation on it being solid, but I'm not a core contributor :stuck_out_tongue_winking_eye: 

Thanks for the offer @forslund! Please let me know when you need something from me, I'm of course available on Mattermost.

forslund:
Not sure I'm a core community contributor anymore but I should be able to do some tests :)

krisgesling:
Thanks Ake!

I would consider you both core community contributors, and Bart - if it was someone else's PR I'd definitely accept your review for a merge. I just believe that it's good practice to have someone other than the original author review and test PRs. Their fresh perspective can often see an assumption that we made or exercise it in a slightly different way than we expected.

bin/mycroft-stop
```diff
@@ -14,8 +14,107 @@
...
+		return 1
+	fi
+}
+
+end_process() {
+	if process_running "$1"; then
+		# Find the process by name, only returning the oldest if it has children
+		pid=$( pgrep -o -f "python3 (.*)-m mycroft.*${1}" )
+		echo "Stopping $1 (${pid})..."
+		kill -SIGINT "${pid}"
```
forslund: For me this doesn't work:
```
53: kill: Illegal option -S
```
Should it be `kill -s SIGINT` ?

PureTryOut: I believe I changed it from that because that didn't work for someone else... I have no problem changing it back though, both options work for me. Changed.

bin/mycroft-stop
```diff
@@ -14,8 +14,107 @@
...
+		return 1
+	fi
+}
+
+end_process() {
+	if process_running "$1"; then
+		# Find the process by name, only returning the oldest if it has children
+		pid=$( pgrep -o -f "python3 (.*)-m mycroft.*${1}" )
+		echo "Stopping $1 (${pid})..."
+		kill -s SIGINT "${pid}"
```
forslund: Ugh, This still gives me issues apparently with my sh (dash) it needs to be `kill -s INT`

Not sure what is the correct thing here if my shell is outdated and shouldn't be used or if `-s INT` is the canonical always safe bet.

bin/mycroft-start
```diff
@@ -14,8 +14,151 @@
...
+	launch_background bus
+	launch_background skills
+	launch_background audio
+	launch_background voice
+	launch_background enclosure
+}
+
+_opt=$1
+_force_restart=false
+shift
```
forslund: This causes an error if there are no arguments. Maybe handling for the empty case should be added Like in [start-mycroft.sh L183](https://github.com/MycroftAI/mycroft-core/blob/dev/start-mycroft.sh#L183) or maybe like in [stop-mycroft.sh](https://github.com/MycroftAI/mycroft-core/blob/dev/stop-mycroft.sh#L91) allowing the switch-case handle the default?

bin/mycroft-stop
```diff
@@ -14,8 +14,107 @@
...
+				result=100
+			fi
+		fi
+	fi
+}
+
+result=0 # default, no change
+
+OPT=$1
+shift
```
forslund: This gives error if there are no arguments, maybe add a check similar to [stop-mycroft.sh L91](https://github.com/MycroftAI/mycroft-core/blob/dev/stop-mycroft.sh#L91)

forslund (commented):
Ugh, the SIGINT signal name doesn't seem to be the correct one for me, sorry. It seems atleast for dash sh (Ubuntu 20.04) it needs to be `-s INT` for it to work properly... (same as current stop-mycroft.sh) Not sure if it's an issue with my setup or if -s INT is the most compatible argument...

Also saw some issues when forgetting to add arguments to the mycroft-start/stop scripts...I don't think they're a big issue but might be nice.

bin/mycroft-stop
PureTryOut: `kill -s INT` works for me :shrug:  Let's just go with it

bin/mycroft-start
PureTryOut: What kind of error? For me `bin/mycroft-start` (so without any extra arguments) works just fine?

forslund: I'm not at the computer but I think it was

```
shift: can't shift that many
```

Edit: at a computer now and yes above is the error message at L122

forslund (approved):
The blocker with the kill command in the stop script has been resolved. There still is the issue with passing no arguments to the start / stop scripts but I don't think it's really a blocker.

bin/mycroft-start
PureTryOut: Interesting. Shells are weird :shrug:  Fixed

PureTryOut:
I fixed that too like you suggested, should be good to go!

PureTryOut:
@krisgesling it's ack'ed by @forslund, should be good to go now.

Issue #2799, opened 2021-01-13
Python 3.9+ issues


forslund:
@j1nx reported that Exceptions were raised during Padatious training on python 3.9 similar to

```
Exception in thread Thread-8:
Traceback (most recent call last):
  File "/usr/lib/python3.9/threading.py", line 950, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.9/threading.py", line 888, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3.9/site-packages/padatious/training_manager.py", line 98, in train
    pool = mp.Pool()
  File "/usr/lib/python3.9/multiprocessing/context.py", line 119, in Pool
    return Pool(processes, initializer, initargs, maxtasksperchild,
  File "/usr/lib/python3.9/multiprocessing/pool.py", line 191, in __init__
    self._setup_queues()
  File "/usr/lib/python3.9/multiprocessing/pool.py", line 343, in _setup_queues
    self._inqueue = self._ctx.SimpleQueue()
  File "/usr/lib/python3.9/multiprocessing/context.py", line 113, in SimpleQueue
    return SimpleQueue(ctx=self.get_context())
  File "/usr/lib/python3.9/multiprocessing/queues.py", line 342, in __init__
    self._rlock = ctx.Lock()
  File "/usr/lib/python3.9/multiprocessing/context.py", line 67, in Lock
    from .synchronize import Lock
ImportError: cannot import name 'Lock' from partially initialized module 'multiprocessing.synchronize' (most likely due to a circular import) (/usr/lib/python3.9/multip
rocessing/synchronize.py)
```

The issue seems to be related to a change in how modules/packages are imported from python 3.9 and onwards See [bug report](https://bugs.python.org/issue41567).

A quick and dirty workaround for testing is available in [here](https://github.com/forslund/mycroft-core/tree/bugfix/mp-workaround). Basically it preloads the modules that may be incomplete when trying to create the Pool object.

It's a bit hard to tell exactly what is happening but one possible culprit is msm, which also use a multiprocessing Pool (threadpool but it's the same base class)...unless training is somehow launched in separate threads as well as processes internally in padatious but that sounds farfetched.

Not sure what the most correct action is a couple of things that can be tested:
- Lock around msm.apply calls and padatious training
- Move the execution of training to the same thread as the skill download (SkillManager thread)

**To Reproduce**
1. Install mycroft with a python3.9 environment
2. Startup mycroft
[...]

PR #2803, opened 2021-01-15
Read skills from XDG Base directory specification locations

PureTryOut:
## Description
~This makes mycroft-core read skills from all locations as specified by the XDG Base Directory specification, so in `/usr/share/mycroft/skills`, `/usr/local/share/mycroft/skills`, and `$XDG_DATA_HOME/.local/share/mycroft/skills` if it exists and otherwise `~/.local/share/mycroft/skills`.~

This makes mycroft-core read skills from the XDG Base Directory location in the user's home folder, so `$XDG_DATA_HOME/.local/share/mycroft/skills` if it exists and otherwise `~/.local/share/mycroft/skills`.

This requires https://github.com/MycroftAI/mycroft-skills-manager/pull/90 for the installation of skills part.

## How to test
1. Apply https://github.com/MycroftAI/mycroft-skills-manager/pull/90 to msm
2. Install some skills
3. Start Mycroft and check if skills in the required directories are being loaded properly

CLA [x]

### DIFF
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -106,13 +106,11 @@
       }
     },
     "upload_skill_manifest": true,
-    // Directory to look for user skills
-    "directory": "~/.local/share/mycroft/skills",
     // Enable auto update by msm
     "auto_update": true,
     // blacklisted skills to not load
     // NB: This is the basename() of the directory where the skill lives, so if
-    // the skill you want to blacklist is in /opt/mycroft/skills/mycroft-alarm.mycroftai/
+    // the skill you want to blacklist is in $XDG_DATA_DIR/mycroft/skills/mycroft-alarm.mycroftai/
     // then you should write `["mycroft-alarm.mycroftai"]` below.
     "blacklisted_skills": [],
     // priority skills to be loaded first
--- a/mycroft/skills/skill_manager.py
+++ b/mycroft/skills/skill_manager.py
@@ -18,6 +18,7 @@ from glob import glob
 from threading import Thread, Event, Lock
 from time import sleep, time, monotonic
 from inspect import signature
+import xdg.BaseDirectory
 
 from mycroft.api import is_paired
 from mycroft.enclosure.api import EnclosureAPI
@@ -264,7 +265,9 @@ class SkillManager(Thread):
 
     def _remove_git_locks(self):
         """If git gets killed from an abrupt shutdown it leaves lock files."""
-        for i in glob(os.path.join(self.msm.skills_dir, '*/.git/index.lock')):
+        for i in glob(os.path.join(
+            xdg.BaseDirectory.save_data_path('mycroft/skills'),
+                '*/.git/index.lock')):
             LOG.warning('Found and removed git lock file: ' + i)
             os.remove(i)
 
@@ -310,7 +313,8 @@ class SkillManager(Thread):
         return skill_loader if load_status else None
 
     def _get_skill_directories(self):
-        skill_glob = glob(os.path.join(self.msm.skills_dir, '*/'))
+        skill_glob = glob(os.path.join(
+            xdg.BaseDirectory.save_data_path('mycroft/skills'), '*/'))
 
         skill_directories = []
         for skill_dir in skill_glob:
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -26,6 +26,8 @@ ROOT_DIRNAME=$(dirname "$0")
 cd "$ROOT_DIRNAME"
 TOP=$(pwd -L)
 
+datadir=${XDG_DATA_HOME:-$HOME/.local/share}/mycroft
+
 function clean_mycroft_files() {
     echo '
 This will completely remove any files installed by mycroft (including pairing
@@ -39,6 +41,9 @@ Do you wish to continue? (y/n)'
         read -rN1 -s key
         case $key in
         [Yy])
+	    rm -r "$datadir"
+
+	    # Legacy locations
             sudo rm -rf /var/log/mycroft
             rm -f /var/tmp/mycroft_web_cache.json
             rm -rf "${TMPDIR:-/tmp}/mycroft"
@@ -108,7 +113,7 @@ fi
 
 for var in "$@" ; do
     # Check if parameter should be read
-    if [[ $param == 'python' ]] ; then
+	if [[ $param == 'python' ]] ; then
         opt_python=$var
         param=""
         continue
@@ -304,20 +309,10 @@ fi" > ~/.profile_mycroft
     # Create a link to the 'skills' folder.
     sleep 0.5
     echo
-    echo 'The standard location for Mycroft skills is under /opt/mycroft/skills.'
-    if [[ ! -d /opt/mycroft/skills ]] ; then
-        echo 'This script will create that folder for you.  This requires sudo'
-        echo 'permission and might ask you for a password...'
-        setup_user=$USER
-        setup_group=$(id -gn "$USER")
-        $SUDO mkdir -p /opt/mycroft/skills
-        $SUDO chown -R "${setup_user}":"${setup_group}" /opt/mycroft
-        echo 'Created!'
-    fi
     if [[ ! -d skills ]] ; then
-        ln -s /opt/mycroft/skills skills
+        ln -s "$datadir"/skills skills
         echo "For convenience, a soft link has been created called 'skills' which leads"
-        echo 'to /opt/mycroft/skills.'
+        echo "to $datadir/skills."
     fi
 
     # Add PEP8 pre-commit hook
--- a/mycroft/skills/msm_wrapper.py
+++ b/mycroft/skills/msm_wrapper.py
@@ -22,6 +22,7 @@ frequently.  To improve performance, the MSM instance is cached.
 from collections import namedtuple
 from functools import lru_cache
 from os import path, makedirs
+import xdg.BaseDirectory
 
 from msm import MycroftSkillsManager, SkillRepo
 
@@ -34,9 +35,8 @@ MsmConfig = namedtuple(
     [
         'platform',
         'repo_branch',
-        'repo_cache',
         'repo_url',
-        'skills_dir',
+        'old_skills_dir',
         'versioned'
     ]
 )
@@ -71,9 +71,8 @@ def build_msm_config(device_config: dict) -> MsmConfig:
     return MsmConfig(
         platform=enclosure_config.get('platform', 'default'),
         repo_branch=msm_repo_config['branch'],
-        repo_cache=path.join(data_dir, msm_repo_config['cache']),
         repo_url=msm_repo_config['url'],
-        skills_dir=path.join(data_dir, msm_config['directory']),
+        old_skills_dir=path.join(data_dir, msm_config['directory']),
         versioned=msm_config['versioned']
     )
 
@@ -95,17 +94,15 @@ def create_msm(msm_config: MsmConfig) -> MycroftSkillsManager:
     msm_lock = _init_msm_lock()
     LOG.info('Acquiring lock to instantiate MSM')
     with msm_lock:
-        if not path.exists(msm_config.skills_dir):
-            makedirs(msm_config.skills_dir)
+        xdg.BaseDirectory.save_data_path('mycroft/skills')
 
         msm_skill_repo = SkillRepo(
-            msm_config.repo_cache,
             msm_config.repo_url,
             msm_config.repo_branch
         )
         msm_instance = MycroftSkillsManager(
             platform=msm_config.platform,
-            skills_dir=msm_config.skills_dir,
+            old_skills_dir=msm_config.old_skills_dir,
             repo=msm_skill_repo,
             versioned=msm_config.versioned
         )
--- a/mycroft/skills/mycroft_skill/mycroft_skill.py
+++ b/mycroft/skills/mycroft_skill/mycroft_skill.py
@@ -124,7 +124,7 @@ class MycroftSkill:
 
         # Get directory of skill
         #: Member variable containing the absolute path of the skill's root
-        #: directory. E.g. /opt/mycroft/skills/my-skill.me/
+        #: directory. E.g. $XDG_DATA_HOME/mycroft/skills/my-skill.me/
         self.root_dir = dirname(abspath(sys.modules[self.__module__].__file__))
 
         self.gui = SkillGUI(self)
--- a/mycroft/skills/settings.py
+++ b/mycroft/skills/settings.py
@@ -99,7 +99,7 @@ def save_settings(skill_dir, skill_settings):
     """Save skill settings to file."""
     settings_path = Path(skill_dir).joinpath('settings.json')
 
-    # Either the file already exists in /opt, or we are writing
+    # Either the file already exists or we are writing
     # to XDG_CONFIG_DIR and always have the permission to make
     # sure the file always exists
     if not Path(settings_path).exists():
--- a/mycroft/skills/skill_updater.py
+++ b/mycroft/skills/skill_updater.py
@@ -55,7 +55,8 @@ class SkillUpdater:
         self.config = Configuration.get()
         update_interval = self.config['skills']['update_interval']
         self.update_interval = int(update_interval) * ONE_HOUR
-        self.dot_msm_path = os.path.join(self.msm.skills_dir, '.msm')
+        self.dot_msm_path = os.path.join(
+                xdg.BaseDirectory.save_data_path('mycroft/skills'), '.msm')
         self.next_download = self._determine_next_download_time()
         self._log_next_download_time()
         self.installed_skills = set()
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,3 +1,5 @@
 [pytest]
 testpaths = test
 norecursedirs = wake_word
+env =
+    XDG_DATA_HOME=/tmp/mycroft-test
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -15,7 +15,7 @@ fasteners==0.14.1
 PyYAML~=5.4
 
 lingua-franca==0.4.2
-msm==0.8.9
+msm==0.9.0
 msk==0.3.16
 mycroft-messagebus-client==0.9.4
 adapt-parser==0.5.1
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -2,6 +2,7 @@ coveralls==1.8.2
 flake8==3.7.9
 pytest==5.2.4
 pytest-cov==2.8.1
+pytest-env==0.6.2
 cov-core==1.15.0
 sphinx==2.2.1
 sphinx-rtd-theme==0.4.3
--- a/test/integrationtests/voight_kampff/test_setup.py
+++ b/test/integrationtests/voight_kampff/test_setup.py
@@ -204,7 +204,10 @@ def create_skills_manager(platform, skills_dir, url, branch):
         MycroftSkillsManager
     """
     repo = SkillRepo(url=url, branch=branch)
-    return MycroftSkillsManager(platform, skills_dir, repo)
+    return MycroftSkillsManager(
+            platform=platform,
+            skills_dir=skills_dir,
+            repo=repo)
 
 
 def main(args):
--- a/test/unittests/base.py
+++ b/test/unittests/base.py
@@ -18,6 +18,8 @@ from shutil import rmtree
 from unittest import TestCase
 from unittest.mock import patch
 
+import xdg.BaseDirectory
+
 from .mocks import mock_msm, mock_config, MessageBusMock
 
 
@@ -54,3 +56,4 @@ class MycroftUnitTestBase(TestCase):
 
     def tearDown(self):
         rmtree(str(self.temp_dir))
+        rmtree(xdg.BaseDirectory.save_data_path('mycroft'))
--- a/test/unittests/skills/test_skill_manager.py
+++ b/test/unittests/skills/test_skill_manager.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 #
 from os import path
+from pathlib import Path
+import xdg.BaseDirectory
 from unittest import TestCase
 from unittest.mock import Mock, patch
 
@@ -95,7 +97,10 @@ class TestSkillManager(MycroftUnitTestBase):
         self.skill_updater_mock = skill_updater_patch.start()
 
     def _mock_skill_loader_instance(self):
-        self.skill_dir = self.temp_dir.joinpath('test_skill')
+        self.skill_dir = (
+            Path(xdg.BaseDirectory.save_data_path('mycroft/skills'))
+            .joinpath('test_skill')
+        )
         self.skill_loader_mock = Mock(spec=SkillLoader)
         self.skill_loader_mock.instance = Mock()
         self.skill_loader_mock.instance.default_shutdown = Mock()
@@ -128,8 +133,9 @@ class TestSkillManager(MycroftUnitTestBase):
         )
 
     def test_remove_git_locks(self):
-        git_dir = self.temp_dir.joinpath('foo/.git')
-        git_dir.mkdir(parents=True)
+        git_dir = Path(
+                xdg.BaseDirectory.save_data_path('mycroft/skills/foo/.git')
+            )
         git_lock_file_path = str(git_dir.joinpath('index.lock'))
         with open(git_lock_file_path, 'w') as git_lock_file:
             git_lock_file.write('foo')
--- a/test/unittests/skills/test_skill_updater.py
+++ b/test/unittests/skills/test_skill_updater.py
@@ -247,10 +247,11 @@ class TestSkillUpdater(MycroftUnitTestBase):
 
     def test_update_download_time(self):
         """Test updating the next time a download will occur."""
-        dot_msm_path = self.temp_dir.joinpath('.msm')
-        dot_msm_path.touch()
-        dot_msm_mtime_before = dot_msm_path.stat().st_mtime
+        skill_updater = SkillUpdater(self.message_bus_mock)
+        skill_updater.dot_msm_path = self.temp_dir.joinpath('.msm')
+        skill_updater.dot_msm_path.touch()
+        dot_msm_mtime_before = skill_updater.dot_msm_path.stat().st_mtime
         sleep(0.5)
-        SkillUpdater(self.message_bus_mock)._update_download_time()
-        dot_msm_mtime_after = dot_msm_path.stat().st_mtime
+        skill_updater._update_download_time()
+        dot_msm_mtime_after = skill_updater.dot_msm_path.stat().st_mtime
         self.assertLess(dot_msm_mtime_before, dot_msm_mtime_after)


devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

mycroft/configuration/mycroft.conf
```diff
@@ -95,8 +95,6 @@
   // General skill values
   "skills": {
     "msm": {
-      // Relative to "data_dir"
-      "directory": "skills",
```
JarbasAl, line 99: we should support this at least during next major version, i personally use a changed .conf to set the data_dir variable and expect my skills there. My use case is having a dev environment with different skills. If they all come from xdg directories i lose this isolation

mycroft/skills/skill_manager.py
```diff
@@ -308,13 +310,18 @@ def _load_skill(self, skill_directory):
         return skill_loader if load_status else None
 
     def _get_skill_directories(self):
-        skill_glob = glob(os.path.join(self.msm.skills_dir, '*/'))
+        skill_glob = []
+        for directory in BaseDirectory.load_data_paths('mycroft/skills'):
```
JarbasAl, line 314: as mentioned in comment above we should support the old mechanism also, ideally core would check and monitor all skill directories instead of a single path, old location (configurable in .conf, can be set by user) + xdg

mycroft/configuration/mycroft.conf
JarbasAl: to expand on this, the default data_dir in .conf should be changed to the correct xdg location

just like this fedora packager did https://gitlab.com/LyesSaadi/spec/-/blob/master/mycroft/mycroft/mycroft.conf

mycroft/skills/skill_manager.py
PureTryOut: What would you set the default location too then? My goal is to get rid of `/opt` by default, using just XDG accomplishes that.

What about `if old location exists in conf, use that (useful for dev purposes), otherwise use xdg`?

mycroft/configuration/mycroft.conf
PureTryOut: > the default data_dir in .conf should be changed to the correct xdg location

There is no such thing though. The "correct" XDG location is either `$XDG_DATA_HOME/mycroft/skills` if `$XDG_DATA_HOME` exists, and otherwise `~/.local/share/mycroft/skills`. The whole point of the specification is that you can't assume `~/.local/share`.

JarbasAl: you can assume whatever place msm is installing things into, its just a default, as i said the point is allowing users to override it / be backwards compatible

mycroft/skills/skill_manager.py
JarbasAl: we should support several directories by design, 
xdg allows :
-system level skills (all users) 
-skills for specific users

and .conf override allows:
- installation specific skills
- multiple non conflicting mycroft-core installs

mycroft/configuration/mycroft.conf
JarbasAl: oh i think i misunderstood, probably the best way is to support a unique string, eg, `data_dir: "{XDG}"` which would be resolved by the code itself

mycroft/skills/skill_manager.py
JarbasAl: also note that this was originally planned but not implemented, the .conf contains

```
// Directory to look for user skills
"directory": "~/.mycroft/skills",
```

since you are touching this maybe add support for that (i still think data_dir is cleaner) or remove it from .conf since it's wrong info

PureTryOut: So I guess set default value to `{XDG}` which code checks and parses to proper XDG directories, and if something else just use that straight up?

PureTryOut: I'm thinking, this kinda conflicts with what I'm doing in #2578 as there I'm getting rid of the `data_dir` variable where possible. However, I do see value in allowing to keep that variable and allowing to set it to `{XDG}` by default so you can keep a dev setup and daily use setup separated. In your dev setup just set the variable to some other location and bam, completely separate configs and data/skills.
That is assuming you can use a different config for a dev setup altogether, which I'm not sure how to handle.

JarbasAl: also for backwards compatibility data_dir should no be deprecated (at least without a warning), i use it on ovos_utils and ovos_skills_manager for example

you raise a valid point about config, i usually edit the default .conf in mycroft-core for the specific dev setup, would be good to allow the user to set the path for user_config also, so by default it would also use `{XDG}` but individual installs could override it, but thats probably a separate PR

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

mycroft/skills/skill_manager.py
PureTryOut: So currently I move any skills found in the location as specified in the config to the new XDG _user_ location. Is that alright with you?

ChanceNCounter: IMO, avoid moving skills and data on existing installs. Too much room for deployment failure. Set it up so future installs will be XDG-compliant by default, and if you're monitoring multiple skill folders anyway, begin installing *new* skills in the XDG-compliant locations.

Just my two cents.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

mycroft/skills/skill_manager.py
PureTryOut: I don't think I agree. That way we can never really get rid of `data_dir` as there might always be people who haven't moved over yet. For configuration I understand, people need to access those directly in the filesystem sometimes to edit them. But for data like this, nobody is manually editing anything in the skills. Maybe skill developers, but they can just find their skill at the new location.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

mycroft/skills/skill_manager.py
PureTryOut: As discussed on Mattermost, using the XDG env variables properly you can keep a dev setup (or other Mycroft instances) separated from your regular installation. This will need proper documentation of course, but we have no need anymore for a config value anymore. Resolving this conversation.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

krisgesling:
Was just glancing at this as we were looking at the PR for MSM and I wondered if we also need to update `scripts/prepare-msm.sh` which seems to also have the `/opt/mycroft/skills` path hardcoded

PureTryOut:
> I wondered if we also need to update `scripts/prepare-msm.sh` which seems to also have the `/opt/mycroft/skills` path hardcoded

Good call, yes it needs to be. Fixed.

scripts/prepare-msm.sh
```diff
@@ -27,39 +30,9 @@ while [ -h "$SOURCE" ] ; do
...
-fi
-
-# change ownership of ${mycroft_root_dir} to ${setup_user } recursively
-function change_ownership {
-    echo "Changing ownership of" ${mycroft_root_dir} "to user:" ${setup_user} "with group:" ${setup_group}
-    $SUDO chown -Rf ${setup_user}:${setup_group} ${mycroft_root_dir}
-}
-
-
 if [[ ${CI} != true ]] ; then
```
PureTryOut, line 33: I wonder, is this condition still required?
No matter if we're running on CI or not, making sure the skills directory exists doesn't hurt.

Getting rid of this means we can get rid of the sourcing and changing directories above this, and _that_ means we can run this script on a pure POSIX shell.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

scripts/prepare-msm.sh
PureTryOut: Actually never mind, this script isn't needed at all anymore. The new skill directories don't require root rights and will be made by the pyxdg dep if they don't exist yet, so no need preparing anything. I'll just remove the script.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

PureTryOut:
I made it depend on msm 0.9.0 now.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

j1nx:
I have tested this as part of #2799 and works perfectly and as expected.

- Skills are downloaded to /home/<user>/.local/share/mycroft/skills/
- Skills repo is initiated within /home/<user>/.local/share/mycroft/skills-repo   (no more hidden directory)

The only little quirck I had was that at a very virgin system where the user does not yet have a "site-packages" within his/her ~/.local/share/lib/ folder, starting mycroft; At the start of Python as the users site-packages are not there yet, it will not be loaded. packages will be installed to it with pip, however mycroft will not see them before a restart as that folder was not initialized at its start. (If this makes any sense) Anyhow, I believe this is more a MSM thing then mycroft-core.

Anyhow: In short. Works as expected!

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2803)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2803/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2803/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2803/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2803/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2803/logs/enclosure.log)

chrisveilleux:
I am curious as to why a skill would go in a data directory.  Skills are not data, they are executable Python code.

PureTryOut:
I agree, and thus I wonder why it was in `/opt` before. The only _right_ place for Python code would be installed as regular Python packages to e.g. `~/.local/lib/python<version>/site-packages`. However, that's not how you guys seem to do things. Thus I opted to act as if it were data, and put it on the right place for that.

krisgesling:
Perhaps the missing bit of context here is the **why**, as I'm sure that it's not just to meet the XDG standard. 
Why do we want to move them from their current location to a new location? 
What problems exist from them being in `/opt`, and/or what will be better once they are moved?

I know there are _better_ places to put things, and there are certainly _wrong_ places to put things, but I don't see `/opt` being in the _wrong_ category. 

The filesystem standard says:
> /opt is reserved for the installation of add-on application software packages.

Skills would seem to fit that bill. Having them here does however mean you have a single set of Skills for all users of a system. For devices like the Mark II this isn't a problem - though I could see a future where you can switch users to get your set of Skills and settings.

Whilst I don't think of them as _data_ files, if I was going to drop them somewhere in my home directory, I can't think of anywhere better than `~/.local/share/mycroft/skills`. At least until Skills can be distributed as complete Python packages.

There's also the potential permissions issues of using `/opt`. Using `$XDG_DATA_HOME` should always be writable by the current user.

Is there anything else?

PureTryOut:
Good comment! I'm thinking mainly in the case of multi-user scenario's, and most importantly permission issues.

Not every user on a system might want to use Mycroft the same, and each could want their own set of skills installed. Originally I had system-wide installed skills alongside user skills planned but that caused some issues, but that would be my ideal scenario: have some common base skills installed/required by the system administrator, and allow every user to choose their own extra skills.

And currently to install Mycroft you need to run some script with root rights, which would be completely unnecessary if every XDG PR was merged alongside this one. Suddenly users without administrator rights would still be able to use Mycroft by just installing it to their home directory instead (`pip install --user mycroft-core`).

In the end it would probably be best to handle skills as regular Python packages but since that is not possible for now, a data location seems to be the most logical. Then again if we want we could also install it to e.g. `~/.local/lib/mycroft/skills` or something since these kind of files are not part of XDG anyway. Then we'll just have to make sure that directory exists ourselves.

ChanceNCounter:
I think handling Skills as code, although using a skill manager renders the point moot for end users, would make it more difficult than it needs to be for brand new user-devs to find and alter them.

> Skills would seem to fit that bill. Having them here does however mean you have a single set of Skills for all users of a system. For devices like the Mark II this isn't a problem - though I could see a future where you can switch users to get your set of Skills and settings.

True! But for standalone devices, the change is painless; the `mycroft` user owns it, or etc. as applicable. Easy peasy.

And I think that's the biggest thing - consider ownership from a Unixy perspective. On a smart speaker, `mycroft` owns Skills. On a general purpose computer, the primary user is the primary owner.

XDG also provides the corresponding "overrides." You and your kid might have completely different Skills, and some of those might conflict with each other. Maybe your kid has a Chatterbox, and they make their own Skills with the visual editor. Maybe their Skills would look very nice on the refrigerator where everyone can see them.

Maybe, when you're using the family computer - and once direct user recognition is implemented 🤞! - you'd rather have the default weather Skill, rather than an 8-year-old's sense of humor. Already done.

chrisveilleux:
> The only _right_ place for Python code would be installed as regular Python packages to e.g. `~/.local/lib/python<version>/site-packages`.

My concern with installing them to `~/.local/lib/python<version>/site-packages` (when everything becomes PIP installable) is that it precludes the utilization of a virtual environment.  Worrying about whether required packages for core or skills conflict with other code utilizing a user-level Python interpreter seems like it could turn into a real pain in the arse.  

This leads to another question... how do we make a virtual environment XDG compliant?  On my laptop, my virtual environments are stored in `~/.local/share/virtualenvs`.

> In the end it would probably be best to handle skills as regular Python packages but since that is not possible for now, a data location seems to be the most logical. Then again if we want we could also install it to e.g. `~/.local/lib/mycroft/skills` or something since these kind of files are not part of XDG anyway. Then we'll just have to make sure that directory exists ourselves.

I would rather see the skills in a directory such as `~/.local/lib/mycroft/skills` than the data directory.  I think it will generally be less confusing than having code in a data directory.

> I think handling Skills as code, although using a skill manager renders the point moot for end users, would make it more difficult than it needs to be for brand new user-devs to find and alter them.

Brand new user-devs should be referring to our documentation on where skills live.  I think it is a fallacy to say any new user-dev would know to look in `/opt/mycroft/skills` without looking at the documentation.  I have a general objection to abandoning best practices to cater to the brand new user-dev.  How else is such a person to learn what best practices are other than seeing them in use?

ChanceNCounter:
> Brand new user-devs should be referring to our documentation on where skills live. I think it is a fallacy to say any new user-dev would know to look in /opt/mycroft/skills without looking at the documentation.

Of course not. The appropriate user's home and XDG directory structure, however, are easy to *track*. So is `/opt/mycroft`, it's just not canonical. If a skill is a module, it ends up at, say, `some_venv/lib/python3.X/site-packages/skill`. This is rough for Newbie.

This brings us to the distinction between dev and usage. User-devs can get around the above by installing their repo --editable in the Mycroft venv. Now they're manually dicking with the venv, as noob devs. Still no good for Newbie. Ideally, somebody ought to understand venvs and pip before they worry about how they've installed code in the venv.

Furthermore, if skills live under the user in general, the dev environment reflects prod. Skills are in userspace during development, and they're in userspace when somebody installs them.

bonus: does not break existing code that expands a path to find skills

>  I have a general objection to abandoning best practices to cater to the brand new user-dev. How else is such a person to learn what best practices are other than seeing them in use?

They're only best practices if you call a Skill a lib. If it's an applet, which is my position, then its location is a design decision. There's an effectively universal spec for stuff users own, and there are differing specs for stuff the system owns. The former is already preferable for the obvious reason, and I contend Skills belong to the system user who runs and/or "is" Mycroft.

chrisveilleux:
> They're only best practices if you call a Skill a lib. If it's an applet, which is my position, then its location is a design decision. There's an effectively universal spec for stuff users own, and there are differing specs for stuff the system owns. The former is already preferable for the obvious reason, and I contend Skills belong to the system user who runs and/or "is" Mycroft.

It seems to me that before a decision is made the default skills directory, a decision needs to be made if they should be stored at the user level or the system level.  I like the extra level of flexibility the user level provides.  I also like that there is less of a concern with needing sudo access to a directory like `/opt/mycroft/skills`.  If there are user-level instances of mycroft, where do the logs go?  Right now they are in the system-level `var/logs/mycroft` directory.  This would not be ideal if there were user-level instances of mycroft.

PureTryOut:
> It seems to me that before a decision is made the default skills directory, a decision needs to be made if they should be stored at the user level or the system level.

As I said before, my personal preference is both at the same time. But since that is impossible for now, the user level.

>  If there are user-level instances of mycroft, where do the logs go? Right now they are in the system-level var/logs/mycroft directory.

That is handled by #2802. It assumes users (rather than devs) will use `bin/mycroft-start` and it'll save the logs to `~/.local/state/mycroft` when `$XDG_STATE_HOME` is not set..

The various XDG things are split out over a total of 3 PR's: #2802, #2794 and this one, #2803. Each PR moves some stuff over to the user's home directory, but together no data is outside of `$HOME` anymore. No `/opt/mycroft`, and no `/var/log/mycroft`.

PureTryOut:
> This leads to another question... how do we make a virtual environment XDG compliant? On my laptop, my virtual environments are stored in ~/.local/share/virtualenvs.

So does that mean you want everything related to that virtual environment, so configs as well as data, to be stored inside that env? In that case exporting `XDG_CONFIG_HOME=~/.local/share/virtualenvs`, `XDG_DATA_HOME=~/.local/share/virtualenvs` and `XDG_STATE_HOME=~/.local/share/virtualenvs` before you launch MyCroft does the trick.

PureTryOut:
I still don't understand the unit test failure :thinking:

krisgesling:
I'm not sure the test is actually using that `temp_dir`. It might be better to set the `dot_msm_path` on the class itself:

This seems to work:
```python
    def test_update_download_time(self):
        """Test updating the next time a download will occur."""
        skill_updater = SkillUpdater(self.message_bus_mock)
        skill_updater.dot_msm_path = self.temp_dir.joinpath('.msm')
        skill_updater.dot_msm_path.touch()
        dot_msm_mtime_before = skill_updater.dot_msm_path.stat().st_mtime
        sleep(0.5)
        skill_updater._update_download_time()
        dot_msm_mtime_after = skill_updater.dot_msm_path.stat().st_mtime
        self.assertLess(dot_msm_mtime_before, dot_msm_mtime_after)
```

PureTryOut:
You seem to be right. I didn't understand why the test was failing as I didn't touch it, but now I actually looked at it it seems `dot_msm_path` was never used by the skill updater. So updating the download time did _something_ but definitely not to the `dot_msm_path`. Your patch makes the SkillUpdater however use the new path and so it works.

Now I just don't understand how this test ever worked before :thinking: 

Anyways, updated.

PureTryOut:
And now our good friend VK is failing again :angry:

PureTryOut:
@krisgesling any clue why VK is being difficult again? For some reason it seems to fail while sending emails, but the log only mentions warnings and no actual errors.

krisgesling:
It looks like this line:
https://github.com/MycroftAI/mycroft-core/blob/c5c44d22d1a1b9c682fb014b5a6c6d3ef50464a5/test/integrationtests/voight_kampff/test_setup.py#L207

should be:
```python
return MycroftSkillsManager(platform, skills_dir=skills_dir, repo=repo)
```

Presuming it shouldn't be `old_skills_dir`?

PureTryOut:
Ah, good find, you're right!

No it shouldn't be `old_skills_dir`, that's just for moving over legacy setups to the new directory, and not something Voight Kampff should care about (seeing it doesn't change behaviour of skills).

Hopefully it passes now, fingers crossed :crossed_fingers:

PureTryOut:
> Hopefully it passes now, fingers crossed crossed_fingers

Nope... Seems the volume skill still failed if I read it right?

krisgesling:
Hmm, if it's volume then #2992 might fix it. I'm at pycon today, but ll try have another look in a break.

krisgesling:
I've just been reading back through the thread and wanted to pull out the questions to make sure we cover everything.

### 1. User level vs system level?
The XDG changes as a whole will (by default) move everything out of the system level to the user level. 

### 2. Directory to use for Skills

There still seems to be concern about using a "data" directory for Skills. Unfortunately there aren't a large number of options to choose from. I think the Arch wiki provides a cleaner [overview of the XDG directories](https://wiki.archlinux.org/title/XDG_Base_Directory#User_directories), but there's also the [XDG specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) itself. Some relevant snippets:

#### `XDG_DATA_HOME`
- Where user-specific data files should be written (analogous to `/usr/share`).
- Should default to `$HOME/.local/share`.
 
####  `XDG_STATE_HOME`
- Where user-specific state files should be written (analogous to `/var/lib`).
- Should default to `$HOME/.local/state`.

> The $XDG_STATE_HOME contains state data that should persist between (application) restarts, but that is not important or portable enough to the user that it should be stored in $XDG_DATA_HOME. It may contain:
> 
>     actions history (logs, history, recently used files, …)
> 
>     current state of the application that can be reused on a restart (view, layout, open files, undo history, …)
> 
> User-specific executable files may be stored in $HOME/.local/bin. Distributions should ensure this directory shows up in the UNIX $PATH environment variable, at an appropriate place. 

There's also a PR on the documentation repo that includes a new page outlining the currently proposed XDG locations:
https://github.com/MycroftAI/documentation/pull/208/files#diff-80549a11e37425500fd598dd99e4d30e002b465510f57ca8836a9b2c8bf96b0a

> - `$XDG_CONFIG_HOME/mycroft` for configuration
> - `$XDG_DATA_HOME/mycroft` for persistent data storage
> - `$XDG_CACHE_HOME/mycroft` for non-essential data

Are there other possibilities that are worth considering?

### 3. How do we make a virtual environment XDG compliant?
Is this more of question for the future when we may have multiple virtual environments for different parts of the system?
The mycroft venv will remain in the mycroft-core directory.

### Any other questions I missed?

PureTryOut:
> 3. How do we make a virtual environment XDG compliant?

I'm not sure why this keeps being asked. I assume people just want their config and data separate depending on the virtual environment used, and that's perfectly possible by setting the env values properly.

With this PR included MyCroft makes use of `$XDG_CONFIG_HOME`, `$XDG_DATA_HOME` and `$XDG_CACHE_HOME` (also `$XDG_STATE_HOME` once #2802 is in). Just set those to the directory you want your settings and data to be saved in and voilá. This also makes it possible to have separate skills per virtual environment as that just follows `$XDG_DATA_HOME` (currently anyway, maybe we decide differently but it has my preference for now).

krisgesling:
Just hit an issue in MSK - we'll need to do an updated release of that to use XDG too. 

Posting here so I don't forget

krisgesling:
Hey Bart, I haven't had time to look at MSK and it's not in our next few sprints so it will likely be some time before I can get to it. 

Is this something you'd have time to look at?

PureTryOut:
If you could link me to the issue in MSK, maybe?

krisgesling:
Thanks! I know working on MSK is not what you set out to do.

I've created a new issue and linked it to this PR:
https://github.com/MycroftAI/mycroft-skills-kit/issues/62

I think it should just be Skill and test creation but we'll want to test all the major functions to be sure.

PureTryOut:
So now the MSK changes are merged, what's left here to do (besides rebasing, will do that in a sec)? If we need to continue the discussion on the proper location for these things then fine, but let's do that so we don't keep this hanging for longer unnecessarily.

PureTryOut:
@krisgesling a few months later and last time I didn't get a response, can we _please_ move this forward?

krisgesling:
Hi Bart, I know it's gone on for a long time. 

What this really needs is someone to comprehensively test it out and stake their reputation on it being solid. Then we can merge it in.

Right now, I can't provide the time it will take to do that. I'm spending every moment I have getting the Mark II's out the door to fulfill a 4 year old promise to many thousands of people. I know that is going to be frustrating, however it's our number one priority. 

If any core contributor community members have time to review and test this it would be much appreciated.

mycroft/skills/skill_manager.py
```diff
@@ -18,6 +18,8 @@
 from threading import Thread, Event, Lock
 from time import sleep, time, monotonic
 from inspect import signature
+import shutil
```
forslund: shutil isn't actually used

forslund (commented):
This seems to work rather well. A part from a nitpick of an unnecessary import, I did run into a possible edge case that may or may not be something...

I had an nonstandardly named folder in the /opt/mycroft/skills (this is allowed but mainly happens in development setups). In my case what normally gets installed in a folder `fallback-query.mycroft` was simply in a folder called `skill-query` (the original name I used when I created the first draft). After startup I got errors such as:
```
Exception in thread Thread-4:
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/home/ake/projects/python/mycroft-core/mycroft/skills/skill_manager.py", line 243, in run
    self._start_settings_update()
  File "/home/ake/projects/python/mycroft-core/mycroft/skills/skill_manager.py", line 198, in _start_settings_update
    self.upload_queue.start()
  File "/home/ake/projects/python/mycroft-core/mycroft/skills/skill_manager.py", line 56, in start
    self.send()
  File "/home/ake/projects/python/mycroft-core/mycroft/skills/skill_manager.py", line 71, in send
    loader.instance.settings_meta.upload()
  File "/home/ake/projects/python/mycroft-core/mycroft/skills/settings.py", line 245, in upload
    self._update_settings_meta()
  File "/home/ake/projects/python/mycroft-core/mycroft/skills/settings.py", line 287, in _update_settings_meta
    skill_gid=self.skill_gid,
  File "/home/ake/projects/python/mycroft-core/mycroft/skills/settings.py", line 190, in skill_gid
    skill = skills[skill_dir]
KeyError: '/home/ake/.local/share/mycroft/skills/skill-query'
```

Doing some checking it seems like msm had both copied the old skill folder and created a folder called `fallback-query.mycroftai`.

This is all first glance analyzis and I need to confirm that this is it. If it is the case I wouldn't consider this a blocker for the issue since it's mostly for devs it would occur. (And not very often at that)

PureTryOut:
I kinda forgot what I pushed 4 months ago, can you still reproduce this @forslund?

forslund:
I have not. It was a pretty crazy set of circumstances that only occurs with messy devs like myself.

PR #2814, opened 2021-01-21
Bug Fix - Don't clear namespace on every system page call

AIIX:
## Description
Calling self.clear() on every show some system page event causes mycroft to remove the namespace from active index and reinsert it constantly with a flood of messages, skills showing a page of text self.gui.show_text() for example should only emit the data changes and not try reinserting and removing skill namespace constantly. If an skill author wants to clear the data they can manually call the self.gui.clear() method before calling self.gui.show_text() 

## How to test
Use self.gui.show_text("some text..") in a while loop and make sure it does not keep removing and re inserting the skill namespace from active skill index only sends the set data message with the changed data values

### DIFF
--- a/mycroft/enclosure/gui.py
+++ b/mycroft/enclosure/gui.py
@@ -259,7 +259,6 @@ class SkillGUI:
                 True: Disables showing all platform skill animations.
                 False: 'Default' always show animations.
         """
-        self.clear()
         self["text"] = text
         self["title"] = title
         self.show_page("SYSTEM_TextFrame.qml", override_idle,
@@ -284,7 +283,6 @@ class SkillGUI:
                 True: Disables showing all platform skill animations.
                 False: 'Default' always show animations.
         """
-        self.clear()
         self["image"] = url
         self["title"] = title
         self["caption"] = caption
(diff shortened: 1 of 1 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2814))

krisgesling:
Hey thanks, this is working as described :)

My only possible concern would be whether Skill B might inadvertently show content from a previous Skill A. However they use a unique namespace for each Skill so this couldn't happen. Is that right?

AIIX:
> Hey thanks, this is working as described :)
> 
> My only possible concern would be whether Skill B might inadvertently show content from a previous Skill A. However they use a unique namespace for each Skill so this couldn't happen. Is that right?

Yes each skill uses its own namespace so there should be no collision even if a skill B wants to tell skill A to display some content, if the skill A is imported in skill B, and skill B calls a method in skill A to display something, skill A will still use the Skill A namespace and not the skill B namespace. In another scenario if skill B imports skill A and fetches values from skill A, and decides to a show a page from skill B from values derived from skill A it will still use the skill B namespace because skill B asked to display a page, so getting content from any skill does not matter each skill will use its own namespace to display content.

krisgesling (approved):

PR #2830, opened 2021-02-08
Use os default gcc in github workflow

forslund:
## Description
GCC 4.8 seems to have been removed from ubuntu latest used by the github workflow. This drops the explicit use of the older version in favor for the default.

GCC 4.8 was previously hardcoded to make mimic build slightly faster. Either newer gcc has improved over gcc 6.x or gitlab has more horse powers but this seems to run just fine.

## How to test
Ensure that the unittests run by github passes for python 3.9 (which builds mimic)

CLA [ Yes ]

### DIFF
--- a/.github/workflows/unit_test.yml
+++ b/.github/workflows/unit_test.yml
@@ -42,7 +42,7 @@ jobs:
     - name: Install compilers
       run: |
         sudo apt-get update -qq
-        sudo apt-get install -y gcc-4.8 g++-4.8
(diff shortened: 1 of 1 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2830))

krisgesling (approved):
LGTM

PR #2847, opened 2021-03-03
feat/set_track_position

JarbasAl:
adds a `set_track_position` option to the AudioService

this is needed by GUI skills that implement a seek bar

### DIFF
--- a/mycroft/audio/audioservice.py
+++ b/mycroft/audio/audioservice.py
@@ -248,6 +248,8 @@ class AudioService:
         self.bus.on('mycroft.audio.service.prev', self._prev)
         self.bus.on('mycroft.audio.service.track_info', self._track_info)
         self.bus.on('mycroft.audio.service.list_backends', self._list_backends)
+        self.bus.on('mycroft.audio.service.set_track_position',
+                    self._set_track_position)
         self.bus.on('mycroft.audio.service.seek_forward', self._seek_forward)
         self.bus.on('mycroft.audio.service.seek_backward', self._seek_backward)
         self.bus.on('recognizer_loop:audio_output_start', self._lower_volume)
@@ -495,6 +497,17 @@ class AudioService:
             data[s.name] = info
         self.bus.emit(message.response(data))
 
+    def _set_track_position(self, message):
+        """
+            Handle message bus command to go to position (in seconds)
+
+            Args:
+                message: message bus message
+        """
+        seconds = message.data.get("seconds")
+        if seconds and self.current:
+            self.current.set_track_position(seconds)
+
     def _seek_forward(self, message):
         """
             Handle message bus command to skip X seconds
@@ -534,6 +547,8 @@ class AudioService:
         self.bus.remove('mycroft.audio.service.next', self._next)
         self.bus.remove('mycroft.audio.service.prev', self._prev)
         self.bus.remove('mycroft.audio.service.track_info', self._track_info)
+        self.bus.remove('mycroft.audio.service.set_track_position',
+                        self._set_track_position)
         self.bus.remove('mycroft.audio.service.seek_forward',
                         self._seek_forward)
         self.bus.remove('mycroft.audio.service.seek_backward',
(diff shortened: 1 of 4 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2847))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2847))

mycroft/audio/audioservice.py
```diff
@@ -495,6 +501,38 @@ def _list_backends(self, message):
             data[s.name] = info
         self.bus.emit(message.response(data))
 
+    def _get_track_length(self, message):
+        """
+        getting the duration of the audio in milliseconds
```
krisgesling, line 506: Seems like everything for core is in seconds, whilst VLC etc may be in ms,  is that right?

mycroft/audio/audioservice.py
```diff
@@ -495,6 +501,38 @@ def _list_backends(self, message):
...
+        if self.current:
+            dur = self.current.get_track_length()
+        self.bus.emit(message.response({"length": dur}))
+
+    def _get_track_position(self, message):
+        """
+        get current position in milliseconds
+
+          Args:
+                seconds (int): number of seconds of final position
```
krisgesling, line 518: could seconds be a float?
This is the expected key on the `message.data` object right?

krisgesling (commented):
Nice, like the method names - super clear.

I assume you're happy for us to do some editing of the docstrings?

We should add test cases to TestAudioServiceControls too

mycroft/audio/audioservice.py
JarbasAl: it could be a float, but vlc and probably most backend require an int representing milliseconds, so in case of a float it introduces imprecisions. as per comment above I think we maybe should use milliseconds which removes the uncertainty

forslund:
I think in the mycroft/skills/audioservice.py AudioService interface class it should be seconds so it'll match the other methods there. On the bus and in audioservices we should go with what makes sense from a technical point of view (ms as @JarbasAI suggests)

mycroft/audio/services/__init__.py
```diff
@@ -116,6 +116,27 @@ def restore_volume(self):
...
+                seconds (int): number of seconds of final position
+        """
+
+    def set_track_position(self, seconds=1):
+        """
+        go to position in seconds
+
+          Args:
+                seconds (int): number of seconds of final position
+        """
```
forslund, line 138: Could you move the oneline description to the same line as the `"""` so it's consistent with the other methods in the class

forslund (commented):
Generally looks good, apart from the ongoing discussion of secs vs ms and tiny nitpicks the only thing I miss is a couple of new unittests :)

ken-mycroft (approved):
Looks good, sorry for the delay

JarbasAl:
docstrings mix seconds and milliseconds, per discussion above things need to be migrated to milliseconds (not seconds) and unittests written

NOTE: i have no time to do this anytime soon

JarbasAl:
you can find logic for get_track_position here https://github.com/HelloChatterbox/HolmesV/commit/bd5e3fa69aae01ddd302e447ba0e71e144f31fe9

example usage: `"skip {timedelta} ahead"`
```python
new_pos = get_track_position() + timedelta
set_track_postion(new_pos)
```

maybe this should be also handled in this PR so we get both setters and getters, it also needs conversion from seconds to milliseconds

JarbasAl:
complete implementation https://github.com/HelloChatterbox/HolmesV/commit/c100886adecff7a23ebe7202a7112bf23e0d7553

devops-mycroft:
Hello, @JarbasAl, thank you for helping with the Mycroft project! We welcome everyone
for Everyone.

you!

krisgesling:
Please ignore our friendly CLA bot - this alternative username was hardcoded in the CLA checker script and I missed lowercasing it along with everything else for comparison. This should flip back by itself shortly.

PR #2852, opened 2021-03-09
TTS cache upgrade

chrisveilleux:
## Description
Replace the cache_handler module with a cache module that does all cache things for TTS

## How to test
Pair device with an empty cache

CLA [Y]


mycroft/tts/cache.py
```diff
@@ -0,0 +1,264 @@
...
+from urllib import parse
+
+import requests
+
+from mycroft.util.file_utils import (
+    ensure_directory_exists, get_cache_directory
+)
+from mycroft.util.log import LOG
+
+MIMIC2_URL = 'https://mimic-api.mycroft.ai/synthesize?text='
```
el-tocino, line 45: Shouldn't this be pulled from the config url?

chrisveilleux: Nice catch.  Will fix.

krisgesling (commented):
All looks great. 

Really like the structure and readability. Seems like there's strong potential to abstract this for all TTS services but certainly not the priority for the internal team at least. 

Can we merge this into dev and then rebase the Mark II feature branch off dev?

NOTE: none of the comments below require immediate changes.

mycroft/tts/cache.py
```diff
@@ -0,0 +1,278 @@
...
+    sentence_hash = hashlib.md5(encoded_sentence).hexdigest()
+
+    return sentence_hash
+
+
+class TextToSpeechCache:
+    def __init__(self, tts_config, tts_name, audio_file_type):
+        self.config = tts_config
+        self.tts_name = tts_name
+        self.persistent_cache_dir = Path(tts_config["preloaded_cache"])
```
krisgesling, line 74: Actually, does this need to use `.get()`?

Because this is called in the super `TTS` class it will get initialized for any TTS engine, and they may not have a `preloaded_cache` attribute in their configuration.

mycroft/tts/cache.py
```diff
@@ -0,0 +1,278 @@
...
+
+        Parse the dialog files in the resource directory into sentences.  Then
+        add the audio for each sentence to the cache directory.
+
+        NOTE: There may be files pre-loaded in the persistent cache directory
+        prior to run time, such as pre-recorded audio files.  This will add
+        files that do not already exist.
+
+        ANOTHER NOTE:  Mimic2 is the only TTS engine that supports this.  This
+        logic will need to change if another TTS engine implements it.
```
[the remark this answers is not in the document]
chrisveilleux: I don't believe so.  The intent was to make something that all engines could use.  I could have gone to the trouble of abstracting out the mimic2-specific bits call but that would have taken a decent amount of time.

mycroft/tts/cache.py
```diff
@@ -0,0 +1,278 @@
...
+
+import requests
+
+from mycroft.util.file_utils import (
+    ensure_directory_exists, get_cache_directory
+)
+from mycroft.util.log import LOG
+
+
+def _get_mimic2_audio(sentence: str, url: str) -> Tuple[bytes, str]:
```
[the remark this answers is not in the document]
chrisveilleux: An idea worth pondering. I struggled with the fact that the call to the Mimic2 API is in two places now.  Ideally this would be better integrated into each TTS engine, but that would have caused API changes that would need to be in a major release.  Maybe a ticket for 21.08?

mycroft/tts/tts.py
```diff
@@ -351,23 +355,42 @@ def _execute(self, sentence, ident, listen):
...
-                phonemes = self.load_phonemes(key)
+            sentence_hash = hash_sentence(sentence)
+            if sentence_hash in self.cache.cached_sentences:
+                audio_file, phoneme_file = self._get_sentence_from_cache(
+                    sentence_hash
+                )
+                if phoneme_file is None:
+                    phonemes = None
+                else:
+                    phonemes = phoneme_file.load()
```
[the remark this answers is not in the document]
chrisveilleux: Thanks.   That is what I was going for!

mycroft/tts/cache.py
chrisveilleux: I fixed this to use a get method.  Good catch.

chrisveilleux:
Closing this PR. Will open a new one against the dev branch as Kris suggested.

PR #2854, opened 2021-03-11
Add ask_confirm method

### LINKED ISSUE #2835: Suggestion: Semi-deprecate ask_yesno (retain it for backwards compat)
Some confusion took place in chat due to the fact that `ask_yesno()` returns a string "yes" or "no", rather than a bool; this hit a skill dev working in German and was very disorienting for them.

From a language support perspective, I'm inclined to agree. The decision to return a string was before my time, but Ake suggests it was a consensus thing. Fair enough, but it's an idiosyncratic barrier for skill devs whose code targets a language other than English.

(Also, imo, most hardcoded confirmation prompts return bools, so this one should, too...)

Even in a major version, changing the return type of an outward-facing function would be unacceptable, but it's possible to shove it in a bool with a different name.

```python3
def ask_confirm(self, ...):
    ...
    return a_bool_if_you_please

def ask_yesno(self, ...):
    warn(DeprecationWarning("yadda yadda bool time, retained for backward compat"))
    return "yes" if self.ask_confirm(...) else "no"

### PR DESCRIPTION
**Pull request by krisgesling**
Adds a new method, `ask_confirm()`, an alternative to `ask_yesno` that returns only a Boolean or None.

closes #2835 
replacing PR #2837 

```python
if self.ask_confirm('question'):
    self.speak("got it")
else:
    self.speak("task failed successfully")
```

This PR also adds an Enum to make `ask_yesno` somewhat more language agnostic which was part of the original issue

```python
result = self.ask_yesno(...)
if result == UserReply.YES:
    self.speak("you agree with me")
elif result == UserReply.NO:
    self.speak("you do not agree with me")
elif result is None:
    self.speak("you did not answer")
else:
    self.speak("stay on topic you fool")
```

The Enum is backwards compatible with raw string comparisons

```python
assert "yes" == UserReply.YES
assert "no" == UserReply.NO
```

### DIFF
--- a/mycroft/skills/mycroft_skill/mycroft_skill.py
+++ b/mycroft/skills/mycroft_skill/mycroft_skill.py
@@ -23,7 +23,7 @@ from os import walk
 from os.path import join, abspath, dirname, basename, exists
 from pathlib import Path
 from threading import Event, Timer
-
+from enum import Enum
 from xdg import BaseDirectory
 
 from adapt.intent import Intent, IntentBuilder
@@ -61,6 +61,11 @@ from ..skill_data import (
     read_value_file,
     read_translated_file
 )
+
+
+class UserReply(str, Enum):
+    YES = "yes"
+    NO = "no"
 
 
 def simple_trace(stack_trace):
@@ -545,11 +550,31 @@ class MycroftSkill:
         resp = self.get_response(dialog=prompt, data=data)
 
         if self.voc_match(resp, 'yes'):
-            return 'yes'
+            return UserReply.YES
         elif self.voc_match(resp, 'no'):
-            return 'no'
+            return UserReply.NO
         else:
             return resp
+
+    def ask_confirm(self, dialog, data=None):
+        """Read prompt and wait for a yes/no answer
+
+        This automatically deals with translation and common variants,
+        such as 'yeah', 'sure', etc.
+
+        Args:
+              dialog (str): a dialog id or string to read
+              data (dict): response data
+        Returns:
+              bool: True if 'yes', False if 'no', None for all other
+                    responses or no response
+        """
+        resp = self.ask_yesno(dialog, data=data)
+        if resp == UserReply.YES:
+            return True
+        elif resp == UserReply.NO:
+            return False
+        return None
 
     def ask_selection(self, options, dialog='',
                       data=None, min_conf=0.65, numeric=False):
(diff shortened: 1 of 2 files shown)


krisgesling:
Just been on a test writing spree and this came to mind so I've extended Ake's new tests to cover `ask_confirm()`

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2854))

ChanceNCounter:
I'm torn as to whether `None` is necessary. On the one hand, it's a natural and useful distinction. On the other hand, in a confirmation prompt, aren't "No" and "Cancel" basically the same operation?

If `ask_yesno()` is being retained, then, imo, the "intended" use cases are like,

`ask_yesno("Overwrite the existing file?")` (True overwrites file, False keeps file, else cancel operation)
`ask_confirm("This operation will overwrite the existing file. Should we continue?")` (True continues, False cancels the operation, clicking the 'X' on the UI popup cancels the operation. Does it need an additional 'cancel' button?)

JarbasAl:
`None` identifies a STT failure

ChanceNCounter:
`None` could identify a number of things, but I think they all constitute a "cancel" condition in a confirmation prompt.

krisgesling:
`None` is still falsey so you can do:
```
if ask_confirm('question'):
    # Confirmed
else:
    # Either explicitly no, or None
```

The bigger difference here is that `ask_yesno()` will return the full utterance if neither "yes" nor "no" is detected. Hence an answer of "bananas" would be a truthy value.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2854)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2854/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2854/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2854/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2854/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2854/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2854)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2854/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2854/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2854/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2854/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2854/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2854))

forslund:

Issue #2806, opened 2021-01-17
Fallback skills v2


forslund:
This is more of an initial discussion.

## Background
The intent service is currently parsing a list of different sources to determine a handler of the intent. The standard way it's handled is that the individual service return a match tuple which is then used at the end of `handle_utterance()` to trigger the specific intent handler. This is what `AdaptService` and `PadataiousService`. However Fallback can't currently follow this structure since it will execute the fallback handler if it can, making it an odd little duckling.

## Suggestion
I suggest a "FallbackSkillV2" class which has the fallback check and execution split up:

```python
class FallbackSkillV2:
    def can_answer(self, utterance, lang):
        """Check if the skill can answer the particular question.


        Arguments:
            utterance (str): text to parse
            lang (str) : lang code
        Returns:
            (bool) True if skill can handle the query
        """

    def answer(self, utterance, lang):
        """Performs the actual skill actions such as speak the answer."""
```

The skill would still have a priority to select between multiple matching fallbacks.

Both the `can_answer()` and `answer()` methods would be registered to be called over the message bus. (some code in the baseclass or a decorator would make these automatic similar to the intent_handler decorator)

So the `FallbackService` would send out a message and all fallbacks would answer whether they can answer the utterance, the best answer would be returned to the `IntentService`. The intent service would then if Adapt or Padatious couldn't answer the query use the skill reported by the `FallbackService`.

## Pros for this approach
- Makes it possible to do the fallback checks in parallell
- Will work in the same way as the other intent services
- Can simplify the `handle_utterance()` method
- Makes it possible to encapsulate Fallback skills in a separate process if we in the future wants to utilize cgroups, filesystem permissions, etc per skill.
- Would be needed for a disambiguation system

## Cons
- Not at all compatible with the current fallback system and would have to live in parallell with the existing approach making the code slightly more complex for a time.

Edit: After @JarbasAI suggestion this could probably be made backwards compatible.

JarbasAl:
this sounds like duplicating common query logic.

i would say instead of returning a confidence the skill just returns True or False, and selection is still made by using the priority mechanism

this also makes it backwards compat in a way, if we make the default return True (can answer) then old skills just keep on working.

forslund:
Yes, we could definitely go with just having a true / false answer. I did intend to keep the priority concept, will add that in clear text. The confidence was mainly an idea to replace the common query skill with a native system. but I'll scale that back to True / False

forslund:
[...]

PR #2855, opened 2021-03-14
Suggestion Skillframework v2.1

forslund:
## Description
Currently the skill is is loaded in several steps where the skill-loader code calls the startup sequence and doesn't provide the working skill until after calling `bind()`, `load_data_files()`, etc. All which really should be part of the skill `__init__()`. Skill creator are left with doing their init code in `initialize` a unique method just used by skills and not common across python.

Current skill system also requires some further boiler plate in the form of the `create_skill()` function instantiating the skill and returning it to the SkillLoader. 

This is a suggestion to allow the skill creators to do all their custom initialization in `__init__()` like a regular python class and remove the extra `create_skill()`boilerplate.

 To be backwards compatible this keeps two skill loading functions in the SkillLoader `_create_v2_skill_instance()` and `_create_v3_skill_instance()`, the v2 version will be used if create_skill is found in the skill module, if no such function is found and there is a MycroftSkill based object the loader will consider this a v3 skill module.

Requirements of the v3 skill:

For this to work the skill creator will need to setup their skill `__init__()` (if used) to forward the keyword arguments to the base class.

Example:
```python
from mycroft import MycroftSkill, skill_handler


class MySkill(MycroftSkill):
    def __init__(**kwarg):
        super().__init__(name='My Cool Skill', **kwarg)  # Name is of course optional
        [ Further init code ]

    @skill_handler('Cool.intent')
    def cool_handler(self, _):
        self.speak('My skill is so cool!')
```

This requirement to pass along kwarg could also be avoided with a custom `__new__()` but I think it's better to have an explicit pass-along so skill creators are aware that Mycroft is doing stuff.

Since we now get a class from the skill and not a loaded object the class can provide compatibility information. In this proposal the skill creator can have several skill classes in their module and use the class method is_compatible to select which of the provided classes to use.

In summary this would reduce some of the boiler plate code needed and make the skill loading more pythonic.

CLA [ yes ]

### DIFF
--- a/mycroft/skills/mycroft_skill/mycroft_skill.py
+++ b/mycroft/skills/mycroft_skill/mycroft_skill.py
@@ -63,6 +63,10 @@ from ..skill_data import (
 )
 
 
+def skill_id_from_path(skill_directory):
+    return basename(skill_directory)
+
+
 def simple_trace(stack_trace):
     """Generate a simplified traceback.
 
@@ -119,13 +123,13 @@ class MycroftSkill:
     def __init__(self, name=None, bus=None, use_settings=True):
         self.name = name or self.__class__.__name__
         self.resting_name = None
-        self.skill_id = ''  # will be set from the path, so guaranteed unique
         self.settings_meta = None  # set when skill is loaded in SkillLoader
 
         # Get directory of skill
         #: Member variable containing the absolute path of the skill's root
         #: directory. E.g. /opt/mycroft/skills/my-skill.me/
         self.root_dir = dirname(abspath(sys.modules[self.__module__].__file__))
+        self.skill_id = skill_id_from_path(self.root_dir)
 
         self.gui = SkillGUI(self)
 
@@ -164,6 +168,24 @@ class MycroftSkill:
 
         # Skill Public API
         self.public_api = {}
+
+        if bus:
+            self._startup(bus)
+
+    def _startup(self, bus):
+        self.bind(bus)
+        try:
+            self.load_data_files()
+            # Set up intent handlers
+            self._register_decorated()
+            self.register_resting_screen()
+            self.initialize()
+        except Exception as e:
+            # If an exception occurs, make sure to clean up the skill
+            self.default_shutdown()
+            log_msg = 'Skill initialization failed with {}'
+            LOG.exception(log_msg.format(repr(e)))
+            raise e
 
     def _init_settings(self):
         """Setup skill settings."""
--- a/mycroft/skills/skill_loader.py
+++ b/mycroft/skills/skill_loader.py
@@ -15,8 +15,8 @@
 """Periodically run by skill manager to load skills into memory."""
 import gc
 import importlib
+from inspect import isclass
 import os
-from os.path import dirname
 import sys
 from time import time
 
@@ -26,6 +26,7 @@ from mycroft.skills.settings import save_settings
 from mycroft.util.log import LOG
 
 from .settings import SettingsMetaUploader
+from .mycroft_skill.mycroft_skill import MycroftSkill, skill_id_from_path
 
 SKILL_MAIN_MODULE = '__init__.py'
 
@@ -123,11 +124,28 @@ def _get_last_modified_time(path):
         return 0
 
 
+def get_v3_skill_class(skill_module):
+    for obj in dir(skill_module):
+        if isclass(obj) and issubclass(obj, MycroftSkill):
+            return obj
+    else:
+        return None
+
+
+def is_skill_v3(skill_module):
+    get_v3_skill_class(skill_module) is not None
+
+
+def is_skill_v2(skill_module):
+    return (hasattr(skill_module, 'create_skill') and
+            callable(skill_module.create_skill))
+
+
 class SkillLoader:
     def __init__(self, bus, skill_directory):
         self.bus = bus
         self.skill_directory = skill_directory
-        self.skill_id = os.path.basename(skill_directory)
+        self.skill_id = skill_id_from_path(skill_directory)
         self.load_attempted = False
         self.loaded = False
         self.last_modified = 0
@@ -241,7 +259,12 @@ class SkillLoader:
             self._skip_load()
         else:
             skill_module = self._load_skill_source()
-            if skill_module and self._create_skill_instance(skill_module):
+            if is_skill_v2(skill_module):
+                self._create_v2_skill_instance(skill_module)
+                self._check_for_first_run()
+                self.loaded = True
+            if is_skill_v3(skill_module):
+                self._create_v3_skill_instance(skill_module)
                 self._check_for_first_run()
                 self.loaded = True
 
@@ -268,6 +291,7 @@ class SkillLoader:
     def _load_skill_source(self):
         """Use Python's import library to load a skill's source code."""
         main_file_path = os.path.join(self.skill_directory, SKILL_MAIN_MODULE)
+        skill_module = None
         if not os.path.exists(main_file_path):
             error_msg = 'Failed to load {} due to a missing file.'
             LOG.error(error_msg.format(self.skill_id))
@@ -277,18 +301,12 @@ class SkillLoader:
             except Exception as e:
                 LOG.exception('Failed to load skill: '
                               '{} ({})'.format(self.skill_id, repr(e)))
-            else:
-                module_is_skill = (
-                    hasattr(skill_module, 'create_skill') and
-                    callable(skill_module.create_skill)
-                )
-                if module_is_skill:
-                    return skill_module
-        return None  # Module wasn't loaded
+        return skill_module
 
-    def _create_skill_instance(self, skill_module):
+    def _create_v2_skill_instance(self, skill_module):
         """Use v2 skills framework to create the skill."""
         try:
+            print('!!!!!! STARTING V2 SKILL !!!!!!!')
             self.instance = skill_module.create_skill()
         except Exception as e:
             log_msg = 'Skill __init__ failed with {}'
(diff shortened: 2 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2855))

krisgesling:
Seems like it would be a good shift, however if we're making such a significant change to how Skills are written, I think we should consider if there are other architectural aspects that we want to modify at the same time. If we can avoid supporting multiple versions of the Skills framework (could easily end up with "v2, v3, v4, v5", ... ) then it will be much easier for us in the long run.

One of the key ones in my mind is decoupling Skills from core. How might we provide an architecture that allows Skills to be completely independent from the rest of the system, communicating with core only via the message bus?

The primary benefit being that we can then isolate Skills more easily. However I think this would also make testing of an individual Skill easier.

JarbasAl:
this doesnt really seem like `v3`, but rather like a general cleanup and making the `create_skill` method optional, it doesnt deprecate anything and is fully backwards compatible, i would go with `v2.1` nomenclature 

i want to point out that the `create_skill` method is potentially very useful if some skill needs to perform some action before the loading stage for some reason, so i think it's good to keep support for it

@krisgesling comment is relevant, but feels like a separate discussion and follow up PR since this isn't really a significant change (for skill devs)

forslund:
@JarbasAI you're correct this is more of a cleanup. I'll drop the 3 and call it `2_1` in the code next time I update. Regarding the the comments about create_skill couldn't you in most cases just run your "before skill loading" the `super().__init__()` call?

You could also implement the __new__() if there are truly things that require that the skill object isn't instantiated at all or have a section in the skill that is run on import?

@krisgesling I do think we'd always need to support 2 versions in some way, The current and the deprecated, when a new current is added the deprecated is removed and the old current is made deprecated.

Regarding the skill decoupling, Do we have a list of these hard mycroft system connections? I know of the fallback system system (issue #2806). Are you thinking of the mycroft utils folder and such?

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -63,6 +63,10 @@
 )
 
 
+def skill_id_from_path(skill_directory):
+    return basename(skill_directory)
+
+
```
chrisveilleux, line 69: rather than making this a separate function, why not make skill_id a property?

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -165,6 +168,32 @@ def __init__(self, name=None, bus=None, use_settings=True):
         # Skill Public API
         self.public_api = {}
 
+        if bus:
```
chrisveilleux, line 171: What happens if the bus is not passed in?

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -165,6 +168,32 @@ def __init__(self, name=None, bus=None, use_settings=True):
...
+            self.load_data_files()
+            # Set up intent handlers
+            self._register_decorated()
+            self.register_resting_screen()
+            self.initialize()
+        except Exception as e:
+            # If an exception occurs, make sure to clean up the skill
+            self.default_shutdown()
+            log_msg = 'Skill initialization failed with {}'
+            LOG.exception(log_msg.format(repr(e)))
```
chrisveilleux, line 194: You don't need to include `repr(e)` when using log.exception.

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -165,6 +168,32 @@ def __init__(self, name=None, bus=None, use_settings=True):
...
+            raise e
```
chrisveilleux, line 195: you don't need to pass the exception to the raise statement

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -196,6 +225,20 @@ def _init_settings(self):
...
+        Skill can read the version info from mycroft.version and determine if
+        it can be loaded. This allows the skill v3 framework to disregard
+        incompatible skill classes in a file and select another one.
+
+        Returns:
+            (bool): True if compatible with the version of Mycroft Core,
+                    else False.
+        """
+        return True
+
```
chrisveilleux, line 241: I think this needs some discussion.  We are counting on skill developers to keep this up to date with each version?  I think this is a ticking time bomb.  Some skill won't implement it, or implement it incorrectly, and chaos will ensue.

mycroft/skills/skill_loader.py
```diff
@@ -123,11 +124,58 @@ def _get_last_modified_time(path):
         return 0
 
 
+def is_skill_v2(module):
```
chrisveilleux, line 127: Another versioning thing to track?  Is the "skill framework version" official?  If so, where is it documented?  Is there a way around this?  Will we just keep adding versions upon versions?  When do old versions get cleaned up?

mycroft/skills/mycroft_skill/mycroft_skill.py
forslund: This is to handle the current (v2) case which does not pass the bus. The bind, etc will need to be called externally after init.

mycroft/skills/mycroft_skill/mycroft_skill.py
forslund: In the general case it should not be overridden by the skill creator, skills should be made compatible with an entire major release and this shouldn't be needed to be implemented.

It is for example for the case if we add a new feature foo in 21.2.7, we want to start using the new feature in the skill. we can then have an `__init__.py`

with

```python
from .skill import SkillWithNewFeature
from .compatibility_skill import CompatibilitySkill
```

compatibility_skill.py

```python
[...]
from mycroft.version import CORE_VERSION_TUPLE

class CompatibilitySkill:
    [...]
    @classmethod
    def is_compatible(cls):
        return CORE_VERSION_TUPLE < (21, 2, 7)
```

And the skill.py:
```python
[...]
from mycroft.version import CORE_VERSION_TUPLE

class SkillWithNewFeature:
```python
from mycroft.version import CORE_VERSION_TUPLE
    [...]
    @classmethod
    def is_compatible(cls):
        return CORE_VERSION_TUPLE >= (21, 2, 7)
```

This has been done earlier with the skill installer skill when the skill-id was added but then the logic was placed in the `create_skill()` function. SInce the suggestion is to remove the need for that method I wanted to suggest a replacement.

The exact same thing can be done with simple imports in the skill __init__

```python
from mycroft.version import CORE_VERSION_TUPLE

if CORE_VERSION_TUPLE < (20, 2, 7):
    from .skill import SkillWithNewFeature
else:
    from .compatibility_skill import CompatibilitySkill
```

My thought was that this was using the classmethod would be more explicit.

mycroft/skills/skill_loader.py
forslund: The current version is referred to v2 in the skill loader.

We can't drop support for all existing skills without having an alternative in place, so we need to support the current version for some time when making changes. I don't think that more than two versions should ever be supported at the same time. Likely the v2 should be kept for a year or so?

mycroft/skills/mycroft_skill/mycroft_skill.py
forslund: Will fix, this part was moved from the old skill loading without any changes

mycroft/skills/skill_loader.py
forslund: Oh also as touched on above, we can probably skip the compatibility checks if we do some hacking in `__new__()` buuut that's a slippery-slope that I'd rather avoid.

JarbasAl:
> @JarbasAI you're correct this is more of a cleanup. I'll drop the 3 and call it `2_1` in the code next time I update. Regarding the the comments about create_skill couldn't you in most cases just run your "before skill loading" the `super().__init__()` call?
> 
Seems like i missed this back then, my usage was returning a different class based on ovos_utils platform detection, a skill optimized for a voice only experience or GUI experience. 

In particular for the onboarding process (select backend/voice + pair), speaking and asking questions in a loop is annoying but needed in a voice only interface, if we have a GUI we can rely on it and be much less verbose

Other use case would be skills that take advantage of extra features unique to platforms, eg OVOS, chatterbox, bigscreen, neon... or that mock features if missing

Part of this can be options in a single skill, but sometimes it might make sense to have different classes

```python

class VoiceOnlySkill:
    ...
    
class GuiSkill(VoiceOnlySkill):
    ...
    
def create_skill():
     if has_gui():
         return GuiSkill()
     return VoiceOnlySkill()    
     
 ```
 
This can also be done in `__new__` and not used in the wild, so don't take this as a criticism or a blocker, just highlighting a use case, this PR is a good change!

**Comment by devops-mycroft** [comment]
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2855)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2855/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2855/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2855/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2855/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2855/logs/enclosure.log)

**Comment by forslund** [comment]
There are a couple of ways of handle what you suggest. My thought was the `is_compatible()` class method could be use to define conditions where the skill is/isn't supposed to be used...Might be named differently maybe?

Another possibility could be using imports and if-statements in the `__init__.py`

```python
if has_gui():
    import GuiSkill
else:
    import VoiceOnlySkill
```

mycroft/skills/skill_loader.py
```diff
@@ -150,7 +150,7 @@ def get_skill_class(skill_module):
     """
     for name, obj in skill_module.__dict__.items():
         if (isclass(obj) and issubclass(obj, MycroftSkill) and
-                obj is not MycroftSkill):
+                obj is not MycroftSkill and obj.is_compatible()):
```
JarbasAl, line 153: what if 2 compatible skills exist in same file? maybe one of them has less features but is still compatible

forslund: Right, something like `Alternative Y is also compatible but will not be used`

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2855))

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2855))

forslund:

PR #2860, opened 2021-03-17
Fix a bug with TTS engine changes not taking effect.

chrisveilleux:
## Description
Use the TTS factory to reload the TTS engine when config changes.

## How to test
Change the TTS engine for the device being tested in Selene.  The new engine should be effective soon after the config change is applied to core.

CLA [Y]

### DIFF
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -182,20 +182,27 @@ def init(messagebus):
     """
 
     global bus
-    global tts
-    global tts_hash
     global config
-
     bus = messagebus
     Configuration.set_config_update_handlers(bus)
-    config = Configuration.get()
     bus.on('mycroft.stop', handle_stop)
     bus.on('mycroft.audio.speech.stop', handle_stop)
     bus.on('speak', handle_speak)
+    bus.on('configuration.updated', _create_tts)
+    _create_tts(None)
 
+
+def _create_tts(_):
+    global config
+    global tts
+    global tts_hash
+    LOG.info('waiting for config reload')
+    time.sleep(1)  # Give the config object time to update.
+    config = Configuration.get()
+    LOG.info('Loading the {} TTS engine'.format(config['tts']['module']))
     tts = TTSFactory.create()
     tts.init(bus)
-    tts_hash = hash(str(config.get('tts', '')))
+    tts_hash = hash(str(config['tts']))
 
 
 def shutdown():


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2860))

mycroft/audio/speech.py
```diff
@@ -182,20 +182,27 @@ def init(messagebus):
...
     Configuration.set_config_update_handlers(bus)
-    config = Configuration.get()
     bus.on('mycroft.stop', handle_stop)
     bus.on('mycroft.audio.speech.stop', handle_stop)
     bus.on('speak', handle_speak)
+    bus.on('configuration.updated', _create_tts)
+    _create_tts(None)
+
 
+def _create_tts(_):
```
forslund, line 195: Maybe making the `_` argument optional `_=None` so it doesn't need to be passed on L192

mycroft/audio/speech.py
```diff
@@ -182,20 +182,27 @@ def init(messagebus):
...
+    global tts
+    global tts_hash
+    LOG.info('waiting for config reload')
+    time.sleep(1)  # Give the config object time to update.
+    config = Configuration.get()
+    LOG.info('Loading the {} TTS engine'.format(config['tts']['module']))
     tts = TTSFactory.create()
     tts.init(bus)
-    tts_hash = hash(str(config.get('tts', '')))
+    tts_hash = hash(str(config['tts']))
```
forslund, line 205: This doesn't look quite correct, shouldn't the reload code at https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/audio/speech.py#L117 be updated. Also shouldn't the hash be checked before doing the load? This would reload the tts on any config change.

I would also probably do the
```
tts.playback.stop()
tts.playback.join()
```

before starting a new TTS. Currently the system is somewhat tangled and not doing this will leak some memory.

I started some restructuring to some other bugs in [feature/tts-independent-playback](https://github.com/forslund/mycroft-core/commit/7f5fe6d86024064688abf3dc6a466164810e604b)

forslund (commented):
Nice to have this done asynchronously, it does seem odd that the synchronous TTS update is still left in there. Shouldn't that update code be removed in favor of this?

mycroft/audio/speech.py
chrisveilleux: I am not exactly sure what you are asking for here.  The existing code did not work.  if you changed your TTS engine in Selene, it was not updated on the device.

PR #2861, opened 2021-03-17
Prioritise remote user config over device

### LINKED ISSUE #2864: "Hey MyCroft" used instead of alternate wake word selected
**Describe the bug**
"Hey Jarvis" was selected as the wake word during device pairing
Device only responds to "Hey MyCroft"

**To Reproduce**
Run the docker version of MyCroft on linux mint
Pair the device and select "Hey Jarvis" as the wake word
Speak "Hey Jarvis" nothing happens
Speak "Hey MyCroft"  "Who was Abraham Lincoln?" - the device speaks/reads the response

**Expected behavior**
I expect "Hey Jarvis" to initiate recognition the way "Hey MyCroft" does

**Log files**
I'm not home right now, but I can submit these later if helpful.

You may also include screenshots, however screenshots of log files are often difficult to read and parse. 

If you are running Mycroft, the [Support Skill](https://github.com/MycroftAI/skill-support) helps to automate gathering this information. Simply say "Create a support ticket" and the Skill will put together a support package and email it to you.

**Environment (please complete the following information):**
 - Device type: [e.g. Raspberry Pi, Mark 1, desktop]
 - OS: [e.g. Ubuntu, Picroft]
 - Mycroft-core version: [e.g. 20.08]
 - Other versions: [e.g. Adapt v0.3.7]

**Additional context**

Please think carefully about whether you have modified anything in Mycroft's code or configuration files.  If so, can you reproduce this on a clean installation of Mycroft? Many "bugs" turn out to be non-standard configuration errors.

krisgesling:
## Description
This re-orders the `mycroft.conf` priority order to use remote settings over device defaults.

Currently the priority order is:
1. Default - from mycroft-core
2. Remote - home.mycroft.ai
3. System - `/etc/mycroft/mycroft.conf`
4. User - `~/.mycroft/mycroft`

This switches 2 and 3 so that devices can declare default values that differ from the default mycroft-core but can also be overridden by the user modifying settings on home.mycroft.ai or their preferred backend.

## How to test
Change the voice on a Mark II Dev Kit - this will not change because the device level settings have Mimic2 as the default voice.

- [x] CLA

### DIFF
--- a/mycroft/configuration/config.py
+++ b/mycroft/configuration/config.py
@@ -230,8 +230,7 @@ class Configuration:
                isinstance(config, LocalConf) and config.path == USER_CONFIG:
                 if user_config_disabled:
                     continue
-                for key in protected_keys:
-                    config = delete_key_from_dict(key, config)
+                prune_config(config, protected_keys)
 
             merge_dict(base, config)
 
@@ -285,3 +284,9 @@ class Configuration:
         """
         Configuration.__patch = {}
         Configuration.load_config_stack(cache=True)
+
+    @staticmethod
+    def prune_config(config, prune_list):
+        for key in prune_list:
+            config = delete_key_from_dict(key, config)
+        return config
--- a/mycroft/configuration/config.py
+++ b/mycroft/configuration/config.py
@@ -20,7 +20,7 @@ import inflection
 from os.path import exists, isfile
 from requests import RequestException
 
-from mycroft.util.json_helper import load_commented_json, merge_dict
+from mycroft.util.json_helper import delete_key_from_dict, load_commented_json, merge_dict
 from mycroft.util.log import LOG
 
 from .locations import (DEFAULT_CONFIG, SYSTEM_CONFIG, USER_CONFIG,
@@ -203,6 +203,8 @@ class Configuration:
         Returns:
             (dict) merged dict of all configuration files
         """
+        # system administrators can define different constraints in how
+        # configurations are loaded
         if not configs:
             configs = [LocalConf(DEFAULT_CONFIG), LocalConf(SYSTEM_CONFIG),
                        RemoteConf(), LocalConf(USER_CONFIG),
@@ -213,10 +215,25 @@ class Configuration:
                 if isinstance(item, str):
                     configs[index] = LocalConf(item)
 
+        # Build maintainers and system administrators may prevent users from
+        # modifying specific keys within the configuration.
+        enclosure_config = LocalConf(SYSTEM_CONFIG).get('enclosure', {})
+        protected_keys = enclosure_config.get('protected_keys', [])
+        user_config_disabled = enclosure_config.get('disable_user_config', False)
+
         # Merge all configs into one
         base = {}
-        for c in configs:
-            merge_dict(base, c)
+        for config in configs:
+
+            # handle system constraints
+            if isinstance(config, RemoteConf) or \
+               isinstance(config, LocalConf) and config.path == USER_CONFIG:
+                if user_config_disabled:
+                    continue
+                for key in protected_keys:
+                    config = delete_key_from_dict(key, config)
+
+            merge_dict(base, config)
 
         # copy into cache
         if cache:
--- a/mycroft/configuration/config.py
+++ b/mycroft/configuration/config.py
@@ -46,6 +46,20 @@ def is_remote_list(values):
     return True
 
 
+def prune_config(config, prune_list):
+    """Delete list of nested keys from the provided config.
+
+    Arguments:
+        config (dict): config to prune
+        prune_list (list(str)): list of keys to delete. Each item may be a
+                                period separated list of nested keys eg
+                                ["nested.dict.key", "listener.sample_rate"]
+    """
+    for key in prune_list:
+        config = delete_key_from_dict(key, config)
+    return config
+
+
 def translate_remote(config, setting):
     """Translate config names from server to equivalents for mycroft-core.
 
@@ -246,7 +260,7 @@ class Configuration:
             elif isinstance(config, LocalConf) and config.path == USER_CONFIG:
                 if user_config_disabled:
                     continue
-                Configuration.prune_config(config, protected_keys)
+                prune_config(config, protected_keys)
 
             merge_dict(base, config)
 
@@ -300,9 +314,3 @@ class Configuration:
         """
         Configuration.__patch = {}
         Configuration.load_config_stack(cache=True)
-
-    @staticmethod
-    def prune_config(config, prune_list):
-        for key in prune_list:
-            config = delete_key_from_dict(key, config)
-        return config
--- a/mycroft/configuration/config.py
+++ b/mycroft/configuration/config.py
@@ -20,11 +20,19 @@ import inflection
 from os.path import exists, isfile
 from requests import RequestException
 
-from mycroft.util.json_helper import delete_key_from_dict, load_commented_json, merge_dict
+from mycroft.util.json_helper import (
+    delete_key_from_dict,
+    load_commented_json,
+    merge_dict
+)
 from mycroft.util.log import LOG
 
-from .locations import (DEFAULT_CONFIG, SYSTEM_CONFIG, USER_CONFIG,
-                        WEB_CONFIG_CACHE)
+from .locations import (
+    DEFAULT_CONFIG,
+    SYSTEM_CONFIG,
+    USER_CONFIG,
+    WEB_CONFIG_CACHE
+)
 
 
 def is_remote_list(values):
@@ -83,6 +91,7 @@ def translate_list(config, values):
 
 class LocalConf(dict):
     """Config dictionary from file."""
+
     def __init__(self, path):
         super(LocalConf, self).__init__()
         if path:
@@ -124,6 +133,7 @@ class LocalConf(dict):
 
 class RemoteConf(LocalConf):
     """Config dictionary fetched from mycroft.ai."""
+
     def __init__(self, cache=None):
         super(RemoteConf, self).__init__(None)
 
@@ -219,7 +229,8 @@ class Configuration:
         # modifying specific keys within the configuration.
         enclosure_config = LocalConf(SYSTEM_CONFIG).get('enclosure', {})
         protected_keys = enclosure_config.get('protected_keys', [])
-        user_config_disabled = enclosure_config.get('disable_user_config', False)
+        user_config_disabled = enclosure_config.get(
+            'disable_user_config', False)
 
         # Merge all configs into one
         base = {}
--- a/mycroft/configuration/config.py
+++ b/mycroft/configuration/config.py
@@ -204,8 +204,8 @@ class Configuration:
             (dict) merged dict of all configuration files
         """
         if not configs:
-            configs = [LocalConf(DEFAULT_CONFIG), RemoteConf(),
-                       LocalConf(SYSTEM_CONFIG), LocalConf(USER_CONFIG),
+            configs = [LocalConf(DEFAULT_CONFIG), LocalConf(SYSTEM_CONFIG),
+                       RemoteConf(), LocalConf(USER_CONFIG),
                        Configuration.__patch]
         else:
             # Handle strings in stack
--- a/mycroft/configuration/config.py
+++ b/mycroft/configuration/config.py
@@ -227,18 +227,23 @@ class Configuration:
 
         # Build maintainers and system administrators may prevent users from
         # modifying specific keys within the configuration.
-        enclosure_config = LocalConf(SYSTEM_CONFIG).get('enclosure', {})
-        protected_keys = enclosure_config.get('protected_keys', [])
-        user_config_disabled = enclosure_config.get(
+        system_config = LocalConf(SYSTEM_CONFIG) or {}
+        remote_config_disabled = system_config.get(
+            'disable_remote_config', False)
+        user_config_disabled = system_config.get(
             'disable_user_config', False)
+        protected_keys = system_config.get('protected_keys', [])
 
         # Merge all configs into one
         base = {}
         for config in configs:
 
             # handle system constraints
-            if isinstance(config, RemoteConf) or \
-               isinstance(config, LocalConf) and config.path == USER_CONFIG:
+            if isinstance(config, RemoteConf):
+                if remote_config_disabled:
+                    continue
+                prune_config(config, protected_keys)
+            elif isinstance(config, LocalConf) and config.path == USER_CONFIG:
                 if user_config_disabled:
                     continue
                 Configuration.prune_config(config, protected_keys)
--- a/mycroft/configuration/config.py
+++ b/mycroft/configuration/config.py
@@ -241,7 +241,7 @@ class Configuration:
                isinstance(config, LocalConf) and config.path == USER_CONFIG:
                 if user_config_disabled:
                     continue
-                prune_config(config, protected_keys)
+                Configuration.prune_config(config, protected_keys)
 
             merge_dict(base, config)
 
--- a/mycroft/util/json_helper.py
+++ b/mycroft/util/json_helper.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 #
 import json
+from copy import copy
 
 
 def merge_dict(base, delta):
@@ -30,6 +31,27 @@ def merge_dict(base, delta):
             merge_dict(bv, dv)
         else:
             base[k] = dv
+
+
+def delete_key_from_dict(key, dictionary):
+    """Recursivily find nested key in a dict and delete it.
+
+    Arguments:
+        key (str): a period separated list of nested keys to remove
+                   eg "nested.dict.keys"
+        dictionary (dict): the dictionary to remove keys from
+    Returns:
+        Dict: original dictionary with specified keys deleted.
+    """
+    modified_dict = copy(dictionary)
+    key_list = key.split('.')
+    if len(key_list) == 1 and modified_dict.get(key) is not None:
+        del modified_dict[key]
+    elif len(key_list) > 1:
+        remaining_keys = '.'.join(key_list[1:])
+        if modified_dict.get(key_list[0]) is not None:
+            modified_dict[key_list[0]] = delete_key_from_dict(remaining_keys, modified_dict[key_list[0]])
+    return modified_dict
 
 
 def load_commented_json(filename):
--- a/mycroft/util/json_helper.py
+++ b/mycroft/util/json_helper.py
@@ -50,7 +50,8 @@ def delete_key_from_dict(key, dictionary):
     elif len(key_list) > 1:
         remaining_keys = '.'.join(key_list[1:])
         if modified_dict.get(key_list[0]) is not None:
-            modified_dict[key_list[0]] = delete_key_from_dict(remaining_keys, modified_dict[key_list[0]])
+            modified_dict[key_list[0]] = delete_key_from_dict(
+                remaining_keys, modified_dict[key_list[0]])
     return modified_dict
 
 
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -18,6 +18,16 @@
   // specific system can set relevant attributes at the SYSTEM level, and users
   // can further customize their experience at the REMOTE or USER levels.
 
+  // REMOTE or USER configurations can be disabled thereby requiring root
+  // access to the system to change any config parameters.
+  "disable_remote_config": false,
+  "disable_user_config": false,
+
+  // Alternatively a set of protected keys can be defined. This prevents
+  // either REMOTE or USER from modifying these settings. Nested keys should
+  // be provided as period separated strings eg ["listener.sample_rate"]
+  "protected_keys": [],
+
   // Language used for speech-to-text and text-to-speech.
   // Code is a BCP-47 identifier (https://tools.ietf.org/html/bcp47), lowercased
   // TODO: save unmodified, lowercase upon demand
@@ -234,8 +244,6 @@
     // Override: SYSTEM (set by specific enclosures)
     // "platform": "picroft",
     // "platform_enclosure_path": "/etc/myenclosure/code.py",
-    "protected_keys": [],
-    "disable_user_config": false,
 
     // COMM params to the Arduino/faceplate
     "port": "/dev/ttyAMA0",
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -2,20 +2,21 @@
   // Definition and documentation of all variables used by mycroft-core.
   //
   // Settings seen here are considered DEFAULT.  Settings can also be
-  // overridden at the REMOTE level (set by the user via
-  // https://home.mycroft.ai), at the SYSTEM level (typically in the file
-  // '/etc/mycroft/mycroft.conf'), or at the USER level (typically in the
+  // overridden at the SYSTEM level (typically in the file
+  // '/etc/mycroft/mycroft.conf'), at the REMOTE level (set by the user via
+  // https://home.mycroft.ai), or at the USER level (typically in the
   // file '~/.mycroft/mycroft.conf').
   //
   // The load order of settings is:
   //   DEFAULT
-  //   REMOTE
   //   SYSTEM
+  //   REMOTE
   //   USER
   //
   // The Override: comments below indicates where these settings are generally
-  // set outside of this file.  The load order is always followed, so an
-  // individual systems can still apply changes at the SYSTEM or USER levels.
+  // set outside of this file.  The load order is always followed, so a
+  // specific system can set relevant attributes at the SYSTEM level, and users
+  // can further customize their experience at the REMOTE or USER levels.
 
   // Language used for speech-to-text and text-to-speech.
   // Code is a BCP-47 identifier (https://tools.ietf.org/html/bcp47), lowercased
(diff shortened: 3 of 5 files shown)


AIIX:
How does this affect values set for a device in /etc/mycroft/mycroft.conf at image creation time, does this now mean that if i have for example a "platform" key value or "recording_timeout" key value set in /etc/mycroft/mycroft.conf it will now get overridden every time by default values from home.mycroft.ai every time mycroft.conf is synced from the remote backend ?

Also the whole point of having the SYSTEM override default values is that any platform shipping mycroft should be able to provide their own values via system expected global path than having to depend on USER override which Linux packaging debs/rpm/etc cannot anyways write to.

ChanceNCounter:
It's roundabout, but the system could first check device settings for something like `NO_REMOTE_OVERRIDE`, and, assuming that setting is `False`, proceed as this PR intends.

j1nx:
200% disagree with changing the order. The order as it is in now is the only logical order.

1) Defaults from the mycroft-core package it self
2) Defaults from the backend when going the pairing route
3) Systemwide changes
4) Local override from the tech users

I understand that 2 and 3 could bite each other in certain cases, but then it means it's needs a problem solving technique, not such a breaking change as "quick fix".

There is a reason, the systemwide config file resides in /etc which is only writable by the root user.


If the remote config is moved downwards in that list, it should only override/set variables not already set within /etc/mycroft or ~/.mycroft

Again, there is a reason variables might be set in either of the local files.

krisgesling:
I think it might be helpful to describe the different levels in more detail, at least as I see them:
1. DEFAULT - set in mycroft-core, default for every installation of mycroft, can only be changed with a PR or intentionally monkey patching it.
2. SYSTEM - set at build time, default for that device type (Mark II, Picroft, BigScreen, OVOS, etc). Can be set at build time or modified by a device administrator with root access to the device.
3. REMOTE - set if pairing with a backend and updates as the user modifies settings. Can be modified by the account owner without root access to the device. Is limited to the fields provided by that backend ie without a change to Selene, a user could not modify their `platform` value.
4. USER - can be modified by any authenticated user on the device.

In this fashion we move from:
1. Mycroft defined
2. User defined (limited by the backend)
3. Developer / Mycroft defined - modifiably by a root user
4. User defined

To:
1. Mycroft defined
2. Developer / Mycroft defined - modifiably by a root user
3. User defined (limited by the backend)
4. User defined

So yes, this would mean that anything set in a remote backend like home.mycroft.ai would override the SYSTEM config. If the backend pushed a `platform` or `recording_timeout` attribute then it would be applied over `/etc/mycroft/mycroft.conf`. However for context this is what the current REMOTE level conf currently includes:
```
{
  "date_format": "DMY",
  "listener": {
    "channels": 1,
    "energy_ratio": 1.5,
    "multiplier": 1,
    "phonemes": "HH EY . M AY K R AO F T",
    "sample_rate": 16000,
    "threshold": "1e-90",
    "wake_word": "hey mycroft"
  },
  "opt_in": true,
  "system_unit": "metric",
  "time_format": "half",
  "tts": {
    "mimic2": {
      "voice": "kusal"
    },
    "module": "mimic2"
  },
  "location": {
    "city": {
      "name": "Darwin",
      "state": {
        "code": "AU.03",
        "country": {
          "code": "AU",
          "name": "Australia"
        },
        "name": "Northern Territory"
      }
    },
    "coordinate": {
      "latitude": -12.46113,
      "longitude": 130.84185
    },
    "timezone": {
      "code": "Australia/Darwin",
      "dst_offset": 34200000,
      "name": "Australia/Darwin",
      "offset": 34200000
    }
  }
  ```

As it stands if you set a default TTS engine or wake word configuration for a device then the user can no longer change this via the backend.

An alternative would be to add in another level so you had: 
1. DEFAULT
2. SYSTEM_DEFAULTS
3. REMOTE
4. SYSTEM_OVERRIDES
5. USER

I just haven't seen the use case for the SYSTEM_OVERRIDES, generally the two configs just avoid each other, hence I thought switching was the cleaner option.

Are there examples of when we would want a SYSTEM config that is set by a device admin to override REMOTE that is set by the end user? Or are there different use cases I'm not thinking about?

**Comment by AIIX** [comment]
> As it stands if you set a default TTS engine or wake word configuration for a device then the user can no longer change this via the backend.

As I see it, Shouldn't any change made by the user in the remote backend ideally be stored at USER level override that being in ~/.mycroft/mycroft.conf rather than changing the logical order of configurations ? Because it seems more logical that any change made by the user is going to override every other configuration default / system for any field be that is coming from the backend or somewhere from a skill. Any changes a user makes to ~/.mycroft/mycroft.conf gets synced to the backend and any changes that are made in the backend by the "USER" should only get synced back to ~/.mycroft/mycroft.conf, otherwise what is the use of ~/.mycroft/mycroft.conf USER override without this not already being the case ?

**Comment by j1nx** [comment]
My point of view;

1 ) DEFAULT - set in mycroft-core, default for every installation of mycroft, can only be changed with a PR or intentionally monkey patching it.

Above is to make sure all values are set. In my view this should not be changed as every variable needs to be set and at development time, a proper default value is choosen.


2) SYSTEM - set at build time, default for that device type (Mark II, Picroft, BigScreen, OVOS, etc). Can be set at build time or modified by a device administrator with root access to the device.

As example: OpenVoiceOS ships with the below configuration;
https://github.com/OpenVoiceOS/OpenVoiceOS/blob/develop/buildroot-external/rootfs-overlay/etc/mycroft/mycroft.conf
With this config file Mycroft will always work as expected, despite the fact the user want to pair or whatever. 
 

3 & 4) REMOTE - set if pairing with a backend and updates as the user modifies settings via the web backend. | USER - can be modified by any authenticated user on the device.


The last two options again should perfectly sync and work both ways.

I think we say the same @krisgesling It is just that the LOCAL and REMOTE architecture should be changed, so that you can change setting either way; The backend, the mycroft-config cli command or manually editting the user mycroft.conf

**Comment by chrisveilleux** [comment]
I like the idea of combining USER and REMOTE.  Both are user-level overrides to settings provided by SYSTEM and DEFAULT configs.  In fact, I was bitten recently by the USER settings overriding what I had set in Selene.

This is a bigger change that can be implemented in the future, but swapping the config order, as this PR does, is essentially the same thing, as long as you don't screw yourself in the USER configs.

If there are any configs in SYSTEM or DEFAULT that should be protected from user overrides, such as "platform", perhaps a configuration file is the wrong place for them.   Why put something in a configuration file if it is not configurable?

**Comment by AIIX** [comment]
> If there are any configs in SYSTEM or DEFAULT that should be protected from user overrides, such as "platform", perhaps a configuration file is the wrong place for them. Why put something in a configuration file if it is not configurable?

I believe the whole idea of System Level Override was that any project is free to override the default configuration provided by Mycroft as seen fit by any open source project, Linux distribution or anyone shipping Mycroft Core like Plasma Bigscreen, Plasma Desktop, Plasma Mobile, Open Voice OS, etc.

If one is asking why put something like "Platform" in a configuration file if it is not configurable then I would like to point to https://github.com/MycroftAI/mycroft-core/blob/dev/mycroft/configuration/mycroft.conf#L233 which clearly is defined in a Mycroft configuration and also as a SYSTEM override while clearly implies that any Platform, Linux distribution or Open source project should be able to freely to set their own Platform or any other default configuration value for their project that comes under the "SYSTEM" override that does not get affected without actual user taking action and is only ever overridden by "USER" override.

If the project does not want any other open source projects to use the configuration file to set their own system level values for system overridable things like "platform" and think it is the wrong place then what is the alternative solution that the project is suggesting that other open source projects should follow ?

**Comment by j1nx** [comment]
DEFAULT and REMOTE are defaults from Mycroft A.I. DEFAULT being the SYSTEM from Mycroft, REMOTE being the USER from Mycroft.

If any other FOSS project want to do things different, above methodology can be shifted to utilizing SYSTEM being the system setting, USER being the users override of SYSTEM

Mycroft != Selene and vica versa.


Switching 2 and 3 changes the original methodology and screws up any other project that want to do there own thing.

**Comment by krisgesling** [comment]
I think what Chris was heading toward with the point about `platform` is that in the future is it worth having a portion of the configs that can't be overridden by the end user? Or in some way extracting these out so they are "platform configs" rather than "user configs". Currently I could edit the USER config on a Picroft to say it's a Mark II. This will probably just cause the Enclosure Service to fail since it won't be able to find the expected hardware. However I don't think anyone is really going to do that, nor is a backend going to try and override this value so I don't really see it being an issue we have to solve right now.

As J1nx kind of suggested, we can also monkey patch the DEFAULT config for the Mark II in the short term. I wouldn't want to change the actual DEFAULT just for the Mark II.

I'm still a little confused as to how the REMOTE config being moved down to 3 will mess with anyone's setups whether that's an individual or a project. If it does, then we can't sync REMOTE and USER or that would presumably also cause the same breakage?

Can you give an example of when or how this would cause an issue?

Two-way sync is certainly the end goal but will take a lot more work to do that. This feels like a step in that direction though. If there are going to be issues having REMOTE (or USER synced with REMOTE) configs override SYSTEM then we need to understand what they are.

**Comment by krisgesling** [comment]
Ok, I wanted to come back and summarize where I think this is at - number 5 being the sticking point.

1. There's general agreement that two-way sync of user settings (if they're using a backend) is ultimately what we want.
2. Projects should be able to continue setting their own defaults that override Mycroft defaults.
3. Users should be able to override project defaults should they choose to.
4. In the future it may be worth exploring the difference between "user" configuration and "device" configuration. But this is well beyond this PR so we're not going there.
5. There is concern that REMOTE configs will override project defaults and break things but I'm still unclear about what those might be. 

The only configs that I've thought might be an issue is for the wake word and TTS settings. However the REMOTE config doesn't provide a definition for "hey mycroft", only that it should be used. Likewise for Mimic2 it gives a "voice" attribute but no config for say url if a project was hosting their own mimic2 server. 

So someone using OVOS and selecting "Hey Mycroft" in Selene would get the OVOS configuration of that hotword. That would differ from the config someone would get on the Mark II OS and selecting the same button on the backend.

If the REMOTE config got moved down as proposed, we would need to ensure any future configs in REMOTE didn't override existing configurations in a way that isn't 100% clear and expected by the user, eg changing wake words as above.

If I'm missing something please point me in the right direction. Currently I can't see anything breaking by merging this PR. I also don't think this is just a quick fix, it aligns the two user defined configs that will happen when we have two-way sync anyway.

**Comment by JarbasAl** [comment]
the `SYSTEM` level config is reserved for administrators/needs `sudo`, the conflict here emerges at the conceptual level, `REMOTE` should not have precedence over something someone with `root` access decided to change

the problem arises when the `SYSTEM` config is not treated as an override but as a default_value provider, meaning things like TTS can not be changed by `REMOTE` when they should

If we embrace the concept that `SYSTEM` is for overrides, not defaults i have a few ideas:
- problems come from usage error -> drop this PR
    - the usage error is using `SYSTEM` for default values
    - correct usage is using `SYSTEM` for overrides
- add more configuration granularity, specifically split the remote config into 2
    - `SYSTEM` takes precedence over `REMOTE`
    - new `USER_REMOTE` config after `SYSTEM`, but only certain keys are allowed (eg, tts, stt)
    	- by default all keys allowed (?)
    - `SYSTEM` controls which keys are allowed in `USER_REMOTE`, ensures that backend can not replace values like `platform` or `auto_updates`
     - only someone with `root` access can change the `SYSTEM` config, a deliberate action is needed for it, in common usage the `SYSTEM` config might even be missing

**Comment by AIIX** [comment]
> Ok, I wanted to come back and summarize where I think this is at - number 5 being the sticking point.
> 
>     1. There's general agreement that two-way sync of user settings (if they're using a backend) is ultimately what we want.
> 
>     2. Projects should be able to continue setting their own defaults that override Mycroft defaults.
> 
>     3. Users should be able to override project defaults should they choose to.
> 
>     4. In the future it may be worth exploring the difference between "user" configuration and "device" configuration. But this is well beyond this PR so we're not going there.
> 

I overall agree with the 4 points mentioned here.

>     5. There is concern that REMOTE configs will override project defaults and break things but I'm still unclear about what those might be.

Defaults can be any key/value set by a project in the future, for example Plasma Bigscreen uses a USB remote microphone and might want to ship different defaults in the SYSTEM config:

```
"listener": {
    "sample_rate": 8000
}
```
This could be any of the ones mentioned here in the remote configuration https://github.com/MycroftAI/mycroft-core/pull/2861#issuecomment-801865141 that any project should have full control over at the time of shipping an image. 

Unless these values are then changed by the user manually in Selene or any backend while pairing, or by logging in and manually changing them for a single device, then and only then the changes logically should get written into REMOTE/USER configuration that overrides SYSTEM configuration, because these are REMOTE/USER changes that the USER is then aware making for any selections and changes.

I as an external project maintainer would hate to later have to file a bug with this issue where I cannot ship configurations out of the box that are optimized for that project, which then largely as a bug report is going to be put on the backlog because it might have no priority above the mark-2.

So for me at least how i see the actual solution:
- SYSTEM config key values are synced to the backend
- USERS makes any change in backend for some key/value
- Backend keeping rest of the unchanged SYSTEM synced key values in order updates the device with only the USER changed key/value to REMOTE/USER config
- only then does the REMOTE/USER config override the SYSTEM config

For the above the first actual thing that needs to be working is the two way sync.

**Comment by j1nx** [comment]
I would like to add that Selene is not the only backend available at the moment and as it is a FOSS project, anyone could develop and/or use their own backend. So it is a bit short sighted to just focus on the current Selene backend parameters.

It is more about the "attitude" of willing to change underlying architecture just to fix a Mark-2 "bug"

Anyhow: Good to see that you reach out to the dev community in this way, before changing these type of things.

**Comment by JarbasAl** [comment]
small note, the 2 way sync isn't really needed in the options i mentioned in my previous comment, it's more of a usability thing than anything else

the advantage of 2-way-sync is that `REMOTE` can send only the keys that actually changed + display accurate data (privacy leak?), the scheme of authorized_overrides I proposed in OPTION 2 removes this need

**Comment by krisgesling** [comment]
Lots of interesting options here... currently Jarbas's option 2.5 is my personally preferred option. This means that a project could protect any and all attributes if they really wanted to.

The listener sample_rate and channels are a very good point, though I think this is also something we'll need to address in the audio stack as I've seen recently that the default sample_rate at least is hardcoded in a number of locations. So changing that value is problematic all around. Also if they're specific to a hotword config, they should be in that specific hotword configuration, not the overall listener config. If they're really for the listener then this should either be software or hardware dependent, hence don't make sense as a user configurable config in REMOTE.

Even if it stayed in the Selene REMOTE config it could be protected with an allow/block list of attributes in the SYSTEM config as suggested.

I do want to push back a little that I'm just thinking about Selene or the Mark II. It's come up because of seeing it in the Mark II but when I looked at the current order it didn't really make sense to me. At least when thinking about dedicated devices and it seems that the current OVOS config shares the same issue for TTS engine. My first assumption was that it was designed for multi-user desktop environments. Thinking of it more as an override than a preference order makes more sense. Either way I think one of the options outlined in this thread is a better way forward. This is also why I wanted to flag this change specifically because it's way too easy to make assumptions that other projects are using something (and thinking about it) in the same way you yourself are.

**Comment by chrisveilleux** [comment]
It seems like the main sticking point here is SYSTEM level configs that should never be changed by REMOTE (or USER) level configs.  This returns me to my previous point.  Why have something in a configuration file that should not be configurable?  

Forget for a second how the config system currently works.  How SHOULD it work?  If we can agree on that, then we can determine a plan to get there.  There are two different concepts here that need to be addressed.  

1. Attributes with default values that can be overridden at different levels
2. Attributes with values that are unique to a system, set by a system administrator, and cannot be changed by a user.

IMO, combining these two concepts into a single implementation is not a good idea.  The business rules behind them are different.  As such, they should be treated as separate entities.

Configuration files:
- the DEFAULT configuration file should define the set of **configurable** attributes and set default values for them.
- the SYSTEM configuration file can
  - override the DEFAULT values to be more suitable to the system
  - define new **configurable** attributes that are specific to the system 
- the USER configuration file can
  - override the values of the **configurable** attributes defined in the DEFAULT and SYSTEM configuration files.
  - can be set locally by logging into the device and manually altering the `~/.mycroft/mycroft.conf` file or set remotely using a backend.

Settings files:
- the SYSTEM settings file contains non-configurable values that cannot be changed by a user.
- attributes expected in the SYSTEM settings file can be defined in core with some sort of settings definition file but no "default" values exist.  They are system specific values.  The "platform" could be considered one of these.
- the SYSTEM settings file can set attributes not defined in core that are specific to the type of device.

Maybe "settings" is not the right term for what I am proposing they do but I hope you get the idea.

@j1nx This is not just an issue for the Mark II.  It is currently an issue for the Picroft as well and could potentially be an issue for any platform.
@JarbasAI The number of config files we have should not be a determining factor in an architecture decision such as this.

**Comment by JarbasAl** [comment]
you are mixing up configuration with permissions

these things are meant to be configurable, they are not hardcoded settings, the thing is that they need `root` to be changed, which can be the user that owns the device or a distribution shipping factory defaults, and usually also the backend but this can be limited

i think #2872 addresses all the concerns raised here

**Comment by j1nx** [comment]
@chrisveilleux Yeah, again bad word choice of me. Indeed OVOS and Picroft run into the same issue. What I meant was, to quick fix the couple of Mark-2 bug reports as linked within this PR.  Anyhow, doesn't matter as it looks like we are all at the same page here.

**Comment by krisgesling** [comment]
An alternative approach based off #2872 

I haven't figured out the best way to test this and the `isinstance` checks are breaking existing tests - but interested if this meets all the criteria we've been talking about?

**Comment by JarbasAl** [comment]
i would like to keep remote and user seperated

there is no official way to disable the backend, so privacy conscious users might want to block remote only, for example, if i do no want to disclose my location i can input garbage data in selene and not have it synced down at all. same goes for all other preferences

i think there is value in allowing selective blocking, but at very list there should be a global remote disable flag, privacy first right?

**Comment by forslund** [comment]
@krisgesling the failing test in #2872 is due to the simple remote mock at [.../test_configuration.py#L70](https://github.com/MycroftAI/mycroft-core/blob/dev/test/unittests/configuration/test_configuration.py#L70), change that to a proper RemoteConf based class and it should work.

Something along the lines of should fix it

```python
class DummyConf(RemoteConf):
    def __init__(self, *arg, **kwarg):
        pass
```

then replace the offending line with

```python
    mock_remote.return_value = DummyConf()
```

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2861)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2861/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2861/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2861/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2861/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2861/logs/enclosure.log)

krisgesling:
@forslund thanks for that!

@JarbasAI - I'm pretty hesitant to provide distinct behaviour for the two now as it seems they are likely to become one and the same in the future. It also seems like every project would just have two lists of the same protected keys for both remote and user.

> there is no official way to disable the backend, so privacy conscious users might want to block remote only, for example, if i do not want to disclose my location i can input garbage data in selene and not have it synced down at all. same goes for all other preferences

You can still put garbage data in REMOTE and USER config will take ultimately priority. 

> i think there is value in allowing selective blocking, but at very least there should be a global remote disable flag, privacy first right?
EDIT: the only change i would make in #2872 would be the usage of delete_key_from_dict , the thinking here is that we may want to disable more stuff for REMOTE than for USER

It seems like people concerned about location or device settings from REMOTE would also be concerned about having any backend at all, particularly for STT and TTS. Instead of building work arounds, I think we add in an official way to disable the backend.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2861)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2861/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2861/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2861/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2861/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2861/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2861)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2861/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2861/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2861/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2861/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2861/logs/enclosure.log)

JarbasAl:
here is an example, in OVOS i want to allow the user to change platform if he needs to, maybe it's a dev and want to pretend its a mk2 in purpose, on the other hand i explicitly want to block selene from sending that value when it adds some fancy new dropdown to the pairing process

those two configs are not the same, one is controlled by you, the other is controlled by a user with physical access, i might want to stop you from imposing system values like the listener_rate, but if a user added some special hat to the raspberry pi i don't want to stop that use

i can think of many more cases where i want to block `REMOTE` only than wanting to block `REMOTE` + `USER`, both at the global .conf level, but also at individual key level

disabling the backend is a great PR also, but in the end not related to this, i want you to stop you setting values like `platform`, not to stop users from using your backend

chrisveilleux:
remote settings are really user settings.  the user sets them, after all.  Selene does not force any settings that are not defined by the user, nor will it ever.  they will probably be combined in the future with some sort of synchronization, similar to skill settings.  also, +1 to a system level config indicating remote configs should be ignored.

mycroft/configuration/config.py
```diff
@@ -268,3 +297,9 @@ def patch_clear(message):
         """
         Configuration.__patch = {}
         Configuration.load_config_stack(cache=True)
+
+    @staticmethod
+    def prune_config(config, prune_list):
+        for key in prune_list:
+            config = delete_key_from_dict(key, config)
+        return config
```
forslund, line 305: Should this be a static method? I think this is mainly an internal thing and doesn't really need to be shipped with the Configuration object to the external caller. To me it can be a module level function. (also docstring)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2861)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2861/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2861/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2861/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2861/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2861/logs/enclosure.log)

mycroft/configuration/config.py
krisgesling: good call on both fronts :)

mycroft/configuration/config.py
```diff
@@ -203,20 +227,39 @@ def load_config_stack(configs=None, cache=False):
...
         # Merge all configs into one
         base = {}
-        for c in configs:
-            merge_dict(base, c)
+        for config in configs:
+
+            # handle system constraints
+            if isinstance(config, RemoteConf) or \
+               isinstance(config, LocalConf) and config.path == USER_CONFIG:
+                if user_config_disabled:
```
JarbasAl, line 258: shouldn't this account for `remote_config_disabled`?

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2861)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2861/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2861/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2861/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2861/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2861/logs/enclosure.log)

mycroft/configuration/config.py
krisgesling: That is a pretty important piece of the puzzle yes... thanks!

Added.

mycroft/util/json_helper.py
```diff
@@ -32,6 +33,28 @@ def merge_dict(base, delta):
...
+
+    Arguments:
+        key (str): a period separated list of nested keys to remove
+                   eg "nested.dict.keys"
+        dictionary (dict): the dictionary to remove keys from
+    Returns:
+        Dict: original dictionary with specified keys deleted.
+    """
+    modified_dict = copy(dictionary)
+    key_list = key.split('.')
```
JarbasAl, line 47: should the separator be made into an argument ?

krisgesling: This is true, but then we need a way to configure that delimiter so we'd be adding another mycroft.conf attribute. 

Should we just use the right data structure to begin with and make it a list of lists of strings...
```python
[['nested', 'key', 'example'], ['second', 'key']]
```

This seems really painful to write out if you want to actually use it but who knows how often it will actually get used and only has to be done once

JarbasAl: see https://github.com/OpenVoiceOS/ovos_utils/pull/66

forslund:

PR #2872, opened 2021-03-23
config for root people

JarbasAl:
replaces https://github.com/MycroftAI/mycroft-core/pull/2861, based on https://github.com/MycroftAI/mycroft-core/pull/2861#issuecomment-804684059

think of this PR as a bunch of ways to require `root` to edit the .conf

Goals:
- allow system administrators to provide default values (eg, `stt`, `tts`)
- allow system administrators to block individual configuration options from `REMOTE` (eg, `enclosure`)
- allow system administrators to block individual configuration options from `USER` (eg, `data_dir`)
- allow system administrators to block configuration completely

### DIFF
--- a/mycroft/configuration/config.py
+++ b/mycroft/configuration/config.py
@@ -203,9 +203,18 @@ class Configuration:
         Returns:
             (dict) merged dict of all configuration files
         """
+        # system administrators can define different constraints in how
+        # configurations are loaded
+        system_conf = LocalConf(SYSTEM_CONFIG).get("system") or {}
+        protected_keys = system_conf.get("protected_keys") or {}
+        protected_remote = protected_keys.get("remote") or []
+        protected_user = protected_keys.get("user") or []
+        skip_user = system_conf.get("disable_user_config", False)
+        skip_remote = system_conf.get("disable_remote_config", False)
+
         if not configs:
-            configs = [LocalConf(DEFAULT_CONFIG), RemoteConf(),
-                       LocalConf(SYSTEM_CONFIG), LocalConf(USER_CONFIG),
+            configs = [LocalConf(DEFAULT_CONFIG), LocalConf(SYSTEM_CONFIG),
+                       RemoteConf(), LocalConf(USER_CONFIG),
                        Configuration.__patch]
         else:
             # Handle strings in stack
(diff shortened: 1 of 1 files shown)


forslund:
Haven't had opportunity to try this. But I think it's a good idea if we're switching the order of the config resolution to have this sort of blocklist feature.

A possible alternative is to add an additional config file but this approach provides much finer control without much extra complexity so I like this alot.

mycroft/configuration/config.py
```diff
@@ -203,9 +203,18 @@ def load_config_stack(configs=None, cache=False):
         Returns:
             (dict) merged dict of all configuration files
         """
+        # system administrators can define different constraints in how
+        # configurations are loaded
+        system_conf = LocalConf(SYSTEM_CONFIG).get("system") or {}
+        protected_keys = system_conf.get("protected_keys") or {}
+        protected_remote = protected_keys.get("remote") or []
+        protected_user = protected_keys.get("user") or []
```
krisgesling, line 211: I can see why we'd split this from a "why not have more choice" perspective but it also seems like we're going to effectively merge remote and user configs anyway. At least in the sense of how a user interacts with it.

If that's the case then it seems like we should do this as a single `protected_keys` attribute, otherwise we're encouraging projects to use them separately, and will then need to deprecate that option soon.

 It feels much cleaner to think of:
- Mycroft defaults
- Platform defaults and requirements (as defined by the protected_keys)
- User settings

mycroft/configuration/config.py
```diff
@@ -203,9 +203,18 @@ def load_config_stack(configs=None, cache=False):
...
+        skip_user = system_conf.get("disable_user_config", False)
+        skip_remote = system_conf.get("disable_remote_config", False)
+
```
krisgesling, line 214: We could also have a `user_configurable_keys` which would act as the inverse of `protected_keys`.
ie if a project wants to protect everything except x and y:
```
"user_configurable_keys": ["x", "y"]
```

This is just an idea though, not necessary in v1 of this. I think let's add it if we see actual need for it.

mycroft/configuration/config.py
```diff
@@ -216,6 +225,17 @@ def load_config_stack(configs=None, cache=False):
         # Merge all configs into one
         base = {}
         for c in configs:
+
+            # handle system constraints
+            if isinstance(c, RemoteConf):
+                if skip_remote:
+                    continue
+                c = {k: v for k, v in c.items() if k not in protected_remote}
```
krisgesling, line 233: I may be reading this wrong, but this will only handle top level keys. For things like the `listener` we likely want to allow `listener.wake_word` but protect others like `listener.sample_rate`

krisgesling (commented):
Yeah, this is broadly what I was thinking too. 

Some comments below but it might be easier for me to PR another possible implementation.

Feels like we're getting close!

JarbasAl:
closed in favor of https://github.com/MycroftAI/mycroft-core/pull/2861

PR #2886, opened 2021-04-27
Cleanup Adapt ContextManager implementation

forslund:
## Description
Makes the Mycroft-core Adapt Context Manager depend on the ContextManager in Adapt. This removes a bunch of duplicated code. The current state of the class reflects _only_ the modifications done to integrate the context with the Mycroft intent system. The handling of timeout has been moved into the earlier weird tuple-thing into the context frame metadata.

The key differences from the default Adapt context manager are
1. Frames are timed out after a set time
1. `get_context()` will only return the latest match
1. methods for removing context (all and specific)

This keeps the current implementation as-is, I need to dig into the history to see why point 2 was added and will be working to integrate some of the functionality in the context manager from adapt. (clear, remove and maybe a canonical way to limit the frame_stack)

I do however think this is a good first step.

## How to test
Ensure unit tests still work, check that wikipedia skill "tell more" functionality works

CLA [ Yes ]

### DIFF
--- a/mycroft/skills/intent_service.py
+++ b/mycroft/skills/intent_service.py
@@ -408,7 +408,7 @@ class IntentService:
         entity['match'] = word
         entity['key'] = word
         entity['origin'] = origin
-        self.adapt_service.context_manager.inject_context(entity)
+        self.adapt_service.add_context(entity)
 
     def handle_remove_context(self, message):
         """Remove specific context
--- a/mycroft/skills/intent_services/adapt_service.py
+++ b/mycroft/skills/intent_services/adapt_service.py
@@ -16,7 +16,7 @@
 from threading import Lock
 import time
 
-from adapt.context import ContextManagerFrame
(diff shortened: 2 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2886))

mycroft/skills/intent_service.py
```diff
@@ -409,7 +409,7 @@ def handle_add_context(self, message):
         entity['match'] = word
         entity['key'] = word
         entity['origin'] = origin
-        self.adapt_service.context_manager.inject_context(entity)
+        self.adapt_service.add_context(entity)
```
JarbasAl, line 412: method does not seem to exist in AdaptService

forslund: Thanks, this was in the first draft but seem to have missed updating here. will fix

forslund:

Issue #2723, opened 2020-10-15
Prep: LF 0.3.0 will require addition to test runner


ChanceNCounter:
Lingua Franca 0.3.0 will not load languages by default (see also: #2703). Hence, `bin/mycroft-skill-testrunner` will need to make sure a test's language is loaded.

This is trivial by either of two routes:

`lingua_franca.load_language(str: lang_code)` before starting a particular language's test suite, or
`lingua_franca.set_default_lang(str: lang_code)` to ensure the language is loaded, and make it LF's default. This way, it would support the omission of the `lang` parameter within localized test files. The latter only works if tests are run consecutively, rather than in order, or they'll clobber each other's invocations of `set_default_lang()`.

ChanceNCounter:
This would need to be done simultaneously with the LF integration, and not before, because (obviously) those functions are not yet available.

ChanceNCounter:
I've just realized, "run consecutively rather than in order" is the most meaningless of typos =P

Testing by setting the default language will only work if tests are run *consecutively* rather than *concurrently or in parallel*. Unittest and pytest, by default, run tests in a manner which *I* certainly can't predict, and getting them to run tests *in* order is challenging (and would be a big hassle for new devs.) I have no idea how VK does it, but when I tried to test Lingua Franca that way, like I said, the tests clobbered each other's `set_default_lang()` calls.

In LF, I solved this by spamming a function in the test file I called `unload_all_languages()` 👅 but that's not necessary here, just important to note that `load_language(...); call_function(data, lang=lang_code)` will be more reliable *within tests* than `set_default_lang(...); call_function(data)`

ChanceNCounter:
aaaand I should further point out *that* only happens *within test files*. LF's test suite is able to set default languages *per file* with no problems. It's files that contain multiple calls to `set_default_lang()`, within different tests, that have problems.

End of nitpick.

ChanceNCounter:
If Mycroft-core goes straight to LF 0.4, there's an antipattern available (👅) that *should* (hopefully) obviate this problem:

`lingua_franca.config.load_langs_on_demand = True` will do some nonsense that's intended only (and specifically) for situations in which the implementation doesn't know ahead of time what languages it will need to invoke. This trades performance - *presently* consisting of a whole lot of negligible calls - for convenience. It can increase *or* decrease heap size, depending on how often LF functions are called on unloaded languages. I don't think it will be a problem for the test runner. I just wish people didn't know it existed 😉

[...]

PR #2890, opened 2021-04-29
Upgrade Lingua Franca to v0.4.1

krisgesling:
## Description
Upgrades Lingua Franca to v0.4.1. 

This includes a major refactor, in which @JarbasAl has reorganized and standardized naming conventions. After that, @ChanceNCounter overhauled the way Lingua Franca handles localized code. The lib now crawls itself whenever a module is imported, caches localized functions in loaded languages, and imports them dynamically.

Furthermore (as implied above) Lingua Franca no longer loads any of its important functions by default. Instead, a given language's functions must be loaded by calling lingua_franca.load_language(<lang code>), or by passing a list of language codes to lingua_franca.load_languages(). Lastly, one can change the "default" language - the one which Lingua Franca will use if a language is not specified when you call a function - using `mycroft.util.lang.set_default_lf_lang(<lang code>)`  or directly with `lingua_franca.set_default_lang(<lang code>)`. This will load the specified language, if it isn't already loaded.

A default language has been added to wrapped methods to ensure they always reflect the configured language unless another language is explicitly requested. This is also because the passing of `lang=None` has been deprecated in Lingua Franca. Skill authors are encouraged to omit the lang parameter entirely, so that their Skills will always use the currently-set default language.

At the moment, up to two languages are loaded by default: the user's configured language from `mycroft.conf`, and English. Both are loaded along with the Skill Manager. The Skill Manager was chosen simply because it's a component that will be running before Mycroft has the opportunity to call other LF functions.

Languages are loaded in individual unit test files in the event that developers run those files in isolation.

Big thanks to @ChanceNCounter for the initial work on this update which can be found in PR #2772. 

## How to test
Unit tests FTW

- [x] CLA


devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2890)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2890/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2890/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2890/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2890/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2890/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2890)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2890/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2890/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2890/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2890/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2890/logs/enclosure.log)

krisgesling:
I'm realising that assigning the default lang in the wrappers is going to prevent the default lang in LF from being used...  Going to pick this up again in the morning but still interested in any comments / reviews.

mycroft/util/parse.py
```diff
@@ -104,8 +107,8 @@ def extract_numbers(text, short_scale=True, ordinals=False, lang=None):
     Returns:
         list: list of extracted numbers as floats, or empty list if none found
     """
-    return lingua_franca.parse.extract_numbers(text, short_scale,
-                                               ordinals, lang)
+    return lingua_franca.parse.extract_numbers(text, short_scale, ordinals,
+                                               lang or get_default_lang())
```
ChanceNCounter, line 111: LF should be handling `lang=None` with a DeprecationWarning. Should Mycroft wrap and fix this? It'll make for cleaner logs at the consumer side, but Skill devs won't catch the warnings.

mycroft/util/parse.py
```diff
@@ -216,7 +225,7 @@ def extract_duration(text, lang=None):
                     be None if no duration is found. The text returned
                     will have whitespace stripped from the ends.
     """
-    lang_code = get_primary_lang_code(lang)
+    lang_code = get_primary_lang_code(lang or get_default_lang())
     return lingua_franca.parse.extract_duration(text, lang_code)
```
ChanceNCounter, line 229: No need to pass a primary lang code here. LF will `extract_duration(lang='en-AU')` and that may even be desirable, when the number scales don't match!

test/unittests/util/test_format.py
```diff
@@ -429,17 +441,19 @@ def test_nice_year(self):
                 self.assertEqual(p['assertEqual'], nice_year(
                     dt, lang=lang, bc=ast.literal_eval(p['bc'])))
                 i = i + 1
+        set_default_lf_lang(default_lang)
 
         # Test all years from 0 to 9999 for all languages,
         # that some output is produced
         for lang in self.test_config:
+            set_default_lf_lang(lang)
```
ChanceNCounter, line 449: This may or may not fail on some runs. Neither unittest nor pytest run tests in a reliable order. However, if it succeeds, this beats the alternative, which is spamming a function that unloads all the other languages.

A third option is setting `config.load_langs_on_demand`, but this is different behavior from what Mycroft users will usually see at runtime, so not ideal for these tests.

ChanceNCounter:
So far, I've only skimmed. Will take another couple passes over the course of the day. But it's looking good!

Reminder: #2723

mycroft/util/parse.py
krisgesling: Yeah - I did it to quiet the 797 warnings you get from the unit tests :scream: 

But it's not necessary and we should be able to drop most (if not all) of the wrappers over time.

test/unittests/util/test_format.py
krisgesling: So far haven't seen any issues but good to note in case that randomly starts ocurring and we can't work out why

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2890)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2890/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2890/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2890/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2890/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2890/logs/enclosure.log)

krisgesling:
This let me shift the new methods to the `mycroft.configuration` module which seems like a better location for configuring a language. Couldn't do it while there was the `get_default_lang` method as it had a circular import issue.

This and your comments about the test runner have gotten me wondering whether Skill or plugin developers would want to call things like `load_language`? If so I wonder if it would be worth including them in the `mycroft.configuration` API too. They could do it via `lingua_franca.load_language` but having everything in the mycroft namespace makes sense to me. As a Skill developer it also tells me that it is a supported method and will be kept compatible within a major release cycle.

Any thoughts?

ChanceNCounter:
I have no particular feelings one way or the other. If wrapping will make things easier for skill devs, go for it!

On the other hand, I think it's almost always going to be Mycroft's problem to load and unload languages. In writing the LF API that way, I was thinking about exactly that, but it needs to be part of a broader algorithm that also ensures STT and TTS are available.

A possible exception is a translation skill, but that would almost require `load_langs_on_demand`, which API will change in LF 0.5.0 if we make the roadmap.

mycroft/util/parse.py
ChanceNCounter: Ironically, that DeprecationWarning is ahead of a custom exception that could be ignored XD

mycroft/skills/intent_service.py
```diff
@@ -153,7 +152,7 @@ def get_skill_name(self, skill_id):
     def reset_converse(self, message):
         """Let skills know there was a problem with speech recognition"""
         lang = _get_message_lang(message)
-        set_active_lang(lang)
+        set_default_lf_lang(lang)
```
ChanceNCounter, line 155: forward-thinking Gez thinks forward and I approve

mycroft/skills/mycroft_skill/mycroft_skill.py
```diff
@@ -596,7 +596,7 @@ def ask_selection(self, options, dialog='',
                 if self.voc_match(resp, 'last'):
                     resp = options[-1]
                 else:
-                    num = extract_number(resp, self.lang, ordinals=True)
+                    num = extract_number(resp, ordinals=True, lang=self.lang)
```
ChanceNCounter, line 599: are there circumstances under which `self.lang != lingua_franca.get_default_lang()`?

krisgesling: Yeah I thought about this briefly when changing it, there's at least one more instance in that file to `join_list()`

My thinking was that the `ask_selection()` method has no `lang` parameter and will therefore always use `self.lang` to select it's dialog. So we should definitely have `join_list(options, "or", lang=self.lang)` as it will be part of the dialog. `if self.lang != default_lingua_franca_lang` then any dialog that requires some `lf.format` is going to be multilingual and likely not what was intended.

For parsing it makes more sense as it will return a computer readable object but I still think it would be better to use the `lang` argument on a method rather than setting the default language globally.

In short - I don't think they should be different, but there's nothing programmatically stopping that from happening. So my gut is saying it's safer to be explicit here.

mycroft/skills/intent_service.py
krisgesling: It was also to avoid people thinking this would set their default language globally across all of Mycroft.

mycroft/skills/mycroft_skill/mycroft_skill.py
ChanceNCounter: Good call, and good catch!

ChanceNCounter:
Looks like several `lang=None`s remain in function signatures. These are probably responsible for most of the DeprecationWarnings. We should get those out of there, cuz those functions *will* throw those warnings in prod.

mycroft/configuration/lang.py
```diff
@@ -19,4 +19,13 @@
 lingua-franca (https://github.com/mycroftai/lingua-franca) selected language
 """
 
-from lingua_franca.lang import set_active_lang, get_active_lang
+from lingua_franca.lang import set_default_lang as _set_default_lf_lang
+
+
+def set_default_lf_lang(lang_code="en-us"):
```
chrisveilleux, line 25: why put `lf` in the function name?  one of the points of wrapping libraries like this is to abstract out the dependency on said library.

mycroft/skills/__main__.py
```diff
@@ -28,15 +28,14 @@
 from mycroft.api import is_paired, BackendDown, DeviceApi
 from mycroft.audio import wait_while_speaking
 from mycroft.enclosure.api import EnclosureAPI
-from mycroft.configuration import Configuration
+from mycroft.configuration import Configuration, set_default_lf_lang
```
chrisveilleux, line 31: I have always been taught that this is an anti-pattern.  Why import it into `configuration.__init__` just to import it here?  This should be importing from mycroft.util.lang

mycroft/skills/intent_service.py
```diff
@@ -16,8 +16,7 @@
 from copy import copy
 import time
 
-from mycroft.configuration import Configuration
-from mycroft.util.lang import set_active_lang
+from mycroft.configuration import Configuration, set_default_lf_lang
```
chrisveilleux, line 19: same comment as in `skills.__main__` import directly from mycroft.util.lang

mycroft/skills/skill_manager.py
```diff
@@ -143,6 +146,9 @@ def __init__(self, bus, watchdog=None):
         self._define_message_bus_events()
         self.daemon = True
 
+        self.lang_code = self.config.get("lang", "en-us")
+        load_languages([self.lang_code, "en-us"])
```
chrisveilleux, line 150: How does `load_languages` differ from `set_default_lang`?  Why do we need both?

mycroft/util/format.py
```diff
@@ -34,7 +34,6 @@
 from lingua_franca.format import (NUMBER_TUPLE, DateTimeFormat, join_list,
                                   date_time_format, expand_options,
                                   _translate_word)
```
chrisveilleux, line 36: importing a private object is a no-no.  If we are importing it, make it public.

mycroft/configuration/__init__.py
```diff
@@ -13,4 +13,5 @@
 # limitations under the License.
 #
 from .config import Configuration, LocalConf, RemoteConf
+from .lang import set_default_lf_lang
```
chrisveilleux, line 16: This should not be imported here if it is not used here.

chrisveilleux (changes requested):

mycroft/skills/skill_manager.py
JarbasAl: see the changelog of the release that introduced this https://github.com/MycroftAI/lingua-franca/releases/tag/release%2Fv0.3.1

and this PR https://github.com/MycroftAI/lingua-franca/pull/108

chrisveilleux: I just looked and there was no reasoning included.  Just that loading languages must now be done.  So my question remains.  Also, it looks like there is `load_language` and `load_languages`.  If we are only loading one language, why use the plural version?

JarbasAl:
```python
lingua_franca.load_language('en')          # 'primary' lang code (package's terminology)
lingua_franca.load_language('pt-pt')       # full lang code
lingua_franca.load_languages(['en', 'es']) # multiple languages at once, first will be default
lingua_franca.set_default_lang('es')       # self-explanatory
```

`default_lang` is the lang to be used if no `lang` is passed to the methods, `load_language` is what actually loads the lang data into memory, the linked PR goes into detail on the rationale of why they arent just all loaded at once (memory saving and performance reasons)

ChanceNCounter: * `set_default_lang` sets the default value of the `lang` parameter in localized functions (anything that accepts that parameter =P)
* `load_language` loads a single language's functions into memory, and expects a string
* `load_languages` loads multiple languages at once, and expects a list, but performs a sanity check so that one should never need to *invoke* them both

In Mycroft's case, many things still crash when English parsers and formatters are unavailable, so when the user's configured language is not English, we need to load both.

JarbasAl: i suppose set_default_lang could be made to automatically load the lang if needed, there is also a config option to load langs on demand. but thats a lingua_franca issue, in the next LF release we can add that, and when it is bumped again in core remove one of these function calls. in the current version of LF both calls are needed

ChanceNCounter: Incidentally, calling `load_language` with no default set will set that language to the default. Similarly, calling `load_languages` with no default set will use the first element of the input.

Put differently, if no default language is set,

* `load_language('en')` will make English the default ('en-us', to be specific, but that's for the next version bump)
* `load_languages(['es', 'en'])` will make *Spanish* the default, because Spanish is the first element in its input.

This is why we don't need to call `set_default_lang` after this `load_languages` call.

mycroft/configuration/lang.py
krisgesling: I wanted to make it clear that it's not the default language for all of Mycroft - so the full name would be something like:
`set_default_formatting_and_parsing_language()`

We could make it:
`set_default_util_lang()`
?

That's more correct than `set_default_lang()`, but less correct than `set_default_lingua_franca_lang()`

mycroft/skills/__main__.py
krisgesling: Assuming you mean directly from `mycroft.configuration.lang`?

This felt cleaner than doing two imports in each file:
```
from mycroft.configuration.lang import set_default_lf_lang
```
But happy to change it.

Unless you meant - why not import directly from `lingua_franca.lang`? In which case the intention was to explicitly say that this is part of the mycroft-core API and it's acting on the methods available in mycroft-core.

Also for context - it was initially in `mycroft.util.lang` however even though it was to act on the `mycroft.util.format` and `parse` modules it felt like a configuration method, so I moved it over here. Again I'm not strongly attached to any particular implementation, this just felt better.

mycroft/util/format.py
krisgesling: Agree.

This was not introduced by this PR. A quick look tells me that 
- this private method is being used for `nice_duration()` and `nice_duration_dt()`. 
- currently `lingua_franca.format.nice_duration()` is not directly compatible with `mycroft.util.format.nice_duration()`
- `nice_duration_dt` (duration between two datetimes) doesn't exist in LF - but presumably we can just get the timedelta and call `nice_duration()`, or add it to LF.
- Making `nice_duration()` compatible will require some work outside the scope of this PR. We possibly want to look at what the available arguments are too. There are a few bools to format it in different ways but I'm not clear on what happens if a dev passes True for conflicting arguments eg `nice_duration(duration, speech=True, clock=True)`
   - (I think speech wins and clock gets ignored)

I've added #TODO's at least. Doing more of the above will be much more refactoring and may require breaking changes to core and/or Lingua Franca.

mycroft/configuration/__init__.py
krisgesling: As above - the intention was to define the `mycroft.configuration` API but happy to put it wherever depending on the outcome of that thread.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2890)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2890/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2890/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2890/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2890/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2890/logs/enclosure.log)

mycroft/util/format.py
ChanceNCounter: `nice_duration_dt` was my mediocre doing, and I'd still have to go back and look at it =P it needs to be migrated or die, probably. Should happen when `nice_duration` is reconciled.

mycroft/skills/skill_manager.py
chrisveilleux: Ok, but `set_default_lang` is called elsewhere in this PR.  Why even expose `set_default_lang` if you need to call `load_language` anyway and it does the work for you?

ChanceNCounter: Drive-by: 

One may wish to change the default language at runtime. Indeed, the three functions were written with Mycroft localization in mind; locales' or entire languages' parsers and formatters can be loaded and unloaded alongside wake words, STT and TTS plugins/models, etc, in one big "change Mycroft language" algorithm.

If I could figure out a way to help users who change to a language they don't speak, and then can't change back, I'd write a skill to do that today, YMMV =P

As to where it's called in this PR, I noticed the one in `reset_converse`, replacing an `active_lang` call. This ensures that the default language didn't somehow get desynced from Mycroft, which is one possible cause of a failure (though unlikely, since the func that would be responsible is wrapped.)

Otherwise, I saw it called in tests. Did you notice it anywhere else?

ChanceNCounter: `load_language` doing the work for you is just sugar. If you're just importing LF to use it in a particular language, there's only one workflow, one function call, off you go.

For everything else there's MasterCard.

ChanceNCounter: It may also be worth explaining what all this is for:

Probably goes without saying, and just in general, Python doesn't usually release memory. It won't grab more, though, if the heap's big enough for its next assignment or import.

Lingua Franca is small, in terms of what a modern computer can do. However, every time a localized function is added, it grows by the number of supported languages. Every time it acquires a new language, it balloons by the number of localized functions. (This isn't entirely accurate, because the guts and helpers of a localized parser don't necessarily resemble the inside of its English counterpart, but that more or less comes out in the wash.)

That config option, `load_langs_on_demand`, makes all this unnecessary for use cases that don't know what language(s) they're using until the functions are called. But for Mycroft, we noticed that you don't actually need *most* of LF at any given moment. You just need the modules particular to your languages.

So now it's got a cache of function signatures for each loaded language and "top-level" module (`parse` and `format`.) The memory footprint is thus reduced in three ways:

* Localized modules that you are using are imported for introspection, then freed for GC
* Localized modules that you aren't using might as well not be there
* When you actually call a localized function, the function signature cache is consulted, and the appropriate function call is pre-assembled. The localized module is then imported, the function called, and once again the localized module is freed for GC

In percentage terms, this was a radical improvement. `load_langs_on_demand` makes the opposite trade; even lower memory consumption, but far more floppydoos.

Now LF goes fast on electronic vegetables, and fits the larger "this is an English language voice stack for Mycroft" paradigm that will surely define language support.

ChanceNCounter: Credit for this goes to three people: Jarbas, for cleaning everything up and standardizing how it's named, which is useful for the localizer-cache thing. Me, for deciding to Try It And See. Ake, for realizing that, no, it *wasn't* an unworkable idea, it just needs to cache function signatures instead of modules. Duh.

mycroft/util/format.py
krisgesling: Have also added an issue in mycroft-core that links to the issue about this in LF.

mycroft/skills/__main__.py
```diff
@@ -199,8 +199,8 @@ def main(alive_hook=on_alive, started_hook=on_started, ready_hook=on_ready,
     # Create PID file, prevent multiple instances of this service
     mycroft.lock.Lock('skills')
     config = Configuration.get()
-    # Set the active lang to match the configured one
-    set_active_lang(config.get('lang', 'en-us'))
+    lang_code = config.get("lang", "en-us")
+    load_languages([lang_code, "en-us"])
```
krisgesling, line 203: Was looking at this with Chris this morning and noticed that we were setting the default language before loading them. 
The setting was done here in the Skills main process, whilst the language loading was in the Skills Manager which is instantiated slightly further down. 
Given loading sets the default anyway, is there an issue with doing both up here?

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2890)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2890/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2890/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2890/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2890/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2890/logs/enclosure.log)

mycroft/skills/__main__.py
ChanceNCounter: Not as long as this will reliably occur before other potential calls to `load_language(s)`, it should be fine.

The oopsie-sugar works both ways - if you set a default which is not loaded, it will load that language. I felt it was important to minimize the number of times Mycroft (or anybody else) would need to invoke the boilerplate.

On the other hand, the language loaders ignore languages which are already loaded. Say that five times fast. This is why the 99% of cases in which that call is `load_languages(['en-us', 'en-us'])` is essentually harmless.

Hence, it shouldn't matter which of those functions you use, nor in which order, except to skip a couple if statements.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2890)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2890/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2890/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2890/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2890/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2890/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2890)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2890/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2890/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2890/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2890/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2890/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2890)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2890/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2890/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2890/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2890/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2890/logs/enclosure.log)

mycroft/configuration/lang.py
ChanceNCounter: FWIW, I reckon this will eventually be the default language for all of Mycroft, in that runtime TTS and STT swapping will either go in this function.

```python
def set_default_lang(lang: str):
    lingua_franca.set_default_lang(lang)
    mycroft.stt.something(lang)
    ...
```

which may not be an argument for calling it `set_default_lang()` *now*, but it might excuse the decision to give this a name directly reflective of what it wraps (like this one.)

ChanceNCounter: However, I would like to discourage direct imports from `lingua_franca.lang`, as the "user"-facing functions are all exposed in `lingua_franca`.

krisgesling: Agreed - I hope for this method to be short-lived and deprecated as soon as we can set all systems languages.

Also thanks for pointing out the import location - have updated that.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2890))

krisgesling:
@chrisveilleux feels like the name of `set_default_lf_lang()` is the only remaining hold up - do you have a preference or can we merge as is with the aim of deprecating that method for a true `set_default_lang()` when it's possible?

ChanceNCounter:
No opinion about that func name <3

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2890)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2890/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2890/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2890/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2890/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2890/logs/enclosure.log)

chrisveilleux (approved):

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2890))

PR #2901, opened 2021-05-06
Skill states

ken-mycroft:
## Description
Adds states to skills

## How to test
Should be backward compatible, use existing tests. For new functionality see example skill here

https://github.com/MycroftAI/skill-npr-news/pull/111

yes

### DIFF
--- a/mycroft/audio/services/simple/__init__.py
+++ b/mycroft/audio/services/simple/__init__.py
@@ -165,6 +165,7 @@ class SimpleAudioService(AudioBackend):
 
     def stop(self):
         LOG.info('SimpleAudioServiceStop')
+        self.bus.emit(Message('SimpleAudioServiceStop', {'data': None}))
         self._stop_signal = True
         while self._is_playing:
             sleep(0.1)
--- a/mycroft/skills/intent_service.py
+++ b/mycroft/skills/intent_service.py
@@ -25,6 +25,8 @@ from .intent_services import (
     AdaptService, AdaptIntent, FallbackService, PadatiousService, IntentMatch
 )
 from .intent_service_interface import open_intent_envelope
+from mycroft.messagebus.message import Message
+import json
 
 
 def _get_message_lang(message):
@@ -82,6 +84,7 @@ class IntentService:
         self.bus = bus
 
         self.skill_names = {}
+        self.skill_categories = {}
         config = Configuration.get()
         self.adapt_service = AdaptService(config.get('context', {}))
         try:
@@ -104,12 +107,24 @@ class IntentService:
         # Converse method
         self.bus.on('mycroft.speech.recognition.unknown', self.reset_converse)
         self.bus.on('mycroft.skills.loaded', self.update_skill_name_dict)
+        self.bus.on('mycroft.skills.list', self.update_skill_list)
+
 
         def add_active_skill_handler(message):
-            self.add_active_skill(message.data['skill_id'])
+            category = 'undefined'
+            if message.data.get("skill_cat", None) is not None:
+                category = message.data['skill_cat']
+
+            self.add_active_skill(message.data['skill_id'], category)
+
+        def remove_active_skill_handler(message):
+            self.remove_active_skill(message.data['skill_id'])
 
         self.bus.on('active_skill_request', add_active_skill_handler)
+        self.bus.on('deactivate_skill_request', remove_active_skill_handler)
         self.active_skills = []  # [skill_id , timestamp]
+
+        # BUG fix this!
         self.converse_timeout = 5  # minutes to prune active_skills
 
         # Intents API
@@ -138,6 +153,11 @@ class IntentService:
     def update_skill_name_dict(self, message):
         """Messagebus handler, updates dict of id to skill name conversions."""
         self.skill_names[message.data['id']] = message.data['name']
+        self.bus.emit(Message('skillmanager.list'))
+
+    def update_skill_list(self, message):
+        for skill in message.data:
+            self.skill_categories[message.data[skill]['id']] = message.data[skill]['cat']
 
     def get_skill_name(self, skill_id):
         """Get skill name from skill ID.
@@ -167,6 +187,7 @@ class IntentService:
             lang (str): current language
             message (Message): message containing interaction info.
         """
+        LOG.debug("do_converse()-utterances:%s, active_skills=%s" % (utterances, self.active_skills))
         converse_msg = (message.reply("skill.converse.request", {
             "skill_id": skill_id, "utterances": utterances, "lang": lang}))
         result = self.bus.wait_for_response(converse_msg,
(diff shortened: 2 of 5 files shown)


mycroft/skills/intent_service.py
```diff
@@ -333,12 +361,23 @@ def _converse(self, utterances, lang, message):
...
+        for skill in tmp_active_skills:
+            if skill[2] == 'system':
+                if self.do_converse(utterances, skill[0], lang, message):
+                    # update timestamp, or there will be a timeout where
+                    # intent stops conversing whether its being used or not
+                    return IntentMatch('Converse', None, None, skill[0])
+
         # check if any skill wants to handle utterance
-        for skill in copy(self.active_skills):
+        for skill in tmp_active_skills:
```
forslund, line 375: Should this exclude the entries where `skill[2] == system`? seems like those gets called twice now...

ken-mycroft: I thought the above loop would cause us to exit the method on a match, are you saying if line 372 gets executed " return IntentMatch('Converse', None, None, skill[0])" we would still fall thru to the below loop?

forslund: I could be reading the code wrong but I mean if none of the active_skills with the `system` category has a match (`do_converse()` returns `False` for all of them)

We would continue and run the whole list (including the skills with a `system` category) since they are still in the list, It wouldn't make any practical error since they'd return `False` again, but it'd take some extra time and isn't really needed to be performed again

ken-mycroft: I agree, I was just saying they would not run twice as the above loop would exit if a match occurred however you are correct that we could check if they were system level in the below loop to avoid wasting time checking them again, so yes i agree

ken-mycroft: Also note this PR is related to 

https://github.com/MycroftAI/skill-npr-news/pull/111

and 

https://github.com/MycroftAI/skill-alarm/pull/112

mycroft/skills/intent_service.py
```diff
@@ -333,12 +361,23 @@ def _converse(self, utterances, lang, message):
                               if time.time() - skill[
                                   1] <= self.converse_timeout * 60]
 
+
+        # first check for system levl skills
+        tmp_active_skills = copy(self.active_skills)
+        for skill in tmp_active_skills:
+            if skill[2] == 'system':
```
krisgesling, line 368: Is this the primary reason for having a category at the moment - to give `system` Skills priority access to converse?

Any Skill could then list themselves as a "system" Skill to gain the same priority access.

It looks like the `active_skill` list inserts any new item to the start of the list, so theoretically the most recently active Skill gets the first attempt at the utterance. If that's not working I'd be curious to know why but we could also the sort `tmp_active_skills` based on the `time.time`  in `skill[1]`.

So if we have a way to hit the mostly recently active Skill first, for the Alarm Skill could we reactivate it ensuring it's at the top of the active list each time the beep is going to get played?

mycroft/audio/services/simple/__init__.py
```diff
@@ -165,6 +165,7 @@ def play(self, repeat=False):
 
     def stop(self):
         LOG.info('SimpleAudioServiceStop')
+        self.bus.emit(Message('SimpleAudioServiceStop', {'data': None}))
```
krisgesling, line 168: Is this meant to be in this PR?

If we want to add messages to each audio service then I think we need to namespace them appropriately

krisgesling (commented):
It feels like we're trying to achieve two distinct but related things here:
1. Simple state management of Skills
2. Skill categories which are partly a hierarchy but may in the future be different categories at the same level.

In both instances, rather than having Skill developers responsible for defining and updating these, it seems that we should be able to derive them. This may also mean they better reflect the actual state / category, or maybe not...

The state implementation here seems focused on whether the Skill is in the active Skills list. If that's the case would it be sufficient to provide a method for Skills to deactivate themselves when they've finished their task? Eg a `make_inactive()` method to complement the existing `make_active()`

From my read of this, if we provide that ability, then the rest of this can be handled in the Skill rather than modifying core. I'm sure that this is only the first piece of a broader piece of functionality but I'd rather get an idea of what we're trying to do and what the scope is of that rather than adding things in a piecemeal fashion. Eg is state going to stay as [active, inactive] or are we looking at [waiting_for_response, thinking, etc]

Given this is going into the Mark II feature branch it's not going to impact other projects and we can pull it back out but we're making the merge of this feature branch more and more difficult by adding things like this. Since it can be done in the Skills for the moment I'm not sure I see the benefits outweighing that cost.

It also seems like the original aim is to improve the stopping functionality so I'd also suggest we look at that method and make it more like `converse()` rather than using converse as a substitute.

mycroft/audio/services/simple/__init__.py
ken-mycroft: I'm tired from haggling. This system has been in this state for many years now. If someone had a better idea they should have fixed it by now. I can appreciate the attitude some of our community members and why they ultimately decide to fork. While there are many different ways these issues could be addressed I am not interested in altering this PR to work around any. Your ideas sound fine, please squash this PR get your ideas working and add me to the review list after you submit your new PR with the working code.

forslund: Hi Ken, as a Swede I don't generally interfere (neutrality and all that), I respect you and I'm sure this came off more aggressive than you intended but this comes of rather rude and I must speak out since I don't think Kris deserves this bite.

(This thread in particular is valid since it's a query about something that's not mentioned in the PR or comment at all.)

The general discussion about the approach is probably the more important thing, we need the open discussion to make better designs. That is one of the strengths of open source, sharing and building upon each other's ideas. Instead of biting back, answer his concerns and make your thoughts on the issue clear to the rest of us. I'm sure you've thought it through. I think the skill state may be an interesting feature, this is the initial commit and will need to be streamlined in someway to fit into the system but will likely be very convenient.

Will now go back to the Swedish way of "making a fist in my pocket", and be silent.

P.S. I think the general frustration in the community is when no response is given, not when response is received. D.S.

ken-mycroft: My message was simply if we have a better approach why have we waited so long to implement it?  I have already put the effort into solving this problem, however, I am always open to alternative solutions and look forward to reviewing any contributions which solve the problem in a different manner. There is, in this case, about a 40 hour difference between a good idea and a working solution. I have other concerns regarding how we handle contributions from the community in general but that is a different topic.

forslund:
How is the intent name stored in the state list? does that use the intent name or the `skill_id:intent_name` format? enable / disable prepends the skill id automatically, maybe it get's doubled?

mycroft/audio/services/simple/__init__.py
ken-mycroft: in the active skills array we don't store intents, we store skill tuples (skill_id, timestamp, category). One thing I should explain is I have not verified if the stop intent handlers are still being called, I was just surprised they showed up in the intent results at all but maybe that is by design and they are not actually getting called.

forslund: the active_skills are only used for converse calls. where active skills may short-circuit the intent service. The active skills list will not in any way block the normal intent handling through adapt, padatious and fallback services unless you changed the behavior (I don't see that here, but may be wrong).

forslund: It will not fire if it's disabled, see my example below. In that case nothing is firing for the OtherAdapt and the "unknown" skill is triggered instead.

PR #2910, opened 2021-05-28
Check for the AVX instruction set during install

krisgesling:
## Description
Add a check for the AVX instruction set on non-ARM processors during install. If this required instruction set is not found it will warn the user that the Precise Engine will not be supported.

Replaces #2269
Thanks to @mathmauney for the original work on this.

## How to test
Run installer on a machine that does not support AVX
or change line 140 so it is searching for something that definitely doesn't exist in the cpuinfo eg change "avx" to "avxzzz"

Yes - both original author and myself.

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -184,7 +184,13 @@ in mycroft.conf.
   Y)es, I want to use the PocketSphinx engine or my own.
   N)o, stop the installation."
       if get_YN ; then
-        sed -i "s/\"use_precise\": true/\"use_precise\": false/" config/
+        if [[ ! -f /etc/mycroft/mycroft.conf ]]; then
+          if [[ ! -e /etc/mycroft/ ]]; then
+            $SUDO mkdir /etc/mycroft
+          fi
+          $SUDO cp $TOP/mycroft/configuration/mycroft.conf /etc/mycroft/
+        fi
+        sed -i "s/\"use_precise\": true/\"use_precise\": false/" /etc/mycroft/mycroft.conf
       else
         echo -e "$HIGHLIGHT N - quit the installation $RESET"
         exit 1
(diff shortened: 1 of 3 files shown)


krisgesling:
Rebased - please also be sure to squash commits if merging.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2910))

dev_setup.sh
```diff
@@ -174,6 +174,30 @@ This script is designed to make working with Mycroft easy.  During this
 first run of dev_setup we will ask you a few questions to help setup
 your environment.'
     sleep 0.5
+    if ! grep -q avx /proc/cpuinfo && [[ ! $(uname -m) == 'arm'* ]]; then
```
ken-mycroft, line 177: Forgive me but I am not sure of the objective. I understand the check for avx via cpuinfo but must admit the arm check is confusing me. What is going on here?

ken-mycroft (commented):
Just confused about the 'arm' conditional.

el-tocino:
@ken-mycroft AVX is an x86 construct, ARM does not have it.

dev_setup.sh
krisgesling: As eltocino replied below, however it is the inevitable question - so I've added a comment to explain for future devs.

The AVX instruction set is an x86 construct. ARM has a range of equivalents, but I'm not sure which are supported or not by TF. So this is intended to catch the bulk of the support issues we get which is from people trying to install Mycroft on older or low-end x86 hardware.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2910))

ken-mycroft (approved):
Looks good. Go ahead.

PR #2927, opened 2021-06-24
stop passing lang=None to Lingua Franca

ChanceNCounter:
LF's only breaking change over the past two versions has been the
deprecation of `lang=None` as a valid parameter. This is because the new
language loading paradigm wants to load certain functions on the fly,
which it cannot do when it is explicitly told to look for a null lang.

I've addressed this by passing `lingua_franca.get_default_lang()` where
the `lang=None` call remained.

Bonus: Gets rid of over 200 DeprecationWarnings in unit tests!

### DIFF
--- a/mycroft/util/format.py
+++ b/mycroft/util/format.py
@@ -31,6 +31,7 @@ from calendar import leapdays
 from enum import Enum
 
 # These are the main functions we are using lingua franca to provide
+from lingua_franca import get_default_loc
 # TODO 21.08 - move nice_duration methods to Lingua Franca.
 from lingua_franca.format import (
     join_list,
@@ -104,6 +105,7 @@ def _duration_handler(time1, lang=None, speech=True, *, time2=None,
     Returns:
         str: timespan as a string
     """
+    lang = lang or get_default_loc()
     _leapdays = 0
     _input_resolution = resolution
     milliseconds = 0
--- a/mycroft/util/parse.py
+++ b/mycroft/util/parse.py
@@ -30,6 +30,7 @@ use in Mycroft Skills.
 from difflib import SequenceMatcher
 from warnings import warn
 
+from lingua_franca import get_default_loc
 from lingua_franca.parse import (
     extract_duration,
     extract_number,
@@ -115,4 +116,7 @@ def extract_datetime(text, anchorDate="DEFAULT", lang=None,
                                 "deprecated. This parameter can be omitted."))
     if anchorDate is None or anchorDate == "DEFAULT":
         anchorDate = now_local()
-    return _extract_datetime(text, anchorDate, lang, default_time)
+    return _extract_datetime(text,
+                             anchorDate,
+                             lang or get_default_loc(),
+                             default_time)


ChanceNCounter:
Whoops. Commit message lied: it's `get_default_loc()`, representing a full BCP-47 language code (`en-US`), not `get_default_lang()`, which provides only the "primary" lang code (`en`)

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2927))

krisgesling:
Hey thanks Chance - I can't remember why I didn't address the rest of these warnings in the LF update PR.

I'm a little bit skeptical about passing config values off to another package and then retrieving them from there. Is there any problem in getting the default lang code straight from the Mycroft config?

eg:
```python
def extract_datetime(... args):
    ...

    if lang is None:
        lang = Configuration.get()["lang"]

    return _extract_datetime(text, anchorDate, lang, default_time)
```

_Note: Unfortunately there is a cyclical import error if you try to import Configuration globally into the parse module._

Unless someone is intentionally changing the LF config then the mycroft config should be the same. And if someone is intentionally using LF's more advanced features they can import and use the package directly since it's guaranteed to be on the system. But even if they do this, another Skill calling `mycroft.util.parse.extract_datetime` will still get its expected default lang.

ChanceNCounter:
In general, I agree. However, it's (conceptually) possible for Mycroft's "active" default not to be the same as the one shown in `mycroft.conf`, and that will only become truer as robust multilingualism is implemented. Config might reflect that Mycroft should spin up in English, but if I tell it to switch to Spanish "for now," it's going to do that at runtime (hopefully) without touching config. It would be nice if these could refer to a value in Mycroft-core reflecting the current, runtime defaults, but imo that's out of scope.

Furthermore, LF's defaults can be swapped on the fly, even by Skills, and I think that's a feature in *very* niche cases. Very, very niche. Super niche. I'd rather people didn't, but I assume they will, and I don't know how you'd stop them in practice =P

This means it's hypothetically possible for a Skill author to do something like this:

```python
lingua_franca.set_default_lang(not_mycroft's_default_lang)
for something:
  foo = format.nice_duration(...)
  ...
lingua_franca.set_default_lang(mycroft's_default_lang)
```

I can't imagine why anyone *would*, for these functions, but so long as Mycroft can't resolve that situation, I think it's better to do it via LF than to crash the Skill.

Finally, but much less important, these functions were supposed to go with the rest of LF. When I submitted the helper, I was a new contributor, and didn't realize the spinoff was in progress. All of which is to say, these really aren't supposed to be here anymore and it's mostly my fault 👅 

All of which is to say, reading from mycroft.configuration makes sense, and I don't mind making the change, but atm I don't think it programmatically represents what it looks like it should mean. It might, someday, but it might also be some other name in the `mycroft.` namespace.

chrisveilleux:
I agree with the points both @krisgesling and @ChanceNCounter made.  Circular imports are usually a sign of bad code organization.  There are a few things that  can cause circular imports with the Config module.  At some point in the future, I would like to find them all and address them.  But I digress.  If it is possible that the default language set in LF can be different than what is configured in core, this is a good change.

chrisveilleux (approved):

Issue #2930, opened 2021-06-24
tts cache creates "None" folder


JarbasAl:
the latest cache refactor is creating a folder named "None" due to the handling of persistent cache added for mimic2 in https://github.com/MycroftAI/mycroft-core/pull/2853

cache is also deleted on boot, the min_percent param is set to 100%, "Remove cache data if disk space is running low." actually means "always remove cache data". 

proposed fix: https://github.com/HelloChatterbox/HolmesV/pull/61

PR #2932, opened 2021-06-25
WIP: Add per TTS config for free disk space

krisgesling:
## Description
This adds configurable values for required free disk space per TTS service.
If free disk space falls below these values, the temporary TTS cache will be curated.

Split from #2930 

**Problem # 1 - tests failing**
This seems to be working in practice but I haven't got the tests to work.
At first I thought it was because the cached files didn't exist, hence didn't take up any space and so the calculation in curate_cache would have seen them as irrelevant. However populating them with some data didn't fix this.

Assuming this is a simple fix and just needs a fresh set of eyes on it to find my mistake...

**Problem # 2 - Are we curating caches properly**
It also raises a question for me as to whether the `curate_cache` should actually use an "OR" eg
```python
if percent_free < min_free_percent or space.free < min_free_disk:
```
rather than the current "AND"
```python
if percent_free < min_free_percent and space.free < min_free_disk:
```

This I think is partially how this issue arose. We currently have a call to the method that only includes the `min_free_percent` argument. It would be assumed that this would then be the factor on whether curation happened or not. However it must also fall below the default disk space argument. So even passing in 99% free disk percentage required, the cache will not be curated unless you also have <50MB of disk space as well.

## How to test
- wait until unit tests work

### DIFF
--- a/mycroft/util/file_utils.py
+++ b/mycroft/util/file_utils.py
@@ -20,8 +20,10 @@ accessing and curating mycroft's cache.
 
 import os
 import psutil
-from stat import S_ISREG, ST_MTIME, ST_MODE, ST_SIZE
 import tempfile
+from pathlib import Path
+from stat import S_ISREG, ST_MTIME, ST_MODE, ST_SIZE
+from typing import List
 
 import mycroft.configuration
 from .log import LOG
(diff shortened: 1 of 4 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2932))

forslund:
I think the test issue is that the `curate_mock.return_value` isn't set in `test_curate_cache_low_disk_space` so the curation. Since you're creating actual files on disk it also works if I remove the mock completely.


Regarding Problem 2, the reason for this (as I recall it) is that it's designed for both small partitions and large ones. Say a 60 MB RAM Disk, in that case we don't want to start curating until we reach the percentage. On the other hand on a large disk where we can have 1% free and still have 1 GB disk space we shouldn't start to curate the cache.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2932)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2932/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2932/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2932/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2932/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2932/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2932)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2932/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2932/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2932/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2932/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2932/logs/enclosure.log)

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2932)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2932/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2932/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2932/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2932/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2932/logs/enclosure.log)

krisgesling:
Was thinking about another approach and it seemed quick enough to warrant some example code. Haven't done any tests for this, or even verified that it runs as expected, interested to hear if I'm on the right path first.

The idea is that we provide a more flexible approach based off a set of limits, again defined per TTS engine. With this new method, all limits should be adhered to at all times. So there will always be the minimum free disk space whether that is defined as a percentage or a fixed MB value. Likewise, the cache will never exceed the max cache size values. A TTS engine or other service may pass as many or as few limits as they like. Passing a single limits dictionary should also make it easier to add any other parameters in the future (eg a whitelist of paths). 

There are currently no sanity checks. 
- I couldn't think of a scenario where min and max values might conflict. 
- If someone sets no min free and their max size exceeds their disk size it will fill it right up.

Worst case scenarios here are that:
- values are too extreme and the cache is constantly cleared
- values are too lax and the disk gets full

In terms of the `tts.{engine}.cache_limits` dict - I tossed up between `cache_limits` and `cache_config` since it may in the future contain things that aren't explicitly limits but do relate to them. Compared to the `preloaded_cache` path which would certainly fit in a config dict, but not in a limits dict... very open to discussion or suggestions here.

mycroft/util/file_utils.py
```diff
@@ -179,6 +181,73 @@ def _delete_oldest(entries, bytes_needed):
...
+    """Reduce cache size if any cache limits are exceeded.
+
+    Args:
+        directory: directory path that holds cached files
+        limits: set of limitations for the cache size.
+                May include one or more of the following values as floats:
+                - min_free_disk_percent
+                - min_free_disk_space
+                - max_usage_disk_percent
+                - max_usage_disk_space
```
forslund, line 194: What unit is the disk space given in? MB?

forslund:
Since this is mainly a per system config. I.e the limits are something that is mainly dictated by the partition size where the cache is stored I think there should preferably be a config for all TTS's that can be overridden by the setting for the specific TTS engine.

 (I assume there are cases where certain TTS engines would like to always re-synthetizise the audio and such so the per TTS may be useful in such cases)

forslund:

PR #2939, opened 2021-07-02
Fix bargein

ken-mycroft:
## Description
Barge-In is not an issue for Precise, however, core has some issues. When a wave file is speaking the recognizer confuses the wav file for the mic. Ducking helps a bit with this but its more useful as an audible confirmation as the recognizer still favors this over the mic input. The proper solution is to pause the audio output until the recognizer has consumed the utterance. 

The root problem arises from the tts module which does not use the audio service but it wants to act like one. It has start and stop method and a queue but no pause or resume. This PR corrects that. It also contains the code which uses this new functionality to help with the barge-in process. 

## How to test
Enable your barge-in config flag and use a long playing dialog skill like the new duck duck go or wiki skill (these return an abstract rather than a single sentence) and then say 'stop'. Without this PR this will not work 90% of the time. 

Yes

### DIFF
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -124,7 +124,7 @@ def mute_and_speak(utterance, ident, listen=False):
         tts.init(bus)
         tts_hash = hash(str(config.get('tts', '')))
 
-    LOG.info("Speak: " + utterance)
+    LOG.debug("Listen=%s, Speak:%s" % (listen, utterance))
     try:
         tts.execute(utterance, ident, listen)
     except RemoteTTSException as e:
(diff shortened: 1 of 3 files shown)


forslund:
Did you intend to include the entire mark-2 branch into this PR? or was it supposed to be against the mark-2 branch? All the commits makes it a bit hard to overview :)

I think this is a very nice addition of functionality to the TTS playback handling. I will look over the implementation in detail during the weekend.

ken-mycroft:
I don't know why it did this. I tried to update it to go against feature/mark-2 not dev.

krisgesling:
Thanks Ake!

I've been asked to deploy this to the Mark II for real-world testing. So have created another PR against the dev branch (#2940). Would be great if we can add review comments there.

PR #2945, opened 2021-07-06
Added comments to document race condition

chrisveilleux:
## Description
[PR2937](https://github.com/MycroftAI/mycroft-core/pull/2937) addressed an issue with a race condition but no comments were added to describe the problem and the work around to address it.  This PR adds those comments.

## How to test
No testing necessary.  Only comments touched.

CLA [Y]

### DIFF
--- a/mycroft/skills/mycroft_skill/mycroft_skill.py
+++ b/mycroft/skills/mycroft_skill/mycroft_skill.py
@@ -411,7 +411,16 @@ class MycroftSkill:
         return False
 
     def __get_response(self):
-        """Helper to get a reponse from the user
+        """Helper to get a response from the user
+
+        NOTE:  There is a race condition here.  There is a small amount of
+        time between the end of the device speaking and the converse method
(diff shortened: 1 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2945))

forslund (approved):
Nice, maybe we should open an issue as well?

chrisveilleux:
> Nice, maybe we should open an issue as well?

There is an issue in the internal ticketing system for this.  I am not sure if we have a clear delineation for what goes into GitHub issues vs. Mycroft's internal issue tracker.

krisgesling:
Until we can setup a syncing system between them, I think we should ideally have a public issue for anything that is of interest to the broader community. This would be particularly useful if downstream projects might have a vested interested in fixing it and contributing that fix back.

I've tried to do this a bit with internal tickets and include a link to the public issue, but it is a pain trying to maintain two listings.

PR #2946, opened 2021-07-06
Improve the speed of waiting for dialogs helper function

chrisveilleux:
## Description
The wait_for_dialogs step function in the Voight-Kampff test suite would loop until timeout regardless of whether or not a match was found.  Also, there was no error condition if no match was found, so the step would always pass.  This change adds logic to break out of the loop after a matching dialog is spoken and provides the status of the matching process.  A new method was added to format a message detailing the error condition.

## How to test
Tests that use this helper function should run faster.  

## Contributor license agreement signed?
CLA [Y]

PR #2951, opened 2021-07-16
Use event handler to capture messages in then_wait()

forslund:
## Description
A restructure on how waiting for events work in VK based on comments from @chrisveilleux in #2946.

This removes the handling of every single event by polling the new_message event, instead it sets up a temporary event handler handling the capture of events.

A check of the already captured messages are also made to avoid certain race conditions when messages are sent after the "when" but before the "then" is set up.

## How to test
Make sure the VK tests still passes.

CLA [ Yes ]

### DIFF
--- a/test/integrationtests/voight_kampff/tools.py
+++ b/test/integrationtests/voight_kampff/tools.py
@@ -15,6 +15,7 @@
 
 """Common tools to use when creating step files for behave tests."""
 
+from threading import Event
 import time
 
 from mycroft.messagebus import Message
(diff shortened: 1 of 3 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2951))

forslund:
Current failure is due to some VK report rights issue, the actual tests passes

test/integrationtests/voight_kampff/tools.py
```diff
@@ -23,8 +24,129 @@
...
+    def reset(self):
+        """Reset the wait state."""
+        self.result.clear()
+
+    # TODO: Remove in 21.08
+    def wait_unspecific(self, timeout):
+        """
+        Wait for a specified time for criteria to be fulfilled by any message.
+
+        This use case is deprecated and only for backward compatibility
```
krisgesling, line 51: Do you know where this is or was being used? Or just an edge case that might exist?

krisgesling (approved):
Much nicer! 

Yay for event driven processes, and the flow of `wait_specific` looks spot on to make sure things don't slip through the cracks.

Minor question that doesn't affect this being merged, just curious

PR #2952, opened 2021-07-16
Adds a GUI media service to handle audio and video playback for Common Play Service

AIIX:
## Description
Adds a GUI media service under Audio Services that establishes full media serving capabilities via Mycroft-GUI's media service backend for Common Play Service

- Supports Audio and Video playback based on Mime Type and Fallback String Checking
- Supports HTTPS, HTTP, RTSP and Local FILE streams
- Supports Ducking

Requires: https://github.com/MycroftAI/skill-playback-control/pull/46
Requires: https://github.com/MycroftAI/mycroft-gui/pull/97

## How to test
Change the local backend to use "guiplayer" type instead of "simple" in mycroft.conf.

```
    "Audio": {
    "backends": {
      "local": {
        "type": "guiplayer",
        "active": true
      },
      "vlc": {
        "type": "vlc",
        "active": false,
        "duck": true
      }
    },
    "default-backend": "local"
```
- play NPR News
- play any video type from CPS based skill
 
### DIFF
--- a/mycroft/audio/services/guiplayer/__init__.py
+++ b/mycroft/audio/services/guiplayer/__init__.py
@@ -1,0 +1,211 @@
+# Copyright 2017 Mycroft AI Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import time
+import mimetypes
+import re
+
+from threading import Lock
+from mycroft.audio.services import AudioBackend
+from mycroft.messagebus.message import Message
+from mycroft.util.log import LOG
+
+
+class GuiPlayerService(AudioBackend):
+    def __init__(self, config, bus=None, name='guiplayer'):
+        super(GuiPlayerService, self).__init__(config, bus)
+        self.config = config
+        self.process = None
+        self.bus = bus
+        self.name = name
+        self._is_playing = False
+        self._paused = False
+        self.tracks = []
+        self.index = 0
+        self.track_lock = Lock()
+        self.track_meta_from_player = None
+        self.track_meta_from_cps = None
+
+        self.bus.on('GuiPlayerServicePlay', self._play)
+        self.bus.on(
+            'gui.player.media.service.sync.status',
+            self.track_media_status)
+        self.bus.on(
+            'gui.player.media.service.get.meta',
+            self.sync_meta_from_player)
+        self.bus.on('play:status', self.sync_meta_to_player)
+
+    def supported_uris(self):
+        return ['file', 'http', 'https']
+
+    def clear_list(self):
+        with self.track_lock:
+            self.tracks = []
+
+    def add_list(self, tracks):
+        with self.track_lock:
+            self.tracks += tracks
+            LOG.info("Track list is " + str(tracks))
+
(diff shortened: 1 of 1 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2952))

mycroft/audio/services/guiplayer/__init__.py
```diff
@@ -0,0 +1,211 @@
...
+                self.bus.emit(Message("playback.display.audio.type"))
+        except BaseException:
+            LOG.debug("Cannot Determine Mime Type Falling Back To List Check")
+            mediatype = self.fallback_type_check(track)
+            if mediatype == "video":
+                LOG.debug("Falling Back To Audio Type")
+                self.bus.emit(Message("playback.display.video.type"))
+            else:
+                LOG.debug("Falling Back To Audio Type")
+                self.bus.emit(Message("playback.display.audio.type"))
```
krisgesling, line 100: I'm not sure what's happening, but using a slightly modified version of Ake's YT Skill, I'm seeing this log message that its falling back to Audio but it then uses the Video player.

I added logs to the `handle_display_video` and `handle_display_audio` methods in the control Skill and it somehow is triggering the Video handler.

AIIX: What type of stream is Ake's YT Skill providing ? Video or Audio ?

AIIX: The debug message was wrong and has been fixed in newer commit

AIIX: If Ake's YT stream URL consist of "videoplayback" in the URL it will fallback to Video type as that is the only sane and logical way to check if the mime type for those type of streaming links is actually an Video.

forslund:
As I recall it would prefer audio streams and fall back to video streams if no pure audio stream existed

AIIX:
Extracted streaming links don't exactly end with known type extensions which does not help with mime detection, I rather have CPS tell the audio service what type of stream it is (Audio/Video) than depending solely on mime detection in the Audio Service but I think that is an issue is for another PR not supposed to be blocking this PR

forslund:
The audioservice interface supports sending the mimetype already so the skill can inform the audioservice of the type already so should definitely not be a blocker

krisgesling:
Sorry if this is a duplicate comment, I thought I'd already posted this, but can't find it... 

It's working well so far, looks like a really solid start. I'm thinking we should add it as a plugin rather than including it in mycroft-core itself. We want to do this with the other Audioservices when we get around to it too. So I've setup a new repo at:
https://github.com/MycroftAI/plugin-playback-gui-player

The docs for audioservice plugins are here:
https://mycroft-ai.gitbook.io/docs/mycroft-technologies/mycroft-core/plugins/audioservice
and definitely shout out if you run into any trouble.

As it is in this PR, it will need to be an "Audioservice" plugin for the moment, but as you can probably tell from the repo name, I think we should revisit that in the near future given we're now handling audio and video.

JarbasAl:
> I'm thinking we should add it as a plugin rather than including it in mycroft-core itself. We want to do this with the other Audioservices when we get around to it too. So I've setup a new repo at:
> [MycroftAI/plugin-playback-gui-player](https://github.com/MycroftAI/plugin-playback-gui-player)
> 
> The docs for audioservice plugins are here:
> [mycroft-ai.gitbook.io/docs/mycroft-technologies/mycroft-core/plugins/audioservice](https://mycroft-ai.gitbook.io/docs/mycroft-technologies/mycroft-core/plugins/audioservice)
> and definitely shout out if you run into any trouble.


audio plugins seem to be broken, see fix here https://github.com/HelloChatterbox/HolmesV/commit/3698485e49b5e81ef8af28235148e22c27c3150d and example plugin to test with here https://github.com/OpenVoiceOS/ovos-vlc-plugin

krisgesling:
Closing in favour of the new plugin that can be found at: https://github.com/MycroftAI/plugin-playback-gui-player

PR #2962, opened 2021-08-02
Make network tests configurable

krisgesling:
## Description
Previously test URIs were hardcoded. They can now be configured in mycroft.conf.

This may be required for offline or air gapped installations - or the user may just wish to decide where their test pings are going.

This also 
- adds some `ERROR` level logs if connections fail.
- adds unit tests.

## How to test
Unit tests included.
Can also set your network_tests object to alternative values and see it fail / succeed accordingly.

- [x] CLA

### DIFF
--- a/mycroft/configuration/mycroft.conf
+++ b/mycroft/configuration/mycroft.conf
@@ -146,6 +146,15 @@
         "ssl": false
   },
 
+  // URIs to use for testing network connection.
+  "network_tests": {
+    "dns_primary": "8.8.8.8",
+    "dns_secondary": "8.8.4.4",
+    "web_url": "https://www.google.com",
+    "ncsi_endpoint": "http://www.msftncsi.com/ncsi.txt",
+    "ncsi_expected_text": "Microsoft NCSI"
+  },
+
   // Settings used by the wake-up-word listener
   // Override: REMOTE
   "listener": {
--- a/mycroft/util/network_utils.py
+++ b/mycroft/util/network_utils.py
@@ -6,6 +6,14 @@ from urllib.error import URLError
 from .log import LOG
 
 
+def _get_network_tests_config():
+    """Get network_tests object from mycroft.configuration."""
+    # Wrapped to avoid circular import errors.
+    from mycroft.configuration import Configuration
+    config = Configuration.get()
+    return config.get('network_tests', {})
+
+
 def connected():
     """Check connection by connecting to 8.8.8.8 and if google.com is
     reachable if this fails, Check Microsoft NCSI is used as a backup.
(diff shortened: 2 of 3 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2962))

mycroft/util/network_utils.py
```diff
@@ -6,6 +6,12 @@
 from .log import LOG
 
 
+def get_network_tests_config():
```
forslund, line 9: Maybe a docstring?

mycroft/util/network_utils.py
```diff
@@ -46,18 +55,27 @@ def _connected_dns(host="8.8.8.8", port=53, timeout=3):
...
     except IOError:
+        LOG.error("Unable to connect to primary DNS server, "
+                  "trying secondary...")
         try:
             s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
             s.settimeout(timeout)
-            s.connect(("8.8.4.4", port))
+            dns_secondary = config.get('dns_secondary')
+            LOG.info(config)
+            LOG.info(dns_secondary)
```
forslund, line 74: These looks like debug outputs? I'd recommend removing them or providing some describing text around the values.

forslund (changes requested):
Super nice!

There were some logs that I think should be removed or get some more context.

mycroft/util/network_utils.py
krisgesling: oops, they are stray debuging will remove

mycroft/util/network_utils.py
krisgesling: Indeed - I did the docstring as an explanation of the function, and a separate inline comment re wrapping to avoid circular import. 

If anyone has strong opinions that these should both be in the docstring I'm happy to amend.

mycroft/configuration/mycroft.conf
```diff
@@ -146,6 +146,15 @@
         "ssl": false
   },
 
+  // URIs to use for testing network connection.
+  "network_tests": {
+    "dns_primary": "8.8.8.8",
+    "dns_secondary": "8.8.4.4",
+    "https_url": "https://www.google.com",
```
krisgesling, line 153: I was just thinking that this isn't explicitly required to use the https protocol so maybe this should be renamed...

forslund: Yeah, but tricky to come up with a good descriptive name. Maybe `web_url` and a comment?

forslund (dismissed):
Looks good, I can accept any replacemenet for `https_url` you come up with :)

krisgesling:
> What about skipping/passing the whole internet connection check for people running locally? This PR looks like the right place to surf along with it?

That seems like a bigger change. Are you meaning, running without a backend or completely offline? These two scenarios probably need different behaviours and UX decisions.

ken-mycroft (approved):

j1nx:
> > What about skipping/passing the whole internet connection check for people running locally? This PR looks like the right place to surf along with it?
> 
> That seems like a bigger change. Are you meaning, running without a backend or completely offline? These two scenarios probably need different behaviours and UX decisions.

Whatever the user wants or is doing, but don't overthink it. Just make a boolean; skip, yes or no. Defaults to no. If a user wants to skip it, it just passes the connectivity check.

This is for power / development users that have their own api keys, backend, TTS, etc.

JarbasAl:
> > > What about skipping/passing the whole internet connection check for people running locally? This PR looks like the right place to surf along with it?
> > 
> > 
> > That seems like a bigger change. Are you meaning, running without a backend or completely offline? These two scenarios probably need different behaviours and UX decisions.
> 
> Whatever the user wants or is doing, but don't overthink it. Just make a boolean; skip, yes or no. Defaults to no. If a user wants to skip it, it just passes the connectivity check.
> 
> This is for power / development users that have their own api keys, backend, TTS, etc.

i think @j1nx means this https://github.com/HelloChatterbox/HolmesV/commit/7e40469a897ad8a3509a5e9677de492de6f457a7

krisgesling:
Yeah cool, have started an issue thread  (#2964) 
TLDR - I agree with the linked commit and don't think we should need an internet connection to load the Skills service at all.

krisgesling:
Test did pass - failed on cleanup due to a file permission error. Merging.

PR #2974, opened 2021-08-11
Use response method for track_info handler

krisgesling:
## Description
The `audioservice.track_info` handler was using a custom reply message, which is an older form of replying to a message.

This required you to know what the `reply_type` is eg:
```
response = self.bus.wait_for_response(
            Message('mycroft.audio.service.track_info'),
            reply_type='mycroft.audio.service.track_info_reply')
```

The new format is automatic:
```
response = self.bus.wait_for_response(
            Message('mycroft.audio.service.track_info'))
```

This is a breaking change as any calls expecting the custom reply_type will break. Creating the PR now so that we can switch it over in the next major release.

## How to test
Using the new standard format of `wait_for_response` as above will currently result in no response being received. With this change, the reply conforms to the new standard and will be returned correctly.

### DIFF
--- a/mycroft/audio/audioservice.py
+++ b/mycroft/audio/audioservice.py
@@ -484,8 +484,7 @@ class AudioService:
             track_info = self.current.track_info()
         else:
             track_info = {}
-        self.bus.emit(Message('mycroft.audio.service.track_info_reply',
-                              data=track_info))
+        self.bus.emit(message.response(track_info))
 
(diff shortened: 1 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2974))

forslund:

PR #2975, opened 2021-08-12
VK: Only stop Mycroft services if running in CI

krisgesling:
## Description
When running tests locally, a reasonable amount of time is spent starting and stopping Mycroft services. Rapidly doing this can also cause broader issues. Shutting down services at the end of a Voight Kampff test run in the CI makes sense, but it doesn't make sense for me personally when running locally. 

This is the simplest implementation which provides one behaviour for CI, and the other behaviour for "not-CI". Is this sufficient or do we need to add in one or more arguments? eg: 
```shell
# assume shutdown by default on all systems
mycroft-start vktest --no-stop-on-finish -t my-skill
```
or
```shell
# shutdown by default on CI, do not shutdown by default on non-CI but provide the option
mycroft-start vktest --stop-on-finish -t my-skill
```

## How to test
Run some VK tests locally and see that Mycroft is still running after the tests complete. 
You can now run the tests repeatedly, much faster than you could previously.

- [x] CLA

### DIFF
--- a/test/integrationtests/voight_kampff/run_test_suite.sh
+++ b/test/integrationtests/voight_kampff/run_test_suite.sh
@@ -18,8 +18,10 @@ ${SCRIPT_DIR}/../../../start-mycroft.sh all
 echo "Running behave with the arguments \"$@\""
 behave $@
 RESULT=$?
-# Stop all mycroft core services.
-${SCRIPT_DIR}/../../../stop-mycroft.sh all
+if [[ -v CI ]]; then
+    # Stop all mycroft core services if running in CI environment.
+    ${SCRIPT_DIR}/../../../stop-mycroft.sh all
+fi
 
 # Reort the result of the behave test as exit status
 exit $RESULT


devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-2975)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-2975/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-2975/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-2975/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-2975/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-2975/logs/enclosure.log)

forslund:
I wonder if the start of core also should be CI only, to keep it symetric? Though it doesn't really make much practical difference since the start will work just fine when already started.

forslund (approved):

krisgesling:
> I wonder if the start of core also should be CI only, to keep it symmetric? Though it doesn't really make much practical difference since the start will work just fine when already started.

Yeah I debated that in my head. My reasoning was that: 
1. you can't _not_ start Mycroft services if you want the tests to run, and
2. if you try start them when they're already started then the command is effectively ignored (if I'm not mistaken).

forslund:
I totally buy that reasoning

PR #2986, opened 2021-08-27
Improve confidence calculation for Common Query

ken-mycroft:
## Description
The calc_confidence() routine is a bit dated. This adds some improvements which also cover some skill blemishes in common query skills. 

## How to test
There should be a difference in the response between "how tall is the eifel tower" and "what is the eifel tower", etc.

Yes


mycroft/skills/common_query_skill.py
```diff
@@ -80,39 +105,76 @@ def __handle_question_query(self, message):
...
+                )
+            )
+
+    def remove_noise(self, phrase):
+        # remove noise to produce essence
+        phrase = " " + phrase + " "
+        for word in self.translated_noise_words:
+            mtch = " " + word + " "
+            if phrase.find(mtch) > -1:
+                phrase = phrase.replace(word, "")
```
JarbasAl, line 138: consider porting the remove_voc method from OVOS

https://github.com/OpenVoiceOS/OVOS-workshop/blob/master/ovos_workshop/patches/base_skill.py#L165

forslund, line 138: Will there be an issue here if there is for example an "and" and the word "andy"? Example "he is called andy and he is cool"

I have previously made good use of `re.sub()` and the "word boundry" `\b`in cases like this.

This function could be:
```python
        for word in self.translated_noise_words:
            phrase = re.sub(r'\b' + word + r'\b', '', phrase)
        return " ".join(phrase.split())
```

Or possibly lever the regex logic even more
```python
        # match any whole word that exists in the list of noise words
        regex = r'\b(' + '|'.join(self.translated_noise_words) + r')\b'
        result = re.sub(regex, '', phrase)
        return ' '.join(result.split())
```

mycroft/skills/common_query_skill.py
```diff
@@ -80,39 +105,76 @@ def __handle_question_query(self, message):
...
-        elif int(level) == int(CQSMatchLevel.GENERAL):
-            return 0.5 + (consumed_pct / 10) + bonus
-        else:
-            return 0.0  # should never happen
+        # extract topic
+        topic = self.remove_noise(match)
+
+        # calculate relevance
+        matches = 0
+        for word in topic:
```
forslund, line 161: Should this be `for word in topic.split():`? Otherwise it would go per character?

forslund (changes requested):
Looks like a good improvement over the original implementation. I think there may exist an issue in how the removal of noise words are done, where it can remove partial words (if I'm reading this correctly).

mycroft/skills/common_query_skill.py
JarbasAl: the method i suggest porting over from OVOS about should account for this and for finding the voc files

```python
    def remove_voc(self, utt, voc_filename, lang=None):
        """ removes any entry in .voc file from the utterance """
        lang = lang or self.lang
        cache_key = lang + voc_filename

        if cache_key not in self.voc_match_cache:
			# this call avoids duplicate logic to load .voc
            self.voc_match(utt, voc_filename, lang)

        if utt:
            # Check for matches against complete words
            for i in self.voc_match_cache.get(cache_key) or []:
                # Substitute only whole words matching the token
                utt = re.sub(r'\b' + i + r"\b", "", utt)

        return utt
 ```

Comment (forslund) [reply mycroft/skills/common_query_skill.py:138]: Yeah it uses that very regex :)

Comment (ken-mycroft) [reply mycroft/skills/common_query_skill.py:138]: I will take a look at it your code J, thanks. FYI in a subsequent version of the code which I have not pushed I have a simple relevance = len(set(answer.split()) - set(noise_words)) / len(answer.split() but I felt this loop was a bit easier for folks to understand.

File: mycroft/skills/common_query_skill.py [hunk 3]
```diff
@@ -49,9 +51,30 @@ class CommonQuerySkill(MycroftSkill, ABC):
...
     """
+
     def __init__(self, name=None, bus=None):
         super().__init__(name, bus)
 
+        data_dir = Configuration.get().get("data_dir")
+        save_root_dir = self.root_dir
+        self.root_dir = data_dir + "/mycroft/"
+        noise_words_filename = self.find_resource("noise_words" + ".list", "res/text")
+        self.root_dir = save_root_dir
```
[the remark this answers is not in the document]
ken-mycroft: Hi Ake, and thanks for the review. The removal of noise if you look in the actual loop uses a w = '  ' + noise_word + '  ' approach which is why variable phrase has a space added to the start and end, but there could always be a subtle bug. The problem with all the routines you reference for finding a resource is that they all assume the root dir is the current skill base but in this case I did not want to clone the noise words file for every skill so I added it once to the base class and use the root to override the base (as you can see i save and restore it). I do use the configured value for 'data_dir' but if you know of a routine for finding resource files in res/text/locale I am all ears. Based on our existing code i assumed mycroft/res/text/locale would always exist so the only issue is if 'data'_dir' is a valid location for resource files will always be located. I am flexible on this.

mycroft/skills/common_query_skill.py
```diff
@@ -49,9 +51,30 @@ class CommonQuerySkill(MycroftSkill, ABC):
...
+        self.translated_noise_words = []
+        try:
+            with open(noise_words_filename) as f:
+                self.translated_noise_words = f.read().strip()
+            self.translated_noise_words = self.translated_noise_words.split()
+        except Exception as e:
```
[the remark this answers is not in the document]
ken-mycroft: Agreed, will fix.

mycroft/skills/common_query_skill.py
```diff
@@ -80,39 +105,76 @@ def __handle_question_query(self, message):
...
+                    {
+                        "phrase": search_phrase,
+                        "skill_id": self.skill_id,
+                        "searching": False,
+                    }
+                )
+            )
+
+    def remove_noise(self, phrase):
+        # remove noise to produce essence
```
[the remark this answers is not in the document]
ken-mycroft: is this OK?

"""remove noise to produce essence"""

mycroft/skills/common_query_skill.py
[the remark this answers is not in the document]
forslund: The actual loop uses the word with the space when finding but without the space when replacing causing it to be able to remove too much. So if a " and " is found it will replace "and" with ""

I see I misunderstood the code a bit but it's still not quite right. The data dir /opt/mycroft/ in the normal case but you want the mycroft-core root folder. The function I refer to is the one in the mycroft.util module, and that refers to the MYCROFT_ROOT/res-folder and not the skill's folder.

mycroft/skills/common_query_skill.py
ken-mycroft: Your code says # PR not yet released, and I have nod idea what voc file you are using, however, I thought I would simply post this to show that we are talking about 6 one way, half dozen another. 


class test():
    def __init__(self):
        self.translated_noise_words = ['where',
'which',
'they',
'them',
'when',
'what',
'that',
'will',
'from',
'that',
'also',
'who',
'how',
'did',
'and',
'but',
'the',
'too',
'why',
'is',
'it',
'do',
'or',
'to',
'of',
'a']

    def remove_noise_alt1(self, utt):
        for i in self.translated_noise_words:
            utt = re.sub(r'\b' + i + r"\b", "", utt)

        return utt

    def remove_noise_dont_preserve_order(self, phrase):
        # remove noise to produce essence
        phrase = " " + phrase + " "
        result_set = set( set(phrase.split(' ')) - set(self.translated_noise_words) )
        return (' '.join(result_set)).strip()

    def remove_noise_preserve_order(self, phrase):
        # remove noise to produce essence
        phrase = " " + phrase + " "
        for word in self.translated_noise_words:
            mtch = " " + word + " "
            if phrase.find(mtch) > -1:
                phrase = phrase.replace(word, "")
        phrase = " ".join(phrase.split())
        return phrase.strip()

    def __calc_confidence(self, match, phrase, level, answer):
        # Assume the more of the words that get consumed, the better the match
        consumed_pct = len(match.split()) / len(phrase.split())
        if consumed_pct > 1.0:
            consumed_pct = 1.0
        consumed_pct /= 10

        bonus = 0

        # extract topic
        topic = self.remove_noise(match)

        # calculate relevance
        matches = 0
        for word in topic:
            if answer.find(word) > -1:
                matches += 1

        answer_size = len(answer.split(" "))
        relevance = 0.0
        if answer_size > 0:
            relevance = float(float(matches) / float(answer_size))

        # extra credit for more words
        wc_mod = float(float(answer_size) / float(100))

        confidence = 0.6

        return confidence

phrases = [
        "is the empre state building the tallest building in the world",
        "which species of hummingbird is the smallest",
        "why did they burn joan of ark at the stake",
        "how tall is the eifel tower",
        "how deep is the marriana trench",
        "how old is paul mccartney"
        ]
t = test()

print("Order Preserved")
for phrase in phrases:
    topic = t.remove_noise_preserve_order(phrase)
    print("%s --- %s" % (topic, phrase))

print('')
print("Order Not Preserved")
for phrase in phrases:
    topic = t.remove_noise_dont_preserve_order(phrase)
    print("%s --- %s" % (topic, phrase))

print('')
print("Alt 1")
for phrase in phrases:
    topic = t.remove_noise_alt1(phrase)
    print("%s --- %s" % (topic, phrase))

Which produces 

species hummingbird smallest --- which species of hummingbird is the smallest
burn joan ark at stake --- why did they burn joan of ark at the stake
tall eifel tower --- how tall is the eifel tower
deep marriana trench --- how deep is the marriana trench
old paul mccartney --- how old is paul mccartney

Order Not Preserved
building state empre in world tallest --- is the empre state building the tallest building in the world
smallest species hummingbird --- which species of hummingbird is the smallest
stake ark at burn joan --- why did they burn joan of ark at the stake
tall eifel tower --- how tall is the eifel tower
trench deep marriana --- how deep is the marriana trench
paul mccartney old --- how old is paul mccartney

Alt 1
  empre state building  tallest building in  world --- is the empre state building the tallest building in the world
 species  hummingbird   smallest --- which species of hummingbird is the smallest
   burn joan  ark at  stake --- why did they burn joan of ark at the stake
 tall   eifel tower --- how tall is the eifel tower
 deep   marriana trench --- how deep is the marriana trench
 old  paul mccartney --- how old is paul mccartney

ken-mycroft: So guys, do you really need me to change this to improve the PR?

JarbasAl: the PR not made comment in that link means i intended to submit that util method to the base MycroftSkill class, i havent made that PR yet.

JarbasAl: you would need to add the `noise.voc` file to `mycroft/res/text/en-us` and then 

```python
clean_utt = self.remove_voc(utterance, "noise.voc")
```

ken-mycroft: Oh, and BTW, from a timing perspective ...

For 10,000 iterations on my laptop

Order preserved took 0.80367398262
Not Order preserved took 0.282736063004
Alt1 took 3.85695791245

Unsurprisingly sets are faster but loops are probably easier to read.

JarbasAl: would be curious what the timing would be with my approach, be sure to also post it if you measure it

forslund: My issue with the implementation in the PR is that it malforms the output

this is an minimal example constructed in the repl using the PR branch:

```python
>> class T(CommonQuerySkill):
...   def __init__(self):
...     pass
...   def CQS_match_query_phrase(self, phrase):
...     pass
...
>>> t = T()
>>> t.translated_noise_words = ['is', 'and']
>>> t.remove_noise("this is andy and he is cool")
'th y he cool'
```

since words in the output is used later I assume it needs to not do partial removes.

But if that is not an issue this version is fine.

mycroft/skills/common_query_skill.py
ken-mycroft: Good catch; ty

mycroft/skills/common_query_skill.py
ken-mycroft: See above output from my previous comment. It does not behave as your example does. The only issue is whether order is preserved and that is not important in this case.

Performance numbers were posted above but just to verify. 

For 10,000 iterations on my laptop

Order preserved took 0.80367398262
Not Order preserved took 0.282736063004
Alt1 took 3.85695791245

So loop easier to understand but twice as slow as using sets. Regex is about 15 times slower than sets, maybe 5 times slower than the loop.

mycroft/skills/common_query_skill.py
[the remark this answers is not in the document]
ken-mycroft: Thanks Ake, I'll update accordingly.

mycroft/skills/common_query_skill.py
forslund: Have you pushed all your code? I get it quite consistently and the logic looks like it _should_ remove partial words.

The looping regex is definitely slow. The second option I proposed with the single regex is slightly faster than the original code but nowhere near the sets.

```
original 0.7861279549542814
for_version 6.605184729909524
regex_version 0.683092060033232
precalculated_regex_version 0.5726282709510997
```

The test file for these is here: https://gist.github.com/forslund/c01636a18ddef0a9bef84dae609f4511

mycroft/skills/common_query_skill.py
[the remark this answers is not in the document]
ken-mycroft: So just so I am not missing something. This method still wants the language embedded in the file path. It does not do it for you - correct?

>>> from mycroft.util.file_utils import resolve_resource_file
>>> print(resolve_resource_file('text/en-us/noise_words.list'))
/opt/mycroft/mycroft/res/text/en-us/noise_words.list
>>> print(resolve_resource_file('noise_words.list'))
None
>>> print(resolve_resource_file('text/noise_words.list'))
None

mycroft/skills/common_query_skill.py
ken-mycroft: No, I have not re-submitted yet, but I did post the source code for my tests above in my reply to Jarbas. See class test():

forslund: Hmm, I still have issues with that version. It works with a slight tweak:

```python
def remove_noise_preserve_order(self, phrase):
        # remove noise to produce essence
        phrase = " " + phrase + " "
        for word in self.translated_noise_words:
            mtch = " " + word + " "
            if phrase.find(mtch) > -1:
                phrase = phrase.replace(mtch, " ")  # Was phrase.replace(word, "")
        phrase = " ".join(phrase.split())
        return phrase.strip()
```

ken-mycroft: You are correct. Thank you for this. I believe my latest commit corrects this.

forslund:
This latest change the root issue I had with the code. I'm quite happy with it now :)

mycroft/skills/common_query_skill.py
ken-mycroft: Thank you for the excellent input!

mycroft/skills/common_query_skill.py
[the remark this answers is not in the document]
ken-mycroft: I had to manually add the language as shown above.

mycroft/skills/common_query_skill.py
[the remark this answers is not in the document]
ken-mycroft: Done

mycroft/skills/common_query_skill.py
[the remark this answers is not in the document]
ken-mycroft: Done

mycroft/skills/common_query_skill.py
ken-mycroft: Addressed in last commit

mycroft/skills/common_query_skill.py
ken-mycroft: Corrected in latest commit.

forslund (approved):
Looks very good now, approved!

There are a couple of minor pep-8 issues (line-lenghts some whitespace nconsistencies) and the updated confidence calculation needs to be taken into account in the unittest:

The following lines:
[test_common_query.py#L99](https://github.com/MycroftAI/mycroft-core/blob/dev/test/unittests/skills/test_common_query_skill.py#L99) 1.0 -> 1.01
[test_common_query.py#L123](https://github.com/MycroftAI/mycroft-core/blob/dev/test/unittests/skills/test_common_query_skill.py#L123) 1.1 -> 1.11

forslund (approved):

krisgesling (approved):

PR #2987, opened 2021-08-31
Add method to remove any entries from a .voc file from a string.

krisgesling:
## Description
The `remove_voc` method complements the existing vocab matching tools.

It is useful to clean utterances of common articles or other noise when parsing for specific information. 

This method is not case sensitive unlike the other vocab matching methods. Is the case sensitivity intentional? Should we make all methods case insensitive?

Context: I was needing this type of method today, then saw the same need had been identified in https://github.com/MycroftAI/mycroft-core/pull/2986 and an existing method to do it [already existed in OVOS](https://github.com/OpenVoiceOS/OVOS-workshop/blob/8d2b274084520154ab9c36dc6c78ed6f7cf13b44/ovos_workshop/patches/base_skill.py#L165)

## How to test
Unit tests included

- [x] CLA

### DIFF
--- a/mycroft/skills/mycroft_skill/mycroft_skill.py
+++ b/mycroft/skills/mycroft_skill/mycroft_skill.py
@@ -669,6 +669,33 @@ class MycroftSkill:
         else:
             return False
 
+    def remove_voc(self, utterance, voc_filename, lang=None):
+        """Removes any entries in a .voc file from the provided string.
+
+        The vocab matching is not case sensitive.
+
+        Args:
+            utterance (str): String to remove entries from
+            voc_filename (str): Name of vocabulary file (e.g. 'yes' for
+                                'res/text/en-us/yes.voc')
+            lang (str): Language code, defaults to self.lang
+        Returns:
+            str: utterance with .voc entries removed
+        """
+        lang = lang or self.lang
+        cache_key = lang + voc_filename
+
+        if cache_key not in self.voc_match_cache:
+            self.voc_match(utterance, voc_filename, lang)
+
+        if utterance:
+            # Check for matches against complete words
+            for entry in self.voc_match_cache.get(cache_key) or []:
+                # Substitute only whole words matching the token
+                utterance = re.sub(r'\b' + entry + r"\b", "", utterance,
+                                   flags=re.IGNORECASE)
+        return utterance
+
     def report_metric(self, name, data):
         """Report a skill metric to the Mycroft servers.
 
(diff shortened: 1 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-2987))

krisgesling:
I think I've missed something... where are the other 2 approaches? 
If there are better options, that's great - let's do that!

I noticed the extraneous spaces thing too but didn't see a reason to remove them. In one way it helps to show that things have been removed, rather than operating on what is now probably at least grammatically incorrect, and potentially has even changed the meaning slightly. This is kind of silly but the best I can come up with at 5am...
```python
cleaned_string = self.remove_voc("john likes the football", "football")
if "john the football" in cleaned_string:
    ...
```

On the flipside, I could see that extra characters may impact a confidence rating based on keyword frequency etc.

In terms of case sensitivity the docstring says "is not case sensitive" but maybe this should be worded as "is case insensitive" to match the flag.

forslund:
There were some discussion around the implementation when I tried to point out a bug in #2986. I couldn't find Ken's versions there now but I put up a gist with a collection of the ones I could remember: https://gist.github.com/forslund/c01636a18ddef0a9bef84dae609f4511

The results for 100000 runs of each:
str_replace 0.8078192858956754
for_regex 6.423368333023973
single_regex 0.6766295690322295
precalculated_single_regex 0.42126812401693314
list_join 0.4742607999360189

The str_replace() is the implementation chosen in above mentioned PR.

for_regex() is basically what is done here and it gets a lot of overhead from the regex compilation. The good thing about of the regex versions is that capitalization of the input string is kept without extra logic.

`list_join()` is a really nifty and fast function if I got it right...To handle capitalization it can be changed to

```python
' '.join(w for w in phrase.split() if w.lower() not in words)
```
The lowering makes it lose about 0.1 seconds in the test case but it's still easy the fastest of the versions above.

krisgesling:
Nice comparison - thanks!

I love the simplicity of the `list_join` comprehension. However I just did some quick testing and there's a few extra pitfalls.

`list_join()` won't handle multi-word vocab
and 
`str_replace()` is case-sensitive which I think we should handle.
Eg:
```python
remove_voc("Never Forget it", "forget it")  # Never
```

I've added a "help wanted" flag to see if anyone else has a good suggestion. You can find unit tests in this PR to validate it. 
Note - I've just removed the spaces that were left in from the original implementation so the unit tests will currently fail.

forslund:

PR #3002, opened 2021-09-14
Extended the idea of the VK Criteriaatcher to handle other types of ma…

chrisveilleux:
…tching

## Description

## How to test

### DIFF
--- a/test/integrationtests/voight_kampff/tools.py
+++ b/test/integrationtests/voight_kampff/tools.py
@@ -15,15 +15,159 @@
 """Common tools to use when creating step files for behave tests."""
 
 from threading import Event
+from typing import Any, Callable, List, Tuple
 import time
 
 from mycroft.audio.utils import wait_while_speaking
 from mycroft.messagebus import Message
 
-
 DEFAULT_TIMEOUT = 10
 
 
+class VoightKampffEventMatcher:
+    """Matches a specified message type to messages emitted on the bus.
+
+    Attributes:
+        message_type: identifier of the message to search for on the bus
+        context: the Behave context from the test utilizing this class
+        match_event: mechanism for knowing when a match is found
+        error_message: message that can be used by the test to communicate
+            the reason for a failed match to the tester.
+    """
+    def __init__(self, message_type: str, context: Any):
+        self.message_type = message_type
+        self.context = context
+        self.match_event = Event()
+        self.match_found = False
+        self.error_message = ""
+
+    def match(self, timeout: int = None):
+        """Attempts to match the requested message type to emitted bus events.
+
+        Use a message bus event handler to capture any message emitted on the
+        bus that matches the message type specified by the caller.  Also
+        checks any messages emitted prior to the handler being defined to
+        protect against a race condition.
+
+        Args:
+            timeout: number of seconds to attempt matching before giving up
+        """
+        timeout = timeout or self.context.step_timeout
+        self.context.bus.on(self.message_type, self.handle_message)
+        self._check_historical_messages()
+        if not self.match_event.is_set():
+            self.match_event.wait(timeout=timeout)
+        self.context.bus.remove(self.message_type, self.handle_message)
+        self.match_found = self.match_event.is_set()
+        if not self.match_found:
+            self._build_error_message()
+
+    def _check_historical_messages(self):
+        """Searches messages emitted before the event handler was defined."""
+        for message in self.context.bus.get_messages(self.message_type):
+            self.handle_message(message)
+            if self.match_event.is_set():
+                break
+        self.context.bus.clear_messages()
+
+    def handle_message(self, message: Message):
+        """Applies matching criteria to the emitted event.
+
+        Args:
+            message: message emitted by bus with the requested message type
+        """
+        self.context.matched_message = message
+        self.match_event.set()
+
+    def _build_error_message(self):
+        """Builds a message that communicates the failure to the test."""
+        self.error_message = (
+            f"Expected message type {self.message_type} was not emitted."
+        )
+
+
+class VoightKampffDialogMatcher(VoightKampffEventMatcher):
+    """Variation of VoightKampffEventMatcher for matching dialogs.
+
+    Attributes:
+        dialogs: one or more dialog names that will constitute a match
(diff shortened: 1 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-3002)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-3002/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-3002/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-3002/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-3002/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-3002/logs/enclosure.log)

test/integrationtests/voight_kampff/tools.py
```diff
@@ -15,15 +15,161 @@
...
+        """
+        timeout = timeout or self.context.step_timeout
+        self.context.bus.on(self.message_type, self.handle_message)
+        self._check_historical_messages()
+        if not self.match_event.is_set():
+            self.match_event.wait(timeout=timeout)
+        self.context.bus.remove(self.message_type, self.handle_message)
+        self.match_found = self.match_event.is_set()
+        if not self.match_found:
+            self._build_error_message()
```
forslund, line 63: I think the match should return the success / failure. Passing this information through state seems contra-intuitive.

test/integrationtests/voight_kampff/tools.py
```diff
@@ -15,15 +15,161 @@
...
+        context: the Behave context from the test utilizing this class
+        match_event: mechanism for knowing when a match is found
+        error_message: message that can be used by the test to communicate
+            the reason for a failed match to the tester.
+    """
+    def __init__(self, message_type: str, context: Any):
+        self.message_type = message_type
+        self.context = context
+        self.match_event = Event()
+        self.match_found = False
```
forslund, line 41: `match_found` is a good candidate to be a property since it reflects the state of the `match_event` Event.

forslund (approved):
A nice evolution. The only thing I'm concerned about is the way status is passed through a member variable instead of a return value.

test/integrationtests/voight_kampff/tools.py
```diff
@@ -15,15 +15,161 @@
...
+            message: message emitted by bus with the requested message type
+        """
+        self.context.matched_message = message
+        self.match_event.set()
+
+    def _build_error_message(self):
+        """Builds a message that communicates the failure to the test."""
+        self.error_message = (
+            f"Expected message type {self.message_type} was not emitted."
+        )
```
forslund, line 86: Might be an idea to include the `reset()` method from the CriteraWaiter as well. so an object isn't left in an unusable state after a matching (unless you know how to reset it)

test/integrationtests/voight_kampff/tools.py
```diff
@@ -15,15 +15,161 @@
...
 import time
 
 from mycroft.audio.utils import wait_while_speaking
 from mycroft.messagebus import Message
 
-
 DEFAULT_TIMEOUT = 10
 
 
+class VoightKampffEventMatcher:
```
krisgesling, line 27: It feels a little confusing to have this called an `EventMatcher`. Could we use `MessageMatcher` instead?

test/integrationtests/voight_kampff/tools.py
chrisveilleux: Could you please elaborate?  I am not understanding what is iffy about using the instance attributes.

test/integrationtests/voight_kampff/tools.py
chrisveilleux: These are intended to be single use classes.  Instantiate it, do the matching and be done with it.  I think reusing the same instance of a class for multiple check could cause trouble.  I could improve the comments to indicate this, or I could make the classes private and write wrapper functions that use them as intended.

test/integrationtests/voight_kampff/tools.py
chrisveilleux: Good call. Will do.

test/integrationtests/voight_kampff/tools.py
chrisveilleux: I think the last change I just pushed to use wrapper functions alleviates this concern.

test/integrationtests/voight_kampff/tools.py
krisgesling: :heart:

forslund: Explicitly stating that it's single use only sounds like a good idea.

test/integrationtests/voight_kampff/tools.py
```diff
@@ -15,15 +15,188 @@
...
+        matcher = VoightKampffDialogMatcher(context, dialogs)
+        match_found, error_message = matcher.match
+        assert match_found, error_message
+
+    Attributes:
+        dialogs: one or more dialog names that will constitute a match
+        speak_messages: bus messages with message type of "speak" captured
+            in the matching process
+    """
+    def __init__(self, context: Any, dialogs: List[str]):
```
krisgesling, line 115: Small nitpick that the args order for the Matcher classes aren't 100% consistent.

Is this an accurate characterization?
```python
VoightKampffMessageMatcher(thing_to_look_for, context)
VoightKampffDialogMatcher(context, thing_to_look_for)
VoightKampffCriteriaMatcher(thing_to_look_for, context, how_to_look)
```

krisgesling (commented):
Did the wrapper functions get removed? I thought they were a good addition.

chrisveilleux:
> Did the wrapper functions get removed? I thought they were a good addition.
> 
> Edit: but loving this overall 👍

I did remove the wrapper functions.  Instead I added instructions on how to use the classes in the class level docstring.  These classes are pretty easy to use.  I am not sure the extra abstraction is necessary.

test/integrationtests/voight_kampff/tools.py
chrisveilleux: Fixed

PR #3003, opened 2021-09-16
Padatious doesn't need to run 3 times

Joanguitar:
## Description
New class PadatiousMatcher created to avoid running Padatious intent service 3 times when a low priority callback happens.
Instead, the new class stores the result from the first call and uses it on subsequent calls with different match priority.

## How to test
There's no new feature, the way to see its impact is by getting faster low priority fallbacks.
So just say something to Mycroft that it shouldn't understand and see that it takes shorter for it to reply.

### DIFF
--- a/mycroft/skills/intent_service.py
+++ b/mycroft/skills/intent_service.py
@@ -21,7 +21,10 @@ from mycroft.util.log import LOG
 from mycroft.util.parse import normalize
 from mycroft.metrics import report_timing, Stopwatch
 from .intent_services import (
-    AdaptService, AdaptIntent, FallbackService, PadatiousService, IntentMatch
+    AdaptService, AdaptIntent,
+    FallbackService,
+    PadatiousService, PadatiousMatcher,
+    IntentMatch
 )
 from .intent_service_interface import open_intent_envelope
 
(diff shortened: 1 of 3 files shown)


forslund:
Nice! I think the `lru_cache` decorator can be removed with this since that was intended to handle the caching of the result to avoid the re-runs. Or are both needed?

Joanguitar:
> 
> 
> Nice! I think the `lru_cache` decorator [here](https://github.com/MycroftAI/mycroft-core/pull/3003/files#diff-a92fea3071d5977679a2aead10ebe8619a1515cd91cb1a546653b6631cb0b655R240) can be removed with this since that was intended to handle the caching of the result to avoid the re-runs. Or are both needed?

True, I just removed it.

devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-3003)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-3003/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-3003/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-3003/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-3003/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-3003/logs/enclosure.log)

PR #3012, opened 2021-10-18
Bugfix/issue 3011

in03:
## Description
Added a 'do later' flag in place of command that sets `use_precise` to false if Precise is unavailable, since the command relied on a dependency installed later in the script. After dependencies are installed, if flag is set, run the original update command. Also fixed `use_precise` wrongly being set to 'true' instead of 'false' after failing the AVX check.

## How to test
Run Mycroft dev_setup.sh on an Intel/AMD machine with no AVX support.
Installer should complete without `jq: command not found` error.
After install check `{ "use_precise": false }` in _/etc/mycroft/mycroft.conf_, and not 'true'.


krisgesling:
It's all north according to the southerners in Sydney/Melbourne :laughing: 

You've got far better weather than we do right now though - just had the hottest ever October day in Darwin :ice_cream:

PR #3015, opened 2021-10-24
[mycroft-config] Fix issue #2991

goldyfruit:
## Description
`mycroft-config set` command still uses old non-existant `mycroft.conf`

## How to test
Execute `mycroft-config set` 

## Contributor license agreement signed?
CLA [ Yes ]

PR #3019, opened 2021-10-29
Refactor/shellcheck bin

forslund:
## Description
Adds shellcheck stage to github actions as discussed in #3015 checking the shellscripts in `bin`. This should be extended to also check the _dev-setup.sh_ and _scripts_ folder etc. but this is a start.

This also fixes the issues detected by shellcheck in the checked folder, the biggest changes were in `mycroft-config`.

The current check excludes the following checks:
- SC1091: Avoids errors when shellcheck can't find sourced file
- SC2034: Unused variables, for example colors that aren't used yet
- SC2012: use of ls, from what I can see in our case this is fine (wc -l)

## How to test
Ensure github action Shellcheck runs ok.

CLA [ x]

### DIFF
--- a/.github/workflows/unit_test.yml
+++ b/.github/workflows/unit_test.yml
@@ -66,3 +66,14 @@ jobs:
       run: if [[ ${{ matrix.python-version }} == 3.9 ]]; then bash <(curl -s https://codecov.io/bash); fi
       env:
         CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+  shellcheck:
+    name: Shellcheck
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Shell check mycroft tools
+      uses: ludeeus/action-shellcheck@ceeca77f6538b97991ca2c2a2ebe1ab64e573b5e
+      env:
+        SHELLCHECK_OPTS: -x --exclude SC2034 --exclude SC2012 --exclude SC1091
+      with:
+        scandir: ./bin
--- a/bin/mycroft-cli-client
+++ b/bin/mycroft-cli-client
@@ -21,4 +21,4 @@ DIR="$( dirname "$SOURCE" )"
 source "$DIR/../venv-activate.sh" -q
 
 # Invoke the Command Line Interface
-python -m mycroft.client.text $@
+python -m mycroft.client.text "$@"
--- a/bin/mycroft-config
+++ b/bin/mycroft-config
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 SOURCE="${BASH_SOURCE[0]}"
-cd -P "$( dirname "$SOURCE" )"
+cd -P "$( dirname "$SOURCE" )" || exit
 DIR="$( pwd )"
 script=${0}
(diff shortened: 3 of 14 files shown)


.github/workflows/unit_test.yml
```diff
@@ -66,3 +66,14 @@ jobs:
       run: if [[ ${{ matrix.python-version }} == 3.9 ]]; then bash <(curl -s https://codecov.io/bash); fi
       env:
         CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+  shellcheck:
+    name: Shellcheck
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Shell check mycroft tools
+      uses: ludeeus/action-shellcheck@master
```
krisgesling, line 75: Can we set the action-shellcheck to a specific commit rather always using master?

This is a [recommended practice by Github](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_iduses) as it removes the ability for the action author or some other malicious actor to add unexpected code to that repo.

`actions/checkout@v2` is fine as it's an official Github Action, but this one I'd say is more vulnerable.

krisgesling (changes requested):
Thanks Ake, this looks great. I'm going to spend a few hours using these scripts just to put them through some manual testing, but it looks right.

Just one small change request

.github/workflows/unit_test.yml
forslund: good suggestion, I'll fix that tonight

forslund:
Manual tests are most welcome, I tried to exercise all updated scripts but may well have missed things. Especially in the more complex scripts.

.github/workflows/unit_test.yml
forslund: Updated locking it to the current commit on master.

krisgesling (approved):
Great thanks!

I did a fairly thorough exercise of all the scripts and didn't find anything out of place so good to go.

The CI is just a bit backed up from the translation updates to the Marketplace.

PureTryOut:
> The current check excludes the following checks:
>    SC1091: Avoids errors when shellcheck can't find sourced file
>    SC2034: Unused variables, for example colors that aren't used yet
>    SC2012: use of ls, from what I can see in our case this is fine (wc -l)

Please let shellcheck ignore those errors with [a directive](https://github.com/koalaman/shellcheck/wiki/Directive) if you want to ignore those errors @forslund

forslund:
I decided against using the directive in the script files since it in my opinion it cluttered up the files unnecessarily but I can make that change if it's preferred.

PureTryOut:
I disagree with it making it cluttered, and right now Shellcheck still spits out warnings which is ugly and imo worrying. There are fine reasons to ignore some warnings but shellcheck should really be told to ignore them. Now people might get a "oh this warning is like one of those others we ignore, I can ignore this one too" feeling while it might this time actually valid complain. Let's be explicit that we allow these particular "mistakes".

forslund:
Your reasoning is valid too I'll make an update to this effect

PR #3022, opened 2021-11-04
Fix race condition in GUI namespace insertion

krisgesling:
## Description
When inserting a new namespace for the GUI, bus events could cause the dict to change while it is being iterated over.

## How to test
Unsure - saw it in OVOS and can see how that could happen.
Change shouldn't have a negative impact.

- [x] CLA

Thanks to @JarbasAI for identifying and fixing it


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-3022))

PR #3025, opened 2021-11-05
Check if GUI is connected rather than maintain list of platforms

krisgesling:
## Description
There are already many Mycroft platforms that have GUIs and this will only grow. We want to know if the device has a GUI connected rather than if it is in a pre-defined list of platforms.

## How to test
With the new latest Wikipedia installed - ask common query a question and ensure it is still displayed.

### DIFF
--- a/mycroft/skills/common_query_skill.py
+++ b/mycroft/skills/common_query_skill.py
@@ -36,8 +36,6 @@ def is_CQSVisualMatchLevel(match_level):
     return isinstance(match_level, type(CQSVisualMatchLevel.EXACT))
 
 
-VISUAL_DEVICES = ['mycroft_mark_2']
-
 """these are for the confidence calculation"""
 # how much each topic word is worth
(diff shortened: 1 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-3025))

forslund (changes requested):
Nice improvement, the test case  needs to be removed / updated otherwise it should be good to go!

krisgesling:
In order to be able to run a scenario that exhibits different behaviour when a GUI is connected, I mocked the GUI class to provide a settable connected attribute. Interested in any opinions on this approach.

This also removes the concept of a "Mark II" from the test as it should be any device with a connected display.

forslund (approved):
Super nice! Approved

JarbasAl (approved):

PR #3033, opened 2021-11-09
Fix config priority

### LINKED ISSUE #3029: remote config overriding system config
bug described here https://community.mycroft.ai/t/remote-conf-overriding-all-other-conf/11476 

this was discussed but consensus not reached https://github.com/MycroftAI/mycroft-core/pull/2861

i was under the impression this was considered an unwanted change, at least without some extra changes accompanying the order change

forslund:
## Description
This resolves #3029 as well as a bug masking the issue by not loading remote config at startup.

## How to test
- Remove all local configs setting "tts"->"module"
- Set the voice to American Male on the web
- Start Mycroft `./start-mycroft.sh all`
- Wait for everything to be up
- Restart the audio service `./start-mycroft.sh audio restart`
- Verify that the Mimic2 voice used
- Add config for the voice used in `/etc/mycroft/mycroft.conf`
```json
{
  "tts" {
    "module": "mimic"
  }
}
``` 
- Restart the audio service again `./start-mycroft.sh audio restart`
- Verify that the mimic voice is used

CLA [ Yes ]

### DIFF
--- a/mycroft/util/log.py
+++ b/mycroft/util/log.py
@@ -86,7 +86,8 @@ class LOG:
         cls.handler = logging.StreamHandler(sys.stdout)
         cls.handler.setFormatter(formatter)
 
-        config = mycroft.configuration.Configuration.get(remote=False)
+        config = mycroft.configuration.Configuration.get(cache=False,
+                                                         remote=False)
         if config.get('log_format'):
             formatter = logging.Formatter(config.get('log_format'), style='{')
             cls.handler.setFormatter(formatter)
--- a/mycroft/configuration/config.py
+++ b/mycroft/configuration/config.py
@@ -236,12 +236,12 @@ class Configuration:
                 _log_old_location_deprecation()
                 configs.append(LocalConf(OLD_USER_CONFIG))
 
+            # Then use the system config (/etc/mycroft/mycroft.conf)
+            configs.append(LocalConf(SYSTEM_CONFIG))
+
             # Then use remote config
             if remote:
                 configs.append(RemoteConf())
-
-            # Then use the system config (/etc/mycroft/mycroft.conf)
-            configs.append(LocalConf(SYSTEM_CONFIG))
 
             # Then use the config that comes with the package
             configs.append(LocalConf(DEFAULT_CONFIG))


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-3033))

mycroft/util/log.py
```diff
@@ -86,7 +86,8 @@ def init(cls):
         cls.handler = logging.StreamHandler(sys.stdout)
         cls.handler.setFormatter(formatter)
 
-        config = mycroft.configuration.Configuration.get(remote=False)
+        config = mycroft.configuration.Configuration.get(cache=False,
```
JarbasAl, line 89: Should this be done in Configuration class directly? If remote make cache always false

forslund: Not sure. For small stand-alone scripts I can see the use of caching but still excluding the remote.

JarbasAl: do any standalone scripts really use Configuration? as in, under what circumstance does a standalone script want to import the whole mycroft-core package? If those are rare and mainly internal scripts (like send bus message) can't that be accounted for a different way?

just thinking out loud, don't feel like this is a blocker in any way, i think the PR is good for merge the way it is :)

forslund: Those are totally valid thoughts...

Another approach would be to cache the different configurations separately and use the cache...When investigating it looked like that at startup there are 4 or 5 uncached calls to the `.get()` before it's cached just to avoid the remote config.

So when calling with remote it would get a cache including the remote and when calling without it would get the cache that was generated with only locals

krisgesling (approved):
Thanks Ake - working great!

I think the discussion above about caching is worth continuing too.

mycroft/util/log.py
krisgesling: Ake have you got examples where you think the cache without remote would be useful? Or is it more theoretical?

Having a single config seems conceptually simpler to me. But then the complexity already exists and this is really just mapping against the somewhat commonly used remote flag, so I can see the benefit of the two caches.

Issue #3044, opened 2021-11-30
XDG `save_config_path` raises `FileExistsError`


krisgesling:
**Describe the bug**
The pyxdg `save_config_path` ensures that a directory exists before using it. This is done using `os.path`:
https://github.com/takluyver/pyxdg/blob/master/xdg/BaseDirectory.py#L57-L58

Based on the Traceback below I'm wondering if we have items starting in separate threads, that are both checking/creating the XDG paths. 

```
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 185, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/usr/lib/python3.8/runpy.py", line 111, in _get_module_details
    __import__(pkg_name)
  File "/opt/mycroft/mycroft/__init__.py", line 17, in <module>
    from mycroft.api import Api
  File "/opt/mycroft/mycroft/api/__init__.py", line 23, in <module>
    from mycroft.configuration import Configuration
  File "/opt/mycroft/mycroft/configuration/__init__.py", line 15, in <module>
    from .config import Configuration, LocalConf, RemoteConf
  File "/opt/mycroft/mycroft/configuration/config.py", line 28, in <module>
    from .locations import DEFAULT_CONFIG, USER_CONFIG, OLD_USER_CONFIG
  File "/opt/mycroft/mycroft/configuration/locations.py", line 26, in <module>
    USER_CONFIG = join(xdg.BaseDirectory.save_config_path('mycroft'),
  File "/opt/mycroft/.venv/lib/python3.8/site-packages/xdg/BaseDirectory.py", line 58, in save_config_path
    os.makedirs(path, 0o700)
  File "/usr/lib/python3.8/os.py", line 223, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/home/mycroft/.config/mycroft'
```

I haven't yet found other projects hitting the same issue.

**Expected behavior**
Create the directory if it doesn't exist, otherwise just use it.

**Environment (please complete the following information):**
 - Device type: Mark 2
 - OS: Mark 2 build dated 2021-11-25 15:34
 - Mycroft-core version: 21.2.1
 - pyxdg v0.26

JarbasAl:
I hit this issue with docker containers and flagged it back then, i recommended to use `xdg_data_home` instead of `save_data_dirs` etc. 

A string definition should be a string definition, not create paths automatically on import

[...]

PR #3045, opened 2021-11-30
Do not create configs folders until writing

forslund:
## Description

This replaces save_*_path with usage of the xdg_*_home when handling
config files as suggested by @JarbasAI in #3044. This means the config system will not create folders unless
actually written.

The check for whether a directory needs to be created is handled behind
a lock to avoid race conditions.

It also changes some usages in strings where the 

The save_config_path() call still exists in the FileSystemAccess-class as a file is opened, which should probably be improved as well.

CLA [ x ]

### DIFF
--- a/mycroft/client/enclosure/mark1/__init__.py
+++ b/mycroft/client/enclosure/mark1/__init__.py
@@ -15,6 +15,7 @@
 import subprocess
 import time
 from alsaaudio import Mixer
+from os.path import join
 from threading import Thread, Timer
 
 import serial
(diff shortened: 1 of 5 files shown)


mycroft/client/enclosure/mark1/__init__.py
```diff
@@ -166,7 +166,7 @@ def process(self, data):
             self.bus.emit(Message("speak", {
                 'utterance': mycroft.dialog.get("reset to factory defaults")}))
             subprocess.call(
-                (f'rm {xdg.BaseDirectory.save_config_path("mycroft")}'
+                (f'rm {xdg.BaseDirectory.xdg_config_home("mycroft")}'
```
JarbasAl, line 169: missing a join ?

forslund: Indeed!

forslund: And the previous path didn't look right either...but should be better now

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-3045))

krisgesling (approved):
Awesome, big thanks to you both! 

Waking up to a solved problem is a very nice way to start the day :smile:

PR #3046, opened 2021-12-01
Do not overwrite new path if it already exists

krisgesling:
## Description
In the migration to XDG compatible paths, we move anything from the old path to the new path. However if for some reason both paths exist the old path would overwrite the new one. 

This should only really happen for devs who are switching back and forth between branches. So just an extra precaution. 

## How to test
With an existing install pre-paired but stopped:
```
mkdir ~/.mycroft/identity
mv ~/.config/mycroft/identity/* ~/.mycroft/identity/
touch ~/.config/mycroft/identity/test
mycroft-start all
```
Prior to this change - the test file will get wiped when the old path overwrites the new one.

- [x] CLA

### DIFF
--- a/mycroft/filesystem/__init__.py
+++ b/mycroft/filesystem/__init__.py
@@ -39,7 +39,7 @@ class FileSystemAccess:
 
         # Migrate from the old location if it still exists
         # TODO: remove in 22.02
-        if isdir(old_path):
+        if not isdir(path) and isdir(old_path):
             if isdir(path):
                 shutil.rmtree(path)
             shutil.move(old_path, path)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-3046))

mycroft/filesystem/__init__.py
```diff
@@ -39,7 +39,7 @@ def __init_path(path):
 
         # Migrate from the old location if it still exists
         # TODO: remove in 22.02
-        if isdir(old_path):
+        if not isdir(path) and isdir(old_path):
             if isdir(path):
```
forslund, line 43: This can never be true now so L43, L44 can be removed

forslund:
This was discussed in the big XDG-PR, the reasoning for leaving it like this was in case a move didn't finish and a corrupted version exists in the new location. Then it would be better to move again to ensure that a valid version gets to the new location.

mycroft/filesystem/__init__.py
krisgesling: Not sure that I follow...
> If the new path doesn't exist but the old path does exist

krisgesling:
Yeah I can get that logic too. 

Just hitting a real issue on the Mark II as presently stable is pre-xdg and latest is post-xdg so if a device switches between them the pairing info gets blown away.

I think this is a very short term problem and happy to just do this on the Mark II feature branch. The whole problem will disappear pretty soon once we no longer have the pre-xdg code on active devices.

mycroft/filesystem/__init__.py
krisgesling: oh the next lines right - this change is saying never remove the new directory :+1:

krisgesling:
I've also added in a symlink at the old directory to try and handle both upgrading from pre-XDG to XDG, and reversion from XDG to non-XDG.
https://github.com/MycroftAI/mycroft-core/commit/294be08efcd89c0e2737e2be6d926d366134bce8

Eg if someone on the Mark II switches from Latest back to Stable - they will be moving from an XDG migrated system, back to a pre-XDG system.

It won't catch if you start on a system that's never migrated from pre-XDG, but I think core devs are the only people doing that anyway. And as I said, very soon this all becomes a thing of the past.

PR #3030, opened 2021-11-08
Use shellcheck directive instead of excludes

forslund:
## Description
This uses shellcheck directives instead of excludes only available in CICD as suggested by @PureTryOut.

This leaves SC1090/1091 as a general excludes, these are set in the .shellcheckrc to make sure the developer gets the same checks as the CI does.

## How to test
Ensure the CICD approves.

## Contributor license agreement signed?
CLA [ Yes ]

PR #3048, opened 2021-12-08
Shellcheck dev_setup.sh

krisgesling:
## Description
Hit an [issue in a Github Actions install of Mycroft-core](https://github.com/MycroftAI/skill-homeassistant/runs/4451242699#step:30:1) where it was unable to evaluate a conditional. This fixes that specific issue along with a range of suggestions from shellcheck.

## How to test
run `shellcheck dev_setup.sh`
install mycroft using `dev_setup.sh`

- [x] CLA

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -571,7 +571,7 @@ if ! pip install -r requirements/tests.txt ; then
 fi
 
 SYSMEM=$(free | awk '/^Mem:/ { print $2 }')
-MAXCORES=$(($SYSMEM / 2202010))
+MAXCORES=$((SYSMEM / 2202010))
 MINCORES=1
 CORES=$(nproc)
 
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -22,7 +22,7 @@ export LANGUAGE=en
 # exit on any error
 set -Ee
 
-cd $(dirname $0)
+cd $(dirname "$0")
 TOP=$(pwd -L)
 
 function clean_mycroft_files() {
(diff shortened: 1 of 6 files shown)


krisgesling:
Also feel free to push additional fixes straight onto this branch

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-3048))

forslund:
In reverse order:
4. The sed expression doesn't need to be spread out on three lines, with a single line this becomes nice and easy
3. Not sure what the best way is but it sounds like `command -v` should be the way to go
2. These seem to be due to using space separated string as commandline arguments, [BashFAQ](http://mywiki.wooledge.org/BashFAQ/050) disapproves of this and recommends an array instead.
1. Easiest and nicest solution is in my opinion to split it into 2 lines

I pushed commits for 1,2 and 4. I suck at shellscripting so another pair of eyes on those changes might be a good idea.

PureTryOut:
`which` is indeed not standard. Debian is even going to drop it soon, there was quite a lively discussion about that recently. `command -v` is the only standard that can be relied upon to work the same across all distros and shells.

krisgesling:
hmm doesn't look like you can add an individual file from outside the included directory. So I've dropped that commit for now.

I'll try find some time to do a clean up of the rest of the scripts since most of it is quite straight forward.

krisgesling:
:facepalm: did a silly on the commit history - will fix it later today

forslund:
May be worth looking at #3030 before doing the rest of the scripts. It adds some shellcheck setup which may be useful. I've been planning to go through the scripts folder but have been waiting for that PR to be reviewed.

krisgesling:
Closing in favor of #3090

PR #3051, opened 2021-12-16
Make TTS queue a singleton

ken-mycroft:
## Description
Fixes issue where when remote tts fails and local gets executed they stomp on each other

## How to test
You would need a special mimic2_tts.py that periodically throws exceptions, otherwise not easy to test

CLA [ yes ] (Whether you have signed a [CLA - Contributor Licensing Agreement](https://mycroft.ai/cla/)

### DIFF
--- a/mycroft/tts/tts.py
+++ b/mycroft/tts/tts.py
@@ -39,19 +39,15 @@ from mycroft.util.plugins import load_plugin
 from queue import Queue, Empty
 from .cache import hash_sentence, TextToSpeechCache
 
-
 _TTS_ENV = deepcopy(os.environ)
 _TTS_ENV['PULSE_PROP'] = 'media.role=phone'
 
-
 EMPTY_PLAYBACK_QUEUE_TUPLE = (None, None, None, None, None)
-
 
 class PlaybackThread(Thread):
     """Thread class for playing back tts audio and sending
     viseme data to enclosure.
     """
-
     def __init__(self, queue):
         super(PlaybackThread, self).__init__()
         self.queue = queue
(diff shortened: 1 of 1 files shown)


forslund (approved):
Super nice, changing this to a singleton is a good way to handle this. Was looking into this the other week when I was debugging the weird zork-skill early bings when running on Mimic2. I was considering injecting the queue but this is a much less intrusive change.

Maybe this single Change could go directly into dev?

mycroft/tts/tts.py
```diff
@@ -355,9 +355,6 @@ def execute(self, sentence, ident=None, listen=False):
         try:
             self._execute(sentence, ident, listen)
         except Exception:
-            # If an error occurs end the audio sequence through an empty entry
-            self.queue.put(EMPTY_PLAYBACK_QUEUE_TUPLE)
-            # Re-raise to allow the Exception to be handled externally as well.
             raise
```
forslund, line 358: Maybe the try-except can be removed completely when the except just re-raises the exception.

forslund:
One thing to note is that we might also need update the handling of the tts-cache. I think in this setup only one of the TTS's cache clearing methods will only be called...

krisgesling:
I was hoping we could merge into both branches, however I've been testing this out on the standard dev channel as well and I think we must have some extra fixes to the mimic1 class in feature/mark-2. If you set your TTS module to mimic it starts playing all sentences in the queue at the same time.

krisgesling:
Merging this now so we can get it onto tonights build but agree with the extra points raised so we should address those as a followup.

forslund:
Ok interesting, I'll try that out and see if I can make this work for dev aswell.

ken-mycroft:
> 

I was worried about this. There might be some mimic1/2 changes i missed. Is this still an issue?

forslund:
I added a couple of things in the backport to dev PR #3055, it has commit for handling the cache clearing and some minor reorganization moving responsibility to the playback-thread. It also has some further fixes where the fallback could misbehave

PR #3052, opened 2021-12-17
Correct incidental findings from review of dev_setup.sh

### LINKED ISSUE #1714: Incidental findings from review of `dev_setup.sh`
_Background_ Did a review of `dev_setup.sh` for #1712 and raised these questions during review, moving to new Issue to separate them out. 

### `dev_setup.sh`

1. In the `show_help()` function, recommend pointing the user to Chat or Forums if they have questions - this isn't a _technical_ thing per se, but it helps to grow the Dev Community

2. just a minor nit-pick, but I'd set the `opt_skipmimic` and `opt_allowroot` flags above `show_help()` and add comments around what they do. 

3. there is an `--allow-root` option, but the code does not allow for the similar `--skip-mimic` - for usability and predictable usage, recommend added `--skip-mimic` as an allowable option. 

4. In the code block that checks whether `--allow-root` is set, and exits if it is not, then I would add a log entry using `syslog` or similar to `mycroft-update.log`, as it makes sense to log this for troubleshooting. 

5. In the function `install_deps()` I'd be inclined to add commentary around which platform each package manager is usually associated with - this would likely help with troubleshooting in the future, particular for non-Debian flavours of Linux. 

6. I might be missing something, but why are we installing `python2` packages with some of the installers, for instance `pacman` for Manjaro? 

7. I think there is a typo in line 119 (of your commit, line 116 of previous commit) - reads `${blue} git python 2 python-setuptools ...`. I _think_ `python 2` should be `python2`. 

8. In the function `install_venv()`, how likely is it that the URL for `get-pip.py` will change in the future? Should it be abstracted into a var at the top of the bash script? 

9. On line 156 I think we should warn the user that building `mimic` can take upwards of an hour. 

10. Do we need to check that the `venv` installs successfully in `install_venv()` and log and exit if it doesn't? It doesn't seem like a common source of failures, but being thorough. 

11. Why do we force the version of `pip`? `pip` is up to about version 18, and I wonder if this causes any confusion? 

12. On line 234 do we need to set the execute bit for `start-venv.sh`? I'm not sure, so asking the question. 

13. why is `/var/log/mycroft` 777? Should it be 775 with `mycroft:mycroft` - again I'm not sure so more "kicking the tyres" / questioning. 

### main.py

* In function `check-dependencies()` we pipe `notify-send` to `/dev/null` - again should we syslog this to one of the logs for troubleshooting? 

END comments

mikejgray:
## Description
Corrects incidental findings from review of dev_setup.sh (fixes #1714). Please let me know if you'd like adjustments to be made - this is an old issue so I hope this PR is still timely!

## How to test
Execute dev_setup.sh in various operating systems

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -110,7 +110,7 @@ for var in "$@" ; do
     fi
     if [[ $var == '-n' || $var == '--no-error' ]] ; then
         # Do NOT exit on errors
-    set +Ee
+        set +Ee
     fi
     if [[ $var == '-sm' ]] ; then
         opt_skipmimicbuild=true
@@ -562,8 +562,8 @@ fi
 
 # install optional python modules
 if [[ ! $(pip install -r requirements/extra-audiobackend.txt) ||
-    ! $(pip install -r requirements/extra-stt.txt) ||
-    ! $(pip install -r requirements/extra-mark1.txt) ]] ; then
+        ! $(pip install -r requirements/extra-stt.txt) ||
+        ! $(pip install -r requirements/extra-mark1.txt) ]] ; then
     echo 'Warning: Failed to install some optional dependencies. Continue? y/N'
     read -n1 continue
     if [[ $continue != 'y' ]] ; then
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -24,6 +24,7 @@ set -Ee
 
 cd $(dirname $0)
 TOP=$(pwd -L)
+LOG_FILE="${PWD}/mycroft-update.log"
 # Whether or not to allow root execution - not recommended if you can help it!
 opt_allowroot=false
 # Whether or not to skip the Mimic build, which can take hours on a slow machine
@@ -120,8 +121,8 @@ for var in "$@" ; do
 done
 
 if [[ $(id -u) -eq 0 && $opt_allowroot != true ]] ; then
-    echo 'This script should not be run as root or with sudo.'
-    echo 'If you really need to for this, rerun with --allow-root'
+    logger -f $LOG_FILE -s 'This script should not be run as root or with sudo.'
+    logger -f $LOG_FILE -s 'If you really need to for this, rerun with --allow-root'
     exit 1
 fi
 
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -358,9 +358,9 @@ Mycroft requires a conflicting package, and will likely uninstall this package.
 On some systems, this can cause other programs to be marked for removal.
 Please review the following package changes carefully."
         read -p "Press enter to continue"
-        $SUDO apt-get install "$APT_PACKAGE_LIST"
+        $SUDO apt-get install $APT_PACKAGE_LIST
     else
-        $SUDO apt-get install -y "$APT_PACKAGE_LIST"
+        $SUDO apt-get install -y $APT_PACKAGE_LIST
     fi
 }
 
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -24,6 +24,13 @@ set -Ee
 
 cd $(dirname $0)
 TOP=$(pwd -L)
+# Whether or not to allow root execution - not recommended if you can help it!
+opt_allowroot=false
+# Whether or not to skip the Mimic build, which can take hours on a slow machine
+opt_forcemimicbuild=false
+opt_skipmimicbuild=false
+opt_python=python3
+param=''
 
 function clean_mycroft_files() {
     echo '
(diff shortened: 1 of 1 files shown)


krisgesling:
Hey Mike, thanks for this - there's some great stuff here!

Just wanted to flag PR #3048 as there's a bit of overlap but definitely a common goal :+1:

mikejgray:
> Hey Mike, thanks for this - there's some great stuff here!
> 
> Just wanted to flag PR #3048 as there's a bit of overlap but definitely a common goal 👍

Ah, hadn't noticed - how do you want to proceed? I don't want to collide with the work you've already done, so please let me know.

mikejgray:
Hey @krisgesling is this still relevant/helpful? I'm happy to refactor or close the PR if not.

dev_setup.sh
```diff
@@ -22,8 +22,16 @@ export LANGUAGE=en
 # exit on any error
 set -Ee
 
-cd $(dirname $0)
+cd $(dirname "$0")
 TOP=$(pwd -L)
+LOG_FILE="${PWD}/mycroft-update.log"
```
krisgesling, line 27: With the current use of `/var/log/mycroft` - I'm wondering if this should go in there, and we can probably drop the `mycroft-` from the filename.

dev_setup.sh
```diff
@@ -62,16 +70,14 @@ Options:
     -p arg, --python arg    Sets the python version to use
     -r, --allow-root        Allow to be run as root (e.g. sudo)
     -sm                     Skip mimic build
+
+Need more help? Please visit:
+Mycroft Chat (https://chat.mycroft.ai/)
+Mycroft Forums (https://community.mycroft.ai/)
```
krisgesling, line 76: Love it!

dev_setup.sh
```diff
@@ -22,8 +22,16 @@ export LANGUAGE=en
...
+cd $(dirname "$0")
 TOP=$(pwd -L)
+LOG_FILE="${PWD}/mycroft-update.log"
+# Whether or not to allow root execution - not recommended if you can help it!
+opt_allowroot=false
+# Whether or not to skip the Mimic build, which can take hours on a slow machine
+opt_forcemimicbuild=false
+opt_skipmimicbuild=false
+opt_python=python3
+param=''
```
krisgesling, line 34: I get the idea of bumping defaults to the top, however I also see these as directly relating to the argument parsing. So in favor of keeping related things together, I'd suggest we leave these where they are.

dev_setup.sh
```diff
@@ -115,8 +121,8 @@ for var in "$@" ; do
 done
 
 if [[ $(id -u) -eq 0 && $opt_allowroot != true ]] ; then
-    echo 'This script should not be run as root or with sudo.'
-    echo 'If you really need to for this, rerun with --allow-root'
+    logger -f "$LOG_FILE" -s 'This script should not be run as root or with sudo.'
+    logger -f "$LOG_FILE" -s 'If you really need to for this, rerun with --allow-root'
```
krisgesling, line 125: I'm not familiar with the shell logger but at least on a Ubuntu 20.04 LTS I seem to be getting a few errors trying to use this.

From a quick look:
- we need to ensure the log file exists first
- doesn't seem like you can set the log file and log to it in the same command?
  `logger: --file <file> and <message> are mutually exclusive, message is ignored`

dev_setup.sh
```diff
@@ -621,7 +627,10 @@ if [[ ! -w /var/log/mycroft/ ]] ; then
     if [[ ! -d /var/log/mycroft/ ]] ; then
         $SUDO mkdir /var/log/mycroft/
     fi
-    $SUDO chmod 777 /var/log/mycroft/
+    setup_user=$USER
+    setup_group=$(id -gn "$USER")
+    $SUDO chown -R "${setup_user}":"${setup_group}" /var/log/mycroft
+    $SUDO chmod -R 664 /var/log/mycroft/
```
krisgesling, line 633: Yeah I assume the intention behind 777 was in the event of multi-user systems. However I'm not sure how common they would really be. 

Perhaps this should be its own PR and anyone interested can have their say.

krisgesling (commented):
Hey Mike, thanks for the nudge.

It's probably worth splitting this into 2 PR's as they seem like distinct changes.
1. Adding an `update.log` - interested in what else should go in this beyond the run as root error.
2. Changing the permissions of the mycroft logs - I think as a standalone PR it has a better chance of being seen by anyone who might be affected by such a change. They can then outline an alternative proposal if they want to.

The extra tidy up of quoting and indents could go in either if they haven't already been covered by previous commits.

mikejgray:
Thanks Kris - I'll break this up into multiple PRs, especially since it's so stale now.

dev_setup.sh
mikejgray: Generally I like to avoid 777 wherever possible...775, as suggested in the open issue, may make sense. But without context of the reasoning behind using 777 I think I'll omit it from the separate PR in favor of addressing it in a more security-focused PR. That's a much larger work effort and probably not one I'd care to undertake without access to more dev environments. Right now I just have Picroft up and running. :)

mikejgray:
One PR is more than sufficient for the changes. Here it is: https://github.com/MycroftAI/mycroft-core/pull/3105

PR #3055, opened 2021-12-18
Backport/playback queue singleton

forslund:
## Description
This backports the TTS Playback queue singleton PR to the dev branch.

This fixes some additional issues (existing on dev at least)
- Handle Mimic2 error on chunk without repetition
- some minor changes where self was used instead of TTS to refer to the singleton for consistency
- Curate cache for all attached TTS's
- Remove use of clear_cache in Mimic (1) TTS implementation, this caused all cache to be removed from Mimic2 when fallback tts was called.

## How to test
The game-zork skill triggers really bad issues when using Mimic2, (500 errors on almost empty utterances) and together with the listening=true it's really noticable. Playing that for a bit should give some idea if it works or not.

CLA [x]

### DIFF
--- a/mycroft/audio/speech.py
+++ b/mycroft/audio/speech.py
@@ -116,10 +116,9 @@ def mute_and_speak(utterance, ident, listen=False):
     # update TTS object if configuration has changed
     if tts_hash != hash(str(config.get('tts', ''))):
         global tts
-        # Stop tts playback thread
-        tts.playback.stop()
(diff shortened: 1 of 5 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-3055))

krisgesling (approved):
Looks great - thanks for this Ake!

PR #3060, opened 2022-01-01
Use double bracket for shell conditional

kleo:
## Description
Fixes `dev_setup.sh: line 471: [: ==: unary operator expected`

## How to test
Start `dev_setup.sh`

## Contributor license agreement signed? 
CLA [x]

PR #3061, opened 2022-01-05
Add initial value for shell variable

luca-vercelli:
Condition

    if [ $disable_precise_later == true ]; 

gives a syntax error if variable is not initialized

Yes

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -70,6 +70,7 @@ opt_forcemimicbuild=false
 opt_allowroot=false
 opt_skipmimicbuild=false
 opt_python=python3
+disable_precise_later=false
 param=''
 
 for var in "$@" ; do


forslund:
This definitely seems like the correct thing to do.

I think #3060 fixes the same issue (in a different way) should both of these fixes be merged...both initializing the value and using double brackets?

Edit: Shellcheck is ok with either solution so no need to change anything

forslund (approved):

krisgesling:
Thanks Luca!

I'm just retriggering fresh builds - I think we're all clear at the moment so should be able to merge this shortly.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-3061))

PR #3069, opened 2022-02-10
Dev setup fix

ExploWare:
==== Fixed Issues ====
replaced single brackets [ ] for double brackets [[ ]] on line 471

=== How to test ===

Start the ./dev_setup.sh and there will not be thrown an error

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -468,7 +468,7 @@ function install_venv() {
 install_deps
 
 # It's later. Update existing config with jq.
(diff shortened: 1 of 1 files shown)


krisgesling:
Hey there, thanks for helping out with the project. 

We have a fix for this change in another PR I've just been away for a number of weeks and need to get back to it. So I'm going to close this pull request and get that change pushed as soon as we can.

Thanks

PR #3076, opened 2022-02-22
Move CLA check to GitHub Action

forslund:
## Description
This runs the CLA check as a github action instead of as a stage in the Jenkins integration test pipeline. The benefit is that the authentication is much simpler to handle.

This can probably be improved somewhat but should hopefully keep the current functionality while allowing the Voight Kampff test suite to run.

CLA [ x ]

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -1,0 +1,42 @@
+name: Contributor License Agreement
+
+on:
+  pull_request:
+    branches:
+      - dev
+      - master
+
+jobs:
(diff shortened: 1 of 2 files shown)


krisgesling:
I've hopefully fixed the immediate issue in #3077 but I think this will be a great way to go and would mean we can easily add it to any repository as it only needs access to itself.

I was saying to Ake privately - it might even be worth us rewriting the script in JS since it's fairly small and contained. Then we can provide it as an actual Github Action.

PR #3073, opened 2022-02-13
dev_setup.sh: Select the proper get-pip url for Python 3.6

forslund:
## Description
The pypi bootstrap script used to update pip no longer supports Python 3.6, instead it recommends Python 3.6 users to use an alternative url. Fixes #3072 

## How to test
Ensure the github action for python 3.6 completes.

## Contributor license agreement signed?
CLA [ yes ]

PR #3080, opened 2022-02-28
Fix pip url for Python > 3.6

krisgesling:
## Description
Fixes the url to fetch pip from on versions of Python > 3.6

## How to test
Ensure all Github Actions complete showing it installs correctly.

- [x] CLA

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -460,10 +460,10 @@ function install_venv() {
     $opt_python -m venv "${VIRTUALENV_ROOT}/" --without-pip
 
     # Check if old script for python 3.6 is needed
-    if "${VIRTUALENV_ROOT}/bin/${opt_python}" --version > /dev/null | grep " 3.6"; then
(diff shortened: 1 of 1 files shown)


krisgesling:
Can't believe I missed it in PR #3073 - but we were grabbing the Python3.6 pip in both branches of the conditional.

Surprising that all the tests were passing still.

devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-3080))

forslund:
Doh!

I assume this worked since python >=3.7 is backwards compatible with older pip code.

It was meant to be

``` if "${VIRTUALENV_ROOT}/bin/${opt_python}" --version | grep " 3.6" > /dev/null; then```

so the output isn't printed.

forslund (changes requested):
I think the redirection to /dev/null should be added to the grep as was originally intended

forslund (approved):

forslund:
Best get this in and hide my copy pasta :). The /dev/null fix can be done separately, it's more important that this works as intended.

PR #3093, opened 2022-04-07
Silence file does not exist error when clearing VK files

krisgesling:
## Description
When clearing VK test files - if some files didn't exist (because they weren't created) the bare `rm` command will report an error that the files didn't exist. We don't care, only that they are removed if they did exist.

## How to test
Run 
```
mycroft-start vktest -t hello-world
mycroft-start vktest clear
```
Should receive an error without this patch. Nicer output with the patch applied.


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-3093))

forslund (approved):
Nice and simple change!

PR #3101, opened 2022-04-25
Issue-3100 - Adding skills methods for padatious

### LINKED ISSUE #3100: Allows skill to dynamically define intents or entities
**Is your feature request related to a problem? Please describe.**
Some skills my require dynamic definition of padatious entities to prevent matching requests intented for other skills.
See https://github.com/MycroftAI/skill-homeassistant/issues/64 and https://github.com/MycroftAI/skill-homeassistant/issues/104

It can currently be done with file within resource directories, but it is not a clean solution.

**Describe the solution you'd like**
New methods should be implemented to dynamically load new entities or intents in padatious with any wanted name. For instance:
```python
   def register_entity_file_as(self, filename, name):
```

**Describe alternatives you've considered**
Currently skill-homeassistant provides absolute path to a temporary file to `MycroftSkill.register_entity_file()`. This method may be updated to only use as name the base name of the file and not the full path.

zeronounours:
## Description
Add two new skills methods to be able to load any file as padatious
intent or entity, even outside of `vocab` and `locale` directories.

Fix #3100

## How to test
I also added unittests on `test_mycroft_skill.py`

### DIFF
--- a/mycroft/skills/mycroft_skill/mycroft_skill.py
+++ b/mycroft/skills/mycroft_skill/mycroft_skill.py
@@ -1015,10 +1015,40 @@ class MycroftSkill:
                          '.intent'
             handler:     function to register with intent
         """
-        name = '{}:{}'.format(self.skill_id, intent_file)
         filename = self.find_resource(intent_file, 'vocab')
         if not filename:
             raise FileNotFoundError('Unable to find "{}"'.format(intent_file))
+        self.register_intent_file_path(filename, intent_file, handler)
+
+    def register_intent_file_path(self, filename, intent_name, handler):
+        """Register an file as an Intent with the intent service.
+
+        For example:
+
+        === food.order.intent ===
+        Order some {food}.
+        Order some {food} from {place}.
+        I'm hungry.
+        Grab some {food} from {place}.
+
+        Optionally, you can also use <register_entity_file>
+        to specify some examples of {food} and {place}
+
+        In addition, instead of writing out multiple variations
+        of the same sentence you can write:
+
+        === food.order.intent ===
+        (Order | Grab) some {food} (from {place} | ).
+        I'm hungry.
+
+        Args:
+            filename:    path of file that contains example queries
+                         that should activate the intent.  Must end with
+                         '.intent'
+            intent_name: name of the Intent to be created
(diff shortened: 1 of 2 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-3101))

krisgesling:
Hey there, thanks for helping to improve Mycroft!

I can definitely see the appeal here. I'm tossing up in my mind whether a separate method is necessary or if we should check to see if the `intent_file` passed into the existing method is a full path?

- In one sense there is simplicity in having a single method to register intents and a single method to register entities. 
- In another sense it adds complexity to that method since it can now accept more variations.  
- Creating the new method also reduces chances of unexpected behaviours being introduced for existing Skills.

Interested in any opinions on this.

If we do go with a separate method, I'm wondering if we can find a more descriptive name. If I'm developing a Skill and see:
- `register_intent_file()`
- `register_intent_file_as()`

It's not immediately clear to me how they differ. I might also interpret the new method as if I'm going to register an intent file as _something else_. Would something like `register_intent_file_path()` provide a better distinction?


_Edit: Just fyi I doubt the VK test failure reported above is related to this PR._

zeronounours:
The current methods allows the use of full path because of `find_resource` relying on `os.path.join` which allows full path components. However, the name of the intent is based on that path and is not sanitized.

Home Assistant skill recently started using full path, but it fails in the end (see https://github.com/MycroftAI/skill-homeassistant/issues/104) as padatious fails to save the network file:
```
FANN Error 2: Unable to open configuration file "/opt/mycroft/.local/share/mycroft/intent_cache/skill-homeassistant.mycroftai:{/tmp/mycroft/cache/HomeAssistantSkill/tracker}.intent.net" for writing.
```

Another solution may be to sanitize the path when generating the name to only keep the basename. However, I think some developpers will have trouble understanding why they fails to load `my_entity.entity` and `/tmp/path/my_entity.entity`.

The name `register_intent_file_path()` is fine for me too.

zeronounours:
I force-pushed a new version with the proposed names `register_intent_file_path` and `register_entity_file_path`.

forslund:

PR #3105, opened 2022-05-21
Improve help message and logging of dev_setup.sh

### LINKED ISSUE #1714: Incidental findings from review of `dev_setup.sh`

mikejgray:
## Description
Corrects incidental findings from review of dev_setup.sh (fixes https://github.com/MycroftAI/mycroft-core/issues/1714). Most of the findings have been corrected already. This PR should close what little remains, aside from a security concern with log file permissions that probably belongs in a separate issue.

## How to test
Execute dev_setup.sh in various operating systems
Execute `shellcheck dev_setup.sh` for linting and best practices

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -75,6 +75,23 @@ Mycroft Forums (https://community.mycroft.ai/)
 '
 }
 
+# Parse the command line
+opt_forcemimicbuild=false
+opt_allowroot=false
+opt_skipmimicbuild=false
+opt_python=python3
+disable_precise_later=false
+param=''
+
+if found_exe sudo ; then
+    SUDO=sudo
+elif found_exe doas ; then
+    SUDO=doas
+elif [[ $opt_allowroot != true ]]; then
+    echo 'This script requires "sudo" to install system packages. Please install it, then re-run this script.'
+    exit 1
+fi
+
 # create and set permissions for logging
 if [[ ! -w /var/log/mycroft/ ]] ; then
     # Creating and setting permissions
@@ -84,14 +101,6 @@ if [[ ! -w /var/log/mycroft/ ]] ; then
     fi
     $SUDO chmod 777 /var/log/mycroft/
 fi
-
-# Parse the command line
-opt_forcemimicbuild=false
-opt_allowroot=false
-opt_skipmimicbuild=false
-opt_python=python3
-disable_precise_later=false
-param=''
 
 for var in "$@" ; do
     # Check if parameter should be read
@@ -145,16 +154,6 @@ fi
 function found_exe() {
     hash "$1" 2>/dev/null
 }
-
-
-if found_exe sudo ; then
-    SUDO=sudo
-elif found_exe doas ; then
-    SUDO=doas
-elif [[ $opt_allowroot != true ]]; then
-    echo 'This script requires "sudo" to install system packages. Please install it, then re-run this script.' | tee /var/log/mycroft/setup.log
-    exit 1
-fi
 
 
 function get_YN() {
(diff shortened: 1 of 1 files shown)


devops-mycroft:
Voight Kampff Integration Test Succeeded  ([Results](https://reports.mycroft.ai/core/PR-3105))

forslund:
Looks like a nice improvement. I think the github action is running into an issue when the log folder hasn't been created and given the correct permissions. Maybe move that stage to the start of the setup?

mikejgray:
Checked locally, should be ok this time.

forslund (dismissed):
I think these changes looks good now and the CI is happy. Do you think it would be a good idea to add a "setup complete" line at the end to explicitly show that the dev-setup completed successfully?

mikejgray:
Probably not a bad idea. We can also explicitly call out that the new logfile exists and may have further information. Should I include that in this PR?

forslund:
I think it's a good idea @krisgesling do you have any opinion?

mikejgray:
I went ahead and added it in.

krisgesling:
Hey sorry, I was away last week but this looks great. 

Agree on the final success message too :+1: 

My only question is whether we need to use the `-a, --append` flag for all the `tee` calls? Currently it seems to overwrite the log file with each new line.

mikejgray:
Thanks @krisgesling , good call-out. I'll be sure to have a better test setup next time. I've submitted that fix.

mikejgray:
@forslund Anything else you'd like me to do on this PR?

krisgesling (approved):
Thanks Mike this is all working great now!

PR #3135, opened 2022-10-13
Lng: Add cs-cz noise words

Tony763:
## Description
Hi, PR contain Czech version of noise words.

### DIFF
--- a/mycroft/res/text/cs-cz/noise_words.list
+++ b/mycroft/res/text/cs-cz/noise_words.list
@@ -1,0 +1,24 @@
+kde
+co je
+který
+jim
+oni
+kdy
+co
+to
+bude
+od
+z
+že
+také
+kdo
+jak
+a
+ale
+také
+proč
+pro
+je
+to
+nebo
+do


devops-mycroft:
Voight Kampff Integration Test Failed ([Results](https://reports.mycroft.ai/core/PR-3135)). 
Mycroft logs are also available: [skills.log](https://reports.mycroft.ai/core/PR-3135/logs/skills.log), [audio.log](https://reports.mycroft.ai/core/PR-3135/logs/audio.log), [voice.log](https://reports.mycroft.ai/core/PR-3135/logs/voice.log), [bus.log](https://reports.mycroft.ai/core/PR-3135/logs/bus.log), [enclosure.log](https://reports.mycroft.ai/core/PR-3135/logs/enclosure.log)

Tony763:
Hi, VK for Wiki failed, what about adding Github action to it (same as in HA skill)?

forslund:
I think the latest version of wiki doesn't have the issue but is held back waiting for a mycroft release.

The wiki-skill is a bit troublesome, the problem is due to wikipedia returning a new answer for a question. It was working on last release of the wiki-skill but has since changed. I think using something like [betamax](https://pypi.org/project/betamax/) in VK providing a set of stable data from the web might be one way to go. The problem is that it means that the tests won't tell the devs whether the skill is actually working alright with real world data.

That said, per skill unit/integrationtests is a great idea :)

forslund (approved):
Google translate tells me this matches well with the english noise words. Nice! (or should I say Noise!)

Tony763:
> I think the latest version of wiki doesn't have the issue but is held back waiting for a mycroft release.
> 
> The wiki-skill is a bit troublesome, the problem is due to wikipedia returning a new answer for a question. It was working on last release of the wiki-skill but has since changed. I think using something like [betamax](https://pypi.org/project/betamax/) in VK providing a set of stable data from the web might be one way to go. The problem is that it means that the tests won't tell the devs whether the skill is actually working alright with real world data.
> 
> That said, per skill unit/integrationtests is a great idea :)

Testing against real integration is always better than just testing against fixed data. That is why I use real Home Assistant installation and not just mocked subset.  Also, CI per repo give you instant feedback when doing changes.

If you want it for wiki skill, let me know and I will prepare it.

JarbasAl:
In ovos and neon we have unittests per skill, i agree it is the way to go and way more useful than VK imho

Eg https://github.com/OpenVoiceOS/skill-ovos-wikipedia/tree/dev/tests/unittests

Tony763:
With HA skill, I run both VK and unittest (allure report is generated).

forslund:
I think using vk in core is valid, even with fixed data. It helps to ensure that a change in core doesn't break. I think both unit tests and the high level tests vk provides are needed to guard against regressions.

For validating skills functionality during (skill) development unittest are probably more valuable.

Tony763:
To wrap it, should I prepare PR with GitHub Action?

forslund:

Issue #3137, opened 2022-10-25
Broken PWD in start-mycroft.sh


Tony763:
Hi commit 56ceb80179a79dfb294f0b3eecdc6612642964c1 break function of `start-mycroft.sh`. 

As pointed by @emphasize, when `mycroft-start` is called, `SOURCE="$0"` will point to `/srv/mycroft-core/bin` and thus break function as everything is run in incorrect directory. If `start-mycroft.sh` is used directly, it works.

The main indicator is: `Please update dependencies by running ./dev_setup.sh again.`

Even if `dev_setup.sh` run again, the message won't go away.

Under hood,  `start-mycroft.sh` try to check MD5 from `.installed` but it is not found (no user alert about missing file)

@krisgesling This is why GitHub Actions stopped working in Home Assistant skill.

```
Autoupdate: false
cat: .installed: No such file or directory
/srv/mycroft-core/bin
Please update dependencies by running ./dev_setup.sh again.
```

forslund:
Closing Issue since we're archiving the repo

PR #3138, opened 2022-10-25
Fix PWD directory of start script

Tony763:
## Description
Hi, I'm proposing a fix for issue #3137. 

For getting directory use `$PWD` and for getting directory of script which was caller use `$0`.

## How to test
Try to start Mycroft by calling `mycroft-start debug` and `sh start-mycroft.sh debug`

### DIFF
--- a/start-mycroft.sh
+++ b/start-mycroft.sh
@@ -14,12 +14,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-SOURCE="$0"
-
 script=${0}
 script=${script##*/}
-cd -P "$( dirname "$SOURCE" )" || exit 1 # Enter scripts folder or fail!
-DIR="$( pwd )"
+
+DIR="$(realpath -- "$PWD")"
 VIRTUALENV_ROOT=${VIRTUALENV_ROOT:-"${DIR}/.venv"}
 
 help() {


forslund:
Hmm, I see that the current code isn't quite working correctly. However the change doesn't seem to work if calling the script from outside the mycroft-core folder (which I believe was the intent of the original code).

Tony763:
If I call `mycroft-start debug` from my home folder, it works normally.

Or You think calling `start-mycroft.sh`?  Should it be called directly?

forslund:
At least it worked previously.

A simple addition seems to be to add
```
cd $(readlink -f "$0" | xargs dirname)
```

before the setting DIR (the old version did something similar)

which will execute the script from the mycroft folder.

The mycroft-start duplicates this but in a bash-specific way.

@PureTryOut do you have any suggestions or comments?

Tony763:
If I change `cd -P "$( dirname "$SOURCE" )"` into `cd $(readlink -f "$0" | xargs dirname)`  in `start-mycroft.sh`  then it can be called from every place.  But If I call `mycroft-start` then it will point to the wrong directory (`/srv/mycroft-core/bin`). `$0` will return path to first called script.

PureTryOut:
> @PureTryOut do you have any suggestions or comments?

Well my suggestion would be merging https://github.com/MycroftAI/mycroft-core/pull/2802 which changes large parts of the script anyway. :stuck_out_tongue_winking_eye: 

The original code wasn't mean to make it possible to run the script from everywhere persé. As far as I can see it will always cd to where the script is located (the real location, not any symlink). I suppose the `bin/mycroft-start` use-case hasn't been tested very well, but with https://github.com/MycroftAI/mycroft-core/pull/2802 I was getting rid of the duplicated entry point anyway.

I'm not entirely sure what to do here besides just getting my PR in, having 2 differently located and named entry points is a bit of a strange situation anyway.

forslund:

